PR #523, opened 2015-01-02
Fix #512 - Add a set_referer method to control where we come back to.

### LINKED ISSUE #512: Login from the issues list
1. On a browser not logged to webcompat.com
2. Go to the issues list page
3. Login
4. Github sends back to the homepage, not the issues list.

miketaylr:
For example, when we go to GitHub to login, we lose the URI of where
we were when we clicked "login". This allows us to keep track of that.

r? @karlcow

### DIFF
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -15,6 +15,8 @@ from ua_parser import user_agent_parser
 from webcompat import app
 from webcompat import github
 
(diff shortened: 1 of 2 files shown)


karlcow:
So the python code looks solid. 
I can't test because my repo is in a mess.

miketaylr:
Thanks. Hopefully you can get your repo sorted out.

Issue #432, opened 2014-12-03
Greatest hits radio select


calexity:
For bug reporting, populate a short list of most-submitted issues so it's easier to select what's going wrong.

As a mobile bug reporter
When I report a bug
Then I want it to be easier to select what's wrong
So I can report it faster.

miketaylr:
Looking at @magsout's commit, I wonder what about the case where the issue is "none of the above"? The current proposed design doesn't give an option for a user to write in their own summary. Would an option like "None of these, I'll write my own" that shows the "Problem in 5 words*" input be possible?

calexity:
@miketaylr @magsout I think that's fine, what about just moving "Add more details" up to right below this to make it more clear that they should just add it there?

miketaylr:
That could work @calexity. Would be cool to see a small mockup when you get a chance.

Would just need to come up with a nice generic bug title "bloop.com - [something generic here]"

calexity:
How about something like this?

Small
![largenewform - upload screenshot](https://cloud.githubusercontent.com/assets/1103366/5716908/12c4060a-9aac-11e4-87fc-6fee9839e097.png)

---
Large
![smallnewform - upload screenshot](https://cloud.githubusercontent.com/assets/1103366/5716909/12c73442-9aac-11e4-9140-a2d9147e5307.png)

For the bug title - maybe "bloop.com - something's wrong"

miketaylr:
Cool. I just realized "Don't know but something wrong" can probably be replaced entirely with "Other - I'll add details below."

miketaylr:
> "bloop.com - something's wrong"

That sounds good. Maybe we take the first few words from the description if it exists, and use that as a default.

magsout:
@miketaylr my PR it' ok with proposal from @calexity ? Just need to manage `form.py`to pupulate radio ?

miketaylr:
/o\, sorry for not looking at this for a while @magsout. I'll check out that PR either at some point today or tomorrow. It would be good to finish this feature soon to be able to tackle image uploading.

miketaylr:
OK, for people not following #524, I screwed up rebasing the branch and ended up closing the PR. Let me see if I can get this guy working and we'll pick it up from there.

miketaylr:
Here's what I see @magsout: ![](https://cldup.com/ZxLn-ZfVnM-2000x2000.png)

Looks good to me. So to finish the first parts of this bug (minus image upload etc.), there's two parts for me:

[...]

PR #524, opened 2015-01-03
Greatest hits radio select

magsout:
I just change the position of the form fields like [screen](http://localhost:5000/) of @calexity  . I have not touched the python or the js

cc @miketaylr


calexity:
Thank you @magsout!

miketaylr:
So this is failing because we're removing the `#summary` input and tests depend on that. Question: do the multiple choice options replace the summary (i.e., 5 words or less title) entirelly, or do they supplement it? 

My thinking is if we don't make it available it may be limiting. I wonder if we could hide it by default and give the option to write in your own summary?

We also need to come up with the best way to distinguish between small viewports (where we want the "greatest hits" and large viewports without it. It could just be that the radio options are hidden by a media query and we ignore them on the backend if there's no value.

magsout:
@miketaylr the goal of the PR was just beginning the development around the new form and iterate on it to get an idea

miketaylr:
Yeah, understood @magsout. It's good--cause it brings up unsolved/undefined issues. I wonder if we should have a long running branch to hack on this in. Pulling this into master as is is probably a bad idea.

magsout:
must not merge, I should specify the beginning.

miketaylr:
Let me see if I can rebase this branch against master. :boar:

miketaylr:
Ahahah, rebase-fail.

miketaylr:
Sorry to screw this up @magsout. I'm going to nuke this branch, make a new one and cherry-pick 493a4c5ba5a3dc4aee88c51e0f421b3718cd2343.

miketaylr:
Let's continue this discussion in #432. I made a `issues/432/2` branch where we can iterate.

PR #530, opened 2015-01-06
Fix #529 - Use closest() to ensure we add is-active to correct wrapper

### LINKED ISSUE #529: Clicking chevron of dropdown doesn't open dropdown in Chrome
Possibly other browsers.

Digging into this, it's because `e.target` in Chrome is the `<span>` that holds the icon, whereas in Firefox it's the `<button>` element that contains the `<span>`. Simple fix coming up.

miketaylr:
r? @magsout 

I guess Chrome and Firefox have different handling of html elements inside of `<button>` :fried_shrimp:


miketaylr:
self r+. :sunglasses:

PR #539, opened 2015-01-15
537/1 Just making sure that people use localhost

karlcow:


run.py
```diff
@@ -40,4 +46,5 @@ def config_validator():
         print("Starting server in ~*TEST MODE*~")
         app.run()
     else:
+        print("[Use http://localhost:5000/]")
```
miketaylr: Did you mean to print `START_MESSAGE` here?

karlcow: rha… my bad. It was a remaining of a previous test.
You may erase the start message.

karlcow:
Fixing my dirty laundry.

miketaylr:
Yay, thanks!

PR #540, opened 2015-01-15
Fix #330 catch importerror message and raise a message

### LINKED ISSUE #330: Suggest re-running pip install on ImportError from run.py
In run.py, if someone forgot to configure BOT_OAUTH_TOKEN we print a nice message to stdout--added by @karlcow  in #284.

Guillaume just ran into an error because we added a new flask extension that requires you to re-run `pip install -r requirements.txt`.

If, when starting the app you run into an `ImportError`, I think we should exist and display a message to the terminal. Something like "It seems like you don't have all dependencies. Please re-run  `pip install -r requirements.txt`."

karlcow:


miketaylr:
Thanks! :hamster:

PR #548, opened 2015-02-03
Fixes #480 - Remove "divider" element from form section of homepage.

### LINKED ISSUE #480: "divider" on bug report form causing contrast error
So, on the home page at the very bottom of the Report Bug Form there is this:

```
<p class="divider" id="submitdiv">___</p>
```

With this CSS:

```
.divider {
display: block;
font-weight: 700;
border-top: 5px solid #FFC900;
width: 30px;
color: #fff;
margin: 1em 0 0;
}
```

It's causing a somewhat false positive contrast error, but it's also a weird way of doing this (with 3 underscores). I'm not even sure of the purpose of the underscores as the font is white and the color of the divider is done via CSS anyways. Could we change this slightly to get rid of this error?

```
<div class="divider" id="submitdiv"></div>
```

```
.divider {
display: block;
border-top: 5px solid #FFC900;
width: 30px;
height: 23px;
margin: 1em 0 0;
}
```

miketaylr:
Taking another look at this, I think we can live without the divider. We used to have a few to break up the page--but this was the only one left.

r? @magsout

### DIFF
--- a/webcompat/static/css/development/page/home.css
+++ b/webcompat/static/css/development/page/home.css
@@ -1,15 +1,5 @@
 /*------Home view------*/
 
-
-.divider {
-  display: block;
-  font-weight: 700;
-  border-top: 5px solid var(--wc-background-dark);
-  width: 30px;
-  color: #fff;
-  margin: 1em 0 0;
-}
-
 /*-------Browse Issues------*/
 #browse-issues-wrapper {
   background-color: #f0f0f0;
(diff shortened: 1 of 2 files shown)


miketaylr:
(The fix for the Travis failures is in the PR for the back button)

magsout:
r+ 

@miketaylr I can merge ?

miketaylr:
Yep, should be OK now--the fixes to Travis were merged into master.

PR #550, opened 2015-02-04
451/1 refactoring Link Header parsing

karlcow:
This refactors a bit the link parsing. I added also two tests to be sure to do the right thing.
There was a bit of back and forth and rewrite. 

r? @miketaylr

### DIFF
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -226,6 +226,21 @@ def sanitize_link(link_header):
     return ', '.join(clean_links_list)
 
 
+def remove_oauth(uri):
+    '''Remove Oauth token from a uri.
+
+    Github returns Oauth tokens in some circumstances. We remove it for
+    avoiding to spill it in public as it's not necessary in Link Header.
+    '''
+    uri_group = urlparse.urlparse(uri)
+    parameters = uri_group.query.split('&')
+    clean_parameters_list = [parameter for parameter in parameters
+                             if not parameter.startswith('access_token=')]
+    clean_parameters = '&'.join(clean_parameters_list)
+    clean_uri = uri_group._replace(query=clean_parameters)
+    return clean_uri
+
+
 def rewrite_and_sanitize_link(link_header):
     return rewrite_links(sanitize_link(link_header))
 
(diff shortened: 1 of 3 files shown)


webcompat/helpers.py
```diff
@@ -194,37 +194,68 @@ def rewrite_links(link_header):
...
+def rewrite_and_sanitize_link(link_header):
+    '''Sanitize and then rewrite a link header.'''
+    return rewrite_links(sanitize_link(link_header))
+
+
+def parse_link_header(link_header):
+    '''Return a structured list of objects for an HTTP Link header.
+
+    This is adjusted for github links it will break in a more generic case.
+    Use something like https://pypi.python.org/pypi/LinkHeader/ instead.
```
miketaylr: > "Use something like https://pypi.python.org/pypi/LinkHeader/ instead."

Is this a TODO?

webcompat/helpers.py
```diff
@@ -194,37 +194,68 @@ def rewrite_links(link_header):
...
+    '''
     links_list = link_header.split(',')
-    clean_links_list = []
+    header_link_data = []
     for link in links_list:
         uri_info, rel_info = link.split(';')
         uri_info = uri_info.strip()
         rel_info = rel_info.strip()
+        rel = rel_info.split('=')
+        rel_value = rel[1][1:-1]
```
miketaylr: Can you leave a comment for what `rel[1][1:-1]` does? I assume it's grabbing foo inside of `<foo>`, but I think my brain won't remember what that does in a few minutes. :smile:

miketaylr:
LGTM with the minor comment nits. Let me run Intern tests locally to make sure there are no unintended side effects.

miketaylr:
Nice, all tests are passing.

miketaylr:
(You can push the comment changes if you want, they're not really deal breakers)

webcompat/helpers.py
karlcow: It was more in case someone was willing to steal that code for something more generic.

webcompat/helpers.py
karlcow: Yes sure. I will do that.

karlcow:
Comments fixed in #551

PR #579, opened 2015-03-06
Fixes #432 - Give common problem types as a choice for the user

### LINKED ISSUE #432: Greatest hits radio select

miketaylr:
I deployed this to https://staging.webcompat.com so you can see it live and report test bugs, etc.

r? @magsout for CSS/HTML/JS bits.
r? @karlcow for Python bits.

### DIFF
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -48,7 +48,7 @@ class IssueForm(Form):
                           [Required(message=summary_message)])
     username = StringField(u'Username',
                            [Length(max=0, message=username_message)])
-    description = TextAreaField(u'How can we replicate this?', [Optional()],
+    description = TextAreaField(u'Give more details (optional)', [Optional()],
                                 default=desc_default)
     site_owner = RadioField(u'Is this your website?', [Optional()],
                             choices=owner_choices)
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -23,9 +23,14 @@ PROXY_REPORT = 'github-proxy-report'
 SCHEMES = ('http://', 'https://')
 
 owner_choices = [(u'True', u'Yes'), (u'False', u'No')]
-problem_choices = [(u'browser_bug', u'Looks like the browser has a bug'),
-                   (u'site_bug', u'Looks like the website has a bug.'),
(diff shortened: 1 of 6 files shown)


magsout:
:+1: 

lgtm..I just have to change some parts of the css, but one can merge before.

webcompat/form.py
```diff
@@ -16,18 +16,22 @@
 from wtforms import TextAreaField
 from wtforms.validators import Length
 from wtforms.validators import Optional
-from wtforms.validators import Required
+from wtforms.validators import InputRequired
```
karlcow: Alphabetical order. Nitpicking. so before "Length"

webcompat/form.py
```diff
@@ -41,19 +45,16 @@
...
     username = StringField(u'Username',
                            [Length(max=0, message=username_message)])
-    description = TextAreaField(u'How can we replicate this?', [Optional()],
+    description = TextAreaField(u'Give more details', [Optional()],
                                 default=desc_default)
-    site_owner = RadioField(u'Is this your website?', [Optional()],
-                            choices=owner_choices)
-    problem_category = RadioField(u'What seems to be the trouble?',
-                                  [Optional()], choices=problem_choices)
+    problem_category = RadioField(u'What seems to be the trouble?*',
```
karlcow: Typo? `?*`

karlcow:
The python looks good. :+1: Just fix the small things and automatic r+ after this.

webcompat/form.py
miketaylr: Good catch.

webcompat/form.py
miketaylr: Nah, we have the `*` to indicate required (Currently on our Problem in 5 words* and Site URL* labels as wel. But now I notice we don't explain the * anywhere (maybe we don't need to?)

miketaylr:
Okey doke, let's pull the trigger. :gun:

PR #586, opened 2015-03-09
Fixes #583 - Check pip dependencies on starting the server

### LINKED ISSUE #583: Check the dependencies before starting the server
Sometimes we change the modules, etc. For example,

```
-WTForms==1.0.5
+WTForms==2.0.2
```

I wonder if we should add a checker before run. Where we would get a pseudo UI like:

```
python run.py

** You need first to update your python/js before starting webcompat.com
@@Instructions to do so@@
```

On the python side, probably something with `pip install --no-install`
or making a diff in between libraries installed and the ones in requirements.txt

For JavaScript, I don't know how it is checked. 
Also when do you need to run again grunt.

miketaylr:
I wonder if we shouldn't put this behind a command line argument, given how rarely we update a dependency.

r? @karlcow

### DIFF
--- a/run.py
+++ b/run.py
@@ -5,6 +5,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import argparse
+import subprocess
 import sys
 
 IMPORT_ERROR = '''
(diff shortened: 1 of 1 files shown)


karlcow:
> I wonder if we shouldn't put this behind a command line argument, given how rarely we update a dependency.

The issue is not to ask for the check, but that the app warns you that there is an issue, because it has been updated. Specifically because it's a rare event, so you forget to check. :cactus:

miketaylr:
@karlcow I like your version better--mine is just doing magical string comparison. Do you want to send a new PR or make your changes in this branch?

miketaylr:
> Specifically because it's a rare event, so you forget to check. 

Yeah, good point. +1

miketaylr:
@karlcow did you want to write a patch with your suggestions?

karlcow:
Let me dive into this again. 
Reminder to self: finish it when it's still hot. :ramen:

karlcow:
OK I send a new pull request. We can close this one.

Issue #587, opened 2015-03-09
Add grunt task to check npm dependencies


miketaylr:
Using https://www.npmjs.com/package/grunt-check-dependencies (or similar).

miketaylr:
```
🐓 grunt
Running "checkDependencies:check" (checkDependencies) task
>> intern: installed: 2.0.0, expected: 2.2.0
>> Invoke npm install to install missing packages
Warning: Task "checkDependencies:check" failed. Use --force to continue.

Aborted due to warnings.

mtaylor at omg in ~/dev/compat/webcompat.com on master*
🐓 npm uninstall intern
unbuild intern@2.0.0

mtaylor at omg in ~/dev/compat/webcompat.com on master*
🐓 grunt
Running "checkDependencies:go" (checkDependencies) task
>> intern: not installed!
>> Invoke npm install to install missing packages
Warning: Task "checkDependencies:go" failed. Use --force to continue.

Aborted due to warnings.

mtaylor at omg in ~/dev/compat/webcompat.com on master*
🐓 npm install
npm http GET https://registry.npmjs.org/intern/2.2.0
npm http 304 https://registry.npmjs.org/intern/2.2.0
...SNIP...

mtaylor at omg in ~/dev/compat/webcompat.com on master*
🐓 grunt
Running "checkDependencies:go" (checkDependencies) task

Running "jshint:tests" (jshint) task
>> 6 files lint free.
```

PR #593, opened 2015-03-10
Issue #587 - Add a task to check npm dependencies and run by default

miketaylr:
This will check npm dependencies if you run any of the following commands.

`grunt`
`grunt checkDependencies`
`make install`
`make build`

r? @magsout


miketaylr:
Ping @magsout. What do you think?

magsout:
oops.. :+1:  I didn't know that module. 

LGTM

miketaylr:
thanks :)

PR #598, opened 2015-03-25
Display list style disc

magsout:
r? @karlcow


karlcow:
Tested. Looks Good. 
Thanks :zap: 
@magsout r+

PR #601, opened 2015-04-01
Fixes #600: remove wyciwyg from URLs (+test).

### LINKED ISSUE #600: return of the wyciwyg
Originally was in web-bugs 831 (not linking to create noise) as `wyciwyg://0/http://bbs.csdn.net/topics/20282413`. I just went to the site and it reproduces when I click Report Site Issue: http://bbs.csdn.net/topics/20282413.

miketaylr:
r? @magsout


miketaylr:
Thanks!

PR #610, opened 2015-04-13
Added line break for better readability/fill-in-ability

tagawa:
And sorry that it's such a super-tiny PR.


miketaylr:
Such a simple change, but makes a big difference. Thanks @tagawa!

tagawa:
That's why they pay me the big bucks!

PR #638, opened 2015-06-24
Fixes #634: load results from q param (for logged in users)

### LINKED ISSUE #634: Search query in URL has no effect
Steps to reproduce:
1. Go to https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=vivaldi

What should happen:
Only issues containing the term "vivaldi" should be displayed.

What actually happens:
All open issues are displayed.

miketaylr:
This is part 1, which allows us to load results from the URL param. Right now it will only work for logged in users. I'll open a bug to handle the unauthed people.

r? @karlcow

### DIFF
--- a/tests/functional/issue-list.js
+++ b/tests/functional/issue-list.js
@@ -226,6 +226,22 @@ define([
           });
     },
 
+    'results are loaded from the query params': function() {
+        var params = '?q=vladvlad';
+        return this.remote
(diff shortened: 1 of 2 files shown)


karlcow:
Tested and working as expected. 
:love_letter: r+

tagawa:
Thank you for this.

PR #643, opened 2015-06-26
Added feature to show QR code for bug URLs

### LINKED ISSUE #636: [feature] Add QR code to bug URLs
For checking in mobile browsers it would be helpful to have a QR code (or button to show one) for the URLs of problematic sites.

tagawa:
Also added font icon so font files are updated.
Fixes #636

webcompat-qrcode


magsout:
Could you prefix className with `wc-` plz

webcompat/static/css/development/components/issue-detail.css
```diff
@@ -172,6 +172,35 @@
   content: "\f00d";
 }
 
+/* QR code launcher */
+.Qr-wrapper {
```
magsout: Could you move all `Qrimage` style in `Qrimage.css` one className per file. 

One file = one Component

tagawa: Sure, done. I copied the structure of the ```LabelEditor``` code so maybe that should be moved to a single file as well.

tagawa:
Thanks for the quick feedback @magsout. I was keeping name consistency with ```LabelEditor```. Do you mean prefix all the classes with ```wc-```?

magsout:
I started to add a prefix to all classes but I have not finished yet. Prefix all new component is a good thing

magsout:
And yes prefix all class by `wc-ClassName`

tagawa:
Thanks @magsout. Done.

webcompat/static/css/development/components/qrcode.css
```diff
@@ -0,0 +1,90 @@
...
+  display:inline-block;
+}
+@media (--viewport-375px) {
+  .wc-QrImage-wrapper {
+    position:static; /* position of label_editor not depend of this class */
+  }
+}
+
+/* QR code */
+.QrImage
```
magsout: Could you add prefix here too ?

miketaylr:
Thanks Daniel. I'll try to pull this down locally and play with it either this afternoon or early tomorrow.

Ideally we want to write some tests for this--I'm happy to help with that.

webcompat/static/js/lib/qrcode.js
```diff
@@ -0,0 +1,67 @@
...
+  events: {
+    'click .wc-QrImage-launcher:not(.is-active)': 'showQr',
+    'click .wc-QrImage-launcher.is-active': 'closeQr'
+  },
+  keyboardEvents: {
+    'q': 'showQr'
+  },
+  render: function() {
+    // Extract page URL from bug description
+    var body = this.model.attributes.body;
```
miketaylr: I think we probably want to use [`Model#get`](http://backbonejs.org/#Model-get) here. I don't think `attributes` is a public API. (I could be wrong though).

webcompat/static/js/lib/qrcode.js
```diff
@@ -0,0 +1,67 @@
...
+  template: _.template($('#qr-image-tmpl').html()),
+  render: function() {
+    var qrDataUrl = qr.toDataURL({value: this.url, size: 8});
+    this.$el.html(this.template({'qrDataUrl': qrDataUrl}));
+    return this;
+  },
+  closeQr: function() {
+    this.issueView.qrButton.removeClass('is-active');
+    // detach() (vs remove()) here because we don't want to lose events if the
+    // user reopens the editor.
```
miketaylr: s/editor/qr code/

webcompat/templates/issue.html
```diff
@@ -127,9 +132,17 @@ <h1 class="wc-IssueDetail-title wc-Title--l">
...
+  <script type="text/template" id="qr-image-tmpl">
+    <div class="wc-QrImage-header">
+      <button class="wc-QrImage-btn">Close</button>
+    </div>
+    <div class="wc-QrImage-main"><img src="<%= qrDataUrl %>" alt="QR code"></div>
+  </script>
+  
 {% endblock %}
 {% block extrascripts %}
+<script src="{{ url_for('static', filename='js/vendor/qr.min.js')|bust_cache }}"></script>
```
miketaylr: Can we go ahead and add the qr.min.js file to the Grunt concat task?

We talked about lazy-loading this script in the future, but for now I think it's better minified and served with the rest.

See https://github.com/webcompat/webcompat.com/blob/master/grunt-tasks/concat.js#L8.

webcompat/static/css/development/components/qrcode.css
tagawa: Ah, missed that one. Thanks.

webcompat/static/js/lib/qrcode.js
tagawa: Thanks, fixed.

webcompat/templates/issue.html
tagawa: Good call. Done.

webcompat/static/js/lib/qrcode.js
tagawa: OK, done.

tagawa:
Thanks for the feedback @miketaylr. I added a test for opening and closing the QR code which seems to be OK, but then I can't deliberately make it fail so something's not right. Would appreciate you giving it a try.

miketaylr:
Awesome. On my TODO list for today.

miketaylr:
Hey @tagawa, I pushed your branch to staging but the QR code doesn't show for me. Can you take a look?

e.g., http://staging.webcompat.com/issues/234

I see `ReferenceError: qr is not defined issues.min.js:6:9156` in the devtools console, after pressing q or clicking on the image.

tagawa:
Er, slight, ahem, user error :scream: 
I hadn't noticed the QR code was ignored by `.gitignore`. Now added so hopefully should be OK.

miketaylr:
Ah. Good news is that .gitignore bugs are some of the easier bugs to fix. ^_^

miketaylr:
This looks great @tagawa --a billion apologies for taking so long to land it. I split up the test you wrote into 4 tests, and they're all passing for me locally. Assuming Travis is happy we'll go ahead and merge.

Some things to note: it would be cool if you could close the QR modal with Esc, perhaps. Maybe also worth adding a hint like the one we have for the GitHub warp. But those can be new issues. :+1:

miketaylr:
Travis looks happy https://travis-ci.org/webcompat/webcompat.com/builds/71330928

miketaylr:
(I'll try to get this deployed tomorrow).

tagawa:
Thank you Taylormeister. Re ESC to close the modal, I tried adding this but that key event is bound to the labels modal. There may be a way to only attach the event listener when the modal is displayed but I haven't looked into it.

A subtle warp-like key hint is a good idea.

miketaylr:
>  I tried adding this but that key event is bound to the labels modal.

Ah true. I bet we could write some kind of generic `closeModalThingy` handler that checks if either the label editor or the qr modal is open, and closes whichever one it is. (If there isn't some built in way to scope the shortcuts).

miketaylr:
OK, all deployed now.

tagawa:
Yay, thank you! :cherries:

PR #651, opened 2015-07-22
Reduced size and top margin of QR code (fixes #649)

### LINKED ISSUE #649: QR popup is cropped if bug report is short
See https://webcompat.com/issues/1425 and open the QR code popup. The bottom part is cropped away because it extends beyond the bug description, and there's an "overflow:hidden" rule or two..

tagawa:
Fixes #649

### DIFF
--- a/webcompat/static/css/development/components/qrcode.css
+++ b/webcompat/static/css/development/components/qrcode.css
@@ -22,7 +22,7 @@
   top: 100%;
   right: 8px;
   /* caret */
-  margin-top: 18px;
+  margin-top: 8px;
   box-shadow: 0px 1px 5px 0px rgba(0, 0, 0, 0.25);
   border-radius: 6px;
   background:  #fff;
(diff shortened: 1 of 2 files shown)


miketaylr:
Thanks for the quick fix!
I pushed this to staging, but you can see it can still be hidden if the bug report is (uselessly) small: http://staging.webcompat.com/issues/255

Is there any way we can make it float above the container, rather than be clipped by it?

magsout:
>Is there any way we can make it float above the container, rather than be clipped by it?

`wc-IssueDetail-details` has overflow x and y. Maybe we can remove its.

miketaylr:
Just goofing around in devtools, if I add `overflow:visible` to both `.wc-IssueDetail-details` & `.wc-Comment-content` then the QR thingy pops up as expected.

Is there an elegant solution for this you would recommend @magsout?

magsout:
hum, I added properties to prevent overflow with elements like <code> or <pre>, maybe need some test.

It's better to remove properties instead of add visible value.

miketaylr:
Should we go ahead and pull this in? Or did somebody want to make some further changes?

magsout:
@miketaylr go. I'll push some fix later (i can't do on @tagawa branch..)

miketaylr:
> @miketaylr go. I'll push some fix later (i can't do on @tagawa branch..)

Apologies, I totally missed this message. Should we close this since #691 is up for review?

magsout:
Yes I guess

PR #659, opened 2015-08-04
Fixes #658. Remove label namespaces so they can be properly checked in editLabels

### LINKED ISSUE #658: Make sure that editLabels converts between namespace and non-namespace
Currently if you have a label with a namespace (`status-` or `browser-`) and open the label editor, it won't know how to check the current labels, so it checks none. If you close, that means it saves the issue with no labels.

miketaylr:
r? @karlcow


karlcow:
Sorry for the mess. Thanks for solving it.

karlcow:
Do we  have a risk to send un-namespaced labels when closing the editor. https://github.com/webcompat/webcompat.com/blob/50dd910e31348bcca4c1f7275a44a092e9eea80c/webcompat/static/js/lib/labels.js#L141

miketaylr:
@ka

Issue #648, opened 2015-07-21
functional tests are dependent of Mike's repo


karlcow:
Most of the functional tests will fail if not configured properly with Mike repo's. I wonder if we should find a way to make this more resilient. Either by having a setup phase or by making the config always pointing to mike setup.

![capture d ecran 2015-07-21 a 16 21 16](https://cloud.githubusercontent.com/assets/505230/8795087/977b2498-2fc4-11e5-8451-61b38218ecd2.png)

For example, there is a hard dependency on `issue/100`

miketaylr:
Yeah, this is not good. And bad for collaboration, IMO.

A few solutions (ordered from easiest -> hardest):

1) A test repo that is owned by webcompat
2) A script that people can run to populate an arbitrary repo with the types of issues we need to test against.
3) We mock out API requests with fake data

karlcow:
Thinking out loud about potential issues.

1. Seems to be a good solution in any cases. It gives a bit of freedom for testers. Though I wonder if it creates issues in terms of tests which requires login aka ownership of the repo. 
2. Solve the issues of ownership, the issue I can see is that the script has to figure out if the repo has already been created and if it's in the right shape for running the tests.
3. The hardest to start up… but maybe the easier to maintain on a longterm with no dependencies on external infrastructure and ownership. 

## About Mocking with intern
* https://www.sitepen.com/blog/2014/07/14/mocking-data-with-intern/
* https://bocoup.com/weblog/effective-unit-testing-with-amd/
* http://rmurphey.com/blog/2014/07/13/unit-tests/
* http://fredkschott.com/post/2014/05/nodejs-testing-essentials/

I wonder if we could find a progressive way to go about it. For mocking maybe we could try with one simple test and figure out the complexity we have not yet figured out.

karlcow:
in #661 @miketaylr asked about the "issue 100". The intern's tests are all dependent of one canonical issue. see for example https://github.com/webcompat/webcompat.com/blob/master/tests/functional/comments.js#L23

```js
.get(require.toUrl(url(100)))
```
it means that when we setup our own test repos we need to have at least 100 issues. And that this one needs to be configured as the ones of mike's repo. 

1. The easiest solution would be to have the test repo be on webcompat.com and force the configurations for people to use this test repo. 
2. Then in the future, make us less dependent on the repo through mocking some of the tests.

karlcow:
The issue is https://github.com/miketaylr/nobody-look-at-this/issues/100
[...]

Issue #272, opened 2014-09-17
tests: labels


miketaylr:
we need intern tests for:

(list modified by @karlcow on 2015-08-07)

* [ ] **setup**: check if we are logged in 
* [x] **tearDown**: erase all labels
* [x] label gear is here
* [x] label widget opens
* [x] setting a label makes it appear in the list of labels
* [x] removing a label makes it disappear from the list of labels
* [ ] label filter works

karlcow:
Started to write a test for this, not there yet. Just a head up.
https://github.com/karlcow/webcompat.com/commit/5e2ae26c225653fd2ade32f9b05e8d69a8d7b1b0

miketaylr:
:heart:

karlcow:
I'm testing with 

```python
    ISSUES_REPO_URI = os.environ.get('ISSUES_REPO_URI') or "miketaylr/nobody-look-at-this/issues"
```

and 

``` js
define([
  // './functional/index',
  // './functional/comments',
  // './functional/issue-list',
  // './functional/issues',
  // './functional/reporting',
  // './functional/contributors',
  // './functional/history-navigation',
  './functional/labels'
], function () {
  'use strict';
});
```

to ease my pain.

karlcow:
> In that case, the automatic login will fail and you then have 10 seconds to manually enter a username and password in the GitHub login screen that appears.

hmm it doesn't seem I get the 10 seconds

karlcow:
ok managed to enter my credentials quick enough.

```
/Users/karl/code/webcompat.com/node_modules/intern/runner.js:67
× firefox 38.0.5 on MAC - labels - setting a label
NoSuchElement: [POST http://localhost:4444/wd/hub/session/cddc1cb6-5f87-4617-830a-a6238fa0446a/element / {"using":"css selector","value":"LabelEditor-launcher"}] Unable to locate element: {"method":"css selector","selector":"LabelEditor-launcher"}
For documentation on this error, please visit: http://seleniumhq.org/exceptions/no_such_element.html
```

huh?
Checking a page https://webcompat.com/issues/100
Indeed I do not see the gear. Exploring the DOM.

``` html
<div class="Label-wrapper">
          <span class="Label Label--title">Labels</span>
          
          <span class="LabelEditor-wrapper">
<button class="LabelEditor-launcher wc-Icon wc-Icon--gear" aria-hidden="true"><span class="wc-sronly">Edit Labels</span></button>
</span>
          
[...]

PR #661, opened 2015-08-04
Fixes #660. Move all label namespace handling to models.

### LINKED ISSUE #660: Handle label namespacing at model level
We should only make namespaced label conversions just before talking to the API or just after receiving data from it.

miketaylr:
I want to add some tests to this before we merge.

r? @karlcow


karlcow:
Ah interesting. I would have thought of putting it on the python side more than the JS side. 
but cool! At least it's a working solution. :heart: 
Tests are to be added in #272 ?

miketaylr:
Hitting every nail with my JS :hammer: 

> Tests are to be added in #272 ?

Yeah, we can do that. How long do you think until you can commit the first simple test? @ https://github.com/karlcow/webcompat.com/commit/5e2ae26c225653fd2ade32f9b05e8d69a8d7b1b0

I can add some more tests specific to the changes in this PR after that.

You mentioned there was a problem with Issue 100, can you explain that again? Maybe I should move the test repo to webcompat...

karlcow:
> Yeah, we can do that. How long do you think until you can commit the first simple test? @ karlcow@5e2ae26 

You mean a Pull request? Because it is committed. :)

> You mentioned there was a problem with Issue 100, can you explain that again? Maybe I should move the test repo to webcompat...

Let me reply in #648

miketaylr:
Yes, a PR.

karlcow:
Here it is sir. :) https://github.com/webcompat/webcompat.com/pull/662

karlcow:
Note for the future. I need to propose something where we handle this on the API side in the python code. More logical and less duplication of code.

miketaylr:
@karlcow is working on tests, so let's go ahead and pull this in so he's not working against a version of the tree that is going to change.

PR #662, opened 2015-08-05
tests for labels

karlcow:
This is a first PR just to test if I'm not doing anything wrong. 
The functional.js has been modified to execute only the labels.js test. (to not loose time)


tests/functional/labels.js
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  registerSuite({
+    name: 'labels',
+
+    'setting a label': function () {
+      return this.remote
+        .setFindTimeout(intern.config.wc.pageLoadTimeout)
+        .get(require.toUrl(url(100)))
+        .sleep(1000)
```
miketaylr: We can probably remove the sleeps(). I have a few in some tests, but @vladikoff likes to remind me how bad I am for using them.

tests/functional.js
```diff
@@ -3,13 +3,14 @@
...
-  './functional/reporting',
-  './functional/contributors',
-  './functional/history-navigation'
+  // './functional/index',
+  // './functional/comments',
+  // './functional/issue-list',
+  // './functional/issues',
+  // './functional/reporting',
+  // './functional/contributors',
+  // './functional/history-navigation',
```
miketaylr: I can merge this if you uncomment them. ^_^

miketaylr:
Looks good, except for the commented out bits. :+1:

karlcow:
Does it work for you? It doesn't for me. :) I will happily uncomment if it is working.

tests/functional.js
karlcow: I explained in the PR, this was done to avoid to have to run all the tests. I'm not yet confident of the test, because they fail for me, but you asked for a PR. :)

karlcow:
I tried with a lot simpler test.

> ConnectionError: HTTPSConnectionPool(host='github.com', port=443): Max retries exceeded with url: /login/oauth/access_token (Caused by <class 'socket.error'>: [Errno 54] Connection reset by peer)

tests/functional/labels.js
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  registerSuite({
+    name: 'labels',
+
+    'label gear is visible': function () {
+      return this.remote
+        .setFindTimeout(intern.config.wc.pageLoadTimeout)
+        .get(require.toUrl(url(100)))
+        .sleep(10000)
```
miketaylr: Let's remove this. I think you were using it to log-in, but pageLoadTimeout is already 10 seconds.

tests/functional/labels.js
```diff
@@ -0,0 +1,51 @@
...
+        .findByCssSelector('.LabelEditor-launcher span.wc-sronly')
+        .getVisibleText()
```
miketaylr: I think this is going to fail because you're trying to get the visible text for an element that is positioned in outer space for screen readers. Why not find the element that is the wrapper for the label editor (`.Label-wrapper`) check [`isDisplayed()`](https://theintern.github.io/leadfoot/Element.html#isDisplayed)? We could then log out and make sure it's not visible, etc.

tests/functional/labels.js
```diff
@@ -0,0 +1,51 @@
...
+        .then(function (text) {
+          assert.strictEqual(text, 'Edit Labels', 'The label gear text is available');
+        })
+        .end();
+    }
+
+    // 'setting a label': function () {
```
miketaylr: Let's remove the commented out stuff here too - I'd prefer to not commit that (and you can get it from your git history if you want it later).

tests/functional/labels.js
karlcow: and it was not enough. It's why I added it. Login from Japan takes time sometimes. Not always immediate.

tests/functional/labels.js
karlcow: Ah! misunderstood this feature. I didn't find a getText. :(

ok I see https://theintern.github.io/leadfoot/Command.html#isDisplayed
and https://theintern.github.io/leadfoot/Command.html#findDisplayedByCssSelector

but then what do you assert?

Do you know a way to have a kind of output on the terminal for understanding what is sent by this code. Instead of relying on black magic (for me. Tests/Failures is not very effective.)

tests/functional/labels.js
karlcow: Do not **merge** this in master. :) It's not expected to be merged yet. I'm still trying to develop a test which will be working. It's just on a PR because you asked for it. :)

tests/functional/labels.js
karlcow: Diving into http://chaijs.com/api/assert/

tests/functional/labels.js
miketaylr: Oh, OK. My question was probably intended to be "when are you going to send a PR (that is ready to be merged)." :p

But it's all good. Ignore (most) of my comments!

Issue #664, opened 2015-08-05
setUp / tearDown for login and logout on functional tests


karlcow:
Should this part of the code be in a login feature or a setUp that we activate at least in any tests library we need. Right now, there is a dependency on having index.js for the rest of the tests to be working.

```js
    'my issues (when logged in)': function() {
      return this.remote
        .setFindTimeout(intern.config.wc.pageLoadTimeout)
        .get(require.toUrl(url))
        .findByCssSelector('.js-login-link').click()
        .end()
        .findByCssSelector('#login_field').click()
        .type(intern.config.wc.user)
        .end()
        .findByCssSelector('#password').click()
        .type(intern.config.wc.pw)
        .end()
        .findByCssSelector('input[type=submit]').submit()
        .end()
        .findByCssSelector('button').submit()
        .end()
```

karlcow:
Ah interesting to ease with the login process when testing only the labels.js library I decided to include it.
See https://gist.github.com/karlcow/3a1ed0435ae3b36532c1

The first HTTP request doesn't go to the right place. 
I get an expected failure for the tests.

![capture d ecran 2015-08-06 a 10 36 17](https://cloud.githubusercontent.com/assets/505230/9101867/214d9d72-3c27-11e5-9809-cf68b38857cd.png)

But more interesting was the first test (login one.)

![capture d ecran 2015-08-06 a 10 31 45](https://cloud.githubusercontent.com/assets/505230/9101884/54d4d2c8-3c27-11e5-853b-e7b0956d93dc.png)

it failed with a 404 and the wrong URI which we can find in 
https://github.com/webcompat/webcompat.com/blob/master/tests/intern.js#L22-L27

karlcow:
Also using 

```
→ npm -v intern
1.4.21
```

karlcow:
Checking `packages.json` I see

```
    "intern": "2.2.0",
```

I wonder why there's no warning about it. when I run intern in command line. That might explain some of my issues?

miketaylr:
:+1:, seems good.

miketaylr:
> Right now, there is a dependency on having index.js for the rest of the tests to be working.

This is true for:

* comments.js
* issue-list.js
* reporting JS

Not true for:
* contributors.js
* history-navigations.js
* issues.js

miketaylr:
@karlcow another thing to keep in mind is once you've successfully logged in once, you have a session cookie. So doing the login again won't require typing in username/pw. 

[...]

PR #666, opened 2015-08-06
Adds two working tests for labels and a setup for the login #272

karlcow:
* setup login
* check the gear is available
* add contactready


karlcow:
I add initially added a test for removing the contactready label, but the slowness of communication with github. Makes it difficult to test. I will create another set of tests.

A note the way we currently test is not perfect because we tested that our UI is working well, not that the label has been sent to github. We should probably do a reload of the page and then see if the test has been setup (which is shouting for mockup ;) )

tests/functional/labels.js
```diff
@@ -0,0 +1,69 @@
...
+
+  registerSuite({
+    name: 'labels',
+
+    setup: function () {
+      // We should be logged before starting the labels test.
+      return this.remote
+        .setFindTimeout(intern.config.wc.pageLoadTimeout)
+        .get(require.toUrl(url(100)))
+        .findByCssSelector('.js-login-link').click()
```
miketaylr: Won't this log out the user if they're already logged in (from a previous test, for example)?

miketaylr:
> A note the way we currently test is not perfect because we tested that our UI is working well, not that the label has been sent to github.

You could test for the presence of a flash error message. We send those if the communication with GitHub fails: https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/models/issue.js#L131

But re-loading the page and checking for the label also seems good.

tests/functional/labels.js
miketaylr: Also, https://github.com/webcompat/webcompat.com/issues/664#issuecomment-128447179

karlcow: Yes you are right. The test has to be better. 
In the setup I should probably pseudo-code

``` js
loggedin: function checkLogStatus();

if (loggedin) { 
  pass;
} elsif { 
  if (sessionCookies) {
    loginLinkClick();
  } else {
    logMePlease(login, passwd);
  }
};
// let's make sure we are logged before proceeding
if !(loggedin) {
  quitWithErrorMsg('Something is rotten');
}
```

miketaylr: A quick and dirty way to check login state from the DOM is to check for a non-empty `data-username` attribute on the `<body>` element, e.g., `<body data-username="miketaylr">`

miketaylr: Ah, one more thing is we have the `/login` and `/logout` routes we can always visit--which redirect to where you came from. So instead of clicking `.js-login-link` you could `.get(require.toUrl("/login"))` to perform a login without risking logging out. Same for /logout, etc.

karlcow:
ok :)

```
PASS: main - labels - label gear is visible (246ms)
PASS: main - labels - label widget is opening on click (352ms)
PASS: main - labels - Label appears once selected (445ms)
PASS: main - labels - Label has been sent to GitHub (3349ms)
PASS: main - labels - Removes a label (5063ms)
0/5 tests failed
0/5 tests failed
```

@miketaylr apart of the setup with the login process, this is PR ready.

miketaylr:
Awesome @karlcow. Can you squash your PR down to just a few commits? 36 is a lot. ^_^

tests/intern.js
```diff
@@ -37,7 +37,7 @@ define(['intern/lib/args'], function (args) {
     functionalSuites: [ 'tests/functional' ],
 
     // A regular expression matching URLs to files that should not be included in code coverage analysis
-    excludeInstrumentation: /./
+    excludeInstrumentation: /\./
```
miketaylr: Also, can you change this back?

vladikoff: in intern3 you can do `excludeInstrumentation: true`

karlcow:
@miketaylr will 6 or 7 commits be acceptable (one per test) :hammer: :smiling_imp:

miketaylr:
@karlcow haha yeah. I should have specified it would be nice to not merge back in the older commits (from myself).

karlcow:
@miketaylr ok :) thanks. I think I will destroy this one and recreate a new branch starting from the last version of master. And re-play the commits. thanks. It seems easier.

karlcow:
Closing this one in favor of #675

Issue #668, opened 2015-08-08
New Issues filter is not working properly


swarnava:
Go to http://webcompat.com
Browse to New issues only.(https://webcompat.com/issues?page=1&per_page=50&state=open&stage=new&sort=created&direction=desc)

the page is also showing needdiagnosis filter.

miketaylr:
Thanks for reporting @swarnava. We've made a few changes to how we handle labels--so this was probably a regression.

magsout:
Bug seems fixed, right ?

miketaylr:
Ah yes--deployed today. Thanks. :)

Issue #24, opened 2014-03-25
Add ability to upload images to backend


miketaylr:
Screenshots are super helpful for bug reports. We should consider adding this ability to our form and hosting images ourselves (or via a 3rd party API like imgur).

miketaylr:
Sorry for all the noise here. Just organizing my thoughts. We have 2 related Milestones:

1. Image uploading
2. Add a screenshot feature

2 depends on 1.

miketaylr:
There will be 3 possible ways to upload an image:

1. When reporting an image, there should be an input[type=file]
2. When commenting on an image, there should be an `input[type=file]` (in addition to dragon drop, eventually) :dragon: :dragon_face: 
3. When reporting an image from one of our addons that supports some sort of a screenshot API

Here's how I see this working for us, roughly:

For issue creation:
1) The file is uploaded and returns a URI pointer to the file
2) That URI is embedded in the issue body using GitHub image markdown syntax: `![file name here](http://path/to/image/)`
3) (I think GitHub uploads all these to their own CDN)

For issue commenting:
Same thing, but included inside a comment rather than the issue body.

![test](http://ecx.images-amazon.com/images/I/41%2BeZQfvcAL._SY355_.jpg)

miketaylr:
Testing something 

![](https://miketaylr.com/bzla/tacocat.jpg)

miketaylr:
>  3) (I think GitHub uploads all these to their own CDN)

So, that's true for true uploads. If you just provide a link to an image, it uses a content proxy:

https://cloudup.com/iSnv2nObVWn

miketaylr:
Thinking of what's left to do:

* [ ] Configure server with location for file uploads (outside of www root)
* [ ] Some form of testing

miketaylr:
So far it's working (on staging): http://staging.webcompat.com/issues/267

miketaylr:
@magsout how hard is it to add a glyph to our symbol font? I want to add the one @calexity suggested here: https://github.com/webcompat/webcompat.com/issues/431#issuecomment-68215433

magsout:
@miketaylr branch origin/issues/24/1 ?

magsout:
@miketaylr just added icon

miketaylr:
@magsout you rock!

magsout:
```    from flaskext.uploads import configure_uploads
ImportError: No module named flaskext.uploads
make: *** [start] Error 1
```

hum, why ?

[...]

PR #674, opened 2015-08-13
Issue #668: Add in status- prefix to "new" query params.

miketaylr:
Maybe the last bug related to label renaming? :smile: 

r? @karlcow

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ cache:
     - $HOME/.selenium/
 
 addons:
-  firefox: "39.0"
+  firefox: "38.0"
 
 env:
   global:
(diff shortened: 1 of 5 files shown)


miketaylr:
```
× firefox 31.0 on LINUX - issues - Comment form not visible for logged out users
× firefox 31.0 on LINUX - issue-list - results are loaded from the query params 
```

I have no idea why these two tests started failing recently--they're unrelated to these changes (they also fail in my #24 branch).

miketaylr:
I wonder if it's somehow a Firefox 31 bug, even though we've specific Firefox 39. I filed https://github.com/travis-ci/travis-ci/issues/4649 for that issue.

miketaylr:
OK, I got this down to 1 failure. Will keep poking at it.

miketaylr:
And I think the other failure is somehow related to an actual bug. Still looking.

miketaylr:
In the interest of getting us (closer) to passing again I'm going to merge this. The change itself is pretty trivial.

PR #675, opened 2015-08-14
Adding labels tests fix #272

### LINKED ISSUE #272: tests: labels

karlcow:
the setup and teardown are not yet written, but ready to fill, once I figured out how to properly do it. And maybe it's something more general. 

The PR can be merged once reviewed.
Right now the tests will execute in the flow, with the assumption that the login has already happened.

@miketaylr r?

### DIFF
--- a/tests/functional.js
+++ b/tests/functional.js
@@ -9,7 +9,8 @@ define([
   './functional/issues',
   './functional/reporting',
   './functional/contributors',
-  './functional/history-navigation'
+  './functional/history-navigation',
+  './functional/labels'
 ], function () {
   'use strict';
 });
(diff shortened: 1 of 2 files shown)


miketaylr:
So it turns out in reporting.js we log out--so these tests all failed.

I did a quick "log in" in the first test @ https://github.com/webcompat/webcompat.com/commit/535ba46b3fb4c31795a9885cc99a99008e7c1223, just as a temporary measure until we use the setup/teardown stuff for all tests.

PR #676, opened 2015-08-14
669/1 Fixing the issue with unknown keywords

karlcow:
We are now handling the response in JSON with a 404.

@miketaylr r?


miketaylr:
Looks good. :+1:

PR #678, opened 2015-08-18
667/1 To square a bit error handling on the API side of our service. Fix #667

### LINKED ISSUE #667: [http] JSON errors for APIs communications
Tested locally.

```
→ http GET http://localhost:5000/api/issues/100/labels 'Accept: application/json'
```

When doing this call, we receive in return a `404 Not Found`, which is fine except for the body which is a full HTML file. Unnecessary bytes on the wire.

We might want to send either nothing in the body `Content-Length: 0` or a very simple JSON body explaining the error.

karlcow:
* Added a couple of tests
* Added error messages in JSON. Handled at the same level than HTML in views.py
* Removed some hard code failures with some URIs

r? @miketaylr

### DIFF
--- a/tests/test_api_urls.py
+++ b/tests/test_api_urls.py
@@ -19,7 +19,8 @@ import webcompat
 # Any request that depends on parsing HTTP Headers (basically anything
 # on the index route, will need to include the following: environ_base=headers
 headers = {'HTTP_USER_AGENT': ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.9; '
-                               'rv:31.0) Gecko/20100101 Firefox/31.0')}
+                               'rv:31.0) Gecko/20100101 Firefox/31.0'),
+           'HTTP_ACCEPT': 'application/json'}
 
 
 class TestAPIURLs(unittest.TestCase):
--- a/tests/test_api_urls.py
+++ b/tests/test_api_urls.py
@@ -78,5 +78,13 @@ class TestAPIURLs(unittest.TestCase):
         self.assertEqual(rv.content_type, 'application/json')
         self.assertEqual(json_body['status'], 404)
 
+    def test_api_category_mine(self):
+        '''API access category/mine without auth returns JSON 401.'''
+        rv = self.app.get('/api/issues/category/mine', environ_base=headers)
+        json_body = json.loads(rv.data)
+        self.assertEqual(rv.status_code, 401)
+        self.assertEqual(rv.content_type, 'application/json')
+        self.assertEqual(json_body['status'], 401)
+
 if __name__ == '__main__':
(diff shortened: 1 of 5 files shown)


miketaylr:
It looks like this breaks the following:

```
FAIL: main - issues - non-existant issues go to 404 (10532ms)
NoSuchElement: [POST http://localhost:4444/wd/hub/session/dfa77d4d-9275-489d-a394-4177c50ba6b2/element / {"using":"css selector","value":"#pageerror h1"}] Unable to locate element: {"method":"css selector","selector":"#pageerror h1"}
```

Can you investigate?

The expected behavior is that going to `/issues/484748493945/` (or whatever) ends up at the 404 page.

miketaylr:
Instead we get a 404 on the response, which means we end up with an empty page.

karlcow:
in https://github.com/webcompat/webcompat.com/blob/master/tests/functional/issues.js#L93-L101

```js
    'non-existant issues go to 404': function() {
      return this.remote
        .setFindTimeout(intern.config.wc.pageLoadTimeout)
        // TODO: uh, update this in the future
        .get(require.toUrl(url(999999)))
        .findByCssSelector('#pageerror h1').getVisibleText()
        .then(function (text) {
          assert.include(text, '(404)', 'We\'re at the 404.');
        });
    },
```

On the CLI, a request returns a 200 HTML file, which it should not, so I guess it's why it's failing.

```
→ http GET http://localhost:5000/issues/1000000
HTTP/1.0 200 OK
Content-Length: 10291
Date: Tue, 18 Aug 2015 21:33:01 GMT
Server: Werkzeug/0.9.4 Python/2.7.6
```

I will investigate later today.

karlcow:
This is what is happening today on webcompat.com  (which is somehow bad ;) )

capture d ecran 2015-08-19 a 06 49 57

And this is what is happening with this PR (which is equally bad HTTP wise)

capture d ecran 2015-08-19 a 06 53 27

I guess the JavaScript client receiving the 404 answer was expecting something different to send back the html 404.

miketaylr:
I think we just have to change the `message` here: https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/issues.js#L223-L225 to the new message that you're sending ("API call not found" (or whatever it is).

karlcow:
Ah yes! That's it! Thanks miketaylr. I was already going through the line of codes. Excellent. 
Thanks. I will add a commit for this.

karlcow:
Ok hopefully the parsing is fine now.

karlcow:
Locally I get now the 404 http://localhost:5000/404

miketaylr:
Awesome! Looks good now, thanks @karlcow.

PR #679, opened 2015-08-18
Fixes #24: Allow images to be uploaded when creating a new bug report.

### LINKED ISSUE #24: Add ability to upload images to backend

miketaylr:
There's a lot going on here, including:

New `upload` blueprint route (we'll make more use of this in the future when we add comment uploads and screenshot uploads).
Adding a `<input type=file>` for our bug report form.
Image uploading to the server.
Use Flask-Uploads for image uploading.
Use Flask-WTF for form handling and image type validation.
Clientside validation for image types.
Functional tests for clientside file validation.
Unit tests for new upload rote.

Thx to @magsout for his help with frontend, as always.

r? @karlcow


tests/test_uploads.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def testGet(self):
+        '''Test that /upload/ doesn't let you GET.'''
+        rv = self.test_client.get('/upload/')
+        self.assertEqual(rv.status_code, 404)
+
+    def testBadUploads(self):
+        # Loop over some files and the status codes that we are expecting
+        for filename, status_code in \
+                (('foo.xxx', 415), ('foo', 415), ('foo.rb', 415)):
```
karlcow: Not very important but this is authorized:

``` python
for filename, status_code in (
        ('foo.xxx', 415),
        ('foo', 415),
        ('foo.rb', 415)):
    blablah
```

tests/test_uploads.py
```diff
@@ -0,0 +1,137 @@
...
+            # is the same as the one being used by the application
+            class TestingRequest(Request):
+                """A testing request to use that will return a
+                TestingFileStorage to test the uploading."""
+                @property
+                def files(self):
+                    d = MultiDict()
+                    d['image'] = TestingFileStorage(filename=filename)
+                    return d
+
```
karlcow: I have read the comment, and I still find that strange. Why the class definition can not be outside of the for loop but still inside the def. Here it redefines the class at each loop.

webcompat/views.py
```diff
@@ -133,10 +135,14 @@ def index():
...
     # Form submission.
-    elif request.method == 'POST' and bug_form.validate():
+    elif bug_form.validate_on_submit():
         # copy the form so we can add the full UA string to it.
         form = request.form.copy()
         form['ua_header'] = ua_header
+        # Do we have an image ready to be uploaded?
+        image = request.files['image']
+        if image:
+            form['image_upload'] = json.loads(upload()[0])
```
karlcow: Both here or in the upload method, we are not checking the size of the file (or maybe I haven't found). I would put a limit on this. Imagine someone trying to upload a 10 Mb image… just for fun.

karlcow: Just noticed that there are flask-uploads… :/

This one
https://github.com/FelixLoether/flask-uploads/
http://flask-uploads.readthedocs.org/en/latest/

and this one
https://pythonhosted.org/Flask-Uploads/
https://bitbucket.org/leafstorm/flask-uploads/
linked from http://flask.pocoo.org/docs/dev/patterns/fileuploads/

There is an issue about that
https://github.com/FelixLoether/flask-uploads/issues/4

Pip seems to install this one.
https://pypi.python.org/pypi/Flask-Uploads/0.1.3
which is the second one. Cool.

The second "official" project has config options for image extensions limiting.
Which there was a size limiting option too.

Though in the doc it says:

>In addition, you can also use patch_request_class to patch your app’s request_class to have a maximum size for uploads. By default, there is no limit, so it’s possible for script kiddies to crash your server by uploading gigantic files. Calling it will install a limit that prevents it from loading more than a certain amount of data.

``` python
patch_request_class(app)        # 16 megabytes
patch_request_class(app, 32 * 1024 * 1024)
                                # 32 megabytes
```

karlcow:
I would love if @magsout could review the JavaScript. upload seems to me a bit too dangerous to just me thinking the JS is fine. I'll focus on the python code.

tests/test_uploads.py
miketaylr: Good to know, will change.

webcompat/views.py
miketaylr: > Both here or in the upload method, we are not checking the size of the file (or maybe I haven't found). I would put a limit on this. Imagine someone trying to upload a 10 Mb image… just for fun. 

Yeah, I have the following in uploads.py:

> # limit image uploads to 4MB
> patch_request_class(app, 4 * 1024 * 1024)

I picked 4MB arbitrarily.

miketaylr:
> I would love if @magsout could review the JavaScript. upload seems to me a bit too dangerous to just me thinking the JS is fine.

Sounds good. The JS doesn't really do much with respect to uploads. Basically it's just duplicating the file validation on the client side that Flask-WTF is doing on the server side--so people don't have to submit an issue and get an error back (if they have JS enabled). But a sanity-check on `checkImageTypeValidity` in bugform.js and the tests would be welcome.

tests/test_uploads.py
miketaylr: I think that's the point. I can't get it to work outside of the loop, but be my guest. ^_^

webcompat/api/uploads.py
```diff
@@ -0,0 +1,53 @@
...
+
+uploads = Blueprint('uploads', __name__, url_prefix='/upload',
+                    template_folder='../templates')
+JSON_MIME = 'application/json'
+images = UploadSet('uploads', IMAGES)
+configure_uploads(app, images)
+
+# limit image uploads to 4MB
+patch_request_class(app, 4 * 1024 * 1024)
+
```
karlcow: houla. I need a second pair of glasses. Thanks Mike. 
Just took a full screenshot of my tiny screen and it is ~400Ko PNG
The limit you set seems entirely reasonable.

tests/test_uploads.py
karlcow: Would that answer your issue?
I simplified the code to just test it outside of the full application.


```python
class TestingRequest():
    """A testing request to use that will return a
    TestingFileStorage to test the uploading."""
    def __init__(self, filename):
        self.filename = filename

    @property
    def files(self):
        d = self.filename
        return d


def testBadUploads():
    # Loop over some files and the status codes that we are expecting
    for filename, status_code in (
            ('foo.xxx', 415),
            ('foo', 415),
            ('foo.rb', 415)):
        foo = TestingRequest(filename)
        print foo.filename

testBadUploads()
```

The result on the console is 

```
>>> testBadUploads()
foo.xxx
foo
foo.rb
```

karlcow: Maybe to explain a bit. When you put a `def __init__(self, blah):` in a `class`. the blah will be set each time you initiate the class. On the other hand, if you were just putting a variable under the `class` it will be shared by all the instances.

karlcow: If you want to read more about this
https://docs.python.org/2/tutorial/classes.html#class-and-instance-variables

webcompat/api/uploads.py
```diff
@@ -0,0 +1,53 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+'''Flask Blueprint for image uploads.'''
+
+import json
+
+from flask import abort
+from flask import Blueprint
+from flask import render_template
```
karlcow: imported but not used.

karlcow:
@miketaylr 
I imported locally to test a couple of things.
Test at https://github.com/webcompat/webcompat-tests/issues/272

I noticed something.  First of all there are two nested uploads folder. Is it what you want?

```
→ find uploads
uploads
uploads/uploads
uploads/uploads/751a802ba7b56cd1f5d1170fda195a06_1433859324.jpg
```

Second, I would encourage to set a kind of dated space for the images inside upload. Something like

```
uploads/2015/08/image_hash_key.jpg
```

The reason is that the image are not stored into a database but in the filesystem. Filesystems have a tendency to become slow when the number of individual items grow into one unique directory. Having a dated space helps to create a partition that will reduce the number of images in one directory. It makes also less necessary to have a long hash_key, but that's secondary.

tests/test_uploads.py
karlcow: Ah no this will not solve your issue. So what the place you copied from is trying to redefine locally a property of a class.

I was looking at the source of `~/.virtualenvs/webcompatcom/lib/python2.7/site-packages/flaskext/uploads.py` which is already patching partly app.request class.

**Let it leave like you initially propose.**

 I'll try to find a more elegant solution later. This one is working, even if not very DRY. I have the feeling that with `super()` we should be able to override a bit things in the same way that he wanted to do it.

tests/test_uploads.py
```diff
@@ -0,0 +1,143 @@
...
+            rv = test_client.post(
+                '/upload/',
+                data=dict(
+                    file=(StringIO('Fake image data'), filename),
+                ))
+            self.assertEqual(rv.status_code, status_code)
+
+
+class TestingFileStorage(FileStorage):
```
karlcow: On the other hand You do not necessary needs this :)

add  at the top

``` python
from flaskext.uploads import TestingFileStorage
```

And then when you need to use it with the modified content_type and the filename you just need to do

```python
d['image'] = TestingFileStorage(filename=filename, content_length=-1)
```
I tried and the tests are working.

webcompat/api/uploads.py
miketaylr: D'oh! Thank you!

tests/test_uploads.py
miketaylr: Oh, how embarassing. >_< :eyes: 

I've changed it (leaving out the `content_length` kwarg because it defaults to -1).

miketaylr:
> I noticed something. First of all there are two nested uploads folder. Is it what you want?

Oh, we can fix that. That's now actually how it's laid out on the server - just a goof up for the localhost stuff.

miketaylr:
> Second, I would encourage to set a kind of dated space for the images inside upload.

Agreed, let me work on that today.

karlcow:
r+ from me for the python side. 
I don't know if you want to wait for @magsout r+ on JS.

Thanks a lot @miketaylr for your patience and the code. :100: 
This is a very nice feature. I think there's an additional item which might come from this, a separate issue. Design work for the bug submitting form.

miketaylr:
Thanks! Let's wait a day or two to see if @magsout can take a look. No real rush for this feature.

> I think there's an additional item which might come from this, a separate issue. Design work for the bug submitting form. 

Cool. Also I need to file bugs (or find the ones that exist) for: adding upload form to issues/comments, drag & drop of images, API upload support for an arbitrary issue (then the stuff around screenshots).

magsout:
lgtm

karlcow:
Thanks @magsout !

And let's pull it with love.

magsout:
Good job everyone :heart:

miketaylr:
Thanks for review @magsout and @karlcow! Will deploy at some point today.

Issue #396, opened 2014-11-14
Unit tests for API endpoints


miketaylr:
We could write some tests around our [API endpoints](https://github.com/webcompat/webcompat.com/blob/master/webcompat/api/endpoints.py) in a similar style to the ["URL" unit tests](https://github.com/webcompat/webcompat.com/blob/master/tests/test_urls.py) that we currently have, to be run with the `nosetests` command. When we have Travis running, we'll have it run both functional Selenium tests as well as our nosetests.

Logged out state will be easiest, perhaps a good first step.

@karlcow would you like to help out with this? I bet you could sniff out more than a few bugs.

karlcow:
In #667 I tested all the error issues. Once the PR is merged they will 
appear as https://github.com/webcompat/webcompat.com/blob/master/tests/test_api_urls.py

I guess I can add tests for success.

karlcow:
ok important comment for this issue. I need to describe this at a point. It took me a lot of time to pull out the right strings.
https://github.com/webcompat/webcompat.com/issues/712#issuecomment-146747836

karlcow:
Experimenting

```diff
diff --git a/tests/test_api_urls.py b/tests/test_api_urls.py
index cb0c9c9..4d6f887 100644
--- a/tests/test_api_urls.py
+++ b/tests/test_api_urls.py
@@ -27,7 +27,7 @@ class TestAPIURLs(unittest.TestCase):
     def setUp(self):
         # Switch to False here because we don't want to send the mocked
         # Fixture data. Which is OK because these don't touch GitHub API data.
-        webcompat.app.config['TESTING'] = False
+        webcompat.app.config['TESTING'] = True
         self.app = webcompat.app.test_client()
 
     def tearDown(self):
```

Then `nosetests -v tests/test_api_urls.py` gives me

```
API issue for a non existent number returns JSON 404. ... ERROR
API issue search with bad keywords returns JSON 404. ... ok
API access to labels without auth returns JSON 200. ... ERROR
API with wrong parameter returns JSON 404. ... ERROR
API setting labels without auth returns JSON 403 error code. ... ok
API access to user activity without auth returns JSON 401. ... ok
API with wrong category returns JSON 404. ... FAIL
API with wrong route returns JSON 404. ... ok

======================================================================
ERROR: API issue for a non existent number returns JSON 404.
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #686, opened 2015-08-28
683/1 fixing the call to 304

karlcow:
This sends a 404 only when necessary.
I need to add a proper test, but I will do it separately in #396 

r? @miketaylr


miketaylr:
Looks good. We probably have another cache fix opportunity here: https://github.com/karlcow/webcompat.com/blob/683/1/webcompat/api/endpoints.py#L153 (I think?).

miketaylr:
:bow:

PR #696, opened 2015-09-06
Change how we compute the user_id in the session.db.

miketaylr:
I have a theory that somehow our `user_id` computation would be off by 1 (or so) resulting in selecting the wrong user. Possibly by restarting the app if a user was logging in? Unsure, but we know this has happened 2 or 3 times in a year. So it's rare, but very very bad.

The plan here is to do away with integer ids and use a salted + hashed github access token as the user id, so it's guaranteed to be unique. And if we do have a hidden bug somewhere we'll find out because `441262a70a96g6c84339af1fb96365df485e32aa75b03e7ce7a79dcb76f48f8a590caf9066c342ce5efbbd5f75ae962f61aca0206x23b4aa5b4f0becfb6571b9` + 1 (or something, if my hunch is true) will just explode rather than select the wrong user. And maybe that will allow us to find the bug.

r? @karlcow

### DIFF
--- a/webcompat/models.py
+++ b/webcompat/models.py
@@ -4,13 +4,14 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from hashlib import sha512
 from sqlalchemy import Column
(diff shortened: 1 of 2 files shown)


karlcow:
that sounds like a good plan.

PR #704, opened 2015-09-14
Adding very simple logging capabilities to webcompat.com

karlcow:
It's now possible to do something such as 

```python
app.logger.info("IP address: {0}".format(request.remote_addr))
```

or really anything we think is useful for debugging, monitoring for performances, etc.

### DIFF
--- a/config.py.example
+++ b/config.py.example
@@ -55,6 +55,14 @@ DEBUG = False
 if not PRODUCTION:
     DEBUG = True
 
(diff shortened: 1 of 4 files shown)


miketaylr:
+1, thanks @karlcow. Before we log anything that might be privacy-sensitive, let's be sure to update our privacy document. And maybe adding a few notes on usage to CONTRIBUTING.md would be a good addition as well.

Issue #501, opened 2014-12-20
Issue comment - Parse hashtag on comment


magsout:
Like 
![capture d ecran 2014-12-20 a 11 06 24](https://cloud.githubusercontent.com/assets/1997108/5514471/4b2a9b52-8838-11e4-8c9c-a6699ec5c5a9.png)

I have to go on Github to see the correct issue.

calexity:
Is this along the lines of what you were thinking @magsout @karlcow @miketaylr?

![commentstatus](https://cloud.githubusercontent.com/assets/1103366/5694487/5145c94e-9913-11e4-8bc8-0b72382f9130.png)

karlcow:
Yes something like this. That's cool.
I would put either icon or text but not both for the status.

calexity:
In that case, I think the text is clearer. I'd drop the icon at least until we've had them for a while.

magsout:
A tooltip or title perhaps? I am concerned if we add the title of issue in the middle of the sentence, becomes difficult to understand, especially if we just add a reference

karlcow:
@magsout yes. good grief, it can be overloaded. Let's imagine how these could be manipulated, changed, etc. The question is more about, how do we want it. :)

Another way of doing it (layout can be changed, just illustrating the concept):

![capture d ecran 2015-01-31 a 09 55 19](https://cloud.githubusercontent.com/assets/505230/5986018/98a4176a-a92f-11e4-8d8f-3a0997b12dc4.png)

Could become something like:

![capture d ecran 2015-01-31 a 09 55 01](https://cloud.githubusercontent.com/assets/505230/5986017/9880a262-a92f-11e4-8482-a0c9c620bd00.png)

magsout:
I like the second :+1:

karlcow:
Note that this is just exploration. I really think we can do better. Any suggestions are welcome. :)

calexity:
I love the tooltip idea! Though it wouldn't work on mobile.

Something like this
http://cssdeck.com/labs/pure-css-tooltip-with-arrow

miketaylr:
- [ ] add marked.js fork to our repo
- [ ] implement issue parsing

miketaylr:
[...]

PR #721, opened 2015-09-30
linkify user names and #issues

hallvors:
Fixes (more or less) #501 - doesn't touch bugzilla links though, if @karlcow wants that I suggest a different bug..


karlcow:
Usually @hallvors for pull requests ask someone specific for the review. so I guess for this one would be 
r? @miketaylr

miketaylr:
Awesome. Thanks @hallvors.

PR #726, opened 2015-10-01
Fix #563 - Make search visible (and usuable) for all users

### LINKED ISSUE #563: the search feature is hidden and not obvious
it took me a while to find webcompat's search feature. a less persistent person would have closed the tab.

as i understand it, in order to search you need to be logged in and viewing the "all issues" page.  neither of these actions are discoverable.

i see that searching is limited to authenticated users due to github api rate limiting issues, which sucks but that's i guess the price of sitting atop a system which you don't control :(

i suggest always adding the search widget to the front page, and if you're not logged in displaying instructions which indicate that you need to do so before you can search.

miketaylr:
r? @magsout


miketaylr:
You can use `node_modules/.bin/intern-runner config=tests/intern.js functionalSuites=tests/functional/search-non-auth.js` if you just want to run the tests I added in this PR.

miketaylr:
travis nooooo

miketaylr:
*cough* my bad.

magsout:
r+ 

LGTM :+1:

magsout:
Travis wakes up, be nice plz

miketaylr:
Travis is being cranky, let's just pull it in. :bomb:

PR #734, opened 2015-10-01
Fixes #728. Set IssueView._urlParams each time we call loadIssues,

### LINKED ISSUE #728: Investigate issue-list test failure
```
FAIL: main - issue-list - Going back in history updates issue list and URL state (1023ms)

AssertionError: Pagination dropdown label is back to where we started: expected 'Show 100' to equal 'Show 25'
```

Unsure if flaky, or I actually broke something.

miketaylr:
r? @magsout 

Otherwise, we have stale state for popstate events.

This should fix Travis perma-failing on that one test. My bad.


magsout:
absolutely-fantastic

PR #737, opened 2015-10-01
Fixed height of SearchBar

magsout:
Regression due to height of `SearchBar`. All elements inside are `position absolute`, so light bug is mixed in the header

r? @miketaylr


miketaylr:
Ah, nice catch.

PR #738, opened 2015-10-01
Added Dropdown to user to Header

magsout:
Need to populate link inside dropdown
When button is clicked, just added as usual `is-active` to `wc-DropdownHeader`
and for accessibility   change value of `aria-pressed`

r? @miketaylr

### DIFF
--- a/webcompat/templates/topbar.html
+++ b/webcompat/templates/topbar.html
@@ -32,12 +32,17 @@
               <span class="wc-Navbar-link-label">Log in</span>
             </a>
             {% else %}
-            <a class="wc-Navbar-link js-login-link" href="{{ url_for('logout') }}">
-              <span class="wc-Navbar-link-icon wc-Icon wc-Icon--arrow-circle-right" aria-hidden="true"></span>
-              <span class="wc-Navbar-link-label">Logout</span>
-            </a>
+            <button class="r-ResetButton wc-Navbar-link js-login-link wc-DropdownHeader">
+              <div class="r-ResetButton" aria-pressed="false">
+                <img class="wc-Navbar-avatar" src="{{ session.avatar_url }}s=50" alt="User Avatar">
+                <span class="wc-Navbar-link-icon wc-Icon wc-Icon--chevron-down" aria-hidden="true"></span>
+              </div>
+              <div class="wc-DropdownHeader-content">
+                <a class="wc-Navbar-linkDropdown" href="#">just a test</a>
+                <a class="wc-Navbar-linkDropdown" href="#">just a test</a>
+              </div>
+            </button>
             {% if session.user_id and session.avatar_url %}
-            <img class="wc-Navbar-avatar" src="{{ session.avatar_url }}s=40" alt="User Avatar">
             {% endif %}
             {% endif %}
           </div>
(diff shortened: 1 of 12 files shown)


magsout:
Ah hum and yes, I need to rewrite some test about logout user.

miketaylr:
:balloon:

magsout:
Just pushed my last commit for these beautiful and awesome days. 
Thanks again @miketaylr @karlcow @hallvors . Love u guys .

karlcow:
aaaaw :heart: Thanks a lot for the amazing work.

miketaylr:
:tada:

miketaylr:
> Ah hum and yes, I need to rewrite some test about logout user.

I'm happy to re-write the logout test stuff, no worries.

miketaylr:
Ah, actually I changed the test yesterday that instead of clicking on a link, it just navigates directly to the `/logout` route -- so no test bustage here.

webcompat/templates/topbar.html
```diff
@@ -32,12 +32,17 @@
...
+                <span class="wc-Navbar-link-icon wc-Icon wc-Icon--chevron-down" aria-hidden="true"></span>
+              </button>
+              <div class="wc-DropdownHeader-content">
+                <a class="wc-Navbar-linkDropdown" href="{{ url_for('logout') }}">My activities</a>
+                <a class="wc-Navbar-linkDropdown" href="{{ url_for('logout') }}">Logout</a>
+              </div>
+            </div>
             {% if session.user_id and session.avatar_url %}
-            <img class="wc-Navbar-avatar" src="{{ session.avatar_url }}s=40" alt="User Avatar">
             {% endif %}
```
miketaylr: We can kill this empty `{% if session.user_id and session.avatar_url %}{%endif%}` block now, I think. :hocho:

miketaylr:
And I guess we can make this change for nav.html as well.

magsout:
ah yes, it's a different view. I forgot it. There is a way to mixe this two views ?

miketaylr:
I was just wondering about that. Looking at them both, we can pull out the "All issues" "login" "logout" links into its own partial and at least share that code.

miketaylr:
I just pulled out the common stuff and made a `shared-nav-links.html` partial, so we're sharing those in top.html and nav.html. Ideally we could just have the same file though, let me see if there's a way in Flask to know what page we're on and we can do some kind of if/else display block (or something).

miketaylr:
@magsout I rebased against master (and force pushed) just now because I need to fix the login test, so please do a `git pull --rebase` or similar before trying to push on this branch. ^_^

miketaylr:
@magsout just a thought, we could land this on master now if we commented out the "My Activity" link. Then we could work on the me-dashboard in a new branch. WDYT?

magsout:
yeah, looks good to me.

miketaylr:
Cool, I'll comment that out and then merge this.

miketaylr:
r+ for @magsout work. I'll ask for review on the JS additions when we land the dashboard (or something).

magsout:
awesome

miketaylr:
ahahahah

magsout:
Call me : Mr GIF

PR #739, opened 2015-10-01
Fixes #729. Non-logged in searches from home page should have all params in URL bar

### LINKED ISSUE #729: All issues page params should be updated, when reached from homepage search
Right now search works, but the model params aren't merged with `q` for some reason.

miketaylr:
:baby: 

r? @magsout

### DIFF
--- a/tests/functional/search-auth.js
+++ b/tests/functional/search-auth.js
@@ -6,8 +6,9 @@ define([
   'intern',
   'intern!object',
   'intern/chai!assert',
-  'require'
-], function (intern, registerSuite, assert, require) {
+  'require',
+  'intern/dojo/node!leadfoot/keys'
+], function (intern, registerSuite, assert, require, keys) {
   'use strict';
 
   var url = function(path) {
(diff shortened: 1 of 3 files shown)


tests/functional/search-auth.js
```diff
@@ -90,7 +91,29 @@ define([
...
-    }
+    },
 
+    'Search from the homepage': function() {
+      return this.remote
+        .setFindTimeout(intern.config.wc.pageLoadTimeout)
+        .get(require.toUrl(url('/')))
+        .findByCssSelector('.js-SearchBarOpen').click()
+        .end()
+        .findByCssSelector('.js-SearchBar input').click()
```
magsout: it's necessary, autofocus doesn't make the job ?

magsout:
rock_star

tests/functional/search-auth.js
miketaylr: Oh yeah, probably not. Forgot!

miketaylr: Hmm, seems like Selenium does it too quickly. I guess we just leave it in. :fried_shrimp:

Issue #712, opened 2015-09-28
Mocking the GitHub communications for functional tests


karlcow:
When we are testing the UI with intern, we want to make sure that the UI reacts accordingly to what we want of the user experience. Right now, the tests are dependent of Github communication but it's not necessary most of the time for the suer experience. Communications issues in between webcompat and github have to be dealt with elsewhere. 

1. Mock all response from Github to WebCompat in intern (when possible)

There is two features which are currently directly talking to Github. 
* search everywhere (in progress)
* load search results from URI bar

decided in Paris september 2015 at webcompat meeting  with @miketaylr

karlcow:
In issue #663 we wanted to make a canonical test repo. But if we mock the correct issues, we kind of make #663 irrelevant. Let's close #663

miketaylr:
I spent a lot of time today looking at mocking modules and servers, and too many of them make the assumption that we want to mock out python unit tests. So instead I came up with the following idea:

We have a `@api_mock('file.json')` decorator for API endpoints that checks if we're in `TESTING` mode, and if so it serves that file instead of requesting it from GitHub. It serves them from `tests/fixtures/`.

Right now this is a READ-only solution, so it won't work for all tests or endpoints. But it's a start.

miketaylr:
Also need to figure out how to mock logins, and how to decide when to send some kind of fixture-auth.js and fixture-non-auth.js, etc.

miketaylr:
Also figure out why this breaks nose:

```
..E......................
======================================================================
ERROR: API access to labels without auth returns JSON 401.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/miket/dev/compat/webcompat.com/tests/test_api_urls.py", line 68, in test_api_labels_without_auth
    json_body = json.loads(rv.data)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 384, in raw_decode
[...]

Issue #165, opened 2014-07-08
issues data dump


karlcow:
It could be useful for statistics purpose to have a dump of the full DB of issues. 
So someone could do matching with a local db, or extract the information about their own system for importing.

Not urgent.

karlcow:
to karl explore to do the data dump using github repo.

hallvors:
https://github.com/webcompat/issue_parser/blob/master/dump_webcompat_to_db.py creates a database with fields id, summary, URL and fills it with all data from our existing issues.

(It needs to be copied along with the extract_id_title_url.py script to somewhere it sees the webcompat module to import the 'engine' database pointer).

The plan is to 
- [x] Write code to import some data about existing issues to db
- [ ] Move the issue tagger bot to run on webcompat server - https://github.com/webcompat/web-bugs-issue-tagger/issues/10
- [ ] Update the issue tagger bot with code that will dump data to our local DB
- [ ] Maybe also generate / update a web-visible JSON file to be used for the JS that will power #307

hallvors:
Should we add a field having the full body to this DB? It's not hard.

hallvors:
This branch:
https://github.com/webcompat/webcompat.com/tree/dbdump
has code that dumps data about new bugs to a local SQLite database file called issues.db

It's tested locally and seems to work fine. It might lack a tiny bit of polish, I'll do pep8 and flake8 and such shortly before doing a pull request.

It has a field called "url" but should perhaps have a "domain" field?

The webhooks stuff on that branch covers *new* issues. Additionally, there's code here:
https://github.com/webcompat/issue_parser/blob/master/dump_webcompat_to_db.py
that can be run at a suitable point in time to grab all the data for existing issues and adding them to the db. This is more of a one-off operation, you would place that file and extract_id_title_url.py next to the webcompat folder and run.py, log in with ssh and run the dump_webcompat_to_db.py script, and remove those two again.

Comments welcome, but feel free to ignore small issues that pep8/flake8 will complain about :)

karlcow:
Hmm. @hallvors took the ownership. Let me change that.

hallvors:
Indeed I took over this, and I didn't even plan to, just happened to think that the stuff in issue_parser repo could be extended slightly to make a DB dump - and went from there.. :)

[...]

PR #757, opened 2015-10-10
Moving database connection stuff to a dedicated db module

hallvors:
I propose we collect the code to create a db session in one place, for example like this.

(This may not be The Best Way To Do It - would be good with a review from for example @karlcow )

### DIFF
--- a/webcompat/models.py
+++ b/webcompat/models.py
@@ -4,13 +4,14 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from db import session_db
+from db import session_engine
 from hashlib import sha512
 from sqlalchemy import Column
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import Integer
 from sqlalchemy import String
 from uuid import uuid4
-from db import db_session, engine
 
 Base = declarative_base()
 Base.query = db_session.query_property()
--- a/webcompat/models.py
+++ b/webcompat/models.py
@@ -12,11 +12,7 @@ from sqlalchemy.orm import scoped_session
 from sqlalchemy.orm import sessionmaker
 from sqlalchemy import String
 from uuid import uuid4
-from webcompat import engine
-
-db_session = scoped_session(sessionmaker(autocommit=False,
-                                         autoflush=False,
-                                         bind=engine))
+from db import db_session, engine
 
 Base = declarative_base()
 Base.query = db_session.query_property()
--- a/webcompat/models.py
+++ b/webcompat/models.py
@@ -8,8 +8,6 @@ from hashlib import sha512
 from sqlalchemy import Column
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import Integer
-from sqlalchemy.orm import scoped_session
-from sqlalchemy.orm import sessionmaker
 from sqlalchemy import String
 from uuid import uuid4
 from db import db_session, engine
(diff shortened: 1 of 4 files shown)


karlcow:
@hallvors 
* Where is the issue this pull request depends on?
* What was the discussion about it?

webcompat/models.py
```diff
@@ -8,15 +8,9 @@
...
-from sqlalchemy.orm import scoped_session
-from sqlalchemy.orm import sessionmaker
 from sqlalchemy import String
 from uuid import uuid4
-from webcompat import engine
-
-db_session = scoped_session(sessionmaker(autocommit=False,
-                                         autoflush=False,
-                                         bind=engine))
+from db import db_session, engine
```
karlcow: we use alphabetical order for import and one import per line.
see https://github.com/webcompat/webcompat.com/blob/master/CONTRIBUTING.md#python

miketaylr:
In spirit this seems fine. But do we want to be adding other tables and data to the "session.db"? Or do we want to create new databases for different things, i.e. issues? I don't actually have an answer, curious what others think.

(If we keep it all in a single db, we should probably rename it to not be session.db)

karlcow:
I have a hunch that it should be a separate db, for at least one reason, that it is easier to un-mess things if we need to fully drop it because it has been polluted. On the other hand, we probably want to keep the bugs data if any in the future.

miketaylr:
I also tend to like the idea of separate databases. And right now I can't think of any relationships between user_id/session and a table of issues... On the other hand, at some point we will probably want a relationship between issues and *usernames* for dashboard stuff. But we don't store usernames in our session table.

@hallvors can you refactor this so we're creating two databases, one for session and one for issues?

hallvors:
@karlcow This is a little refactoring that is meant to make the fix for #165 easier and cleaner. Is it mandatory to have an Issue and a Pull Request even for small fixes? IMHO a PR is a good place to discuss things ;)

@miketaylr Sure, will change this to have two data files.

I also forgot to do the pep8 thing..

miketaylr:
> Is it mandatory to have an Issue and a Pull Request even for small fixes?

Probably not mandatory, usually when I do this kind of refactoring I'm already inside a branch working on a feature (that has an issue). So it would just be a single commit like `Issue NNN. Move database connection stuff to a dedicated db module` (or something). 

I would say for simple fixes a PR is fine, but for stuff that would benefit from discussion Issues are a good thing™ (even though yes there are comment boxes in this PR :P).

hallvors:
I've fixed the variable names to be session_db and session_engine, so whenever other parts import those it should be pretty clear what DB we're using. When we need other files and tables we can add them in db/__init.py__ . OK this way?

hallvors:
Hm.. not sure why the Travis build fails now, this runs OK locally as far as I can tell.

miketaylr:
> Hm.. not sure why the Travis build fails now, this runs OK locally as far as I can tell.

It looks like some unrelated label test failures. Somehow a label got left on the test issue, so everything that expects it to not have a label fails.

Those tests have been flaky in the past, so I wouldn't worry about it. Hopefully once #712 lands, we'll be less affected by these kinds of failures.

miketaylr:
> OK this way?

Looks good to me. In a perfect world the commit messages would start with `Issue #165.`, since these commits get you closer to solving that issue.

(Also that way they are cross-linked in the issue and make it easier to track down the commits/history in the CHANGELOG (we list PR and Issue)).

But I'll take the contribution over perfection. ^_^

miketaylr:
I deployed this to staging and logging in and out still works, so that's good. :balloon:

PR #759, opened 2015-10-10
don't depend on cat, fixes #758

### LINKED ISSUE #758: run.py depends on cat, why?
Trying to run run.py on Windows fails because it tries to call the cat command on line 63:

    req = subprocess.check_output(["cat", "requirements.txt"]).splitlines()

The easiest workaround is to do ```run.py -t``` to launch "test mode", and I guess we'll never ever run this on a system where 'cat' is not available for production. But why depend on cat in the first place?

    for req in open("requirements.txt"):
        pkg_resources.require(req)

should do, I think.

hallvors:


karlcow:
When creating commit message think about putting the issue number. :)

karlcow:
This is about #758 I guess

karlcow:
Thanks!

hallvors:
As the summary says, it's #758 yes ;) but OK, so it should be in the commit message..

Issue #317, opened 2014-11-01
Reporting new issues with a link


karlcow:
Currently to report a new issue, one has to go to the home page of Webcompat.com, click on the arrow and then fill the form. 

We might be able to make it a bit more straightforward for people, companies, etc. To say report any Web Compatibility issues on webcompat, by having a route for posting to the form.

```
https://webcompat.com/issues/new
```

It provides the form to fill the issue and once fill send a POST to our site.
That could be `text/html` or `application/json` depending on the `Accept` of the client.

karlcow:
I think we can separate the notion of open form (currently `?open=1`) and the POST on the route (currently `/`). 
The HTML form in the home page could have `action="/

```html
<form method="post" enctype="multipart/form-data" class="js-ReportForm wc-ReportForm wc-Form" 
           action="/issues/create" 
           novalidate="">

</form>
```

Currently if we change it manually in the form before submitting the request, it appropriately does:

```
127.0.0.1 - - [10/Sep/2015 16:24:41] "POST /issues/create HTTP/1.1" 405 -
127.0.0.1 - - [10/Sep/2015 16:24:41] "GET /img/loader.gif HTTP/1.1" 200 -
```

And it displays:

```
Method Not Allowed
The method is not allowed for the requested URL.
```
We just need to create a new route to handle these requests. 

The benefits:

* Doesn't change anything to the current home page.  `?open=1` still opens the form
* Makes it possible to simplify the current `@app.route('/', methods=['GET', 'POST'])` in `views.py`
* Makes it possible to have a proper template for this route (including a simplified page for plug-ins)

miketaylr:
This will get fixed by #1193, so I'm going to close this in favor of that.

PR #770, opened 2015-10-15
Issue #317. Add /issues/new route for issue creation.

miketaylr:
Leaving out "fixes" from PR message because this is only a partial fix for #317. More work to do, but this first part helps me with screenshot uploading.

I'm happy to tackle the `GET` route for `/issues/new` at a later point, or someone else can take the reins once this lands.

r? @karlcow


karlcow:
:heart_eyes: thanks.

PR #780, opened 2015-10-20
add incoming issues to local db, fixes #788

### LINKED ISSUE #165: issues data dump

### LINKED ISSUE #788: Set up live "backup" database on webcompat.com server, updated from suitable webhooks
Spin off from #165, we are considering having a "backup" database where we store copies of all web-bugs issues when they are created. The idea is to let GitHub send information to our webhooks and use that to update the database. See #780 for code.

The purpose of this database is twofold - one important purpose and one nice-to-have:

1) We can massage the data a little more and for example add a "domain" field to index on - this lets us easily generate RSS feeds and such by domain, see #60, with less traffic to GitHub.
2) It's nice to have the data around locally, especially if we ever wish to move away from GitHub as a backend.

hallvors:
r?@karlcow who seems to have read the code already


hallvors:
Hm. I handled Karl's second comment before the first, and now  GitHub shows my commit associated with wrong comment - oh well :)

webcompat/db/__init__.py
```diff
@@ -17,3 +24,47 @@
...
+
+    def __init__(self, issue_id, summary, url, body):
+        self.issue_id = issue_id
+        self.summary = summary
+        self.url = url
+        self.body = body
+
+Base.metadata.create_all(bind=issue_engine)
+
+Base2 = declarative_base()
```
miketaylr: nit: can we use more descriptive names than `Base` and `Base2`? Maybe `SessionBase` and `IssueBase` or similar?

miketaylr:
I wasn't asked for review, but this looks good. Will defer to @karlcow.

hallvors:
(Quite likely we should have some error handling if the content of a report doesn't fit those DB column sizes..)

hallvors:
Two things to explore:
* Size limits for various fields
* If we should track deletion of issues, edits to issue description

hallvors:
> I wasn't asked for review

I'm just following the guidelines ;) - "don't ask more than one person for review"

miketaylr:
@hallvors heh, no issues with what you've done or the process. I was just adding my 2 cents. :moneybag:

karlcow:
I think we should track deletion. 
I would be happy to give r+ if we are not using the DB for now for production stuff. So write only and not read, for the purpose of testing on the rest of the code. 

Does that sound reasonable?

miketaylr:
> I think we should track deletion. 

I guess that's probably a new issue, to sync with GitHub API (deletion) changes.

> So write only and not read, for the purpose of testing on the rest of the code

+1. Let's let it collect data for a few weeks and come back and inspect how it does. Also if we want deletion, we don't want to rely on anything until that's ready.

miketaylr:
> I would be happy to give r+ if we are not using the DB for now for production stuff. So write only and not read, for the purpose of testing on the rest of the code.
> 
> Does that sound reasonable?

@hallvors want to respond and then fix the merge conflicts? Assuming you agree with Karl I'll be happy to merge at that point.

hallvors:
Rebased and fixed conflicts

karlcow:
Let's see if we get interesting results.

hallvors:
Wohoo!! :)

PR #782, opened 2015-10-20
Fixes #773. Prevent caching empty labels responses from GitHub

### LINKED ISSUE #773: Empty Label response can possibly be cached from GitHub?
Sadly our existing labels tests aren't good enough. This came up trying to get #712 working.

Right now on production if you open the label gear you get an empty label w/ the following error in the console: `no element found`

Looks like we're sending an empty response?

This doesn't happen in dev, but the other bug is that if you open the label editor on an issue with labels, the checkboxes aren't selected. That's a regression, likely from all the namespace mangling.

I'm beginning to want to rip out all the special namespace handling code and just show them everywhere. 

@karlcow any good reasons to not show `browser-foo` and `status-foo` in the webcompat UI? Or were we just trying to have a seamless transition?

miketaylr:
r? @karlcow

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 Babel==1.3
 Flask==0.10.1
-Flask-Cache==0.13.1
 Flask-Limiter==0.7.4
 Flask-SQLAlchemy==1.0
 Flask-Uploads==0.1.3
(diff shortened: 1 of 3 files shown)


karlcow:
excellent. 
Thanks @miketaylr

PR #790, opened 2015-10-22
Extra quoting in CSS selector, fixes #789

### LINKED ISSUE #789: Label containing spaces will break label editor filtering
Since we don't have any labels with spaces at the moment we haven't noticed, but search/filter won't work because ```$('[name=' + labelName + ']')``` will throw on input with spaces. The selector needs some extra quoting.

hallvors:


miketaylr:
Good catch.

PR #794, opened 2015-10-22
Fixes #792: Clicking on a suggested label should actually work.

### LINKED ISSUE #792: "Filter by popular labels" does nothing
If you do a search that matches nothing, you get a list of ways you might want to improve your search. 

One of them is "filter by popular labels" with a small list of clickable labels. Nothing happens when you click them.

(Aside: it is a bit strange to suggest improving a search that matches nothing by "filtering" it further. Are those links meant to apply the labels as a filter to the existing search with no results??)

@miketaylr  @karlcow  @magsout @calexity - thoughts?

miketaylr:
Let's see if the tests actually pass.

r? @hallvors

### DIFF
--- a/tests/functional/search-non-auth.js
+++ b/tests/functional/search-non-auth.js
@@ -34,6 +34,21 @@ define([
         .end();
     },
 
+    'Clicking on label search suggestion works': function() {
+      var params = '?q=dfjdkfjdkfjkdfjdkjf';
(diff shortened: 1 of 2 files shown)


miketaylr:
This will be obsolete.

miketaylr:
(I'll update the PR once #796 lands)

Issue #784, opened 2015-10-21
Labels editor forgets selected labels when you've made changes once


hallvors:
1) Go to issue, for example https://webcompat.com/issues/1
The issue should have two labels: "firefox" and "sitewait"

2) Open label editor and apply another label, for example "windows". Click "Save and close"

3) Open label editor again

Bug: no labels are selected. If you close the editor unaware of this state, it will remove all labels from the issue.

magsout:
Hum yes confirmed... 

But if you refresh page and open Editor labels are selected.

miketaylr:
Yeah, this is likely some kind of regression. It's worked in the past.

hallvors:
The code just throws the response model at the issue. I'm not sure if that ever worked? Anyway, I have a fix.

hallvors:
(Bug is here: https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/models/issue.js#L219 - not sure what exactly the response object is, but it takes a little more work to get the labels array..)

karlcow:
What I see myself.

1. Go to a bug with already a label
2. Click on label widget
3. Select an additional label
4. The previous ones disappear from the list
5. clicking out of the box, only the one you have selected is saved, the old one has disappeared.

This doesn't always happen. I haven't identified if it's a specific label or if it's an HTTP issue.

hallvors:
The specific label firefox-mobile will always disappear because we don't have it in the current set of labels. What probably happens here is that it's the only legacy label with a hyphen in it, and since we use hyphens as a "namespace" separator, the code will handle "firefix-mobile" as a label called "mobile" in the namespace "firefox". Since we AFAIK also have a label called just "mobile", we most likely end up applying that label instead.

I think we should kill the firefox-mobile label. It's going to be very confusing to have some labels with hyphens that are *not* intended to be namespaced, although if we move to a hard-coded set of labels it won't be impossible to do so.

karlcow:
all new bugs are written with `firefox-mobile` so webcompat-bot is setting the firefox-mobile maybe we need to change that.

karlcow:
See https://github.com/webcompat/web-bugs/labels/browser-firefox-mobile

hallvors:
hm.. that's browser-firefox-mobile. OK, I need to debug this.. #798 is probably invalid.

PR #796, opened 2015-10-22
New label code - fixes #783, #784, #797, #799

### LINKED ISSUE #783: [discussion] Simplify label/prefix handling with a proposed new LabelList model
From IRC by @hallvors 

> I have a proposal, sort of: https://gist.github.com/hallvors/9cdf228aeb72598b78eb for a `LabelList` model initialized by either URL or array. Just `list.get('labels')` to get an array of objects fit to present in the page (`list[i].name` and `list[i].color` ready for your templates). just `list.save()` to send stuff to backend - prefixing should just happen. But it's not well tested yet :-p. Didn't really use the Vanilla.js framework ;). (when adding labels - when new things are checked, for example - it accepts `list.set('labels', listOfObjects)`). If you like that, I'll rewrite all code touching label stuff to use `LabelList`.

> @miketaylr said: "i'll leave some comments in the [gist](https://gist.github.com/hallvors/9cdf228aeb72598b78eb)"

hallvors:
r?@miketaylr


miketaylr:
https://github.com/webcompat/webcompat.com/blob/master/tests/test_api_urls.py#L67 needs to be updated.

webcompat/static/js/lib/labels.js
```diff
@@ -4,31 +4,119 @@
...
+*
+* new issues.LabelList({labels: [{name:'status-worksforme', url:'...',
+*    color:'cccccc'}]});
+*
+* Or a URL to a JSON file describing the labels:
+*
+* new issues.LabelList({url:'/path/to/labels.json'});
+*/
+
+issues.LabelList = Backbone.Model.extend({
```
miketaylr: Now that the label model is more than 5 lines (or whatever), can we split this out and add to `/lib/models/labels.js` or `/lib/models/label-list.js`?

miketaylr:
Looks good to me, just a few things to fix:

- [ ] move label model to own file
- [ ] update failing test
- [ ] teach @api.route('/issues/labels') how to do a proxy request for non-auth users

hallvors:
(On the /api/labels requests, we're probably moving towards having a set of labels "hard-coded" in config, if we can land that soon we won't be using Twitter API here at all I think. But I'll figure out how to make it a "proxy call" meanwhile.)

miketaylr:
Take a look at the `/rate_limit` endpoint, that's a pretty simple example of how to do it.

hallvors:
Ready to merge!

hallvors:
I'm waiting to see if GitHub is smart enough to auto-close all three issues mentioned in summary when @miketaylr merges this :)

miketaylr:
Awesome, thanks @hallvors.

PR #801, opened 2015-10-27
 Fixes #749. Add a "My Activity" page.

### LINKED ISSUE #749: Make a "My Activity" page
From #442.

This issue is more specific. We want a page that shows:

1) Bugs that @mention you
2) Bugs that you reported.

miketaylr:
And refactor a bunch of stuff to make it work.

Note there's more work to be done to make our components composiable/modular. This is a fest step towards making that happen with pagination. I'll likely re-do some of what I've done here with this Mixin pattern when we add sorting/per page dropdowns. But it works, so let's put it up for review.

r? @karlcow for Python/HTML/CSS bits
r? @hallvors for JS/HTML/CSS bits

And anybody who has feedback on visual design is welcome to provide that. I'll push to http://staging.webcompat.com/me


miketaylr:
Tests are failing, so it's possible I screwed up the merge conflicts. Let me push some fixup commits.

miketaylr:
Alright, tests are green. Fire ze (review) missles. :rocket:

magsout:
👍🎉💕

magsout:
@miketaylr an error on staging: 

capture d ecran 2015-10-27 a 19 43 25

miketaylr:
Hmm, thanks. Will look at it.

miketaylr:
(silly mistake)

config.py.example
```diff
@@ -101,3 +102,20 @@ else:
...
+# 'new', 'needsdiagnosis', 'needscontact', 'contactready' , 'sitewait', 'close'
+# Creating the model
+Category = namedtuple('Category', ['name', 'dataAttribute', 'label'])
+CATEGORIES = []
+cat_labels = [('new', 'new', 'New Issues'),
+              ('needsDiagnosis', 'needsdiagnosis', 'Needs Diagnosis'),
+              ('needsContact', 'needscontact', 'Needs Contact'),
+              ('ready', 'contactready', 'Ready for Outreach'),
+              ('sitewait', 'sitewait', 'Site Contacted'),
+              ('close', 'closed', 'Closed')]
```
hallvors: Why doesn't this code include invalid and duplicate? Just curious. Are these "all statuscategories we want" or "all statuscategories we show in the UI"?

config.py.example
```diff
@@ -101,3 +102,20 @@ else:
...
+# populating the categories for issues status
+for cat_label in cat_labels:
+    CATEGORIES.append(Category(name=cat_label[0],
+                               dataAttribute=cat_label[1],
+                               label=cat_label[2]))
```
hallvors: Also, is there any benefit to using namedtupe and the for loop instead of just defining an object directly?

    CATEGORIES = [ {"name":"new", "dataAttribute": "new", "label": "New Issues"}, ...]

webcompat/api/endpoints.py
```diff
@@ -91,6 +91,28 @@ def proxy_issues():
...
+    This is only used for "creator" and "mentioned" right now.
+
+    Any logged in user can see details for any other logged in user. We can
+    extend this to non-logged in users in the future if we want.
+    '''
+    if not g.user:
+        abort(401)
+    # copy the params so we can add to the dict.
+    params = request.args.copy()
+    params['state'] = 'all'
```
hallvors: Won't this override a query - if I go to /issues/hallvors?state=closed or something?

webcompat/static/js/lib/mixins/pagination.js
```diff
@@ -0,0 +1,162 @@
...
+    nextButton.removeClass('wc-hidden')
+              .prop('disabled', isLastPage())
+              .toggleClass('is-disabled', isLastPage());
+    prevButton.removeClass('wc-hidden')
+              .prop('disabled', isFirstPage())
+              .toggleClass('is-disabled', isFirstPage());
+
+    if (nextPage) {
+      // chop off leading "/api" and set @href
+      nextButton.attr('href', issuesCollection.getNextPage().slice(4));
```
hallvors: I'd prefer .substring rather than .slice to make it (more) obvious we're dealing with a string and not an array.. but maybe that's just me :-p

config.py.example
miketaylr: These map to our current "filter" buttons in the UI. We've discussed being able to track dupes in the future, but for now I don't think we want to add that yet.

config.py.example
miketaylr: A good question for @karlcow, I think (he proposed this code, or maybe it was @magsout ?). 

My intuitive sense is that the list of dicts is more readable, but I don't really have a strong preference:
```
CATEGORIES = [
     {"name":"new", "dataAttribute": "new", "label": "New Issues"},
     {"name":"new", "dataAttribute": "new", "label": "New Issues"},
     {"name":"new", "dataAttribute": "new", "label": "New Issues"}
]
```

webcompat/api/endpoints.py
miketaylr: Right now `/issues/hallvors` won't do anything, so it doesn't matter much I think. 

But if we want to allow this kind of fine-grained info in the future we'll need to update this. We're trying to keep the scope of this PR small and make it more interesting in the future. :)

webcompat/static/js/lib/mixins/pagination.js
miketaylr: Heh, could be just you. But I can change it. ^_^

config.py.example
miketaylr: Actually, I think I do prefer the explicit list of dicts to the building up of the list programmatically. Probably less pythonic, but just seems cleaner. @magsout, thoughts?

config.py.example
hallvors: So it's basically "UI categories". I suggest clarifying the comment slightly, it's not about "status categories used in the project", it's only about enumerating the labels we want to flag in the UI as particular bug "states".

karlcow: status of the bug != other type of labels. These are really status.
the `cat_labels` is probably not appropriate and should be `status_labels`.

Note also that this was initially designed to answer a particular need for @magsout in its templating use, with the idea of extending to other parts of the code in a second strike.

This can evolve, but we have to be careful, this is used by @magsout

config.py.example
karlcow: I don't have preferences apart of the possibility of evolving them. Right now the dict proposed is short. 
The namedtuple was chosen according to what magsout was trying to do in the template section. And calling the variable with a `Object.value` syntax.

karlcow: It's used here. https://github.com/webcompat/webcompat.com/blob/issues/749/2/webcompat/templates/web_modules/filter.html#L1

miketaylr: I say let's pull it in and evolve it later. :+1:

miketaylr:
(I just rebased against master and force pushed).

config.py.example
magsout: >I don't have preferences apart of the possibility of evolving them.

Same thing for me. I just need to browse the information. An object, an array .. Do what you want

miketaylr:
One last failure, related to the fixture file `labels.6665c4820456e6a813dcbbdddd690ffc.json` being missing. Will take a look tonight or tomorrow, my brain is too jet lagged right now :confused:

miketaylr:
Ah, now I get it. I rebased and master has the `?per_page=100` hack for the labels endpoint. Fixture files want a hashed encoding of GET params (to be able to serve different states). Let me see if renaming the file doesn't fix it.

miketaylr:
OK, it seems like everyone is on board with these changes, and tests are passing. I'm gonna go ahead and merge this now. We can clean up the few issues noted here when we iterate on the activity page and add in the other components as mixins.

PR #807, opened 2015-10-29
Update state when labels change, fixes #806

### LINKED ISSUE #806: "state" of open issue is not updated in the UI when new labels are set
1) Open some issue
2) Apply a label like "sitewait"

Bug: the UI indicating the state of the bug doesn't change to "Ready for Outreach". If we are a dynamic, client-side MVC app let's do it properly ;)

hallvors:


miketaylr:
:+1:

PR #808, opened 2015-10-29
bug status labels should be mutually exclusive, fixes #448

### LINKED ISSUE #448: Bug status label should be mutually exclusive
Currently we have 

1. needdiagnosis
2. (needcontact) See #434 
3. contactready
4. sitewait
5. fixed, duplicate, wontfix, worksforme, ?

It should not be possible to have these mutually selected. The proposal would be that when you select one, you unselect the other(s).

hallvors:


miketaylr:
Looks good, we should probably write tests for this so we don't break it in the future. But that can be in a follow up PR.

PR #810, opened 2015-10-29
Somewhat better Windows instructions, I think. Fixes #364

### LINKED ISSUE #364: Have better instructions for dev setup in a windows environment
> I had loads of trouble getting the environment set up on windows for local testing. >_<
mentioned in #363

hallvors:


CONTRIBUTING.md
```diff
@@ -216,24 +220,23 @@ We use `pip` to install other Python packages. You may need to install `pip` if
 To determine if you need to, type the following command into the terminal:
 
 ```bash
-type pip
```
miketaylr: (I assume this means `type` isn't available in Windows?)

miketaylr:
Looks good. For now I'll assume it works and make a point to try to get it running on my Windows machine (tomorrow, or in a few weeks) and report bugs on the docs if I run into issues.

CONTRIBUTING.md
hallvors: There is a ```type``` command on Windows, it does basically what ```less``` does on *nix I guess. So not what these instructions expect ;) Rather than writing two sections I figured out trying to call it with --version would be reasonable and cross-platform.

PR #812, opened 2015-10-29
Static labels - fixes #781, fixes #799

### LINKED ISSUE #781: [discussion] Config-level list of accepted labels?
Previous discussions:

https://github.com/webcompat/webcompat.com/issues/436#issuecomment-72563771

https://github.com/webcompat/webcompat.com/issues/773#issuecomment-149421222

Pros: 
- Simplicity. 
- We know the list in advance and don't need to make requests to GitHub. 
- Would solve some caching/empty response bugs (and probably introduce new ones :p)

Cons: 
- Requires process for adding new labels (might be a pro...). 
- We have to build it.

Discuss!

### LINKED ISSUE #799: The browser-firefox-mobile label is always removed when editing labels
See https://github.com/webcompat/webcompat.com/issues/784#issuecomment-151056209

hallvors:


webcompat/api/endpoints.py
```diff
@@ -264,14 +265,8 @@ def modify_labels(number):
...
-    request_headers = get_request_headers(g.request_headers)
-    if g.user:
-        path = 'repos/{0}/labels'.format(REPO_PATH)
-        labels = github.raw_request('GET', path, headers=request_headers)
-    else:
-        path = 'https://api.github.com/repos/{0}/labels'.format(REPO_PATH)
-        labels = proxy_request('get', uri=path, headers=request_headers)
-    return (labels.content, labels.status_code, get_headers(labels))
+    with open( STATIC_PATH + '/config_data/labels.json') as f:
+        return (f.read(), 200, {'content-type': 'application/json'})
```
miketaylr: We have a constant `JSON_MIME` in this file that you can use instead of the `application/json` string.

webcompat/static/config_data/labels.json
```diff
@@ -0,0 +1,182 @@
+[
+  {
+    "url": "https://api.github.com/repos/webcompat/web-bugs/labels/status-duplicate",
+    "name": "status-duplicate",
+    "color": "cccccc"
```
miketaylr: I think we probably don't want to be in the business of serving hand-written JSON. 

What about building this file programatically when the app starts from something like labels-config.py. We could hash the file to see if it changed that way we don't have to re-build the file unless we know there are new labels.

All we need in labels-config.py is a label name and a color.

hallvors: Well, it's not exactly hand-written - just downloaded from GitHub :-p

Anyway. To build this file on startup, I should add code to webcompat/__init__.py - right? Or run.py?

miketaylr: > Well, it's not exactly hand-written - just downloaded from GitHub :-p

Heh, yeah. But in the future we'll edit it and I'll leave out a quote mark and the app will explode. :bomb: 

Another option is instead of building the file, we *just* download it from GitHub from time to time? Maybe every 24 hours? (labels don't get added that frequently...)

> To build this file on startup, I should add code to webcompat/init.py

Yeah. We've also got some code for doing checksums @ https://github.com/webcompat/webcompat.com/blob/master/webcompat/helpers.py#L58-L68 if you want to go down the conditional builds route.

miketaylr: > Another option is instead of building the file, we just download it from GitHub from time to time? Maybe every 24 hours? (labels don't get added that frequently...)

Or do something cool like set up a webhook handler for a "new label was created" event (should such a thing exist).

miketaylr:
Gonna close this, see proposed way forward in https://github.com/webcompat/webcompat.com/issues/781#issuecomment-165495273.

PR #820, opened 2015-11-05
Fixes #819. Update ua-parser to 0.5.0.

### LINKED ISSUE #819: Tablet/Android identification issues
So it seems we had a couple of bugs with `Firefox 45` on `Android 6.0` which are not identified as such.

```
<!-- @browser: Android 6.0 -->
<!-- @ua_header: Mozilla/5.0 (Android 6.0; Tablet; rv:45.0) Gecko/45.0 Firefox/45.0 -->
```

We might to fix that. https://github.com/webcompat/web-bugs/issues/1885

miketaylr:
This will fix the broken detection of Firefox tablets as Android browser. However, they will be identified as Firefox Mobile.

r? @karlcow

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,6 @@ Flask-Uploads==0.1.3
 Flask-WTF==0.12
 GitHub-Flask==0.3.4
 WTForms==2.0.2
-ua-parser==0.3.5
+ua-parser==0.5.0
 nose==1.3.1
 blinker==1.3


karlcow:
With 💖. Thanks.

miketaylr:
For the history books, this updated fine locally -- but totally broke production for about 10 minutes. :bomb: 

I assume version 0.51 will fix this: https://github.com/ua-parser/uap-python/issues/34

I don't have the error in front of me (I didn't think to copy it down), but installation was failing because it was looking for the regexes.yaml inside a uap-core directory that wasn't there (it was inside ua_parser, iirc).

These are the commands I used to fix it, from bash history:

```
1992  git clone git@github.com:ua-parser/uap-core.git
 1993  ls
 1994  mv uap-core/ webcompat.com/build/ua-parser/
 1995  sudo mv uap-core/ webcompat.com/build/ua-parser/
 1996  ls
 1997  cd webcompat.com/
 1998  sl
 1999  ls
 2000  sudo pip install ua-parser
```

miketaylr:
(lol @ `sl`)

Issue #818, opened 2015-11-05
Add new iOS meta for correct viewport on iOS9


magsout:
https://forums.developer.apple.com/thread/13510

magsout:
![Uploading image.png…]()

PR #823, opened 2015-11-05
Fixed overflow title in small device

magsout:
Relative to #818 . It's was a mistake by me, not an error about this bug https://forums.developer.apple.com/thread/13510 but just an overflow with long title.

r? @miketaylr 

simulator screen shot 5 nov 2015 12 46 33


miketaylr:
Thanks for the fix @magsout. I'm about to fly back to the US from Taiwan, so I probably won't deploy until Monday. :heart:

PR #842, opened 2015-12-01
Print a more helpful error if there is no config.py file. Fixes #841

### LINKED ISSUE #841: Be more informative if config.py is missing
If the user has forgotten to create a local config.py, the run.py script detects an ImportError and helpfully suggests to run ```pip install -r requirements.txt``` again. That's pretty helpful but still not exactly the most helpful advice you could give..since it won't actually fix the problem ;)

hallvors:
r? @karlpro

### DIFF
--- a/run.py
+++ b/run.py
@@ -18,10 +18,24 @@ Please re-run:
 ==============================================
 '''
 
+NO_CONF_ERROR = '''
+==============================================
+The config.py file seems to be missing. 
+
+Please create a copy of config.py.example and customize it accordingly.
+For details, please see 
+https://github.com/webcompat/webcompat.com/blob/master/CONTRIBUTING.md#configuring-the-server
+==============================================
+'''
+
 try:
     from webcompat import app
 except ImportError, e:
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -18,10 +18,25 @@
...
+'''
+
 try:
     from webcompat import app
 except ImportError, e:
-    raise ImportError('{0}\n\n{1}'.format(e, IMPORT_ERROR))
+    if e.import_name == 'config':
+        # config not found, somebody forgot to make a copy of config.py.example and modify?
+        print(NO_CONF_ERROR)
+        quit()
```
karlcow: We probably should keep the consistency in the way we raise the message. Or was there a reason to use print/quit?

```python
raise ImportError('{0}\n\n{1}'.format(e, NO_CONF_ERROR))
```

karlcow:
Also would be good to have the issue number in the commit message. #841

miketaylr:
@hallvors can you squash/fixup these 2 commits into 1? After that, LGTM to merge.

hallvors:
Happy now, @miketaylr ?

miketaylr:
LGTM. :+1:

Issue #847, opened 2015-12-10
Add parameters parsing for cssfixme


karlcow:
`/cssfixme.php?url=http%3A%2F%2Fhallvord.com%2Ftemp%2Fmoz%2FcompatTesterTesting%2Fstyle.css`

should be able to inject the CSS stylesheets inside the textarea.

Be careful with security issues.

Depends on #12

magsout:
it is possible to explain this issue? :heart:

miketaylr:
@magsout we're moving http://hallvord.com/temp/moz/cssfixme.php to webcompat.com/tools/cssfixme (it's a tool to add standards/unprefixed CSS to -webkit-prefix only code).

One of the features is that you can give it a URL via a get param, and it will automatically fetch that and put it in the tool for you -- that's what this bug is for.

hallvors:
Basically, we want this URL to work:
https://webcompat.com/tools/cssfixme/?url=http%3A%2F%2Fhallvord.com%2Ftemp%2Fmoz%2FcompatTesterTesting%2Fstyle.css

karlcow:
Let's not deploy this for now. There are some security/usability issues with the code.  I found new ones.

karlcow:
* http://localhost:5000/tools/cssfixme?url=http://localhost gives the home page of the server you are on if any or a 504 Gateway issue for nginx
* http://localhost:5000/tools/cssfixme?url=http://127.0.0.1 gives the home page of the server you are on if any or a 504 Gateway issue for nginx
* http://localhost:5000/tools/cssfixme?url=http://www.mozilla.org injects the full HTML, we should check the content-type of the return document  is `text/css` before injecting it
* http://localhost:5000/tools/cssfixme?url=www sends for me the home page of my router, because it seems the domain is defined locally. fun stuff.
* suggest to send in the template instead of an empty area an error message saying that the URL is invalid.

miketaylr:
> Let's not deploy this for now. There are some security/usability issues with the code. I found new ones.

It's too late, we already deployed. I'm going to disable it on master now. @karlcow, please file new issues.

miketaylr:
Actually, no need for new issues @karlcow. This is the right place.

miketaylr:
OK, fix deployed. I'm not even sure this feature is worth the trouble, given its ability to take down our server if we don't do it correctly:

```
[...]

PR #851, opened 2015-12-10
Fixes #843. Update selector for failing GitHub test.

miketaylr:
r? @karlcow

Issue #850, opened 2015-12-10
Add actual scripting to make cssfixme working


karlcow:
Depends on #12

hallvors:
Three simple steps: we need to put the two js files in https://github.com/webcompat/css-fixme/tree/master/js in a suitable location (/static/js/lib/cssfixme/ .. was proposed) and add corresponding SCRIPT tags to the page.

Finally, we need a small script to hook up the UI and the rest of the JS. (We can remove the inline onclick handler from the button in the template).

The body of the onclick handler on the button should be somewhat like this:

    var csscode = document.getElementsByTagName('textarea')[0].value; // or give it an ID..
    var pre = document.getElementById('fixedcss');
    var compact = document.getElementsByName('compact')[0].checked;
    doTheBigStyleFixing(csscode, pre, compact);

I also recommend adding something like this:

```
document.getElementsByTagName('pre')[0].addEventListener('dblclick', function(e) {
    window.getSelection().removeAllRanges();
    var rng = document.createRange();
    rng.selectNodeContents(document.getElementsByTagName('pre')[0]);
    window.getSelection().addRange(rng);
    e.preventDefault();
}, false);
```

hallvors:
If we use css-fixme as a git submodule, we presumably get all the junk in the repo including the .php file. What about making the build process just fetch those two files and stick them in the right place when we build webcompat.com ?

hallvors:
For example using https://github.com/mCzolko/grunt-downloadfile

miketaylr:
Seems OK as long as we're not trying to download the files every single grunt build (same problem as image optimization). That might get annoying if you're offline.

Another option is to move the files from the submodule during the build process (if they don't exist, or are newer than what we have).

hallvors:
That sounds like a good idea.

I almost forgot: the "hooking it all up" JS should also include an onload handler that checks if the TEXTAREA contains code and if so triggers the style fixup. This is mostly for when the server-side processing adds CSS code inside the TEXTAREA.

PR #852, opened 2015-12-10
Handle ?url query string for css-fixme, #847

hallvors:


miketaylr:
Looks great! Any feedback, @karlcow?

webcompat/views.py
```diff
@@ -5,6 +5,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import json
+import requests
```
karlcow: this should come after, separate from the standard library. It is an imported module.

webcompat/views.py
```diff
@@ -274,7 +275,16 @@ def contributors():
 @app.route('/tools/cssfixme')
 def cssfixme():
     '''Route for CSS Fix me tool'''
-    return render_template('cssfixme.html')
+
+    url = request.args.get('url')
```
karlcow: Missing test suite, for testing the URLs.
For example, this is crashing the site
http://localhost:5000/tools/cssfixme?url=blah
because blah is not an url. See below.

webcompat/views.py
```diff
@@ -274,7 +275,16 @@ def contributors():
...
+    escaped_csscode = ''
+    if url:
+        req = requests.get(url)
```
karlcow: Probably is not that much if url than probably a `try:` and if the request fails act accordingly.

karlcow: The cool thing is that the message in the console is quite interesting.

```
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/requests/models.py", line 338, in prepare_url
    "Perhaps you meant http://{0}?".format(url))
MissingSchema: Invalid URL u'blah': No schema supplied. Perhaps you meant http://blah?
```

MissingSchema.

karlcow: Another crash

```
http://localhost:5000/tools/cssfixme?url=%3Ciframe%20href=%27http://hallvord.com/temp/moz/compatTesterTesting/style.css%27%3E
```
This time

```
requests.exceptions.InvalidSchema

InvalidSchema: No connection adapters were found for '<iframe href='http://hallvord.com/temp/moz/compatTesterTesting/style.css'>'
```

webcompat/views.py
karlcow: You can add tests in https://github.com/webcompat/webcompat.com/blob/master/tests/test_urls.py

hallvors:
Should be OK to merge, only failure on Travis is the one about wrong selector in Github test which is fixed in #851

miketaylr:
> This branch has conflicts that must be resolved

@hallvors can you rebase and fix conflicts before we pull this in, please?

miketaylr:
> The problem is that TEXTAREA will render those entities in the content - literally, so escaping everything will mangle any CSS code that contains <>&. Hence div > p would show up as div &gt; p in the generated output.

Is that true? Every browser I tested does the "right" thing here, and if I copy and paste I get unescaped entities (which is what I would expect).

`data:text/html,<textarea>&lt;script&gt;alert(0)&lt;/script&gt;</textarea>`:




I'd feel a lot better if we just escape all instances of `<`, `>` and `&` -- XSS paranoia seems like a good policy.

miketaylr:
(I guess this PR is out of sync with #850, which was just merged -- rebasing against master should help here. These patches still have the inline click handlers, etc.)

miketaylr:
https://www.owasp.org/index.php/XSS_%28Cross_Site_Scripting%29_Prevention_Cheat_Sheet#RULE_.231_-_HTML_Escape_Before_Inserting_Untrusted_Data_into_HTML_Element_Content recommends escaping `'` and `"` as well.

hallvors:
> Is that true? Every browser I tested does the "right" thing here

Strange. I'm sure I saw the opposite while testing in Orlando - I'm even sure @karlcow was sitting next to me and saw it too. Had I somehow added a double escaping? :-o

hallvors:
We're using the Flask templating engine and it's escaping input by default (unless you jump through some hoops to avoid that by passing in Markup class objects or something). So yes, we ended up with a double escape which confused me.

hallvors:
@miketaylr should be fine to merge now - not sure what the two failures on Travis are about but they are definitely not related to this patch..

miketaylr:
> We're using the Flask templating engine and it's escaping input by default (unless you jump through some hoops to avoid that by passing in Markup class objects or something). So yes, we ended up with a double escape which confused me.

Ah yeah, that seems confusing, thanks for looking into it. :+1: 

> not sure what the two failures on Travis are about but they are definitely not related to this patch

Looks like the PR failure was a pep8 failure. But I see you've fixed it so the push build passed. Let me just retrigger the PR build then we should be good to merge.

miketaylr:
(oops, was looking at wrong build. yes, unrelated failures indeed)

PR #853, opened 2015-12-10
Cssfixme scripts, fixes #850

### LINKED ISSUE #850: Add actual scripting to make cssfixme working

hallvors:


webcompat/static/js/lib/css-fixme-ui.js
```diff
@@ -0,0 +1,26 @@
...
+    doTheBigStyleFixing(csscode, pre, compact);
+  }
+
+  document.getElementById('btn_do_fixup').addEventListener('click', do_fixup, false);
+
+  if (document.getElementsByTagName('textarea')[0].value) {
+    do_fixup();
+  }
+
+  document.getElementById('fixedcss')[0].addEventListener('dblclick', function(e) {
```
miketaylr: Is this a typo? `document.getElementById('fixedcss')[0]` will be undefined, I think.

webcompat/static/js/lib/css-fixme-ui.js
```diff
@@ -0,0 +1,26 @@
...
+    window.getSelection().removeAllRanges();
+    var rng = document.createRange();
+    rng.selectNodeContents(document.getElementById('fixedcss')[0]);
```
miketaylr: Same here, right?

miketaylr:
Can we add a section to CONTRIBUTING.md in the dev env setup about how to pull down the submodule?

I guess it doesn't need to be much more than: `git submodule init && git submodule update`

webcompat/static/js/lib/css-fixme-ui.js
hallvors: d'oh

hallvors:
(At least some of these test failures were about stuff already fixed elsewhere, I did a rebase to latest master so we will see if that fixes things..

webcompat/templates/cssfixme.html
```diff
@@ -9,10 +9,11 @@
       <p>Paste CSS in this box to see what needs fixing:</p>
 
       <textarea style="width:100%;height:20rem;"></textarea><br/>
-      <button type="button" onclick="doTheBigStyleFixing(document.getElementsByTagName('textarea')[0].value)">Add standard equivalents for prefixed CSS</button>
+      <button type="button" id="btn_do_fixup">Add standard equivalents for prefixed CSS</button>
       <label>
         <input type="checkbox" name="compact">Compact output (whitespace only after closing brace)</label>
       <pre id="fixedcss"></pre>
+      <script src="/js/cssfixme.min.js"></script>
```
miketaylr: Traditionally we've done the following for scripts that have build processes:

```
{%- block extrascripts -%}
{%- if config.PRODUCTION or config.DEVELOPMENT -%}
<script src="{{ url_for('static', filename='js/diagnose.min.js')|bust_cache }}"></script>
{%- else -%}
<script src="{{ url_for('static', filename='js/lib/models/label-list.js') }}"></script>
<script src="{{ url_for('static', filename='js/lib/labels.js') }}"></script>
<script src="{{ url_for('static', filename='js/lib/models/issue.js') }}"></script>
<script src="{{ url_for('static', filename='js/lib/diagnose.js') }}"></script>
{%- endif -%}
{%- endblock %}
```

This way we're doing cache busting and we can more easily debug in dev environments. Any reason to not do something similar here?

miketaylr: (Sorry, I should have noticed this earlier).

hallvors:
I made a change to use the non-minified version of css-fixme.js but it's probably not quite consistent with the other stuff since this file is still built by combining three others during a build process. The reason for that is the fact two of those files comes from a Git submodule and thus live outside of the tree originally - they are copied into the tree *by* the build process. I'm not sure if we can do something to link this up without a build step..

miketaylr:
> I made a change to use the non-minified version of css-fixme.js but it's probably not quite consistent with the other stuff since this file is still built by combining three others during a build process

I think that's a good compromise. Thanks.

PR #859, opened 2015-12-12
Add a new QueryParams model

hallvors:
This is just introducing a new QueryParams model. Nothing in this commit actually *uses* the new model yet, so if you want to test it you can only do so from the console:

    var qp = new issueList.QueryParams();
    qp.fromQueryString(location.search);
    qp.setParam('label', 'firefox');
    qp.setParam('stage', 'contactready');
    qp.toBackendURL();

etc..

(BTW: you first have to add a SCRIPT element to the DOM that actually loads the new code..

    document.body.appendChild(document.createElement('script')).src='/js/lib/models/query-params.js'

or something like that..)

### DIFF
--- a/webcompat/static/js/lib/models/query-params.js
+++ b/webcompat/static/js/lib/models/query-params.js
@@ -139,7 +139,12 @@ issueList.QueryParams = Backbone.Model.extend({
     var issuesAPICategories = ['closed'].concat(this.bugstatuses);
 
     // Rules for when to use GitHub directly and when Webcompat
-    if (this.get('q') || searchAPICategories.indexOf(this.get('stage')) > -1) { // We have a query that needs search API
+    var q = this.get('q');
+    // TODO: can be simplified if we bring back AppliedLabels view, this next line should go
+    // we know that the issue API can handle queries with labels in, we don't want a labels:foo
+    // in the search field to force the search API.
+    q = q.replace(/labels:[^ ]+/g, '');
+    if (q || searchAPICategories.indexOf(this.get('stage')) > -1) { // We have a query that needs search API
       if ('withCredentials' in XMLHttpRequest.prototype && !loggedIn) { // CORS support, not logged in - talk directly to GH
         url = this.configUrls._githubSearch + '?' + this.toSearchAPIParams();
       } else {
--- a/webcompat/static/js/lib/models/query-params.js
+++ b/webcompat/static/js/lib/models/query-params.js
@@ -1,3 +1,4 @@
+/*globals issues*/
 var issueList = issueList || {};
 
 /*
@@ -50,8 +51,8 @@ issueList.QueryParams = Backbone.Model.extend({
     _githubSearch: 'https://api.github.com/search/issues'
   },
   bugstatuses: ['contactready', 'needsdiagnosis', 'needscontact', 'sitewait'],
-  initialize: function(){
-    this.on('change', function(e){
+  initialize: function() {
+    this.on('change', function(e) {
       // When the query/parameters change, we want to
       // * update the drop-down menu if required
       // * update the URL and push a history entry
@@ -60,17 +61,17 @@ issueList.QueryParams = Backbone.Model.extend({
       // However, first we want to check if the change is significant.
       var significant = false;
       var changelist = Object.keys(e.changed);
-      for(var i=0, change; change = changelist[i]; i++) {
+      for (var i = 0, change; change = changelist[i]; i++) {
         // I think we only get notified of one property at a time so looping is
         // maybe not necessary here.
         var newvalue = e.changed[change];
         var oldvalue = e.previousAttributes()[change];
-        if(newvalue.toString().trim() === oldvalue.toString().trim()) {
+        if (newvalue.toString().trim() === oldvalue.toString().trim()) {
           continue; // just whitespace change, let's ignore this
         }
         // If a user clicks a label we're already filtering by, we can likewise ignore it
-        if(change === 'labels') {
-          if (this.get('labels').indexOf(newvalue)>-1) {
+        if (change === 'labels') {
+          if (this.get('labels').indexOf(newvalue) > -1) {
             continue;
           }
         }
@@ -95,16 +96,16 @@ issueList.QueryParams = Backbone.Model.extend({
   /*
   * This creates a query suitable for the URL bar from the current params
   */
-  toDisplayURLQuery: function(){
+  toDisplayURLQuery: function() {
     var urlParams = [];
-    _.each(this.attributes, function(value, key){
-      if(value) {
+    _.each(this.attributes, function(value, key) {
+      if (value) {
         urlParams.push(key + '=' + value);
       }
     });
     return urlParams.join('&');
   },
-  toBackendURL: function(loggedIn){
+  toBackendURL: function(loggedIn) {
     /*
     * This method returns either a GitHub API URL or a webcompat.com/api URL
     * depending on the login status and what type of query we run.
@@ -148,7 +149,7 @@ issueList.QueryParams = Backbone.Model.extend({
       } else {
         url = '/api/issues/search' + '?' + this.toSearchAPIParams();
       }
-    }else { // Seems like this query is so simple we only need the issues API.
+    } else { // Seems like this query is so simple we only need the issues API.
       if (_.contains(issuesAPICategories, this.get('stage'))) {
         url = '/api/issues/category/' + this.get('stage') + '?' + this.toIssueAPIParams();
       } else {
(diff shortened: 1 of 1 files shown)


hallvors:
Trying to figure out those failures.. this is weird :( Adding [not ready] to title for now.

hallvors:
Travis changed his mind when build was restarted. Weird stuff I saw locally was just a caching issue. OK to merge.

miketaylr:
Tests are passing :+1:, will review once I get through e-mail backlog.

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+*
+* The query itself can contain name:value parts like label:state-needsdiagnosis
+* but we have no control over paging and number of results when querying
+* GitHub directly.
+*/
+issueList.QueryParams = Backbone.Model.extend({
+  defaults:{
+    stage:'all',
+    state: 'open',
+    page:1,
```
miketaylr: nit: can we have consistent whitespace here? I suggest a single space after the `:` -- I suppose we should try to enforce this via eslint.

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+    per_page:50,
+    sort:'created',
+    direction: 'desc',
+    q: '',
+    creator: '',
+    mentioned: '',
+    labels: [] // Using 'labels' plural form would be nice, but needs to be
```
miketaylr: This comment confuses me, it says using 'labels' would be nice but we are using labels here?

webcompat/static/js/lib/models/query-params.js
miketaylr: Bah, I just realized our `grunt eslint` command isn't reaching all the JS we want. Ignore this, we can land some CLEANUP commands in #860.

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+              // singular for the backend - or we'd have to translate name on send
+  },
+  configUrls: {
+    _githubSearch: 'https://api.github.com/search/issues'
+  },
+  bugstatuses: ['contactready', 'needsdiagnosis', 'needscontact', 'sitewait'],
+  initialize: function(){
+    var self = this;
```
miketaylr: Doesn't look like `self` is used in `initialize`.

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+    this.on('change', function(e){
+      // When the query/parameters change, we want to
+      // * update the drop-down menu if required
+      // * update the URL and push a history entry
+      // * update the relevant views (list of issues, search box)
+      // To do so, this method will fire various events.
+      // However, first we want to check if the change is significant..
```
miketaylr: nit: double period here.

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+      var significant = false;
+      var changelist = Object.keys(e.changed);
+      for(var i=0, change; change = changelist[i]; i++) {
+        // I think we only get notified of one property at a time so looping is
+        // maybe not necessary here. However..
```
miketaylr: nit: Double period here. Supposed to be ellipsis?

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+        var newvalue = e.changed[change];
+        var oldvalue = e._previousAttributes[change];
```
miketaylr: I guess `e.previousAttributes` is an undocumented internal API? Fingers crossed it doesn't disappear :p

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+    // we know that the issue API can handle queries with labels in, we don't want a labels:foo
+    // in the search field to force the search API.
+    q = q.replace(/labels:[^ ]+/g, '');
+    if (q || searchAPICategories.indexOf(this.get('stage')) > -1) { // We have a query that needs search API
+      if ('withCredentials' in XMLHttpRequest.prototype && !loggedIn) { // CORS support, not logged in - talk directly to GH
+        url = this.configUrls._githubSearch + '?' + this.toSearchAPIParams();
+      } else {
+        url = '/api/issues/search' + '?' + this.toSearchAPIParams();
+      }
+    }else { // Seems like this query is so simple we only need the issues API..
```
miketaylr: nit: double period.

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+    // Some names are different for the search API..
+    if (this.get('direction')) {
+      paramsToSend.order = this.get('direction');
+    };
+    // These properties must end up as ?q=name:value with name slightly translated
+    var qMap = {
+      creator:   'author',
+      mentioned: 'mentions'
+    };
+    for(key in qMap) {
```
miketaylr: Does this leak a global `key`?

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+  deleteLabel: function(labelStr){
+    var currentLabels = this.get('labels');
+    if(currentLabels.indexOf(labelStr) > -1) {
+      currentLabels.splice(currentLabels.indexOf(labelStr), 1);
+      this.set('labels', {}, {silent:true}); // hack: to trigger change event on *next* set..
+      // hack explained: currentLabels is a reference to an array, so updates cause immediate change.
+      // However, when we're manipulating the array directly, the model does not fire any change events
+      // because JS gives Backbone no way to observe changes to an array. If we set label to a reference
+      // to the array it already references, it won't fire change events either, even if array content
+      // was updated in the meantime. Hence the workaround sets it to {} and then back to the array.
```
miketaylr: Weird!

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+      this.setParam('labels', currentLabels);
+    }
+  },
+  fromQueryString: function(str, silent){
+    var self = this, namevalues, pair, name, value;
```
miketaylr: I think eslint will yell at you for this, might be good to run `eslint path/to/this/file.js` to get a jump on those.

webcompat/static/js/lib/models/query-params.js
```diff
@@ -0,0 +1,360 @@
...
+      name = pair[0];
+      value = pair[1];
+      // just in case..
+      if(pair[0] === 'label') {
+        pair[0] = 'labels';
+      }
+      self.setParam(pair[0], pair[1]);
+    });
+  },
+  // silentReset is intended used when a request to the backend fails
```
miketaylr: Are we failing lots of requests? I've seen a few comments like this, but it would be good to understand the failure conditions (and file bugs).

miketaylr:
I say let's pull this in once you address the comments -- I admit I don't have the full picture of how this works in my head yet. Which means you volunteer to fix most of the bugs for `/issues` until I do. :p

miketaylr:
Once #860 lands, I suggest rebasing and re-running grunt to catch the eslint errors.

webcompat/static/js/lib/models/query-params.js
hallvors: This is mainly included because I noticed such bugs very well when at Mozlando :)

webcompat/static/js/lib/models/query-params.js
hallvors: Seems like there should be a better way to do this..

webcompat/static/js/lib/models/query-params.js
miketaylr: I just started noticing random 500s on the home page like so:

`127.0.0.1 - - [15/Dec/2015 14:52:50] "GET /api/issues/category/new HTTP/1.1" 500 -`

So it seems there some more bugs to uncover. :beetle:

miketaylr:
@hallvors want to fix the eslint errors then we can pull this in?

hallvors:
eslint saved me some real debugging there :+1: 

Anyway, the "push" build still has this failure:

> FAIL: main - Search (non-auth) - Search works (3588ms)
> AssertionError: The search results show up on the page.: expected 'Issue 280: w3.org - desktop site instead of mobile site' to include 'vladvlad'

But I don't get why.. adding some code that's not even used anywhere yet shouldn't cause failures, right? What's the "push" build anyway?

miketaylr:
> What's the "push" build anyway?

Travis does two builds: push (when you push) and pr (if you make a pr). Your pr build passed, but the push one failed. I've seen that particular test fail in the past, so there's likely a bug somewhere. It seems to be somewhat random... so that's fun. :bomb:

miketaylr:
Just retriggered the build (you can do that on Travis is you suspect weird results) and it's green now. Let's pull this one in.

Issue #870, opened 2015-12-15
Intermittent 500 for /api/issues/category/new


miketaylr:
STR:

Fire up local server
Visit localhost:5000/

```
127.0.0.1 - - [15/Dec/2015 14:52:50] "GET /api/issues/category/new HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/helpers.py", line 360, in wrapped_func
    return func(*args, **kwargs)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/api/endpoints.py", line 136, in get_issue_category
    'content': filter_new(json.loads(issues[0])),
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
[...]

PR #871, opened 2015-12-15
Issue #870. Return 304 responses before trying to run filter_new.

miketaylr:
r? @karlcow 

Thinking of how I can write a test for this to prevent breaking in the future.


karlcow:
The test is maybe. if '304', then body is empty.

miketaylr:
> The test is maybe. if '304', then body is empty.

I can't think of a way to trick the 304 response from and API endpoint though, unless we add a special condition to @mockable_response to return a 304 (maybe if it's in test mode and sees a ?response=304 param or something)?

Issue #879, opened 2015-12-17
React to add-on opening new issue with screenshot


miketaylr:
cf. #452.

I'm thinking we'll have the add-on do a postMessage w/ the screenshot image as an argument to the home page. Once it receives that, we construct an img element and put it on the page, and then we'll put it in the `FormData` object that we POST to the server. :rainbow: :racehorse:

miketaylr:
Possibly need to add base64 image support in the image uploading backend. Not sure right now.

miketaylr:
OK, with #867 done, this should be pretty simple. :tada:

miketaylr:
![](https://cldup.com/c8gAJ1JHeU.gif)

Looks cool!

magsout:
Oh yeaah 💯

hallvors:
Excellent!

karlcow:
nicely done! :cake:

Issue #717, opened 2015-09-30
Set up service to delete logs older than 14 days


karlcow:
Once we start to collect we need to rotate and delete logs with a cron on the server side

miketaylr:
We still want to do this, every 2 weeks to begin with.

miketaylr:
ping @karlcow

karlcow:
Requirements:

* `/usr/sbin/logrotate`
* `cron` working, probably in `/etc/crontab` for the handlers
* a directory of your choice on the machine for the logs only accessible to admins. calling it `/tmp/secretlogs/` for the purpose of the demo.
* The file for the logs should be only accessible to admins let's call it `webcompat.log`

Do we have already there this?

```
/etc/cron.daily/logrotate
```

This means that every day logrotate is started and we can create a logrotate conf file for it.

in `/etc/logrotate.conf`, there might be already other configurations, so you can add at the end.

```
/tmp/secretlogs/webcompat.log {
    rotate 14
    daily
    copytruncate
    create 700 webc_user webc_group
    dateext
    compress
}
```

You can test with 
`logrotate -s /var/log/logstatus logrotate.conf`

This should run daily and keep only the last 14 files, compress them, and put a date in the filename.

More options at http://www.linuxcommand.org/man_pages/logrotate8.html

karlcow:
r? @miketaylr 
If you prefer me to do it on the server itself. 
I just need to remember how I setup my access there. :) no issue.

miketaylr:
> Do we have already there this?
> 
> /etc/cron.daily/logrotate

yep! so it looks like we're all set to rotate logs, we just need to log stuff (which is what https://github.com/webcompat/webcompat.com/pull/1125 is for).

I'm happy to set this up on the server -- thanks!

miketaylr:
> I'm happy to set this up on the server -- thanks!

Yeah right.

miketaylr:
OK. This is set up. Thanks again for the guidance @karlcow. 🍰

Issue #710, opened 2015-09-28
Optimizing images when uploading to the server


karlcow:
We may want to optimize an image when it's being uploaded:
* reduce the size of the image (performance concerns)
* remove EXIF data from the image (privacy concerns)

Discussed at Paris WebCompat meeting on September 2015 with @miketaylr 
See https://github.com/webcompat/webcompat.com/blob/master/webcompat/api/uploads.py#L53

karlcow:
We will be probably using pillow for dealing with images.
http://python-pillow.github.io/

karlcow:
```
Many of Pillow’s features require external libraries:

    libjpeg provides JPEG functionality.
        Pillow has been tested with libjpeg versions 6b, 8, and 9 and libjpeg-turbo version 8.
        Starting with Pillow 3.0.0, libjpeg is required by default, but may be disabled with the --disable-jpeg flag.
    zlib provides access to compressed PNGs
        Starting with Pillow 3.0.0, zlib is required by default, but may be disabled with the --disable-zlib flag.
```

karlcow:
```(webcompatcom)17:57:31 ~/code/webcompat.com
→ brew install libjpeg webp
Warning: You are using OS X 10.11.
We do not provide support for this pre-release version.
You may encounter build failures or other breakage.
==> Downloading https://homebrew.bintray.com/bottles/jpeg-8d.el_capitan.bottle.2.tar.gz
######################################################################## 100,0%
==> Pouring jpeg-8d.el_capitan.bottle.2.tar.gz
🍺  /usr/local/Cellar/jpeg/8d: 18 files, 760K
==> Installing dependencies for webp: libpng
==> Installing webp dependency: libpng
==> Downloading https://homebrew.bintray.com/bottles/libpng-1.6.18.el_capitan.bottle.tar.gz
######################################################################## 100,0%
==> Pouring libpng-1.6.18.el_capitan.bottle.tar.gz
🍺  /usr/local/Cellar/libpng/1.6.18: 17 files, 1,2M
==> Installing webp
==> Downloading https://homebrew.bintray.com/bottles/webp-0.4.3.el_capitan.bottle.tar.gz
######################################################################## 100,0%
==> Pouring webp-0.4.3.el_capitan.bottle.tar.gz
🍺  /usr/local/Cellar/webp/0.4.3: 32 files, 1,7M
(webcompatcom)17:59:51 ~/code/webcompat.com
```

[...]

PR #883, opened 2015-12-22
Fixes #882 - Add our own upload backend, remove Flask-Uploads

### LINKED ISSUE #882: Remove Flask-Upload and implement our own upload backend
Splitting #879 into two bugs here to make it easier to review.

The new backend should be able to handle regular file uploads, as well as incoming screenshots from the add-ons, do image format validation, and save them to the filesystem.

miketaylr:
So, in order to manually verify the screenshots are working, you'll need to install some WIP add-ons. The easiest is probably Opera or Chrome: https://github.com/webcompat/webcompat-reporter-extensions/tree/master/opera (go to extensions, enable developer mode, point to that directory).

You also need to cherry-pick a WIP commit `git fetch origin && git cherry-pick e275f75f8a86f891215260434aaaf7d5872e1e8f`.

It's admittedly a mess, so feel free to ignore that entirely and we can verify screenshots when #879 is up for review.

"regular" file uploads should continue to work in the same way.

I haven't had time yet to figure out the testing story for the screenshot stuff, so that will come in follow up commits or issues.

r? @karlcow


webcompat/api/uploads.py
```diff
@@ -42,21 +99,23 @@ def upload():
...
-            data = {
-                'filename': filename,
-                'url': images.url(filename)
-            }
-            return (json.dumps(data), 201, {'content-type': JSON_MIME})
-        except UploadNotAllowed:
-            abort(415)
+    if 'image' in request.files and request.files['image'].filename:
+        imagedata = request.files['image']
+    elif 'screenshot' in request.form:
```
miketaylr: `screenshot` comes from an add-on `postMessage` https://github.com/webcompat/webcompat.com/blob/e275f75f8a86f891215260434aaaf7d5872e1e8f/webcompat/static/js/lib/bugform.js#L89-L117, if you're curious.

tests/test_uploads.py
```diff
@@ -22,6 +21,47 @@
 from webcompat import app
 
 
+# taken from https://github.com/srusskih/flask-uploads/blob/master/flaskext/uploads.py#L476 # nopep8
+class TestingFileStorage(FileStorage):
```
miketaylr: This change isn't very interesting, it's just copying and pasting from Flask-Upload so we can ditch the dependency and have our tests still working.

run.py
```diff
@@ -31,8 +31,8 @@
 try:
     from webcompat import app
 except ImportError, e:
-    if e.import_name == 'config':
-        # config not found, somebody forgot to make a copy of config.py.example and modify?
+    if 'import_name' in e and e.import_name == 'config':
+        # config not found, did you forget to copy config.py.example?
```
karlcow: nitpick: I was about to fix this in another commit. Here the comment says what the dev should do (which is the error message), instead of documenting the code. ^_^ I would put instead. 

```python
# Catching the missing config.py file
```
The `NO_CONF_ERROR` explains what the dev has to do.

webcompat/views.py
```diff
@@ -245,7 +245,7 @@ def download_file(filename):
         Python app.
         '''
         return send_from_directory(
-            app.config['UPLOADS_DEFAULT_DEST'] + '/uploads', filename)
+            app.config['UPLOADS_DEFAULT_DEST'], filename)
```
miketaylr: `/uploads` was added to `[UPLOADS_DEFAULT_DEST]`, because that's specific to how someone wants to set up their system.

run.py
miketaylr: :+1:

tests/test_uploads.py
karlcow: nothing forbids you to put the comment under the class. :)

webcompat/api/uploads.py
```diff
@@ -26,11 +23,71 @@
...
+ALLOWED_FORMATS = ('jpg', 'jpeg', 'jpe', 'png', 'gif', 'bmp')
 
-# limit image uploads to 4MB
-patch_request_class(app, 4 * 1024 * 1024)
+
+class UploadNotAllowed(Exception):
+    """
+    This exception is raised when we try to save a file that is not
+    allowed. Allowed files are defined in ALLOWED_FORMATS
+    """
```
karlcow: I guess I can fix it later. but I would rather have something slightly different here. Probably something derived from TypeError. But your code will work anyway here. Just a reminder for me.

webcompat/api/uploads.py
```diff
@@ -26,11 +23,71 @@
...
+        '''Method to return the file extension.'''
+        if self.is_filestorage:
+            return imagedata.filename.split('.')[-1].lower()
+        if self.is_base_64:
+            # We only expect base64 png screenshot uploads,
+            # so we do a double check to verify that, and raise otherwise
+            # data:image/png;base64, <base64 string>
+            if imagedata.split('/')[1].split(';')[0] != 'png':
+                raise UploadNotAllowed()
+            return 'png'
```
karlcow: here I don't understand. Why do we allow only png. Will not it get into the way of attach screenshots in the comment and someone uploading a jpeg.

karlcow: Also Pillow provides specific way to test the right file format instead of the extension which is a ok test for validity but not a real one. 

```python
>>> from PIL import Image
>>> im = Image.open('pacifique-fuji.jpg')
>>> im.format
'JPEG'
>>> 
```

Also when the file is not correct, we get an error.

``` python
>>> im2 = Image.open('foobar.pdf')
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/PIL/Image.py", line 2286, in open
    % (filename if filename else fp))
IOError: cannot identify image file 'foobar.pdf'
```

webcompat/api/uploads.py
```diff
@@ -26,11 +23,71 @@
...
+        return UploadNotAllowed()
+
+    def save(self):
+        '''Check that the file is allowed, then save to filesystem.'''
+        if self.file_ext not in ALLOWED_FORMATS:
+            raise UploadNotAllowed()
+
```
karlcow: See above.

karlcow:
ok some preliminary comments done. I haven't installed it yet locally. This should take a bit of time to review. Hopefully this afternoon.

webcompat/views.py
karlcow: did you modify `config.py.example`? I don't see in this pull request?

miketaylr: Oops, will do, thanks.

miketaylr:
Thanks Karl, great feedback. I'm traveling tomorrow, but maybe will work on addressing comments in the evening.

miketaylr:
> I haven't installed it yet locally.

One more comment, if you do install the Opera add-on (or any of them), you'll need to change `var prefix` from https://github.com/webcompat/webcompat-reporter-extensions/blob/master/opera/background.js#L5 to:

`var prefix = 'http://localhost:5000/?open=1&url=';`

Then once you have it installed, you can click the new lightbulb logo button in your browser and it'll open the local version w/ a placeholder screenshot below the form (still waiting on UI). Submitting that form should result in a new issues w/ the screenshot embedded just like an upload.

karlcow:
ok let's see.

```
→ git checkout -b miketest upstream/issues/882/1
Branch miketest set up to track remote branch issues/882/1 from upstream.
Switched to a new branch 'miketest'
```

ok after pointing Opera Desktop to the extension I was able to make the extension inside the chrome. The issue is that the screenshot doesn't happen at all. I guess it's working only on mobile?

OK let's see with a normal upload.

```
→ find uploads -ls
48462835        0 drwxr-xr-x    3 karl             staff                 102 22 déc 16:25 uploads
48462836        0 drwxr-xr-x    3 karl             staff                 102 22 déc 16:25 uploads/2015
48462837        0 drwxr-xr-x    3 karl             staff                 102 22 déc 16:25 uploads/2015/12
48462838        8 -rw-r--r--    1 karl             staff                1309 22 déc 16:25 uploads/2015/12/e267f789-8010-401b-931e-29147dcbb268.png
```

so far so good. 

With your current code, renamed a PDF file as `foo.png`
https://github.com/webcompat/webcompat-tests/issues/314

No errors. The log 
```
127.0.0.1 - - [22/Dec/2015 16:27:02] "GET /?open=1&url=https%3A%2F%2Fwww.mozilla.org%2Ffr%2F HTTP/1.1" 200 -
127.0.0.1 - - [22/Dec/2015 16:27:08] "GET /api/issues/labels?per_page=100 HTTP/1.1" 200 -
127.0.0.1 - - [22/Dec/2015 16:27:08] "GET /api/issues/category/new HTTP/1.1" 200 -
127.0.0.1 - - [22/Dec/2015 16:27:19] "POST /issues/new HTTP/1.1" 302 -
127.0.0.1 - - [22/Dec/2015 16:27:19] "GET /thanks/314 HTTP/1.1" 200 -
```

The upload 

```
48462835        0 drwxr-xr-x    3 karl             staff                 102 22 déc 16:25 uploads
48462836        0 drwxr-xr-x    3 karl             staff                 102 22 déc 16:25 uploads/2015
48462837        0 drwxr-xr-x    4 karl             staff                 136 22 déc 16:27 uploads/2015/12
48462914      200 -rw-r--r--    1 karl             staff               99045 22 déc 16:27 uploads/2015/12/7e952d67-4413-4c1d-851e-eefbdf325d6a.png
48462838        8 -rw-r--r--    1 karl             staff                1309 22 déc 16:25 uploads/2015/12/e267f789-8010-401b-931e-29147dcbb268.png
```

But I think the previous code had the same caveat. So it just shows that it is "working" the same as before. ^_^ for this part at least so that's good. We can start to improve from here.

webcompat/api/uploads.py
```diff
@@ -26,11 +23,71 @@
...
+                           imagedata.startswith('data:image/'))
+        self.is_filestorage = isinstance(imagedata, FileStorage)
+        self.file_ext = self.get_file_ext(imagedata)
+
+    def get_file_ext(self, imagedata):
+        '''Method to return the file extension.'''
+        if self.is_filestorage:
+            return imagedata.filename.split('.')[-1].lower()
+        if self.is_base_64:
+            # We only expect base64 png screenshot uploads,
```
karlcow: ah interesting.

```
127.0.0.1 - - [22/Dec/2015 16:40:15] "GET /?open=1&url=https%3A%2F%2Fwww.mozilla.org%2Ffr%2F HTTP/1.1" 200 -
127.0.0.1 - - [22/Dec/2015 16:40:17] "GET /api/issues/labels?per_page=100 HTTP/1.1" 200 -
127.0.0.1 - - [22/Dec/2015 16:40:18] "GET /api/issues/category/new HTTP/1.1" 200 -
self.is_base_64    False
self.is_filestorage True
self.file_ext           jpg
127.0.0.1 - - [22/Dec/2015 16:40:39] "POST /issues/new HTTP/1.1" 302 -
127.0.0.1 - - [22/Dec/2015 16:40:39] "GET /thanks/316 HTTP/1.1" 200 -
```

The raise `UploadNotAllowed` is never raised when `self.is_base_64  =  False` in `def get_file_ext(self, imagedata):`.

karlcow:
ok I start to have a better understanding of the code. Either we receive from `image` or from `screenshot`, in both cases we get `imagedata`. That could be nice to log this for seeing if we have a lot of screenshot vs image.

Then the code is going to the initialization of `Upload` under a `try:` statement.

The `init` computes

* save a local imagedata
* check if it's is_base_64
* check if it's is_filestorage
* extract the extension
  * is_filestorage based on the filename
  * is_base_64 based on the string

I guess we could replace with PIL. easily, after the 

``` python 
from PIL import Image
try:
    im = Image.open(blah)            
except IOError:
    # Not a valid format
    abort(415)
try:
    file_ext = im.format.lower()
    if file_ext not in ALLOWED_FORMATS:
        #not an allowed format
        raise TypeError("Image file format not allowed")
```

and you could return `file_ext`. That way we can test for anyone and it would reduce your code, also it would removed the need of a specific Exception.

miketaylr:
> The issue is that the screenshot doesn't happen at all.

Did you do this step? 

>  You also need to cherry-pick a WIP commit git fetch origin && git cherry-pick e275f75f8a86f891215260434aaaf7d5872e1e8f.

This pulls in a commit from another branch that puts the screenshot into the UI (but in a very crude fashion). Sorry, this is a huge pain in the butt with so many moving pieces. But thanks for reviewing so far!

miketaylr:
> But I think the previous code had the same caveat.

Heh, yeah. I'm maintaining bug compatibility here :see_no_evil: 

> That could be nice to log this for seeing if we have a lot of screenshot vs image.

That's a good point. I guess once we have logging in place (depending on #717), let's hook that up. I'll file a follow-up issue.

> I guess we could replace with PIL.

Yeah, this is cool. And then I guess it would make it very simple for you to extend this to the image optimization work in #710.

webcompat/api/uploads.py
miketaylr: > Why do we allow only png. Will not it get into the way of the feature "attach screenshots" in the comment and someone uploading a jpeg

The base64 image from the add-on is only a PNG (for now, we could extend it in the future to allow the user to choose the image encoder, if that seems useful). This codepath is different than the "regular" image upload.

miketaylr: Pillow can give us the extension from a decoded base64 image, so that's what I moved it to -- no more hard-coding. \o/

miketaylr:
I'm going to rebase fabc25e out before merging, so ignore that -- it just makes it simpler to test.  747e77b has the changes from Karl's feedback thus far.

@karlcow, what do you think of 747e77b? I decided to convert everything to a PIL Image object so we can use its `save` method, rather than special-case it for base64 stuff.

karlcow:
## About testing screenshot

> This pulls in a commit from another branch that puts the screenshot into the UI (but in a very crude fashion). Sorry, this is a huge pain in the butt with so many moving pieces. But thanks for reviewing so far!

I think I had done. But it's ok, because it doesn't really change the rest of the code. given that after screenshot and image it's basically the same code.

miketaylr:
> A couple of things possible.

Yeah, good point. I don't know why/how someone would be sending a base64 JPEG, but let's take the safer/more flexible option.

Thanks for your review @karlcow-sama :pray: :clap: !

miketaylr:
Also TODO: Add PIL to requirements.txt and write any docs about getting that set-up (I had to `brew install` a handful of things, etc.)

karlcow:
@miketaylr About 

>Also TODO: Add PIL to requirements.txt and write any docs about getting that set-up (I had to brew install a handful of things, etc.)

This might help you.
https://github.com/webcompat/webcompat.com/issues/710#issuecomment-147659048

miketaylr:
Hm, those first two commits shouldn't have been there. Looking.

miketaylr:
> This might help you.

Thanks Karl. I did the lazy thing and just linked to the Pillow deps install docs.

miketaylr:
> Hm, those first two commits shouldn't have been there. Looking.

Rebased against master, so they went away.

miketaylr:
OK, I've added tests both for "regular" file uploads and screenshot uploads, and fixed a bug that the tests revealed. If Travis is green, I'm gonna merge.

PR #889, opened 2015-12-29
Fixes #786. Create a config module to hopefully break less things.

### LINKED ISSUE #786: improvement about config.py
I don't know if it's me, but each time `config.py` is updated (and of course I don't see it) I spend many times to understand the errors in my terminal. 

Plus, with the @karlcow 's idea about export some information like the list of `Categories` (which is the great idea,)  can may complicate maintenance of this file because it does not track by git. I wonder if it would not be better to have two files, maybe something like: `config_app.py`  and `config_access.py`

- `config_access.py` has only information about github acces (`call_back_url`, `client_secret` etc) is not tracked by git
- config_app.py is exactly same as `config.py` but withtout variable (`callback_url`, `secret key` etc..) but import `config_access.py` to have access for `callback_url, client_secret` etc) is tracked by git. We can add information without break the config and the website.

It's just a thought. What do you think ?

miketaylr:
Let's see how this does on Travis before requesting review.


miketaylr:
```
File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/flask_wtf/csrf.py", line 43, in generate_csrf

    raise Exception('Must provide secret_key to use csrf.')

Exception: Must provide secret_key to use csrf.
```

Looks like the empty string for `SECRET_KEY` in secrets.py.example breaks stuff (before it was "meow"). Will update after dinner.

CONTRIBUTING.md
```diff
@@ -282,24 +282,24 @@ grunt
...
 # Windows
-copy config.py.example config.py
+copy config/secrets.py.example config/secrets.py
 ```
 
-You can now edit `config.py` and
+You can now edit `secrets.py` and
 
-1. Add the right values to the [repo issues URIs](https://github.com/webcompat/webcompat.com/blob/master/config.py.example#L68-L73). `ISSUES_REPO_URI = "<user>/<repo>/issues"`. For example, miketaylr's setup needs to say `ISSUES_REPO_URI = "miketaylr/test-repo/issues"`
+1. Add the right values to the repo issues URIs. `ISSUES_REPO_URI = "<user>/<repo>/issues"`. For example, miketaylr's setup needs to say `ISSUES_REPO_URI = "miketaylr/test-repo/issues"`
```
miketaylr: Once we land this, I'll come back and add the hyperlinks.

config/env_constants.py
```diff
@@ -0,0 +1,17 @@
...
+
+import os
+
+# Define the application directory
+BASE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))
+
+# the PRODUCTION and DEVELOPMENT environment variables are set in uwsgi.conf
+# on the webcompat server. If they're not set, they will default to None
+PRODUCTION = os.environ.get('PRODUCTION')
+DEVELOPMENT = os.environ.get('DEVELOPMENT')
```
miketaylr: I kind of want to rename `DEVELOPMENT` to `STAGING`, but I'd rather not mess with the uwsgi.conf on the server. How weird would it be to have `STAGING = os.environ.get('DEVELOPMENT')`? Maybe something for a "good-first-patch" bug.

config/env_constants.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+
+# Define the application directory
+BASE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))
```
miketaylr: One notable difference here is the change to BASE_DIR to keep it the same, otherwise we'd end up with `session.db`, etc. in `/config/`.

config/env_constants.py
karlcow: no issue for me.

config/env_constants.py
karlcow: Is it not giving the same result?

```python
>>> os.path.abspath(os.pardir)
```

miketaylr: Oh, probably. >_<

config/env_constants.py
miketaylr: Cool, let's do that in the PR as well.

miketaylr:
Look OK now, @karlcow?

karlcow:
ok let's see 

```
→ git checkout -b deleteme upstream/issues/786/1
→ git status
On branch deleteme
Your branch is up-to-date with 'upstream/issues/786/1'.
nothing to commit, working directory clean
→ mv config.py ~/Desktop/
→ cp config/secrets.py.example config/secrets.py
→ python run.py

The OAUTH_TOKEN is not configured in your config file.
You will need to set up one on github for testing your
local developments.
Read Instructions at
https://github.com/webcompat/webcompat.com/blob/master/CONTRIBUTING.md#configuring-the-server
```

so far so good

```
→ subl config/secrets.py
# adding OAUTH token
→ python run.py
 * Running on http://localhost:5000/ (Press CTRL+C to quit)
 * Restarting with stat
```

joly good.

```
→ nosetests
..F...............................
FAIL: API access to labels without auth returns JSON 200.
  File "/Users/karl/code/webcompat.com/tests/test_api_urls.py", line 71, in test_api_labels_without_auth
    self.assertEqual(rv.status_code, 200)
AssertionError: 404 != 200
-------------------- >> begin captured logging << --------------------
requests.packages.urllib3.connectionpool: INFO: Starting new HTTPS connection (1): api.github.com
requests.packages.urllib3.connectionpool: DEBUG: "GET /repos//labels HTTP/1.1" 404 None
--------------------- >> end captured logging << ---------------------

Ran 34 tests in 4.129s

FAILED (failures=1)
```

Weird.
Also checking the files.

Rha my bad. Apologies. I tested both. You want your initial solution.

``` python
BASE_DIR = os.path.abspath(os.pardir)
# secrets /Users/karl/code
# db /Users/karl/code

BASE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))
# secrets /Users/karl/code/webcompat.com
# db /Users/karl/code/webcompat.com
```

Be sure to check that the BASE_DIR plus issues.db  is really the path you want.

Some other nitpicking comments: see in the code

config/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+'''This module handles application configuration and secrets.'''
+
+import os
+from datetime import datetime
+from collections import namedtuple
+
+from environment import *
+from secrets import *
```
karlcow: don't like global import like these, but don't have a better solution. So let's fly with it. ;)

config/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+'''This module handles application configuration and secrets.'''
+
+import os
+from datetime import datetime
+from collections import namedtuple
```
karlcow:
```
import os
from datetime import datetime
```

can be removed

karlcow:
r+ with you initial solution.

config/__init__.py
miketaylr: If you think of a better solution at some point, send a PR. :cookie:

miketaylr:
> Be sure to check that the BASE_DIR plus issues.db is really the path you want.

For now, yea. If there's a better canonical place to stick smallish dbs like this, let's move it there in another issue.

config/__init__.py
miketaylr: Updated, thanks (btw which linter are you using to show you this? pep8 doesn't seem to care--would be nice to catch this stuff before committing it)

miketaylr:
(I don't get that weird nosetest failure, but let's see if Travis does and investigate more if that's the case).

miketaylr:
OK, let's pull the trigger.

Remember next time you need to start the server you probably need to copy over the secrets from `config.py` to `config/secrets.py.example`, and then rename that to `config/secrets.py`. File bugs as you find 'em. 

(I'm going to deploy this to staging before production, just to be safe™).

magsout:
:+1: nice work..

PR #887, opened 2015-12-29
Issue #848 - Add style to cssfixme service

magsout:
## Before:

capture d ecran 2015-12-29 a 17 49 00

## After:

capture d ecran 2015-12-29 a 17 40 19

We can improve it,it's a lick of paint

r? @karlcow @miketaylr @calexity @hallvors @tagawa

Issue #848, opened 2015-12-10
Add style to cssfixme service. 


karlcow:
We're including a tool to fix CSS code, which will go live on https://webcompat.com/tools/cssfixme . The tool was developed in its own repo, which is https://github.com/webcompat/css-fixme/ .

This tool needs some more style/design work. Can the styling be more integrated with the rest of this site? 
This task might also involve changing/moving the text and explanations on the page, if you as a designer would like to do so.

Depends on #12
See also #847

hallvors:
An important part here is the equivalent of this:

    pre strong{
        background: yellow;
        color:black;
    }

to highlight the style rules the too adds to the output.

karlcow:
It needs a bit of reorg of the text, and better styling. 

What it currently looks like as of today.
![capture d ecran 2015-12-18 a 12 10 40](https://cloud.githubusercontent.com/assets/505230/11888413/7871347a-a580-11e5-885d-91bfc7df2eaf.png)

magsout:
let's add some webcompat style on this tools

miketaylr:
:+1:

magsout:
@hallvors why `strong` element in your selector ? If i inspect DOM, I can't see any `<strong>` element

magsout:
Let's close. Open new issue if we need something more..

PR #891, opened 2015-12-30
Issue #848 - improve layout CSSFixme

magsout:
Added some improvement:
 - https://github.com/webcompat/webcompat.com/pull/887#issuecomment-167939419
 - https://github.com/webcompat/webcompat.com/pull/887#commitcomment-15193558

capture d ecran 2015-12-30 a 12 51 19

r? @karlcow @hallvors


hallvors:
@magsout what does code that has been modified look like? See http://hallvord.com/temp/moz/cssfixme.php?url=http%3A%2F%2Fhallvord.com%2Ftemp%2Fmoz%2FcompatTesterTesting%2Fstyle.css for an example.

hallvors:
Very minor language point: I wonder if "CSS code" will be easier to understand than "CSS instructions"..?

magsout:
@hallvors aaaaahh, ok, :-1: my bad... So, I return to the previous style for the `<pre>`, and I only highlights the strong elements

magsout:
capture d ecran 2015-12-30 a 14 24 07
capture d ecran 2015-12-30 a 14 24 13


@hallvors @karlcow @miketaylr Which one do you prefer ?

karlcow:
@magsout The 1st one (at the top). Many thanks. Super travail!

magsout:
@hallvors is it possible to detect if there is content in the `<pre>` element? If so, you could add a class `is-active` because by default, when there is no content we still see the style of the <pre> because of padding (which I can delete it, but it's less pretty)

hallvors:
We can certainly set an is-active style from the JS when inserting the generated code. Or just generate the PRE element too on the fly..? You can just add a DIV id=result-parent or something to put the generated PRE inside, I think that may be cleaner than toggling PRE style based on a class :)

hallvors:
(I also vote for the 1st one - much of the point is to show very clearly what we're changing and it's more obvious that way)

magsout:
>Or just generate the PRE element too on the fly..? You can just add a DIV id=result-parent or something to put the generated PRE inside, I think that may be cleaner than toggling PRE style based on a class :)

@hallvors looks good :+1:

webcompat/templates/cssfixme.html
```diff
@@ -20,7 +28,13 @@
           <button type="button" class="Button Button--default" id="btn_do_fixup">Add standard equivalents for prefixed CSS</button>
         </div>
         <div class="wc-CSSFixme-item">
-          <pre class="wc-CSSFixme-resultat" id="fixedcss"></pre>
+          <div id="js-CSSFixme"></div>
```
magsout: @hallvors I removed `<pre>`  element and add this parent. So you can add <pre> directly in this div. if you can add a class : `class="wc-CSSFixme-resultat"` would be awesome ;)

magsout:
Ok, updated

magsout:
build passed

miketaylr:
Same, top style looks great.

hallvors:
I sort of liked the way the code in the second screenshot stands out with the yellow background though.. I'm testing the JS changes now, when the background is plain it is too much a part of the page, somehow. What about adding for example 

```
.wc-CSSFixme-resultat{
  border-left: 8px solid var(--wc-background-dark);
  padding-left: 5px;
}
```
To make it stand out a bit more?
/me dabbling in designy stuff - feel free to ignore suggestions that betray my lack of taste :p

hallvors:
@magsout I hope you're happy with me pushing directly here..?

magsout:
@hallvors I added the changes to the style. What makes good

capture d ecran 2015-12-30 a 21 26 31

>@magsout I hope you're happy with me pushing directly here..?

it's fine to like that .. :+1: 

I'm reading your email and try to understand the issue

miketaylr:
> It's only a global because the script for some reason is included twice in the page

I'm about to push a fix for that. I don't understand 100%, but the `extrascripts` template block was included *inside* the `body` template block, and I guess it looped over itself... twice? I don't know. But moving extrascripts outside of body (like all the other templates) fixes it. Pushing fix now.

magsout:
👍 @miketaylr

magsout:
👍 @miketaylr

miketaylr:
:+1: :+1: ^_^

magsout:
oups .. ;)

magsout:
Awesome.

@hallvors or @karlcow if everything looks good;)

hallvors:
Great, now we can undo the global-ness of that variable. If I do that quickly enough perhaps I can make it before Mike merges ;)

hallvors:
and @magsout :+1: to the extra styling - stands out now

miketaylr:
Heh, I'll let you pull the merge trigger @hallvors -- and I can deploy this afternoon as well.

miketaylr:
Great work @magsout and @hallvors!

PR #892, opened 2015-12-30
Fixes #878. Return 401 Unauthorized for non-logged in users at /me

### LINKED ISSUE #878: /me doesn't redirect to login then /activity/username for logged-out users
cf. #877 

Code is here: https://github.com/webcompat/webcompat.com/blob/master/webcompat/views.py#L209-L215, for whatever reason it's not prompting the user to login.

miketaylr:
Rather than redirecting for a login, send a 401.

r? @karlcow


miketaylr:
`NameError: global name 'abort' is not defined` oops

miketaylr:
Will fix failing test when #889 lands.

karlcow:
r+ @miketaylr

Issue #867, opened 2015-12-15
Create UI for including screenshot in bug report


miketaylr:
Issue #452 tracks adding the ability to grab screenshots of a problematic tab, via browser add-ons (and built into Fennec, eventually). I hope to make good progress the implementation details of that this week (knock on :deciduous_tree:).

We need a UI to possible show a small preview of the screenshot and an unchecked-by-default checkbox that the user can check if they want to include it. This is to prevent us from automatically submitting images that might contain sensitive information (or maybe NSFW stuff or whatever).

Here's one take from @calexity: https://github.com/webcompat/webcompat.com/issues/431#issuecomment-68215120

miketaylr:
I want to ship a minimal/kinda-crappy-but-not-super-terrible version of this so we can have the feature before we have the perfect UI. I think a minimum requirement is that we show a preview of the image, so the user can see what's about to be sent. 

One thing to keep in mind is this UI will only be shown if a user sends a postMessage'd base64 image from the "https://webcompat.com" origin -- which means they have our new add-on versions installed. Otherwise, the UI stays the same.

Mobile seems like a pretty easy leap from @calexity's design. Desktop is trickier...

Some ideas:

@magsout @karlcow @calexity  (+anybody else) thoughts? In the interest of shipping v1 sooner than later, minimal tweaks (or alternative mockups welcome!) would be awesome. Anything more complex would be OK for v2 wish-list (like a lightbox w/ highlight/blackout capability).

![](https://cldup.com/ne823xTcbU-3000x3000.png)
![](https://cldup.com/gqXTd99NDl-3000x3000.png)

miketaylr:
Also cc @Raphseck, if you have any ideas. It might be helpful to go to "https://webcoompat.com/?open=1" to see what this currently looks like (basically the same w/o the inline screenshot and the "give more details" bit extends the entire width of the viewport for the desktop version".

miketaylr:
I'm going to take silence as consent here, and probably start building this out tomorrow. We can iterate on better designs if we ever come up with them. :p

Raphseck:
@miketaylr and guys!
Sorry about the late and sorry about my english!
I like this layout and had some ideas about the screenshot part.
Maybe some more intuitive here.
Hope you like!
![file_1](https://cloud.githubusercontent.com/assets/14264676/12027189/cb316ad2-ada9-11e5-88ab-32173fc35725.jpg)
[...]

PR #893, opened 2015-12-30
Issue #867 - UI for including screenshot in bug report

magsout:
Ok, so @miketaylr 

capture d ecran 2015-12-30 a 21 26 31

### DIFF
--- a/webcompat/templates/form.html
+++ b/webcompat/templates/form.html
@@ -44,7 +44,7 @@
         </div>
       </div>
       <div class="r-Grid-cell r-Grid-cell--alignBottom r-all--1of2 r-maxS--2of2">
-        <div class="wc-Form-group  {% if form.image.errors %}wc-Form-error{% endif %}">
+        <div class="js-image-upload wc-Form-group {% if form.image.errors %}wc-Form-error{% endif %}">
           <label class="wc-Form-upload" for="image">
             <div class="wc-Form-upload-wrapper">
               <div class="wc-Form-upload-icon" aria-hidden="true">
--- a/webcompat/templates/form.html
+++ b/webcompat/templates/form.html
@@ -58,7 +58,7 @@
               {% endif %}
               {{ form.image(class_='ButtonUpload', accept='.jpe,.jpg,.jpeg,.gif,.png,.bmp') }}
             </div>
-            <button class="wc-Form-upload-button is-hidden r-ResetButton" role="button">Remove image upload</button>
+            <button class="wc-Form-upload-button is-hidden r-ResetButton" role="button" type="button">Remove image upload</button>
           </label>
         </div>
       </div>
--- a/webcompat/templates/form.html
+++ b/webcompat/templates/form.html
@@ -44,7 +44,7 @@
         </div>
       </div>
       <div class="r-Grid-cell r-Grid-cell--alignBottom r-all--1of2 r-maxS--2of2">
-        <div class="js-image-upload wc-Form-group {% if form.image.errors %}wc-Form-error{% endif %}">
+        <div class="js-image-upload wc-Form-group wc-pos-relative {% if form.image.errors %}wc-Form-error{% endif %}">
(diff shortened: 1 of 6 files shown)


webcompat/templates/form.html
```diff
@@ -37,19 +37,28 @@
...
-          {% endfor %}{% endif %}
-          {{ form.image(class_='ButtonUpload', accept='.jpe,.jpg,.jpeg,.gif,.png,.bmp') }}
+      <div class="r-Grid-cell r-Grid-cell--alignBottom r-all--1of2 r-maxS--2of2">
+        <div class="wc-Form-group  {% if form.image.errors %}wc-Form-error{% endif %}">
+          <label class="wc-Form-upload" for="image">
+            <div class="wc-Form-upload-wrapper">
+              <div class="wc-Form-upload-icon" aria-hidden="true">
+                <span class="wc-Icon wc-Icon--cloud-upload"></span>
+              </div>
+              {{ form.image.label(class_='wc-Form-label wc-Form-label--upload') }}
```
magsout: If you can remove the attribute `for`  for this `<label>`

miketaylr: I'll see if I can figure that out. ^_^

miketaylr:
Whoa, speedy. :zap:  Will play with this in a bit.

webcompat/templates/form.html
miketaylr: We might be able to just hand-write the `<label>` and get rid of the python part here (I think). Probably simplest.

miketaylr:
This looks awesome @magsout. I'm gonna add an image preview via `FileReader` (if the browser supports it), otherwise it's hard to know if you've selected something. Once that's in place, I think we can land this and I'll extend it to work with the add-on screenshot stuff (and add drag & drop, etc.)

miketaylr:
(that push was just to rebase against master)

miketaylr:
We can improve the UX in future issues, a couple ideas:

1) ability to clear out file selection, right now you have to just select a new image
2) remove/modify "Add screenshot image" + icon
3) other stuff...

miketaylr:
(it can look bad depending on the image, but that's OK I think for now... incremental improvements, etc.):

magsout:
@miketaylr we can improve UX before merge if you want. Not a big deal.

miketaylr:
@magsout if you have any suggestions, feel free to hack on them. I don't really have any amazing ideas, hence my "let's do it later". :p

When you get a chance, can you review the JS part?

miketaylr:
I think I have one idea, let me try to come up with a quick mockup.

miketaylr:
Here's one idea, stealing from the original design:

Once we have the upload "read" and the background-image set, we hide the label w/ icon and show a yellow bar on the bottom with a link "Click here to remove image" (I guess it would technically be a button, but I think styling as a link would be good). Once the user clicks that, we remove the bg-image, clear out the file input, and show the original "Upload screenshot image">

We could re-use that same bottom bar for the add-on use case where a user needs to check the checkbox (something like "[ ] Select to include screenshot image")

screen shot 2015-12-31 at 3 33 53 pm

What do you think?

magsout:
wooot @miketaylr awesome :+1:

magsout:
I added a small animation and replaced `<div>` by `<button>`.

I have one issue when I removed image uploaded:

capture d ecran 2016-01-05 a 12 35 06

miketaylr:
Yay, thanks for cleaning up my markup/CSS ^_^. Let me try to reproduce that error, it's kinda wierd.

miketaylr:
Ah, the weird view function error is because it's a button -- it's trying to submit or navigate or something. Will sprinkle some JS to fix it, then merge.

miketaylr:
> Will sprinkle some JS to fix it

Or just use regular HTML `type=button`.

magsout:
Ah yes, I writed `role` instead of type. Good catch @miketaylr

PR #909, opened 2016-02-02
[WIP] Webpack

magsout:
OK, I started to add webpack instead of grunt. For now, I prepared this POC with some npm plugin (webpack babel, cssnext-loader etc..) 
- image-loader: to optimise media like jpeg, png (only if media are cosummed by background directy in the css), for media use with element  we can use a npm runtime with imagmin directly.
- svg-loader: to optimise svg 
- cssnext-loader: for now I use this plugin but in the futur I have to removed it and use postcss-cssnext with postcss-loader
- babel-loader: I installed only babel core, for require directive. If we need more ES6 code it's simple to add es6-preset (import, let, const etc...)
- json-loader: consumme json for script like markdon-it

For now my webpack script is simple, but we need more something like that https://gist.github.com/magsout/707406e3773092fbadce 

We need now to thinking about how to add  our JavaScript codebase to webpack. I follow this article http://stackoverflow.com/questions/28969861/managing-jquery-plugin-dependency-in-webpack to handle jQuery as a plugin dependency in webpack. 

Now I see two way for our script

1) we add all plugin with import directive, not sure that it is so easy to particular causes of dependencies like jQuery and Backbone
2) we try to make our codebase more "commonjs" with ES6 module. Class it's not necessary, fonctionnel javscript is better I guess.

Remember it's just a POC (it's take me 5 minutes to do it). Why use webpack ? Because I think webpack is better than grunt and gulp. Logical loader is interesting for JS of course, but for asset too, like CSS, svg (I'll replace all icon-font by svg near), img etc... it's fast and modern. We can  started to make our codebase more modular. I know it's less important than features for webcompat. But the project grow up every day so it's important to prepare our codebase to the future..

What do you think @hallvors @miketaylr @karlcow ?


miketaylr:
In principal I'm in favor of modernizing our workflow -- will take a closer look at some point today.

miketaylr:
> 2) we try to make our codebase more "commonjs" with ES6 module. Class it's not necessary, fonctionnel javscript is better I guess.

This would be cool. Not sure how long it would take, but I'd like to eventually move in this direction.

magsout:
Super useful http://stackoverflow.com/questions/28969861/managing-jquery-plugin-dependency-in-webpack

magsout:
Closing for now. @miketaylr I have some issue with some scipt that I don't understand right now.

zoepage:
@magsout I'd love to reopen the issue and work on upgrading our workflow (making it simpler).
Should I re-open or start a new issue?

magsout:
@zoepage as you want. If you think this PR is a good start for you.

 We can work together if you want. Maybe we can work witth webpack 2.

zoepage:
@magsout perfect, thank you! Yes, a collaboration sounds like a very good idea. I'd love to fix the contribute.md first and document everything before we go on with refactoring. So there won't be any surprises ;)

magsout:
@zoepage

>I'd love to fix the contribute.md first and document everything before we go on with refactoring.

sounds good

miketaylr:
I think we should probably close this PR, and open a new issue (we probably want to be using webpack 2 anyways), and link back to this. If a PR is going to sit for a number of weeks (or 1year+), it's probably not ready for review. ^__^

@zoepage can you do that?

zoepage:
@miketaylr yes. New issue in #1352 and closing this one.

PR #919, opened 2016-02-08
Fixes #912. Remove FB spamblocker measures.

### LINKED ISSUE #912: Remove Facebook anti-spammer measures
It's been a while since #688 was in place. Let's remove that.

miketaylr:
Let's see what happens.

r? @karlcow


karlcow:
r+ @miketaylr Let's see :volcano:

miketaylr:
> Maybe in the future we should put that behind a feature flag (if it comes back)

Ah yes, good idea. Would be fun to sketch that out an API for that with you at some point.

PR #928, opened 2016-03-01
Fixes #880. Allow visiting /activity/username route directly.

### LINKED ISSUE #880: Error when log in at /activity/magsout and no data provide
Not logged, go to https://webcompat.com/activity/magsout, log in.  There is a flash message about `API CALL NOT FOUND`   Label in dropdown is wrong (avatar user)

capture d ecran 2015-12-19 a 09 09 10

Plus, I cant's see my activity, but If I go to Homepage et back to the previous url, it's ok.

miketaylr:
r? @karlcow 

(sorry for taking so long here @magsout)

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -209,7 +209,24 @@ def me_redirect():
 
 @app.route('/activity/<username>')
 def show_user_page(username):
-    return render_template('user-activity.html', user=username)
+    '''The logic for this route is as follows:
+
+    (this dupes some of the functionality of /me, but allows directly visiting
+    this endpoint via a bookmark)
+
+    If the user is not logged in, send back a 401.
+    Make sure we have username and avatar details from Github
+    If the username matches, render the template as expected.
+    If it doesn't match, abort with 403 until we support looking at
+    *other* users activity.
+    '''
+    if not g.user:
+        abort(401)
+    get_user_info()
+    if username == session['username']:
+        return render_template('user-activity.html', user=username)
+    else:
+        abort(403)
 
 
 @app.route('/rate_limit')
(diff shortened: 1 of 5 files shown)


miketaylr:
Let's not merge yet, I should write a test or two for this.

karlcow:
r+ @miketaylr 

not to self. Need to refactor this errorhandler think. We are reusing the same code over and over again. 
Probably an opportunity for a function around this code:

```python
    if (request.path.startswith('/api/') and
       request.accept_mimetypes.accept_json and
       not request.accept_mimetypes.accept_html):
        message = {
            'status': 400,
            'message': 'API call. ' + message,
        }
        resp = jsonify(message)
        resp.status_code = 400
```

More something like (but to think not completely satisfied with the code below)

```python
ERROR_DICT = {
    400: 'Bad Request.',
    401: 'Unauthorized. Please log in.',
    403: 'Forbidden. Maybe that looking at private stuff?',
    404: 'Not Found. Lost in Punk Cat Space',
    }


@app.errorhandler(400)
@app.errorhandler(401)
@app.errorhandler(403)
@app.errorhandler(404)
def custom_error_handler(err):
    if api_call(request):
        return api_message(err.code)
    return render_template(
        'error.html',
        error_code=err.code,
        error_message=ERROR_DICT[err.code])


def api_call(request):
    '''Checks if it's an API call'''
    if (request.path.startswith('/api/') and
       request.accept_mimetypes.accept_json and
       not request.accept_mimetypes.accept_html):
        return True
    else:
        return False


def api_message(code):
    '''Prepares HTTP response for API calls.'''
    message = {
        'status': code,
        'message': ERROR_DICT[code],
    }
    resp = jsonify(message)
    resp.status_code = 404
    return resp
```

miketaylr:
thanks @karlcow -- I agree, we could abstract the error handling into a nice little method. Will fix the failing nose test and write some new ones before merging.

miketaylr:
OK, tests are passing -- let's merge. Thanks for review @karlcow.

webcompat/views.py
```diff
@@ -278,6 +295,24 @@ def jumpship(e):
     return redirect(url_for('index'))
 
 
+@app.errorhandler(400)
+def unauthorized(err):
```
karlcow: Ah my bad!!!!
Need better eyeballs.

@miketaylr The function is being redefined You need to change the name when you cut and paste :p

karlcow: 3 times the same function :D for 400, 401 and 403

miketaylr: ooops, will fix tomorrow!

miketaylr: (also, thanks -- good catch, even if a day late :p)

PR #933, opened 2016-03-02
Markdown

magsout:
I added new Component `Markdown` to handle _element_ from Github without ClassName. 

r? @miketaylr


magsout:
Oops, neeed to rebase

magsout:
@miketaylr I need you. I'm not an expert of git. When I try to fixed conflict after rebased I'm lost...

miketaylr:
@magsout np! let me try to fix it up and push.

miketaylr:
OK, should be good to review now. :+1:

miketaylr:
Hm, weird error:

```
Running "cssnext:dist" (cssnext) task

Warning: /home/travis/build/webcompat/webcompat.com/webcompat/static/css/development/main.css:54:1: Failed to find 'components/Markdown' from /home/travis/build/webcompat/webcompat.com
```

`cssnext:dist` task works for me locally, maybe I need to kick the Travis cache.

webcompat/static/css/development/main.css
```diff
@@ -51,3 +51,4 @@
 @import "components/label";
 @import "components/upload-form";
 @import "components/contributors";
+@import "components/Markdown";
```
miketaylr: Maybe Travis doesn't like `Markdown` vs `markdown` here? (I guess Linux cares about case sensitivity and OSX doesn't? :confused:)

magsout:
Ah yes, I usually write Markdown instead of markdown on my other project. But in this case `@import component/markdown` is good

miketaylr:
LGTM. :shipit:

magsout:
miketaylr  thanks, but curious, how did you do ?

miketaylr:
Sure!

So to rebase against master from a branch (not a fork) I usually do something like this:

```
git fetch
git pull --rebase origin master
```

And then when there's a conflict I do `git status` to see which files were "both modified" (or whatever the message is).

I open those in my editor and search for `<<<<` to jump to the spot where the conflict is. This is the tricky part -- you have to decide which parts should stay and which parts go. This can be dead simple, or it can be confusing if patches change things around a lot.

https://help.github.com/articles/resolving-a-merge-conflict-from-the-command-line/#creating-an-edit-collision should help here a little. But it can still be a pain in the butt to get right.

Once you think you've got it right, you have to `git add file-that-you-just-edited` to let git know that you're done with the conflict resolution. And after that, `git rebase --continue`. 

You might get another conflict from the next patch it applies in the rebase. So rinse and repeat until it tells you you're done.

magsout:
>Once you think you've got it right, you have to git add file-that-you-just-edited to let git know that you're done with the conflict resolution. And after that, git rebase --continue. 

OK, what I missed when I tried to rebase... Thanks for the explanation, :+1:

Issue #857, opened 2015-12-12
Some browsers start inline find on keypress, cancel default action on [g]


hallvors:
Minor annoyance: we have a ```[g]``` shortcut key to take you to GitHub. However, if the browser is configured to start inline find on a keypress it will do so when the user hits G, and start highlighting various letters in the text while navigation is happening. 

The simple fix is to add 
```
e.preventDefault();
```
here: https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/issue-list.js#L544

miketaylr:
@hallvors which platform does this happen on (just curious)? On mac it's `/`

deepthivenkat:
#938

PR #938, opened 2016-03-03
Issue #857 - prevent the default action inline find on keypress

deepthivenkat:
Added prevent default for the key press event


miketaylr:
Thank you @deepthivenkat!

PR #941, opened 2016-03-04
Fixes #821: Add Tablet info to browser name

### LINKED ISSUE #821: Add Tablet info to browser name
See https://github.com/webcompat/webcompat.com/issues/819#issuecomment-153974462.

The idea would be if we have `device.model == 'Tablet'`, we add a `(Tablet)` after the browser name. Just need to investigate how other tablet browsers are currently displayed so we don't do something silly like `Chrome Tablet (Tablet)`.

dshgna:
Initial commit.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -30,12 +30,13 @@ GITHUB_SEARCH_LINK_HEADER = '<https://api.github.com/search/issues?q=taco&page=2
 REWRITTEN_ISSUES_LINK_HEADER = '</api/issues?per_page=50&page=3>; rel="next", </api/issues?per_page=50&page=4>; rel="last", </api/issues?per_page=50&page=1>; rel="first", </api/issues?per_page=50&page=1>; rel="prev"'  # nopep8
 REWRITTEN_SEARCH_LINK_HEADER = '</api/issues/search?q=taco&page=2>; rel="next", </api/issues/search?q=taco&page=26>; rel="last"'  # nopep8
 PARSED_LINKED_HEADERS = [{'link': 'https://api.github.com/repositories/17839063/issues?per_page=50&page=3', 'rel': 'next'}, {'link': 'https://api.github.com/repositories/17839063/issues?per_page=50&page=4', 'rel': 'last'}, {'link': 'https://api.github.com/repositories/17839063/issues?per_page=50&page=1', 'rel': 'first'}, {'link': 'https://api.github.com/repositories/17839063/issues?per_page=50&page=1', 'rel': 'prev'}]  # nopep8
-NON_TABLET_UA = "Mozilla/5.0 (Android; Mobile; rv:40.0) Gecko/40.0 Firefox/40.0"
-TABLET_UA = "Mozilla/5.0 (Android 4.4; Tablet; rv:41.0) Gecko/41.0 Firefox/41.0"
+NON_TABLET_UA = "Mozilla/5.0 (Android; Mobile; rv:40.0) Gecko/40.0 Firefox/40.0"  # nopep8
+TABLET_UA = "Mozilla/5.0 (Android 4.4; Tablet; rv:41.0) Gecko/41.0 Firefox/41.0"  # nopep8
 PARSED_NON_TABLET_BROWSER_NAME = "firefox mobile"
 PARSED_TABLET_BROWSER_NAME = "firefox mobile tablet"
 PARSED_NON_TABLET_BROWSER = "Firefox Mobile 40.0 "
 PARSED_TABLET_BROWSER = "Firefox Mobile 41.0 (Tablet)"
+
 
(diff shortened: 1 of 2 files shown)


miketaylr:
Looks like you're on the right track, yes!

It might be good to add a few tests to https://github.com/webcompat/webcompat.com/blob/master/tests/test_helpers.py to make sure the right UA strings get the Tablet treatment and some non-Tablet ones don't.

karlcow:
hehe @miketaylr ping.

But to help. The science of UA strings is black magic. So basically there's no perfection. But we want to make sure that indeed we get the right label when there is a tablet hitting webcompat.

So currently get_browser_name is used in
https://github.com/dshgna/webcompat.com/blob/issue/821/webcompat/views.py#L129-L133

which calls itself the templating engine for the index page and prepopulate the form:

capture d ecran 2016-03-08 a 12 39 13

```html
<div class="wc-Form-group js-Form-group">
            <label class="wc-Form-label" for="browser">Browser / Version</label> <input class="wc-Form-item" id="browser" name="browser" value="Firefox 47.0" type="text">
          </div>
```

When we actually submit the form we get the value of the UA from 

```python
```
https://github.com/dshgna/webcompat.com/blob/issue/821/webcompat/views.py#L162

Then we head to `report_issue`

https://github.com/dshgna/webcompat.com/blob/issue/821/webcompat/issues.py#L19

which is using `build_formdata`
https://github.com/dshgna/webcompat.com/blob/issue/821/webcompat/form.py#L139

which is reading the `browser` parameter
https://github.com/dshgna/webcompat.com/blob/issue/821/webcompat/form.py#L181

but also triggers the labels:
https://github.com/dshgna/webcompat.com/blob/issue/821/webcompat/form.py#L178

aka `get_labels()`
https://github.com/dshgna/webcompat.com/blob/issue/821/webcompat/form.py#L106

SO I guess the final question is what type of labels do we want in the bug reporting system when there is a tablet :)

miketaylr:
> hehe @miketaylr ping.

Oh! So sorry @dshgna, this mention escaped me. :see_no_evil:

miketaylr:
@dshgna if you click on the Details link of the travis-ci build (see https://cloudup.com/cy5FSvXC0WB), you'll see the following error (scroll down and look for the red text):

`The command "pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example" failed and exited with 1 during .`

Can you re-run pep8 and fix any errors there (and push the fixes)?

miketaylr:
@karlcow do you think it makes sense to have "firefox mobile tablet" label in addition to "firefox mobile"? Or is the name in the bug report "Firefox Mobile (Tablet)" sufficient? I don't have any strong feelings, I'm sort of leaning towards just one "firefox mobile" label...

miketaylr:
@dshgna a few commit nits to take care of if you don't mind:

When you fix up pep8 errors, can you add "Issue #821. " to the beginning of the commit message (for the test commit)? Likewise, for the first commit -- can you add a hash mark in front of the issue number, e.g., "Fixes #821"?

That will help link all the commits to the issue. If you need help on how to re-write a commit message just let me know!

karlcow:
@miketaylr  about

>@karlcow do you think it makes sense to have "firefox mobile tablet" label in addition to "firefox mobile"? 

I would rather have two differents labels such as:

* browsername-tablet
* browsername-mobile

and/or an information about the screen size. The reason it has happen often that I didn't know if the bug reporter was talking about tablets and was testing on a small viewport device. In some cases, Web sites choose to send desktop sites to the tablet form factor.

I suspect you guess the form factor by looking at the UA in general and because you spent most of your time in github instead of webcompat ;) :p (Use the force, mike, it helps to improve the product hmm no laser sabre in emojis. Maybe this will do ⚡️🗡🐸)

miketaylr:
> I would rather have two differents labels such as:

OK, that's what this should do. :+1: 

> and/or an information about the screen size. The reason it has happen often that I didn't know if the bug reporter was talking about tablets and was testing on a small viewport device. In some cases, Web sites choose to send desktop sites to the tablet form factor.

Sounds like a good feature request. File a bug? This would be good info to put in the bug body.

> I suspect you guess the form factor by looking at the UA in general and because you spent most of your time in github instead of webcompat ;)

Yes and no. :p

dshgna:
@miketaylr Just a bit confused here on the merge :) 

> I would rather have two different labels such as:

I got the understanding that I had to change the 'browser mobile tablet' to 'browser tablet' and just made a [commit](https://github.com/dshgna/webcompat.com/commit/aa99978839e723072c088690023d8269aa2931cb) as such. 

So, is this okay as is?

miketaylr:
Personally I think "firefox mobile tablet" is ok! Let's see how it feels and we can change it up as needed.

dshgna:
@miketaylr sure thanks! :+1:

Issue #942, opened 2016-03-04
Split up Travis tests into 2 auth and non-auth tasks


miketaylr:
That way forks w/o access to the encrypted ENV VARs can run the non-auth tasks (and nosetests, etc.).

See https://docs.travis-ci.com/user/pull-requests#Security-Restrictions-when-testing-Pull-Requests

miketaylr:
So, despite merging myself and  breaking things for a few minutes, this is working now:

![](https://cldup.com/8ko8z89q1F-3000x3000.png)

(I can hear @karlcow going *tsk tsk* all the way from Tokyo...)

miketaylr:
> this is working now:

Heh, sort of. Have to make that a non-exiting conditional.

PR #946, opened 2016-03-05
Issue #942. Don't exit 1 if the conditional fails.

miketaylr:


miketaylr:
(just fixing a bash syntax error here -- gonna self-merge so we don't break any future PR builds over the weekend)

miketaylr:
Half green, fail due to intermittent search weirdness. Seems good!

Issue #929, opened 2016-03-01
Make error handler code in views.py more DRY


miketaylr:
See @karlcow's comment at https://github.com/webcompat/webcompat.com/pull/928#issuecomment-190514243.

miketaylr:
@deepthivenkat do you want to take a look at this one?

deepthivenkat:
Yes..I will try this

deepthivenkat:
yes! I will try this

deepthivenkat:
#949

deepthivenkat:
#958   r? @karlcow   r? @miketaylr 
Sorry for the delay. @karlcow : I have considered the suggestions you have made for my previous pull.I have added generic methods(api_call(request), api_message(code)) for the following error codes : 400,401,403,404 and 500. However, I had to maintain separate methods for error handling messages for 429 and GitHubError. Is there any specific reason to make use of json.dump() instead of jsonify in app.errorhandler(429) ?

karlcow:
Thanks @deepthivenkat 
A couple of issues.

* nosetests reports failure. We need to understand why.
* When you commit things, you need to make sure that your commit explains what it means. See for example

```
3240856 #710 Adds parameters to save all frames for an animated GIF
577dcca #710 Optimizes images further for JPEG and PNG
3ab1512 #710 Adjusts to TypeError. e not needed
```

deepthivenkat:
r? @karlcow Thanks for the review comments.Sorry for committing though the tests failed.I will make sure I run all the tests before performing commit from now on.
Three tests failed. Two tests failed because I hard coded 404 in api_message method instead of passing code as argument.
The third test was written by me. I intended to assert if I get 403 when I try to check other users activity.But for that the user needs to be logged in first.So I was getting 401 instead of 403.
Since this test was not correct, I removed it.
Now all 3 tests are passing.

I have maintained the same branch since I needed the code changes that I have done previously.

karlcow:
Thanks @deepthivenkat 

The `GitHubError` was introduced by @miketaylr  for issue #32 I guess sending back whichever error Github message was.
https://github.com/deepthivenkat/webcompat.com/commit/d3676adfca4d7dea85bc112f64c8d8ed41b5ff74

I want to have a second opinion for this. Because if it's just about 401, maybe this piece of code can disappear given we are handling it.

deepthivenkat:
@karlcow  So we throw 401 for the following two cases?
1) Unauthorized, log in.
[...]

PR #949, opened 2016-03-07
Issue #929 - Generic error handler based on status code for status 40…

deepthivenkat:
…0,401,403,404


miketaylr:
r? @karlcow 

(I can review too, but Karl suggested this approach -- would be interested in his feedback).

miketaylr:
@deepthivenkat can you run pep8 and fix up any lint errors there?

`The command "pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example" failed and exited with 1 during .`

deepthivenkat:
Sure..checking

karlcow:
@deepthivenkat Thanks for starting working on this.

Let's discuss a bit more. ^_^

So in my approach I realized that I mixed two ways of making it more DRY, and it's probably over-doing it.
There's room for discussions on how to improve in a step by step.

* If we are sure that all `4**` HTTP codes follow the same pattern. We do not to abstract the two functions (`api_call(request)`, `api_message(code)`). We just need to have the appropriate decorators on top of one generic function.
* If we have different treatments in between the `4**` HTTP codes then the abstraction of the two methods is worth. 

We still need a dictionary for error messages. We also need to figure out if we want to have separate messaging for HTML (Human calls) and JSON (API calls).

Opinions? If what I say above is not clear enough, I can go a bit more in details. :)

karlcow:
There's a need to clean this pull request with the input of #958 and also it removes some data. etc.
Maybe easier would be to restart with a clean branch and delete this pull request and the #958 
If you need guidance tell us.

deepthivenkat:
I accidentally deleted requirements.txt. there were few more issues. So I deleted the repository and cloned it again and started with a clean branch(the one I submitted yesterday is from this new branch). Is it okay if we continue with the new pull request?

magsout:
@deepthivenkat so let's closed this issue in favor of https://github.com/webcompat/webcompat.com/pull/958, sound good ?

deepthivenkat:
Yes..Thanks :)

PR #951, opened 2016-03-07
Fix #767: display proper URL hostname when scheme is missing slashes

### LINKED ISSUE #767: when http URL is missing a slash in the report
This is minor. It happened once in 
https://github.com/webcompat/web-bugs/issues/1807

Or logic for parsing the URL and displaying the domain in the title of an issue fails when the URL is having an incorrect http. I'm merely filling the issue to have  a record of the problem. It good be a good first patch. 

To make it clearer.

The person typed/reported:

```
https:/amazon.com
```

And then our code parsed it as 

```
https: 
```

creating a title which is `https: - layout is messed up` instead of `amazon.com - layout is messed up`

deckycoss:
this updates webcompat.form.normalize_url to account for missing slashes (e.g., "http:/example.com"), and adds basic unit tests for normalize_url and domain_name.

r? @miketaylr

### DIFF
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+'''Tests for form validation.'''
+
+import unittest
+from webcompat import form
+
+
+class TestForm(unittest.TestCase):
+
+    def test_normalize_url(self):
+
+        r = form.normalize_url('example.com')
+        self.assertEqual(r, 'http://example.com/')
+
+        r = form.normalize_url('http:/example.com')
+        self.assertEqual(r, 'http://example.com/')
+
+        r = form.normalize_url('https:/example.com')
+        self.assertEqual(r, 'https://example.com/')
+
+        r = form.normalize_url('http:example.com')
+        self.assertEqual(r, 'http://example.com/')
+
+        r = form.normalize_url('https:example.com')
+        self.assertEqual(r, 'https://example.com/')
+
+
+    def test_domain_name(self):
+        
+        r = form.domain_name("http://example.com")
+        self.assertEqual(r, "example.com")
+
+        r = form.domain_name("https://example.com")
+        self.assertEqual(r, "example.com")
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -30,7 +30,7 @@ class TestForm(unittest.TestCase):
         self.assertEqual(r, 'http://example.com')
 
     def test_domain_name(self):
-        
+
         r = form.domain_name("http://example.com")
         self.assertEqual(r, "example.com")
 
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -11,6 +11,12 @@ class TestForm(unittest.TestCase):
 
     def test_normalize_url(self):
 
+        r = form.normalize_url('http://example.com')
+        self.assertEqual(r, 'http://example.com')
+
+        r = form.normalize_url('https://example.com')
+        self.assertEqual(r, 'https://example.com')
+
         r = form.normalize_url('example.com')
         self.assertEqual(r, 'http://example.com')
 
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -12,20 +12,22 @@ class TestForm(unittest.TestCase):
     def test_normalize_url(self):
 
         r = form.normalize_url('example.com')
-        self.assertEqual(r, 'http://example.com/')
+        self.assertEqual(r, 'http://example.com')
 
         r = form.normalize_url('http:/example.com')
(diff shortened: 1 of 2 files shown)


tests/test_form.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+'''Tests for form validation.'''
+
+import unittest
+from webcompat import form
+
+
+class TestForm(unittest.TestCase):
```
miketaylr: :balloon: Awesome tests.

webcompat/form.py
```diff
@@ -118,9 +119,24 @@ def get_labels(browser_name):
...
     '''normalize URL for consistency.'''
     url = url.strip()
-    if not url.startswith(SCHEMES):
+    parsed = urlparse.urlparse(url)
+
+    if url.startswith(BAD_SCHEMES):
+        # if url starts with a bad scheme, parsed.netloc will be empty,
+        # so we use parsed.path instead
+        path = parsed.path.lstrip('/')
+        url = '%s://%s' % (parsed.scheme, path)
```
miketaylr: nit: can we use `str.format` style (and elsewhere) here instead of the old `%` `sprintf` style of string interpolation here?

(I know `normalize_url` was already using it, but generally I think `'{0://{1}'.format(parsed.scheme, path)` is easier to read).

See https://docs.python.org/2/library/stdtypes.html#str.format

tests/test_form.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def test_normalize_url(self):
+
+        r = form.normalize_url('example.com')
```
miketaylr: Can we add a test for the instances where nothing should happen, i.e., `http://example.com` and `https://example.com`? (Right now those will fail, will comment below in a sec)

webcompat/form.py
```diff
@@ -118,9 +119,24 @@ def get_labels(browser_name):
 def normalize_url(url):
     '''normalize URL for consistency.'''
     url = url.strip()
-    if not url.startswith(SCHEMES):
+    parsed = urlparse.urlparse(url)
+
+    if url.startswith(BAD_SCHEMES):
```
miketaylr: This won't do what we expect because `'http://'.startswith('http:/')` => `True`. Possibly we could do something with a RegExp or splitting on `//`?

webcompat/form.py
```diff
@@ -118,9 +119,24 @@ def get_labels(browser_name):
...
+        if parsed.fragment:
+            url += '#' + parsed.fragment
+    elif not parsed.scheme:
         # We assume that http is missing not https
-        url = 'http://%s' % (url)
+        if url.startswith("//"):
+            url = "http://%s" % (url[2:])
+        else:
+            url = 'http://%s' % (url)
+
```
miketaylr: nit: we can remove this empty line here

webcompat/form.py
```diff
@@ -118,9 +119,24 @@ def get_labels(browser_name):
...
+    parsed = urlparse.urlparse(url)
+
+    if url.startswith(BAD_SCHEMES):
+        # if url starts with a bad scheme, parsed.netloc will be empty,
+        # so we use parsed.path instead
+        path = parsed.path.lstrip('/')
+        url = '%s://%s' % (parsed.scheme, path)
+        if parsed.query:
+            url += '?' + parsed.query
+        if parsed.fragment:
```
miketaylr: I don't think we need to worry about the query or fragment since we just want the domain name. Those bits will be contained in the `url` of the form.

miketaylr:
This is a good start @deckycoss, let me or @karlcow know if you have any other questions.

webcompat/form.py
deckycoss: i'd like that; i was only using % to be consistent with the code that was already there.

webcompat/form.py
deckycoss: @miketaylr if i take that bit out then normalize_url("http:example.com/example#section") will return "http://example.com/example", whereas normalize_url("http://example.com/example#section") will return "http://example.com/example#section". is that acceptable?

webcompat/form.py
miketaylr: Yeah, that's what I thought. No big deal, I just prefer `format`

webcompat/form.py
miketaylr: Hmmmmmmm... I'm trying to think here. This should only matter in the case where the `if domain` part doesn't succeed (which if it did, we only display a domain). 

I think you're right, let's leave this as you've originally written it. Sorry!

deckycoss:
@miketaylr i believe this last commit should address your feedback

miketaylr:
LGTM. :+1:

PR #956, opened 2016-03-09
Fixes #557: Create installation script for labels in a test repo

### LINKED ISSUE #557: Installation script for labels in a test github repo
This is related to #436 
Currently someone who want to hack on the project would have to create by hand all the labels.
It has a couple of consequences:
* cumbersome
* prone to errors
* not easy for first timers

We may want to create an initialization script for it.

dshgna:
r? @miketaylr @karlcow 

Tested successfully on my [test repo] (https://github.com/dshgna/test-repo/labels)

### DIFF
--- a/labels.py
+++ b/labels.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import requests
+import json
+import sys
+from webcompat import app
+
+API_URI = 'https://api.github.com/repos/'
+ISSUES_URI = app.config['ISSUES_REPO_URI'].rpartition('/issues')[0]
+USER_LABELS_URI = API_URI + ISSUES_URI + "/labels"
+WEBCOMPAT_LABELS_URI = API_URI + 'webcompat/webcompat-tests/labels'
+AUTH_HEADERS = {'Authorization': 'token {0}'.format(app.config['OAUTH_TOKEN']),
+                'User-Agent': 'webcompat/webcompat-bot'}
+
+
+def get_issue_labels(labels_uri):
+    '''Get the issue labels from the given repo.'''
+    response = requests.get(labels_uri)
+    if not response.status_code == 200:
+        response.raise_for_status()
+    return response.json()
+
+
+def create_label(labels):
+    '''Create a label in the user repo'''
+    response = requests.post(
+                            USER_LABELS_URI,
+                            json.dumps(labels),
+                            headers=AUTH_HEADERS)
+    if not response.status_code == 201:
+        response.raise_for_status()
+
+
+def delete_label(name):
+    '''Delete a label in the user repo'''
+    delete_labels_uri = USER_LABELS_URI + "/" + name
+    response = requests.delete(delete_labels_uri, headers=AUTH_HEADERS)
+    if not response.status_code == 204:
+        response.raise_for_status()
+
+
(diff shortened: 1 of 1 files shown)


labels.py
```diff
@@ -0,0 +1,47 @@
...
+
+def get_issue_labels(labels_uri):
+    '''Get the issue labels from the given repo.'''
+    r = requests.get(labels_uri)
+    return r.json()
+
+
+def create_label(data):
+    '''Creates a label in the given repo'''
+    requests.post(USER_LABELS_URI, json.dumps(data), headers=AUTH_HEADERS)
```
karlcow: @dshgna You might probably want to forbid the create_label when the repo is `webcompat/webcompat-tests` and `webcompat/webcompat.com`. So a user with the right privilege doesn't by accident erase and re-create the labels. Typos and sleepless nights are so quick.

Maybe to be less magic, I would probably give `def create_label(labels, repo_uri)` in the spirit of `what` and `where`

Also it would be good to test the result of your request here and return a code/string for the user to know if it really happened. try/except probably.

labels.py
```diff
@@ -0,0 +1,47 @@
...
+USER_LABELS_URI = API_URI + ISSUES_URI + "/labels"
+WEBCOMPAT_LABELS_URI = API_URI + 'webcompat/webcompat-tests/labels'
+AUTH_HEADERS = {'Authorization': 'token {0}'.format(app.config['OAUTH_TOKEN']),
+                'User-Agent': 'webcompat/webcompat-bot'}
+
+
+def get_issue_labels(labels_uri):
+    '''Get the issue labels from the given repo.'''
+    r = requests.get(labels_uri)
+    return r.json()
```
karlcow: Here also you need to know if the request really happened. What is happening in the code if there's a network failure and the data are empty.

labels.py
```diff
@@ -0,0 +1,47 @@
...
+for label in user_labels:
+    delete_label(label.get('name'))
+
+'''get all labels from the webcompat tests repo'''
+webcompat_labels = get_issue_labels(WEBCOMPAT_LABELS_URI)
+
+'''re-create labels in user issues repo'''
+for label in webcompat_labels:
+    data = {"name": label.get('name'), "color": label.get('color')}
+    create_label(data)
```
karlcow: create functions. for everything you need and put the main logic of your code in a `main` function. Tell me if you need more details.

labels.py
```diff
@@ -0,0 +1,47 @@
...
+'''delete all existing labels from the user issues repo'''
+user_labels = get_issue_labels(USER_LABELS_URI)
+for label in user_labels:
+    delete_label(label.get('name'))
+
+'''get all labels from the webcompat tests repo'''
+webcompat_labels = get_issue_labels(WEBCOMPAT_LABELS_URI)
+
+'''re-create labels in user issues repo'''
+for label in webcompat_labels:
```
karlcow: This should happen only if the  previous request has been successful. If not successful, we should warn there was an issue.

dshgna:
r? @karlcow
Thanks a lot for the very detailed review above. 
Added exception handling code. Please let me know if I have missed anything :)

karlcow:
@dshgna Thanks for the modifications
There is still a couple of things. I will comment into the last commit.

karlcow:
OK I have the feeling that labels.py should not be at the root and we start to have multiple scripts for management. webcompat.sh, Makefile, and now labels.py, We need to find a better strategy for those. Probably a separate issue. @magsout @miketaylr 

Maybe a tools/ directory.

dshgna:
@karlcow: Thanks again for the feedback, I'm learning a lot. I've made the changes and committed.

I also think a tools directory would be great!

karlcow:
r+ Thanks a lot @dshgna

Issue #906, opened 2016-01-26
[ui] upload image in comment


magsout:
We have this 
![capture d ecran 2016-01-26 a 09 51 04](https://cloud.githubusercontent.com/assets/1997108/12576209/577f2564-c412-11e5-9882-4a376e5573d8.png)

But the inital design of @calexity is better IMHO.

https://github.com/webcompat/design/blob/master/2_Current%20Designs/Issues/Sketch%20and%20PSDs/IssuesStyle.psd

Any thought? @karlcow @hallvors @miketaylr @Raphseck

miketaylr:
@magsout agreed. Right now we only support file selection (I plan on adding DnD support this quarter), so we could just say "<a>Select an image</a> to attach." or keep it as "Attach a screenshot image" but use that style.

cch5ng:
hi, I'm interested in applying to the upcoming outreachy internship. I'd like to work on this issue as part of my application.

cch5ng:
@magsout do you want this change to apply to only new comments for existing issues?
(i.e. like http://localhost:5000/issues/252)

should this also apply to the form for new issues?
(i.e. like http://localhost:5000/?open=1)

![screen shot 2016-03-11 at 3 32 11 pm](https://cloud.githubusercontent.com/assets/4755432/13718455/e9e7eada-e79e-11e5-8ceb-6e93aca9e03c.png)

magsout:
Hello @cch5ng , thanks for your help. 

Yes,  it is to improve the uploader only in a new comment for existing issue.

For now we just keep the ability to sent an image after a click on a input. Drag and drop Will be add later.

miketaylr:
Thanks again @cch5ng -- this looks so much better.

<img width="838" alt="screen shot 2016-03-25 at 6 31 49 pm" src="https://cloud.githubusercontent.com/assets/67283/14056270/deca4976-f2b7-11e5-8874-f363b373ceef.png">

PR #961, opened 2016-03-12
fix: #906 update style for new comment of old bug

cch5ng:
r? @magsout (also I had a question is there a checklist of browsers/versions I am expected to test?)

### DIFF
--- a/webcompat/static/css/development/components/comment.css
+++ b/webcompat/static/css/development/components/comment.css
@@ -84,6 +84,14 @@
     background : #fff;
   }
 
+    /* modifier new comment of existing bug */
+    .wc-Comment-wrapper-new {
+      border-bottom: 1px dashed var(--Comment-color);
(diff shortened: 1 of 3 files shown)


magsout:
@cch5ng hum,  we support ` browsers: ['ff >= 20', 'ie >= 10', 'safari >= 7', 'opera >= 12', 'chrome >=30'],`

We should support less browser version for Firefox, opera and Chrome. Maybe we need to check some metric with analytics. I would say last 2 version of browser is a good idea.

cch5ng:
@magsout thanks. sorry, let me test for IE (and older versions for other browsers). I will send a review request afterwards. 

should I be concerned for the failing travis tests? I ran unit tests locally and most were passing. I think a couple failed but were not related to my changes (I think due to my environment setup)

cch5ng:
closing this PR and opening a new one. had additional one line change for the fix.

Issue #707, opened 2015-09-15
add label for Edge browser


karlcow:
We need to add browser-edge as a label.

karlcow:
we need to adjust the bot. See repo for the bot.

karlcow:
https://github.com/webcompat/web-bugs-issue-tagger/issues/9

deepthivenkat:
I should be modifying this [file](https://github.com/webcompat/webcompat.com/blob/60637b1b5885869c42a98e7513ddf8821455b9f9/webcompat/templates/list-issue/issue-list.html)  ?

Any border color preference?

deepthivenkat:
#962 r? @karlcow

miketaylr:
This is fixed.

PR #962, opened 2016-03-13
Issue#707

deepthivenkat:


karlcow:
@deepthivenkat Some issues to fix ^_^

1. Here you are trying to create a pull request in the same branch. See there are commits for the other issue you were trying to fix before. When working on issues, create a new branch for each issue and always starts from the master. The way I do it.

```bash
git checkout master # be sure to be in master
git fetch origin master # grab the latest version of master
git checkout -b 707/2 # create my second version of work on the issue 707. See below
# working on your branch 707/2 with simple commits.
# and meaningful messages
git commit -m "#707 Do this and that."
git commit -m "#707 Check for blah."
# You can push to your own fork on github
git push origin 707/2 # create a branch with the name 707/2 on github in your repo.
# You can request for comments.
# When you are done, create a pull request with the right message against master on webcompat.
```

Branch names: If I work on issue 707. I usually name the first branch `707/1`, then if I realize it was not what I intended, or it was not a good work iteration or I took too long working on it and I need to adjust to a master completely different. I go back to master and create a new branch `707/2` which would my second iteration about this work. I will delete `707/1` when I'm not sure I don't need anything else there. It's @miketaylr style that I have adopted. @magsout has a different style. :)

2. Pull requests need an explicit wording so we know what they are about. A better title would be `Fix #707 - Adds label for Edge browser`

I propose we close this Pull request. And you restart a branch with just the needed commits. 

Thanks.

PR #969, opened 2016-03-16
Fixes #865: Backend implementation of "detect bugs with similar domain names"

### LINKED ISSUE #865: Backend implementation of "detect bugs with similar domain names"
This issue is about coding a backend for the feature described in  #307 
This work can probably start before the design in #307 is ready. It will most likely depend on the issues database created in #788 
- [x] Use tldextract to get domain name in webhooks/helpers.py
- [x] Update db/**init**.py to add domain name field
- [x] Create new endpoint 'domainsearch' in api/endpoints.py 
- [ ] Dependency on Migration script for database. See #968

dshgna:
r? @hallvors @karlcow 

- [x] Use tldextract to get domain name in webhooks/helpers.py
- [x] Update db/__init__.py to add domain name field
- [x] Create new endpoint 'domainsearch' in api/endpoints.py

I successfully tested the functionality against my local test repo.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,4 +11,4 @@ Pillow==3.1.1
 requests==2.9.1
 ua-parser==0.6.1
 WTForms==2.1
-tldextract==1.7.5
+tldextract>=1.7.5
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,4 @@ Pillow==3.1.1
 requests==2.9.1
 ua-parser==0.6.1
 WTForms==2.1
+tldextract==1.7.5
--- a/webcompat/api/endpoints.py
+++ b/webcompat/api/endpoints.py
@@ -10,10 +10,12 @@ This is used to make API calls to GitHub, either via a logged-in users
 credentials or as a proxy on behalf of anonymous or unauthenticated users.'''
 
 import json
+import sys
 
 from flask import abort
 from flask import Blueprint
 from flask import g
+from flask import jsonify
 from flask import request
 from flask import session
 
@@ -26,6 +28,8 @@ from webcompat.helpers import mockable_response
 from webcompat.helpers import normalize_api_params
 from webcompat.helpers import proxy_request
 from webcompat.issues import filter_new
+from webcompat.db import issue_db
+from webcompat.db.helpers import domain_search
 
 api = Blueprint('api', __name__, url_prefix='/api')
 JSON_MIME = 'application/json'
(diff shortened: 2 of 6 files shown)


webcompat/api/endpoints.py
```diff
@@ -10,10 +10,12 @@
 credentials or as a proxy on behalf of anonymous or unauthenticated users.'''
 
 import json
+import sys
```
hallvors: Is the sys import used for anything?

webcompat/db/helpers.py
```diff
@@ -0,0 +1,21 @@
...
+    '''Returns the ten most recent issues with a similar domain name'''
+    search_domain += "%"
+    session = issue_db()
+    query = (
+        session.query(WCIssue)
+        .filter(WCIssue.domain.like(search_domain))
+        .limit(10).all()
+        )
+    result_dict = []
+    for row in query:
```
hallvors: Perhaps this code should be in a "to_dict" helper method or something?

webcompat/webhooks/helpers.py
```diff
@@ -48,8 +49,22 @@ def set_label(label, issue_number):
     payload = [label]
     api_post('labels', payload, issue_number)
 
+def extract_domain_name(url):
+    # List of major sites where using only the domain results in losing valuable information
+    blackList = r'(\.google\.com|\.live\.com|\.yahoo\.com|go\.com|\.js$)'
```
hallvors: A simplification I was trying to suggest on IRC was something like 

```
    prefix_blacklist = ['www']
    parts = tldextract.extract(url)
    if parts[0] in prefix_blacklist:
        return '.'.join(parts[1:2])
    else:
        return '.'.join(parts[0:2])

..or something like that. Well, perhaps keep the strip('www') to handle www.mail.yahoo.com and friends ;) For bonus points, strip all entries in prefix_blacklist..

File: requirements.txt [hunk 4]
```diff
@@ -11,3 +11,4 @@ Pillow==3.1.1
 requests==2.9.1
 ua-parser==0.6.1
 WTForms==2.1
+tldextract==1.7.5
```
hallvors: If tldextract is updated it presumably adds better data for the server name public prefix lists. I'd like to make the dependency >= so we get those updates.

webcompat/api/endpoints.py
```diff
@@ -26,6 +28,8 @@
 from webcompat.helpers import normalize_api_params
 from webcompat.helpers import proxy_request
 from webcompat.issues import filter_new
+from webcompat.db import issue_db
```
hallvors: Is this import required?

webcompat/webhooks/helpers.py
dshgna: Managed it :) Small question here on the domain blacklist: Do we need to include the \.js?

webcompat/api/endpoints.py
dshgna: Sorry, fixed it and the other unused import.

webcompat/webhooks/helpers.py
karlcow: `s/parts[0]/parts.subdomain/`
`s/parts[0:2]/parts[:]/`

but yes much better.

PR #971, opened 2016-03-17
[WIP] Management/Admin scripts for the project

magsout:
r? @karlcow 

wip, just for starting cleaned root folder, what do you think ?

- removed `Makefile` because it's device agnostic (unix only) and add all recipes directly in `package.json` I like this convention to add all recipes in one place
-  removed `.eslint`  and added all rules in `package.json`
- moved `requirement.txt` for now in `config` folder
- removed `webcompat.sh`


karlcow:
ah interesting. One thing I don't know the answer is that if there is a constraint for requirements.txt to be in the root directory. apart of that it's fine for me. 

@magsout once you moved requirements.txt, does the project restart well? I wonderf what is happening for example with 
https://github.com/webcompat/webcompat.com/blob/master/run.py#L71

magsout:
maybe just managed path here https://github.com/webcompat/webcompat.com/blob/master/run.py#L77 ? no ? Let me test.. Honestly I ddin't test anyhting about it, recipes only ... :rabbit:

magsout:
@karlcow I modified path in run.py and that's ok

screen recording 2016-03-17 at 08 51 pm

miketaylr:
Moving `requirements.txt` shouldn't be an issue -- we'll just need to update CONTRIBUTING.md docs to point to the new path when using `pip install -r path/to/requirements.txt`

miketaylr:
But +1 from me, I've been using npm scripts in another small project.

.travis.yml
```diff
@@ -36,7 +36,7 @@ before_install:
 install:
   - travis_retry npm install -g grunt-cli
   - travis_retry npm install
-  - travis_retry pip install -r requirements.txt --download-cache $HOME/.pip-cache
+  - travis_retry pip install -r config/requirements.txt --download-cache $HOME/.pip-cache
```
miketaylr: We had a PR update the Travis.yml file to not use this manual `--download-cache` thing, so there's probably a conflict here. Let me know if you need help resolving.

miketaylr:
@magsout the docs that need to be updated are here: https://github.com/webcompat/webcompat.com/blob/master/CONTRIBUTING.md#installing-project-source-code

r+ from me. Let's let @karlcow give the final +1.

magsout:
Need to figure out how npm travis script can be used to test the installation works well. perhaps decompose even in the script package.json

>@magsout the docs that need to be updated are here: https://github.com/webcompat/webcompat.com/blob/master/CONTRIBUTING.md#installing-project-source-code

Yes, of course, I waited that my idea looks good for you ;)

karlcow:
@magsout yup looks good. 


I'm not a big fan of having to use package.json for the recipes, but I can live with it. :) I would have personally used something in python, but not very important.
How is it used now that the Makefile is gone?

Thanks! :cake:

magsout:
>I'm not a big fan of having to use package.json for the recipes, but I can live with it. :) I would have personally used something in python, but not very important.
How is it used now that the Makefile is gone?

In many project with have node and package.json, So I guess wen can say it's universal ? ;p

> How is it used now that the Makefile is gone?

just replace all `make`  by `npm run`

```shell
npm run watch
npm run build
npm run start
```

magsout:
Updated @karlcow @miketaylr 

What do you think about all recipes ? Revelant ? Clear ?

package.json
```diff
@@ -38,5 +38,64 @@
     "postcss-url": "^5.1.1",
     "suitcss-utils-align": "^0.2.0",
     "suitcss-utils-display": "^0.4.0"
+  },
+  "scripts": {
+    "init": "npm run module && npm run virtualenv && && npm run pip && npm install && npm run config",
```
miketaylr: `npm run virtualenv && && npm run`

Extra `&&`?

miketaylr:
Yep, LGTM except for the one small `&&` issue.

karlcow:
@magsout looks good to me. 
Thanks :)

magsout:
OOoOps

miketaylr:
> @magsout looks good to me.

Karl gave his r+. So let's merge.

PR #973, opened 2016-03-17
Fixes #741: help messages in github issues

### LINKED ISSUE #741: Help feature for first-time visitors who land on a Github issue
If we @mention GitHub-users, they are going to get a notification sending them to a GitHub issue URL. This is extremely useful for us and one of the main reasons we are sticking to GitHub as a backend. 

However, it is not necessarily easy for an unsuspecting first-time visitor to some issue to understand what this project is about when arriving on a random bug report.

I suggest we add some standard text at the top of each bug report (we can hide this when we are *on* webcompat.com) - perhaps a "What's this about?" link?

deckycoss:
the text for the help message is a placeholder for now--it can be changed at form.GITHUB_HELP.

r? @karlcow


webcompat/form.py
```diff
@@ -25,11 +25,8 @@
 AUTH_REPORT = 'github-auth-report'
 PROXY_REPORT = 'github-proxy-report'
 SCHEMES = ('http://', 'https://')
-<<<<<<< HEAD
 BAD_SCHEMES = ('http:/', 'https:/', 'http:', 'https:')
-=======
-GITHUB_HELP = '<p class="is-hidden">This issue was filed via webcompat.com</p>'
->>>>>>> Issue 741: experimenting with hiding help messages via is-hidden class
+GITHUB_HELP = '<p class="github-help">This issue was filed via webcompat.com</p>'
```
karlcow: Merge conflict which was not fixed.

webcompat/form.py
```diff
@@ -26,6 +26,8 @@
 PROXY_REPORT = 'github-proxy-report'
 SCHEMES = ('http://', 'https://')
 BAD_SCHEMES = ('http:/', 'https:/', 'http:', 'https:')
+GITHUB_HELP = \
+    '<p class="github-help">This issue was filed via webcompat.com</p>'
```
karlcow: If you need a long line. two strategies:

```python
GITHUB_HELP = '''
<p class="github-help">This issue was filed via webcompat.com</p>
'''

or 
GITHUB_HELP =  '<p class="github-help">This issue was filed via webcompat.com</p>' #no_pep8

**Comment by miketaylr** [comment]
@karlcow does this OK to you know?

(Aside, I wonder if we should hide the `.github-help` class by default on webcompat.com -- it's redundant there)

**Comment by deckycoss** [comment]
@miketaylr @karlcow although my most recent push technically works, it still doesn't solve the problem discussed [here](https://github.com/webcompat/webcompat.com/issues/741#issuecomment-199144662)--namely, that the help message's visibility on webcompat depends, for some reason, on whether or not the paragraph contains any attributes.

it seems that when the help message paragraph has an attribute defined, it is parsed as part of the "browser info" comment text node, and thus is caught by the filter along with the browser info. however, if the p has no attributes, it is parsed as html, and is not caught by the filter. i've hit a wall trying to figure out why this happens. if i could be pointed to the code where js-Issue-markdown is filled with the github API data, then i might be able to work with that.

**Comment by miketaylr** [comment]
@hallvors can you give some pointers on the Markdown stuff? I know you've had some success working with that lib in the past.

**Comment by hallvors** [comment]
Hm.. it might or might not be similar to the IMG problem I looked into.. https://github.com/webcompat/webcompat.com/issues/764#issuecomment-150831121 Will investigate.

**Comment by karlcow** [comment]
> (Aside, I wonder if we should hide the .github-help class by default on webcompat.com -- it's redundant there)

Yes. It's necessary to modify the stylesheet or to add the is-hidden class value aka:

```html
<p class="github-help is-hidden">…
```

hallvors:
Here:
https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/vendor/markdown-it-sanitizer-0.4.1.js#L106
the regular expression allows only a given number of tags, and the *only* attribute it permits on any of those is a ``start="n"`` on OL (where n is one or more digits). You're not allowed to add class on a p, it won't be tokenized as markup if you try.

hallvors:
Since this creates basically the effect we're looking for, I suggest we might simply ship this. I'm not really impressed by the quality of the MarkdownIt library, so in the longer term we'd probably like to evaluate a few alternatives and find something else. If you add a corresponding CSS rule to the style sheets to hide this message, it will presumably remain hidden if and when we update the Markdown library.

miketaylr:
> or to add the `is-hidden` class

I like this idea. @deckycoss want to make this quick change? I think we're good to merge after that.

deckycoss:
@miketaylr sure thing

hallvors:
(My only concern with that is GitHub might have or at some point add an is-hidden class - in which case the message we want to see only on GitHub is hidden there. Can we at least make it wc-is-hidden or something?)

karlcow:
@hallvors it's not at some point… :)

https://assets-cdn.github.com/assets/github-c485dee0bed13cea79d66e0aad503b98c18ac78897b2d4eccf5bcf59af3babf3.css

```css
.is-hidden,.has-removed-contents {
	display:none
}
```

miketaylr:
OK -- so we've got some gross conflicts here (likely caushed by pushing the branch without rebasing against a later master). I think maybe the best thing to do is cherry pick the commits. We also need to add the actual `.wc-is-hidden` class.

webcompat/form.py
```diff
@@ -25,7 +25,11 @@
 AUTH_REPORT = 'github-auth-report'
 PROXY_REPORT = 'github-proxy-report'
 SCHEMES = ('http://', 'https://')
+<<<<<<< HEAD
```
miketaylr: Ah, this explains why my attempt at fixing conflicts was so hard. :)

deckycoss: sorry about that. i made sure to rebase against upstream before pushing 
my last commit. but if there's still conflicts, i must have missed 
something...

miketaylr: Yeah, it's creating conflicts as it replays the patch history for some reason. Not a huge issue, I can clean it up.

miketaylr:
Let's close this in favor of https://github.com/webcompat/webcompat.com/pull/991. (I squashed your commits into a commit over there @deckycoss -- none of our work is lost :+1: )

PR #984, opened 2016-03-29
Fixes #983: New issues not getting browser labels

### LINKED ISSUE #983: New issues aren't getting browser labels applied to them.
New issues aren't getting the browser labels applied to them.

For example, https://github.com/webcompat/web-bugs/issues/2394 -- that should have a "browser-firefox-mobile" label.

dshgna:
r? @miketaylr


miketaylr:
Ah interesting, so the bug was the extra space was screwing up the format the bot wanted?

dshgna:
@miketaylr: yep :( So there's a small issue now- (Tablet) devices will not be getting the browser label as such. We'll either have to introduce a new label or add a specific case in the label-handling code.

miketaylr:
@dshgna hmm, ok. i think ideally a Firefox Mobile tablet would have a "browser-firefox-mobile" label and continue to say "Firefox Mobile (Tablet)" in the UI. So.. that probably means some special-casing in the label code. Want to file an issue and work on that?

Issue #985, opened 2016-03-29
Report Site Issue failed with Firefox Nightly


karlcow:
It seems that at least @lissyx has issues with Report a Site issues on Firefox Nightly when reporting bug to webcompat. It would be good to know a bit more the circumstances. He tried for example yesterday to fill a bug for ryanair Web site and it failed. 

I wonder if the sequence is 
1. "report a site issue", then login dance to github
2. or already logged in and report a site issue.

cc @miketaylr who made the feature in Firefox Android.

karlcow:
It could be our side, it could be the side of Firefox.

lissyx:
I used the "anonymous" reporting feature. I got the spinner after tapping on it. Then it stays like this until this morning (yes, the whole night).

Circumstances were that I wanted to report broken Ryanair website on Nightly/Android

miketaylr:
It sounds like the same thing as #986. Possibly we broke something -- going to take a look with @adamopenweb shortly.

miketaylr:
Adam posted a stacktrace from the error that's throwing in https://github.com/webcompat/webcompat.com/issues/986#issue-144337969.

miketaylr:
@lissyx by chance were you using the Webcompat Reporter add-on?

https://addons.mozilla.org/en-US/firefox/addon/webcompatcom-reporter/

miketaylr:
Actually, it sounds like @lissyx was using the built-in "Report Site Issue" menu from Firefox for Androd. Can you confirm?

lissyx:
Yes, I was using the built-in report site issue.

miketaylr:
Thanks this should be fixed now. Let me know if you run into any other issue.

PR #988, opened 2016-03-29
Issue #985. Fix busted form submission (when form has screenshot)

miketaylr:
I'm going to see if tests are green, then r=me (due to how bad this bug is).


miketaylr:
What is confusing is this bug only happens in Firefox (deskop or mobile)... Chrome add-on works fine. I wonder if this wasn't a regression in behavior (I'm 100% sure this worked at the time I built the feature). :disappointed:

PR #990, opened 2016-03-31
Fixes #710 - Add optimization for Image formats

### LINKED ISSUE #710: Optimizing images when uploading to the server

karlcow:
* Adds a different set of optimizations for files format.
* There's a known issue in Pillow which is related to a simple Image.save() and has impact on animated gif. We could hack the Pillow library or wait how the discussion unfold. but the issue already exists in our current setup. 

This PR just  improves things, so I propose we merge.
I'm in the process of discussing Pillow issues on bug there.

### DIFF
--- a/webcompat/api/uploads.py
+++ b/webcompat/api/uploads.py
@@ -53,8 +53,8 @@ class Upload(object):
                 # Chop off 'data:image/.+;base64,' before decoding
                 imagedata = re.sub('^data:image/.+;base64,', '', imagedata)
                 return Image.open(BytesIO(base64.b64decode(imagedata)))
-            raise IOError('Not a valid image format')
-        except IOError as e:
+            raise TypeError('TypeError: Not a valid image format')
+        except TypeError:
             # Not a valid format
             abort(415)
 
--- a/webcompat/api/uploads.py
+++ b/webcompat/api/uploads.py
@@ -72,6 +72,7 @@ class Upload(object):
 
     def save(self):
         '''Check that the file is allowed, then save to filesystem.'''
+        save_parameters = {}
         if self.file_ext not in self.ALLOWED_FORMATS:
             raise TypeError('Image file format not allowed')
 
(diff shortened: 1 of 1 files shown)


karlcow:
@miketaylr r?

webcompat/api/uploads.py
```diff
@@ -53,8 +53,8 @@ def to_image_object(self, imagedata):
                 # Chop off 'data:image/.+;base64,' before decoding
                 imagedata = re.sub('^data:image/.+;base64,', '', imagedata)
                 return Image.open(BytesIO(base64.b64decode(imagedata)))
-            raise IOError('Not a valid image format')
-        except IOError as e:
+            raise TypeError('TypeError: Not a valid image format')
```
miketaylr: Just to learn, why do you propose `TypeError` over `IOError`?

webcompat/api/uploads.py
```diff
@@ -82,7 +83,15 @@ def save(self):
         dest_dir = os.path.dirname(file_dest)
         if not os.path.exists(dest_dir):
             os.makedirs(dest_dir)
-        self.image_object.save(file_dest)
+        # Optimize further the image compression for these formats
+        if self.image_object.format in ['JPEG', 'JPG', 'JPE', 'PNG']:
```
miketaylr: Hm, I wonder if (in a future patch) we can somehow merge this list and the `ALLOWED_FORMATS`. I'm sure there's a nicer way to do that.

(I also see something about BMP compression here: http://pillow.readthedocs.org/en/3.1.x/handbook/image-file-formats.html, but I doubt we'll be getting many of those files anyways).

webcompat/api/uploads.py
karlcow: Yes. Sure. Let's read the documentation for errors

* [IOError](https://docs.python.org/2/library/exceptions.html#exceptions.IOError)  
> Raised when an I/O operation (such as a print statement, the built-in open() function or a method of a file object) fails for an I/O-related reason, e.g., “file not found” or “disk full”.
* [TypeError](https://docs.python.org/2/library/exceptions.html#exceptions.TypeError)  
>  Raised when an operation or function is applied to an object of inappropriate type. The associated value is a string giving details about the type mismatch.

My feeling is that semantically we are bailing out because our code didn't get the appropriate format of image, more than the issue with the file being existent or not on the disk.

miketaylr: Thanks, helpful.

Issue #995, opened 2016-04-05
Make a nicer looking error message for "image is too big" error


miketaylr:
Right now it looks hilariously bad, but it's still an improvement over a nginx 413 page.

![](https://cldup.com/8Y6MXypyYa-2000x2000.png)

@magsout, would it be possible to make a red banner with white text similar to the "Remove Image Upload" banner we have? 

![](https://cldup.com/6fzIepWFt6-3000x3000.png)

(It wouldn't need to be a button, just plain text should be OK i think?)

karlcow:
Heh. Hilariously bad could be moved to hilariously funny. :question: 
Like the fly from the lightbulb who would have eaten too much. 
But as a first step. The submit button should be blocked and maybe the message should be inside the yellow area.

miketaylr:
Hehe, I like the idea of using a stuffed fly. :pizza: :bee: 


> But as a first step. The submit button should be blocked and maybe the message should be inside the yellow area.

Agreed. #996 makes the submits disabled. I was also thinking we could maybe change the yellow area to be red, with white text containing the error message.

magsout:
Message error in red at the bottom.
Close button on top/right


![capture d ecran 2016-04-08 a 09 03 31](https://cloud.githubusercontent.com/assets/1997108/14376421/b50f0178-fd69-11e5-880b-b4c7268d6182.png)

miketaylr:
Looks nice! I wonder if we can remove the close image, or at least make "choose an image" a hyperlink (or something that looks like a hyperlink, could be a button) that opens the image picker.

Or maybe the close image button also opens the file picker too?

magsout:
>Looks nice! I wonder if we can remove the close image, or at least make "choose an image" a hyperlink (or something that looks like a hyperlink, could be a button) that opens the image picker.

Ah yes, good point. Hum So icon is not enough and revelant..

PR #996, opened 2016-04-05
Fixes #993. Add clientside "image is too freaking big" validation.

### LINKED ISSUE #993: Prevent form submission on too-large images
STR:

1) Go to staging.webcompat.com
2) Try to Upload a massive image (like 5MB)

Expected: probably a client-side message saying "this is too big!" and I can't submit it yet
Actual: I get an ugly 413 page

So this bug deals with client-side detection of too-large images.

miketaylr:
Nothing fancy, but this prevents people from even hitting the server (which has similar size detection logic) when their images are too large.

Follow up bug for a better looking UI in #995.

r? @magsout

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -264,6 +264,7 @@ function BugForm() {
     // file is unreasonably large.
     if (img.size > 1024 * 1024 * 4) {
       this.makeInvalid('img_too_big');
+      return;
     }
 
     // One last image type validation check.
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -32,6 +32,12 @@ function BugForm() {
       // image should be valid by default because it's optional
       'valid': true,
       'helpText': 'Please select an image of the following type: jpg, png, gif, or bmp.'
+    },
+    'img_too_big': {
+      'elm': this.uploadField,
+      // image should be valid by default because it's optional
+      'valid': true,
+      'helpText': 'Please choose an image that is smaller than 4GB.'
     }
   };
 
@@ -222,6 +228,10 @@ function BugForm() {
       inlineHelp.insertAfter('.wc-Form-label--upload');
     }
 
+    if (id === 'img_too_big') {
+      inlineHelp.insertAfter('.js-image-upload');
+    }
+
     this.disableSubmits();
   };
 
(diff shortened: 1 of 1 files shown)


karlcow:
I see in #995, you have an already uploaded image in the UI. So I guess you already read it by that time.
https://developer.mozilla.org/en-US/docs/Web/API/Blob/size
https://w3c.github.io/FileAPI/#dfn-size
Does the File API gives the possibility to read the size before any upload, and so to bail out in advance? Maybe I'm making non sense.

webcompat/static/js/lib/bugform.js
```diff
@@ -250,7 +260,13 @@ function BugForm() {
     // We can just grab the 0th one, because we only allow uploading
     // a single image at a time (for now)
     var img = event.target.files[0];
-    // One last validation check.
+    // The limit is 4GB (which is crazy big!), so let the user know if their
+    // file is unreasonably large.
+    if (img.size > 1024 * 1024 * 4) {
```
miketaylr: > Does the File API gives the possibility to read the size before any upload, and so to bail out in advance? 

@karlcow yep! that's what we're doing here.

webcompat/static/js/lib/bugform.js
```diff
@@ -250,7 +260,13 @@ function BugForm() {
...
+      this.makeInvalid('img_too_big');
+    }
```
miketaylr: BUT, I did forget to `return` here. ^_^

magsout:
:+1:  r+

webcompat/static/js/lib/bugform.js
```diff
@@ -32,6 +32,12 @@ function BugForm() {
       // image should be valid by default because it's optional
       'valid': true,
       'helpText': 'Please select an image of the following type: jpg, png, gif, or bmp.'
+    },
+    'img_too_big': {
+      'elm': this.uploadField,
+      // image should be valid by default because it's optional
+      'valid': true,
+      'helpText': 'Please choose an image that is smaller than 4GB.'
```
magsout: hum, 4GB ?  4MO, no ?

miketaylr: ahahaha, you're right. I'll push a follow up command.

webcompat/static/js/lib/bugform.js
miketaylr: Also 4MB here, not GB.

PR #1003, opened 2016-04-09
Fixes #979. Indicate minimum supported Node (LTS)

### LINKED ISSUE #979: Determine minimum supported Node version and update docs
(Also it would be good to see if everything worked on a modern version, like v4 or v5)

miketaylr:
r? @magsout


magsout:
👍

PR #1013, opened 2016-04-22
Fixes #1010. Make flash messages use fixed positioning.

### LINKED ISSUE #1010: flash message should be "sticky" or fixed
If you're on mobile and hit an error, you'll likely not see the flash message up top. If we make it fixed position, or similar, that should solve the problem.

miketaylr:
This way they can be visible for smaller viewports that are scrolled down.

r? @magsout


magsout:
👍

miketaylr:
Thanks!

Issue #1000, opened 2016-04-07
Design a nicer image loader progress indicator


magsout:

miketaylr:
(this might actually be #1011, we have one but it's hidden on mobile at least).

(or would this be about tracking upload progress?)

miketaylr:
We can show upload progress to the server easily, btw. Just need a sweet design.

magsout:
Upload progress would awesome, but simple loader is good. The point is to show that something happen

miketaylr:
(oops, used wrong issue in my commit message)

miketaylr:
Changing the title since we already have a loader image... it's just not as awesome as it could be.

miketaylr:
cf. #492

miketaylr:
<img width="228" alt="screen shot 2016-04-22 at 5 30 50 pm" src="https://cloud.githubusercontent.com/assets/67283/14756570/fcbae7fc-08af-11e6-8e40-36a49d7e2cf6.png">

Maybe we could do something like have the bg of the lightbulb fill from light to dark (if that makes sense, sort of like "fill up" animation). But it would have to work no matter where you were scrolled, or what viewport size you have.

softvision-sergiulogigan:
I was about to suggest an improvement on the Loading Spinner, and I found this thread. I definitely think we should do something about that spinner. It doesn't quite match the website's sleek design.

zoepage:
Cross-referencing bug #1433 (so we don't forget) ;)

karlcow:
This one needs to be solved. I don't know if it's the same from elsewhere. But from Japan, if I didn't know better I would abandon the image upload most of the time. It takes around 10+s  without any visual indication that we are waiting for an image upload.

karlcow:
We also need to check what takes so much time.

karlcow:
See https://github.com/webcompat/webcompat-tests/issues/1266

miketaylr:
> But from Japan, if I didn't know better I would abandon the image upload most of the time. It takes around 10+s without any visual indication that we are waiting for an image upload.

We currently show a spinner here:

![d5lut52d7f](https://user-images.githubusercontent.com/67283/31908056-417ed53c-b7fb-11e7-96ab-91d6f03572fb.gif)

But either it's not appearing for you (possible bug), or you didn't know to look there (UX bug). Does it not appear for you @karlcow?

miketaylr:
[...]

PR #1014, opened 2016-04-22
Fixes #1011. Bump z-index of loader so it will always be on top.

### LINKED ISSUE #1011: loader spin gif below screenshot image
It should be above, so it's less confusing.

miketaylr:
The z-index of the image upload was also 2, so I just bumped it to 10 (in case something else gets a 3 in the future? -- probably unlikely)

r? @magsout


magsout:
Nice r+ 👌

magsout:
Close #1000

miketaylr:
Thanks!

PR #1016, opened 2016-04-23
Fixes #994. Better handle 413 errors when uploading an image.

miketaylr:
This PR has 3 related changes:

1) Show a better flash message when the server returns a 413 (image too large)
2) Fix the z-index issue where the flash message was behind the image preview
3) Fix bugs in "image too large" validation that meant submit buttons were never re-enabled if you removed or changed the image.

r? @magsout

PR #1018, opened 2016-04-25
Fixes #995. Nicer error styles for image errors (type + size)

### LINKED ISSUE #995: Make a nicer looking error message for "image is too big" error

miketaylr:
(We should land #1016 first, then I'll see if I need to rebase or fix any conflicts)

r? @magsout

### DIFF
--- a/webcompat/static/css/development/components/form.css
+++ b/webcompat/static/css/development/components/form.css
@@ -2,6 +2,7 @@
 
 :root {
   --Form-fontSize: 1rem;
+  --Form-helpMessageColor: #ff0000;
 }
 
 .wc-Form {}
@@ -12,7 +13,7 @@
     margin-bottom:1.6em;
   }
 
-  /* information: wrapp label and helpMessage */
+  /* information: wrap label and helpMessage */
   .wc-Form-information {
     display: flex;
   }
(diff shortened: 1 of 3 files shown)


webcompat/static/css/development/components/form.css
```diff
@@ -49,10 +50,29 @@
...
+      padding: 1em 0;
+      font-weight:700;
+      background-color: var(--Form-helpMessageColor);
+      border-bottom-left-radius: 20px;
+      border-bottom-right-radius: 20px;
+      color: #fff;
+      cursor: pointer;
+      text-align: center;
+      text-decoration: underline;
+      transform: translateY(0);
```
miketaylr: Hm, this is possibly copy-pasta that should be removed. I'm not using a `is-hidden` class to transition from.

miketaylr:
(Redirecting review to Adam, I think @magsout is swamped at work right now -- no issues ❤️ )

r? @adamopenweb

adamopenweb:
:+1:

miketaylr:
Thanks @adamopenweb!

PR #1033, opened 2016-04-30
Fixes #1031. Handle non-HTTPError exceptions

### LINKED ISSUE #1031: 500 handler shouldn't assume code property
Part of addressing #1009.

See https://github.com/webcompat/webcompat.com/issues/1009#issuecomment-215828564

miketaylr:
...by throwing a 500.

r? @karlcow

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -302,7 +302,10 @@ def cssfixme():
 @app.errorhandler(500)
 def custom_error_handler(err):
     # log the exception stack trace
-    app.logger.exception("!!!")
+    # (but don't bother for localhost because the
+    # Flask debugger is already enabled)
+    if not app.config['LOCALHOST']:
+        app.logger.exception("Exception thrown:")
     if api_call(request):
         return api_message(err.code)
     return render_template(
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -306,12 +306,14 @@ def custom_error_handler(err):
     # Flask debugger is already enabled)
     if not app.config['LOCALHOST']:
         app.logger.exception("Exception thrown:")
-    if api_call(request):
-        return api_message(err.code)
-    return render_template(
-        'error.html',
-        error_code=err.code,
-        error_message=ERROR_DICT[err.code]), err.code
+    try:
+        if api_call(request):
+            return api_message(err.code)
+        return render_template('error.html', error_code=err.code,
+                               error_message=ERROR_DICT[err.code]), err.code
+    except AttributeError:
+        # Somethign bad happened, we're not dealing with an HTTPError
+        abort(500)
 
 
 def api_call(request):


miketaylr:
```
FAIL: main - Search (auth) - Search from the homepage (11878ms)

StaleElementReference: [POST http://localhost:4444/wd/hub/session/4da9de69-cf9f-4099-980b-32be62bfc263/element/212/value / {"value":["vladvlad"]}] Element not found in the cache - perhaps the page has changed since it was looked up
```

Meh.

webcompat/views.py
```diff
@@ -302,13 +302,18 @@ def cssfixme():
...
+    if not app.config['LOCALHOST']:
+        app.logger.exception("Exception thrown:")
+    try:
+        if api_call(request):
+            return api_message(err.code)
+        return render_template('error.html', error_code=err.code,
+                               error_message=ERROR_DICT[err.code]), err.code
+    except AttributeError:
+        # Somethign bad happened, we're not dealing with an HTTPError
+        abort(500)
```
karlcow: Do we want to log somewhere the Exception?
And this seems recursive. abort(500) will call the custom error handler. Basically called twice?

Note that I don't have a better idea for now.

karlcow:
> The command "if [ "$TRAVIS_SECURE_ENV_VARS" == "true" ] ; then node_modules/.bin/intern-runner reporters=console config=tests/intern user="$USER" pw="$PW" ; fi" exited with 1.

webcompat/views.py
karlcow: or maybe something along:

```python
if isinstance(err, HTTPException):
    # Here the usual dance
else:
    # logging
    error_code = 500
    error_message = ERROR_DICT[err.code]
```

webcompat/views.py
```diff
@@ -302,13 +302,18 @@ def cssfixme():
...
-    if api_call(request):
-        return api_message(err.code)
-    return render_template(
-        'error.html',
-        error_code=err.code,
-        error_message=ERROR_DICT[err.code]), err.code
+    # (but don't bother for localhost because the
+    # Flask debugger is already enabled)
+    if not app.config['LOCALHOST']:
+        app.logger.exception("Exception thrown:")
```
karlcow: not for localhost. So for the rest. Do you really want to go through this for every error requests?

miketaylr: Personally I don't mind collecting errors, even if it's noisy (and maybe we never look at them). It can be useful to come back to when investigating other things.

But like I said elsewhere, we should consider using Sentry which would mean removing this at some point anyways.

webcompat/views.py
miketaylr: > Do we want to log somewhere the Exception?

The exception would get logged above (and be labeled as an AtributeError in the exception trace), or do you mean with a more specific message?

miketaylr: > And this seems recursive. abort(500) will call the custom error handler. Basically called twice?

I tested this by adding a `print('custom_error_handler')` at the top of `custom_error_handler`:

```
127.0.0.1 - - [02/May/2016 12:43:24] "GET / HTTP/1.1" 500 -
Error on request:
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/werkzeug/serving.py", line 180, in run_wsgi
    execute(self.server.app)
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/werkzeug/serving.py", line 168, in execute
    application_iter = app(environ, start_response)
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/app.py", line 1410, in handle_exception
    return handler(e)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/views.py", line 317, in custom_error_handler
    abort(500)
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/werkzeug/exceptions.py", line 605, in __call__
    raise self.mapping[code](*args, **kwargs)
InternalServerError: 500: Internal Server Error
```
If I actually call `custom_error_hander(500)`, rather than `abort(500)` I get the recursion error:

```
[snip a billion lines]
    custom_error_handler(500)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/views.py", line 317, in custom_error_handler
    custom_error_handler(500)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/views.py", line 317, in custom_error_handler
    custom_error_handler(500)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/views.py", line 317, in custom_error_handler
    custom_error_handler(500)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/views.py", line 317, in custom_error_handler
    custom_error_handler(500)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/views.py", line 317, in custom_error_handler
    custom_error_handler(500)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/views.py", line 317, in custom_error_handler
    custom_error_handler(500)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/views.py", line 317, in custom_error_handler
    custom_error_handler(500)
  File "/Users/miket/dev/compat/webcompat.com/webcompat/views.py", line 311, in custom_error_handler
    if api_call(request):
  File "/Users/miket/dev/compat/webcompat.com/webcompat/views.py", line 322, in api_call
    if (request.path.startswith('/api/') and
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/werkzeug/local.py", line 338, in __getattr__
    return getattr(self._get_current_object(), name)
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/werkzeug/local.py", line 297, in _get_current_object
    return self.__local()
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/flask/globals.py", line 18, in _lookup_req_object
    top = _request_ctx_stack.top
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/werkzeug/local.py", line 163, in top
    return self._local.stack[-1]
  File "/Users/miket/dev/compat/webcompat.com/env/lib/python2.7/site-packages/werkzeug/local.py", line 69, in __getattr__
    return self.__storage__[self.__ident_func__()][name]
RuntimeError: maximum recursion depth exceeded
```

miketaylr:
So it seems if we abort(500) at the end here, that doesn't hit custom_error_message -- which is good because it doesn't blow the recursion stack, but bad because it shows the ugly default 500 page, rather than our custom page.

Let me re-arrange some things to make this work (and be more DRY).

miketaylr:
I think we should move all the HTTP error handling into its own module/Blueprint. It's kind of messy in views.py right now. So let me hack on that and open a new PR.

Issue #713, opened 2015-09-29
Better management of Errors for webcompat API Endpoint


karlcow:
[API, error management](http://flask.pocoo.org/docs/0.10/patterns/apierrors/). Abort and flask best practices. 

* Separate the errors message in their own errors.py
* See if we really need to use `abort` or be better at handling these
Discussed in Paris - September 2015 with @miketaylr - Webcompat team meeting

karlcow:
Closing fixed by https://github.com/webcompat/webcompat.com/pull/1036

PR #1036, opened 2016-05-02
Fixes #1031. Handle non-HTTPError exceptions

### LINKED ISSUE #1031: 500 handler shouldn't assume code property

miketaylr:
And also move all error handlers into their own blueprint module.

r? @karlcow

### DIFF
--- a/webcompat/__init__.py
+++ b/webcompat/__init__.py
@@ -28,9 +28,10 @@ import webcompat.views
 # register blueprints
 from api.endpoints import api
 from api.uploads import uploads
+from error_handlers import error_handlers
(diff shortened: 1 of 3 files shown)


miketaylr:
See https://github.com/webcompat/webcompat.com/pull/1033 for previous discussion.

karlcow:
ok let's try with this.
And you just fixed #713 ;)

karlcow:
r+ @miketaylr

PR #1041, opened 2016-05-04
Fixes #1012. Downscale large images that come from the add-on.

### LINKED ISSUE #1012: Resize gigantic screenshots that come from the add-on
Right now it's possible to use the "Report Site Issue" functionality of Fennec on a Nexus 6P and get an image that is like 7MB -- the screen is sort of insane: 2560x1440 px (518 ppi)

I imagine in the future we'll see more and more of this.

It should be possible to take that image and downscale it via an off-screen canvas if it's too large.

miketaylr:
Tested on my Nexus 6P here: https://github.com/webcompat/webcompat-tests/issues/379

(before that screenshot was too big to get submitted)

r? @adamopenweb

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -15,6 +15,7 @@ function BugForm() {
   this.screenshotData = '';
   // by default, submission type is anonymous
   this.submitType = 'github-proxy-report';
+  this.UPLOAD_LIMIT = 1024 * 1024 * 4;
 
   this.inputMap = {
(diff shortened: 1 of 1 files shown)


adamopenweb:
Looks good @miketaylr thanks!
Can we make this more future proof at some point?

miketaylr:
> Can we make this more future proof at some point?

Yeah, let me file a follow up bug. Seems silly that it would fail with an 8MB screenshot, and I suspect tablets and phones will get there one day.

PR #1052, opened 2016-05-19
fix #1051 Converting PNGs to JPEG

### LINKED ISSUE #1051: Optimizing images on upload is too slow
Pillow's optimization is slow. :(

I need to measure more but right now local full-size screenshots are taking forever:

`save function took 20237.716 ms`
Commenting out the following block in upload.py:

```
# if self.image_object.format in ['JPEG', 'JPG', 'JPE', 'PNG']:
        #     save_parameters['optimize'] = False
        # # If animated GIF, aka duration > 0, add save_all parameter
        # if (self.image_object.format == 'GIF' and
        #    self.image_object.info['duration'] > 0):
        #         save_parameters['save_all'] = False
```

`save function took 1704.766 ms`

cc @karlcow

karlcow:
r? @miketaylr 

This convert the PNG to JPEG for optimizing the size and the time, and bypass the issue with PNG poor encoding performances.
This is not necessary to integrate. but it would be worth testing at least. This is a simple fix following up my tests in #1051

### DIFF
--- a/webcompat/api/uploads.py
+++ b/webcompat/api/uploads.py
@@ -86,6 +86,8 @@ class Upload(object):
         # Optimize further the image compression for these formats
         if self.image_object.format in ['JPEG', 'JPG', 'JPE', 'PNG']:
             save_parameters['optimize'] = True
+            # Convert PNG to JPEG. See issue #1051
+            file_dest = 'jpg'.join(file_dest.rsplit('png', 1))
         # If animated GIF, aka duration > 0, add save_all parameter
         if (self.image_object.format == 'GIF' and
            self.image_object.info['duration'] > 0):


webcompat/api/uploads.py
```diff
@@ -86,6 +86,8 @@ def save(self):
         # Optimize further the image compression for these formats
         if self.image_object.format in ['JPEG', 'JPG', 'JPE', 'PNG']:
             save_parameters['optimize'] = True
+            # Convert PNG to JPEG. See issue #1051
+            file_dest = 'jpg'.join(file_dest.rsplit('png', 1))
```
miketaylr: Ha, the conversion is surprisingly simple. ^_^

miketaylr:
I think we should go for it -- I'll merge and test on staging.

webcompat/api/uploads.py
```diff
@@ -86,6 +86,8 @@ def save(self):
         # Optimize further the image compression for these formats
         if self.image_object.format in ['JPEG', 'JPG', 'JPE', 'PNG']:
```
miketaylr: Oh, did we want to remove 'PNG' from the `save_parameters['optimize']` condition?

miketaylr: > Oh, did we want to remove 'PNG' from the save_parameters['optimize'] condition?

I suppose it doesn't matter if we convert everything to JPEG, so not really a big deal. Something to clean up later.

webcompat/api/uploads.py
karlcow: yes, due to pillow. Note that we could be more explicit. 
What it does here, is take the destination path, and change the `png` suffix by `jpg`. Because pillow is using the suffix for choosing the conversion format (at least it's one way it does it). If the file alreay has a `jpg`, it  doesn't change the string.

webcompat/api/uploads.py
karlcow: the save_parameters['optimize'] is useful in the case of JPEG and doesn't have a strong impact on performance. If we exclude the 'PNG', we will need to add it for making sure we are saving to JPG.

miketaylr: Got it, thanks.

PR #1060, opened 2016-05-22
Fixes #1059 - Update contributing.md to address Ubuntu specific Node.js and Virtual Environment Issues 

### LINKED ISSUE #1059: Update contributing.md for node installation and virtual environment related issues for Ubuntu
1) Node.js is a prerequisite for building webcompat locally.
After installing node, node -v doesn't work.

2) In Ubuntu, even after successful installation of the following commands 

[sudo] virtualenv env
source env/bin/activate

and completing installation of everything in requirements.txt, the user still gets the following error:

  File "run.py", line 38, in <module>
    raise ImportError('{0}\n\n{1}'.format(e, IMPORT_ERROR))
ImportError: No module named flask.ext.github

Please re-run : pip install -r requirements.txt
    
deepthivenkat:


CONTRIBUTING.md
```diff
@@ -277,8 +294,11 @@ source env/bin/activate
 #  Linux: http://pillow.readthedocs.org/en/3.0.x/installation.html#linux-installation
 # install rest of dependencies
 pip install -r requirements.txt
+#In Ubuntu, if ImportError: No module named flask.ext.github occurs, it means the dependencies in requirement.txt is installed in /usr/lib instead of <project_repository>/env/lib/python<version>/site-packages.
```
miketaylr: nit: can we get a space after `#` for this and other lines?
nit: please change to "are installed" (rather than "is installed")

s/requirement.txt/requirements.txt/

CONTRIBUTING.md
```diff
@@ -211,6 +211,23 @@ sudo apt-get update
 sudo apt-get install python2.7 python2.7-dev
 ```
 
+In Ubuntu, sometimes even after installing Node.js, the command 
+
+```
+node -v
```
miketaylr: I think here you can just use an inline code comment, like so:

"...installing Node.js, the command `node -v` does not show...`. 

To do that, you just need 1 set of backticks \`node -v\`

miketaylr:
Great, thanks!

OK, so one final step. We're not really interested in an "Correction" commit, so it would be good to rebase (or fixup) these two commits into 1. 

See https://help.github.com/articles/about-git-rebase/

What you want to do is start an interactive rebase and mark the second commit as a fixup commit:

`git rebase -i HEAD~2` (this means only consider the last 2 commits that are different from HEAD)

Then you'll mark the 2nd ("Correction") commit as a fixup commit in your editor, and save and close the editor. Once that succeeds, you need to force push the branch again:

`git push -f`

And then we can merge. :)

deepthivenkat:
Sure Mike! Thanks

- Deepthi

deepthivenkat:
Hello Mike,

I have rebased and made it a single commit. 

One small question: is it advisable to use [ci skip] with git commit in such cases since I am only editing documentation?

miketaylr:
> One small question: is it advisable to use [ci skip] with git commit in such cases since I am only editing documentation?

Yeah, I think in general that's good -- saves Travis's computers from having to do work. ^_^

PR #1078, opened 2016-06-01
Fixes #1074 - Added clear filter event to LabelSearch method to trigg…

### LINKED ISSUE #1074: Clicking on label from /issues page does not update URL params
Steps to reproduce:

Go to [issues](https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc) page. Click on a label for any of the issues. Eg : firefox-mobile

Expected Result:

In url: [https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=label%3Abrowser-firefox-mobile](url)
In search box : label:browser-firefox-mobile	
															 	

Actual Result: 

In url: [https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc](url)    **[The url is not the same as expected]**
In search box: label:browser-firefox-mobile

deepthivenkat:
…er URL change

r? @miketaylr

### DIFF
--- a/webcompat/static/js/lib/issue-list.js
+++ b/webcompat/static/js/lib/issue-list.js
@@ -394,6 +394,7 @@ issueList.IssueView = Backbone.View.extend(
       var labelFilter = 'label:' + clickedLabel;
       issueList.events.trigger('search:update', labelFilter);
       issueList.events.trigger('issues:update', {query: labelFilter});
+      issueList.events.trigger('filter:clear', {removeQ: false});
       e.preventDefault();
     },
     resetStageFilter: function(options) {


miketaylr:
Yay! So the fix was pretty simple.

In the future, when you make a PR, can you explain the patch a little bit? That helps the reviewer to get up to speed quicker.

For this bug, something like "This patch triggers `filter:clear` in `labelSearch` (which ends up trigger `filter:reset-stage` which ends up calling `updateModelParams` (which calls `updateURLParams` -- which is what we want). It has the side effect of clearing filters in a search, which is a required constraint of our set up right now."

@deepthivenkat do you intend to write an intern test for this?

miketaylr:
> @deepthivenkat do you intend to write an intern test for this?

Deepthi filed #1079 to track that.

miketaylr:
Also, thanks for the patch and bug fix! 🎈

PR #1081, opened 2016-06-04
Fixes #981 - adding a link for the labels in individual issue page

### LINKED ISSUE #981: [bug] non clickable labels
capture d ecran 2016-03-25 a 13 49 32

On issue individual pages, the labels are non clickable 

1. Go to https://webcompat.com/issues/707
2. Click on ie

Nothing happens. All the ie issues should pop up.

deepthivenkat:
r? @miketaylr

### DIFF
--- a/webcompat/static/css/development/components/labels.css
+++ b/webcompat/static/css/development/components/labels.css
@@ -1,25 +1,21 @@
 /** @define Labels */
 
-:root {
-  --Labels-colorHover: var(--base-colorDark);
-}
-
 .wc-Labels {
   display: inline-block;
-  border-width:0 0 3px 0;
-  border-style:solid;
-  font-weight:700;
-  border-color:transparent;
-  opacity:.7;
-  text-decoration:none;
+  font-size: .8rem;
+  font-weight: 700;
+  border-color: transparent;
+  opacity: .7;
+  text-decoration: none;
   color: inherit;
+  margin-left: .5em;
+  padding: 0.2em;
+  border-radius: 4px;
 }
   /* state */
   .wc-Labels:hover,
   .wc-Labels:focus {
     text-decoration:none;
+    color: inherit;
     opacity:1;
-    border-color: #414042;
-    color: var(--Labels-colorHover);
-    border-bottom-color: #FFFFFF;
   }
--- a/webcompat/static/css/development/components/labels.css
+++ b/webcompat/static/css/development/components/labels.css
@@ -21,4 +21,5 @@
     opacity:1;
     border-color: #414042;
     color: var(--Labels-colorHover);
+    border-bottom-color: #FFFFFF;
   }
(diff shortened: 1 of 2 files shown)


miketaylr:
Functionally, this is great!

I think we need to tweak the design though,

Default:
screen shot 2016-06-09 at 1 19 45 pm

Hover:
screen shot 2016-06-09 at 1 19 34 pm

We can override the opacity 0.7 that's getting inherited, I think it would be better as 1. And probably the underline can just be white instead of black.

Can you also add an Intern test to this PR, so we don't break this in a refactor at some point in the future?

magsout:
@deepthivenkat good 👍 

But I think, you can keep the default className and replace `<span>` by the <a>` like :

```html
<a 
  href="/issues?q=label:<%= label.remoteName %>"
  class="wc-Label wc-Label--badge js-Label" 
  data-remotename="<%= label.remoteName %>"
  title="Labels : <%= label.name %>"
  style="background-color:#<%=label.color%>"
>
  <%= label.name %>
</a>
```

We need to add `text-decoration: none` in `wc-Label`

miketaylr:
@deepthivenkat ping

deepthivenkat:
@miketaylr  @magsout I have to make the underline color as white. This is normally done by adding text-decoration-color property which is deprecated.

Instead I have made use of border-bottom-color property for hover event.

miketaylr:
> This is normally done by adding text-decoration-color property which is deprecated.

So it turns out this isn't deprecated, but is only supported in Firefox and Safari (with a prefix).

deepthivenkat:
Attaching a gif to show the css changes
label

r? @magsout

miketaylr:
I don't think we want the labels to change size when they're hovered over, it's sort of jarring. 

Also the gray and yellow text doesn't look very good (it's hard for me to read).

deepthivenkat:
label2

@miketaylr how about now?

miketaylr:
Much better! Why are we changing the text color from white to gray?

deepthivenkat:
animation
Now?

deepthivenkat:
@miketaylr Grey was the default color. The CSS property used for font color:

color: inherit

miketaylr:
> @miketaylr Grey was the default color. The CSS property used for font color:

Ah, gotcha. This looks better to me -- let's let @magsout review. :)

deepthivenkat:
Oki!

miketaylr:
Ah, I'm sorry -- I goofed here. We intentionally changed the color from white in https://github.com/webcompat/webcompat.com/pull/1094. >___<

deepthivenkat:
No problem! changed it back to grey

magsout:
@miketaylr @deepthivenkat 

I think he previous style is better:

http://cl.ly/170o3B2Y4122

`Hover` with opacity is much better IHMO.

I have remove, border and decrease font-size:

```css
.wc-Labels {
    display: inline-block;
    font-size: .8rem;
    font-weight: 700;
    border-color: transparent;
    opacity: .7;
    text-decoration: none;
    color: inherit;
    margin-left: .5em;
    padding: 0.2em;
    border-radius: 4px;
}
```

miketaylr:
@magsout agreed, seems nicer.

miketaylr:
Let's go ahead and merge this, thanks @deepthivenkat!

PR #1099, opened 2016-06-17
Fixes #1098. Change the path for requirements.txt

### LINKED ISSUE #1098: Change path of requirements.txt in Contributing.MD
Currently we have
pip install -r requirements.txt in the file but the path of the file is
config/requirements.txt so we need to change it accordingly.

tsl143:
#1098


miketaylr:
Ah! Sorry if this caused you troubles and thanks for the patch!

👍

miketaylr:
(in the future you can write a commit message like `Issue #1098. Change path for requirements.txt in CONTRIBUTING.md`)

PR #1106, opened 2016-06-23
Fixes #1058 - Added code for fetch first page(30 Issues) and render, …

### LINKED ISSUE #1058: Handle comment pagination on the server side
See point 1 from https://github.com/webcompat/webcompat.com/issues/1044#issuecomment-218825394.

Basic idea is to request and lump together comments until `rel=last` if we detect that there are more than 1 page of comments.

Once we have this we can remove the hack from #1044.

deepthivenkat:
…fetch remaining issues and render

1) Removed per_page=100 from model/comment.js
2) Fetch last page from model/comment.js
3) Loop over each pageNumber and fetch from model passing pageNumer as argument and render

r?@miketaylr 

I have a question here. When I fetch by passing page number - I am getting the values two times for each fetch.

I cant figure out why it is happening. Any ideas?

### DIFF
--- a/webcompat/static/js/lib/comments.js
+++ b/webcompat/static/js/lib/comments.js
@@ -7,7 +7,16 @@ var issues = issues || {};
 issues.CommentsCollection = Backbone.Collection.extend({
   model: issues.Comment,
   url: function() {
-    return '/api/issues/' + issueNumber + '/comments?per_page=100';
+    return '/api/issues/' + issueNumber + '/comments?page=' + this.defaultPage;
+  },
+  initialize: function(param) {
+    this.defaultPage = param[0].page;
+  },
+  fetchPage: function(Commentpage, options) {
+    if (Commentpage.page) {
+      this.defaultPage = Commentpage.page;
+    }
+    return this.fetch(options);
   }
 });
 
--- a/webcompat/static/js/lib/issues.js
+++ b/webcompat/static/js/lib/issues.js
@@ -361,7 +361,8 @@ issues.MainView = Backbone.View.extend({
     $(document.body).addClass('language-html');
     var issueNum = {number: issueNumber};
     this.issue = new issues.Issue(issueNum);
-    this.comments = new issues.CommentsCollection([]);
+    this.commentPage = {page: 1};
+    this.comments = new issues.CommentsCollection([this.commentPage]);
     this.initSubViews();
     this.fetchModels();
   },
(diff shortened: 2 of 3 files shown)


miketaylr:
@deepthivenkat can you push a commit fixing the eslint errors?

webcompat/static/js/lib/comments.js
```diff
@@ -7,7 +7,16 @@ var issues = issues || {};
 issues.CommentsCollection = Backbone.Collection.extend({
   model: issues.Comment,
   url: function() {
-    return '/api/issues/' + issueNumber + '/comments?per_page=100';
+    return '/api/issues/' + issueNumber + '/comments?page=' + this.defaultPage;
+  },
+  initialize: function(param) {
+    this.defaultPage = param[0].page;
+  },
+  fetchPage: function(Commentpage, options) {
```
miketaylr: The case for `Commentpage` here is probably better as `commentPage` (typically in JS something starting with a capital should be a constructor function).

webcompat/static/js/lib/issues.js
```diff
@@ -411,16 +412,17 @@ issues.MainView = Backbone.View.extend({
 
       // If there are any comments, go fetch the model data
       if (this.issue.get('commentNumber') > 0) {
-        this.comments.fetch(headersBag).success(_.bind(function() {
-          this.addExistingComments();
-          this.comments.bind('add', _.bind(this.addComment, this));
-
+        var that = this;
```
miketaylr: Is this needed? We should be able to use `_.bind` instead (and it seems like fetch success handler is already bound to the outer `this`.

webcompat/static/js/lib/issues.js
```diff
@@ -411,16 +412,17 @@ issues.MainView = Backbone.View.extend({
...
+        this.comments.fetch(headersBag).success(_.bind(function(InitialResponse) {
```
miketaylr: Same here for the function argument => `initialResponse`. `response` might also be OK?

webcompat/static/js/lib/issues.js
```diff
@@ -436,6 +438,24 @@ issues.MainView = Backbone.View.extend({
         wcEvents.trigger('flash:error', {message: msg, timeout: 4000});
       }
     });
+    var i;
+    for (i = 2; i <= this.lastPageNumber; i++) {
```
miketaylr: You can move `i` initialization inside the for loop => `for (var i = 2....`

Can you also add a comment why you're starting with 2?

webcompat/static/js/lib/models/comment.js
```diff
@@ -25,5 +25,38 @@ issues.Comment = Backbone.Model.extend({
       createdAt: moment(response.created_at).fromNow(),
       rawBody: response.body
     });
+    if (this.parseHeader(jqXHR.xhr.getResponseHeader('Link')).last) {
+      response.lastPageNumber = this.parseHeader(jqXHR.xhr.getResponseHeader('Link')).last.split('\?page\=')[1];
+    }
+  },
+  parseHeader: function(linkHeader) {
```
miketaylr: We should probably abstract this method out into a mixin, but we can do that in a follow up commit. Can you file a bug for that? And put a `//TODO: blah blah see Issue #XXXX.` comment as well?

That way we don't end up making changes in one of the models, but not the other.

webcompat/static/js/lib/issues.js
```diff
@@ -436,6 +438,24 @@ issues.MainView = Backbone.View.extend({
...
+      this.commentPage.page = i;
```
miketaylr: If this property is only used in the `fetchPage` call below, can we just do `{page: i...}`?

webcompat/static/js/lib/comments.js
```diff
@@ -7,7 +7,16 @@ var issues = issues || {};
 issues.CommentsCollection = Backbone.Collection.extend({
   model: issues.Comment,
   url: function() {
-    return '/api/issues/' + issueNumber + '/comments?per_page=100';
+    return '/api/issues/' + issueNumber + '/comments?page=' + this.defaultPage;
+  },
+  initialize: function(param) {
```
miketaylr: nit: can we rename to `params`?

miketaylr:
@deepthivenkat please re-open this PR from your fork of webcompat.com (ideally with review feedback addressed), thanks!

PR #1107, opened 2016-06-25
Fixes #1088. Add pointer-events:none to font icons.

### LINKED ISSUE #1088: search using search icon from /issues page seems unresponsive
While troubleshooting one of the broken automated functional tests (authenticated search, "Search Works"), I'm seeing an unresponsive search button.
(update this is a Chrome issue only, works on Firefox); tested manually with Chrome50 and with 
[Chromedriver v2.22 for OS X] (http://chromedriver.storage.googleapis.com/index.html?path=2.22/) with the intern tests

Context: this issue was discovered while adding Chrome testing support to intern.js (see issue #980)

To reproduce:
* Click All Issues
* Enter search terms (Search by keyword input)
* Click the search icon (far right)
* The list doesn't update. (if you click the Return key, the URL gets updated with additional search parameters and the list does update)

I'm not sure if this is related to issue #1023. 

These are a few of the files I looked at to see what may be the cause:
* https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/issue-list.js
  * the section around line 166 looked OK in terms of the event trigger but it didn't seem as though the event handler got called
* https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/models/query-params.js
* https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/models/issue.js

for the files under /models, I couldn't easily detect what might be wrong. but it seemed like one of those triggers was getting called for the successful keyword search (using Return instead of the search button).

also I don't know if this is useful info but after a successful keyword search (clicking Return after entering terms)... I also then tried clicking the search icon even though this is not a usual use case. The result was a console error like:
```
jquery-1.11.2.min.js:4 GET http://localhost:5000/api/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=vladvlad 404 (NOT FOUND)
```

although if I manually create a url like "http://localhost:5000/api/issues/search?q=vladvlad", then I seem to get the expected JSON results for the record in search. hope this helps and let me know if I should do anything else. troubleshooting search might be a little over my head at this moment.

miketaylr:
This is to prevent them from consuming the click event in Chrome,
which apparently doesn't bubble up to the parent `<button>`?

r? @magsout 

cc @cch5ng

### DIFF
--- a/webcompat/static/css/development/components/icon-font.css
+++ b/webcompat/static/css/development/components/icon-font.css
@@ -21,6 +21,10 @@
   vertical-align:middle;
 }
 
+.wc-Icon:not(button) {
+  pointer-events: none;
+}
+
 .wc-Icon--cloud-download:before {
   content: "\f0ed";
 }


miketaylr:
Whoops, the label editor icon is a `<button>` with `wc-Icon` class. Let me try something else.

magsout:
r+ 👍

PR #1119, opened 2016-07-06
Fixes #1058 - Fetching and rendering all comments

### LINKED ISSUE #1058: Handle comment pagination on the server side

deepthivenkat:
1) Fetch and Render page 1 (30 comments)
2) Get last page number from response header of page 1
3) Fetch and render remaining comments from page 2 to last page

### DIFF
--- a/webcompat/static/js/lib/comments.js
+++ b/webcompat/static/js/lib/comments.js
@@ -7,7 +7,16 @@ var issues = issues || {};
 issues.CommentsCollection = Backbone.Collection.extend({
   model: issues.Comment,
   url: function() {
-    return '/api/issues/' + issueNumber + '/comments?per_page=100';
+    return '/api/issues/' + issueNumber + '/comments?page=' + this.param;
+  },
+  initialize: function(param) {
+    this.param = param[0].page;
+  },
+  fetchPage: function(commentPage, options) {
+    if (commentPage.page) {
+      this.param = commentPage.page;
+    }
+    return this.fetch(options);
   }
 });
 
--- a/webcompat/static/js/lib/comments.js
+++ b/webcompat/static/js/lib/comments.js
@@ -7,14 +7,14 @@ var issues = issues || {};
 issues.CommentsCollection = Backbone.Collection.extend({
   model: issues.Comment,
   url: function() {
-    return '/api/issues/' + issueNumber + '/comments?page=' + this.param;
+    return '/api/issues/' + issueNumber + '/comments?page=' + this.pageNumber;
   },
-  initialize: function(param) {
-    this.param = param[0].page;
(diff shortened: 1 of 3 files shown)


webcompat/static/js/lib/issues.js
```diff
@@ -371,7 +371,8 @@ issues.MainView = Backbone.View.extend({
     $(document.body).addClass('language-html');
     var issueNum = {number: issueNumber};
     this.issue = new issues.Issue(issueNum);
-    this.comments = new issues.CommentsCollection([]);
+    this.commentPage = {page: 1};
+    this.comments = new issues.CommentsCollection([this.commentPage]);
```
miketaylr: I'm confused why we're passing in an array here, and then over in comments.js, we're just grabbing the 0th index of that. So why do we need an array at all?

Also, `this.commentPage` isn't used anywhere else, so why not just pass in the object directly to the collections, i.e., 

`this.comments = new issues.CommentsCollection({page: 1});`

webcompat/static/js/lib/comments.js
```diff
@@ -7,7 +7,16 @@ var issues = issues || {};
 issues.CommentsCollection = Backbone.Collection.extend({
   model: issues.Comment,
   url: function() {
-    return '/api/issues/' + issueNumber + '/comments?per_page=100';
+    return '/api/issues/' + issueNumber + '/comments?page=' + this.param;
+  },
+  initialize: function(param) {
+    this.param = param[0].page;
```
miketaylr: See my comment below, `param[0].page` should just be `param.page` (unless there's some reason I don't understand why this needs to be an array).

miketaylr:
This looks good!

I think we can simplify some things further (and rename some stuff so it's more clear, e.g. `param` -> `pageNumber`.

Here's a diff of how we can make this a bit easier to read/understand: https://gist.github.com/miketaylr/01b9f090dda8c6e928d40c7eb08081c5

miketaylr:
(Also, I don't see the double download issue anymore -- cool that you fixed it!)

screen shot 2016-07-11 at 10 42 01 am

deepthivenkat:
@miketaylr  I have added the review changes

miketaylr:
@deepthivenkat can you please change the commit message to match conventions (or rebase/fixup it into the previous commit)? (there's also a typo "reiew comments")

deepthivenkat:
Done @miketaylr

miketaylr:
Looks great, thanks!

Issue #980, opened 2016-03-24
move functional tests to BrowserStack (+ Chrome testing)


cch5ng:
(update 02.18.17) checklist for this isssue
* [ ] 1 either clean up current intern tests which are failing (or at least know clearly which tests are currently failing as expected, locally and on travis)
* [ ] 2 update travis and intern.js configurations to run automated browser tests on BrowserStack.com (as of 07.2016 configurations were working for firefox, chrome on BrowserStack but was having some challenge configuring separate scripts to run different configurations)
* [ ] 3 make better sense of current failing tests resulting on BrowserStack (step 1 should greatly help with this)
* [ ] 4troubleshooting 2 functional authenticated search tests (one is done)
* [x] 5 add a thank you to BrowserStack to the webcompat.com site (and README), per their request from sales rep for giving free test accounts to webcompat
  * my main contacts from 07.2016 were Ady (Adithya Chadalawada) I believe in sales/marketing role and Ashwin from support side
* [x] 6 move the BrowserStack account from my personal email to a general webcompat email account (so I don't become a bottleneck in the future)

(updated 06.13) this issue/automated test enhancement requires
- updating travis and intern.js configurations to run automated browser tests on BrowserStack.com
- troubleshooting 2 functional authenticated search tests (one is done)
- adding a thank you to BrowserStack to the webcompat.com site (and README), per their request from sales rep for giving free test accounts to webcompat

some of the discussion of requirements is in [this old PR](https://github.com/webcompat/webcompat.com/pull/1026)

(old March 2016) hi (I'm one of the outreachy applicants). while I was testing css changes for a PR, I decided to check with Browser Stack whether they would offer a free testing account for this project. one of their sales reps replied yes as long as there would be a mention of Browser Stack on the project github page and website.

I was thinking there would be impacts to these files: about page/view, README.md, and CONTRIBUTING.md. I think other people who need to test the UI would be able to request an account for testing in the future. it seems handiest for IE testing (older versions). 

does this sound OK? hope this is the right place to ask.

for ref: http://www.browserstack.com/
https://www.browserstack.com/pricing  (scroll to free for open source)

miketaylr:
Hi @cch5ng! 

This would be cool. Currently we run our Intern tests on a Firefox instance via Travis CI (which has a built-in Firefox plugin). But more browser coverage would be awesome.

(It would also require figuring out how to get the Intern tests to run there, and how that interacts with Travis).

magsout:
[...]

PR #1124, opened 2016-07-07
[WIP] #980 fork browserstack2 update intern/travis configs for chrome…

cch5ng:
(commit comment cont'd)… testing
separate intern configs by test env
add npm run scripts by test env
add timeout to login helper for manual github 2fa auth code entry
test browserstack sup fix for firefox breaking test issues-non-auth.js
update intern version because of browserstack tunnel err

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,10 +28,18 @@ env:
     - FAKE_ID=e80a666fbf1fa9ea18db
     - FAKE_SECRET=688c4546f09624f8c44773b22268064dfca19a59
     - TRAVIS_NODE_VERSION="0.12.7"
+    # BrowserStack (SELENIUM_HOST, SELENIUM_PORT, BS_AUTOMATE_PROJECT, BS_AUTOMATE_BUILD, secure - 2 lines)
+    - SELENIUM_HOST=hub.browserstack.com
+    - SELENIUM_PORT=80
+    - BS_AUTOMATE_PROJECT="$TRAVIS_REPO_SLUG"
+    - BS_AUTOMATE_BUILD="Travis build No. $TRAVIS_BUILD_NUMBER for $TRAVIS_REPO_SLUG"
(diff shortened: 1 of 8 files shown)


cch5ng:
@miketaylr hi, I have one known issue on travis-browserstack as of last test before 070416 (firefox pressing 'g' to redirect page). I'm not sure about multiple local test failures (both firefox and chrome). but I think this is mainly what is necessary for the browserstack configs.

need to update CONTRIBUTING.md but these are new test commands:
`npm test user='a user' pw='a pwd' ` (for local firefox all)
`npm run test-chrome -s user='a user' pw='a pwd' ` (for local chrome all)
`npm run test-functional-nonauth -s` (for local firefox non auth)

cch5ng:
@miketaylr for the script conditions at the bottom of .travis.yml should I expect that

if [ "$TRAVIS_SECURE_ENV_VARS" == "false" ] ===> this case should run the tests on travis/firefox?
if [ "$TRAVIS_SECURE_ENV_VARS" == "true" ] ===> this case should run the tests on browserstack?

cch5ng:
I need to fix the travis script pointer for non auth tests

also need to update configs for new selenium, firefox versions in readme

cch5ng:
closing b/c this was a test to verify non auth test trigger
also needs more changes

cch5ng:
> for the script conditions at the bottom of .travis.yml should I expect that
> if [ "$TRAVIS_SECURE_ENV_VARS" == "false" ] ===> this case should run the tests on travis/firefox?

I will double check w/ browserstack support what the expectation should be. I thought the non auth tests should run on browserstack but did not observe this in earlier logs.

miketaylr:
> I thought the non auth tests should run on browserstack but did not observe this in earlier logs.
 
Hm, that's what I would expect as well.

cch5ng:
> I thought the non auth tests should run on browserstack but did not observe this in earlier logs.

> Hm, that's what I would expect as well.

found: it's a travis policy not to transmit encrypted data for fork PR's so the browserstack tunnel is not getting the browserstack auth credentials. 

https://docs.travis-ci.com/user/pull-requests#Security-Restrictions-when-testing-Pull-Requests
https://docs.travis-ci.com/user/encryption-keys/#Usage

(2nd link is what was used to encrypt the browserstack auth info)

but as with the last fork test, the local firefox intern config can be used. so tests will still get run (on travis but not on chrome).

so pushes from webcompat members will go to browserstack (including the non auth tests) but community members' PR's will only go to travis (is that OK?). sorry if it was not worded clearly initially but I think you had alluded to this, I just didn't understand why this was happening.

Issue #719, opened 2015-09-30
Log issues submission for tracking abuse


karlcow:
log the info via flask about issues submission 
* Timestamp
* URI
* IP address

Decided in Paris. WebCompat meeting - September 2015

karlcow:
For this to happen, I need first to solve the deletion of logs. So depends on #717

miketaylr:
ping @karlcow

PR #1125, opened 2016-07-08
#719 Adds tracking ip logs

karlcow:
Not putting a Fix on this pull request. 
I would like if possible to deploy this in staging for testing and be sure we do the right thing (specifically with proxies in the middle).

My guts is that more than one way of doing it. This one is just very basic. Maybe it's enough for what we want now, but could be revisited later.


miketaylr:
r+

Issue #1123, opened 2016-07-07
Investigate recent Travis errors


miketaylr:
`11/67 tests failed`

A recent example: https://travis-ci.org/webcompat/webcompat.com/builds/142939936

These tests pass for me locally, so something is up.

miketaylr:
Bumping Firefox and Selenium versions to the latest --- I suspect that a bug that had us on old versions was fixed by Travis and uncovered some racy or bad tests (maybe).

miketaylr:
Our most recent run just has 1/67 failing... `FAIL: main - Issue-list - Loading URL with stage param loads issues (10401ms).`

That makes me wonder if the failures from a few days ago weren't just GitHub issues.

miketaylr:
Might as well add the missing fixtures that make reading Travis logs messy.

miketaylr:
So... Travis is green again. Let's close this and open a new bug if things get wacky.

PR #1129, opened 2016-07-12
Issue #1123. Add missing fixture files.

miketaylr:
This *seems* to have made Travis happy: https://travis-ci.org/webcompat/webcompat.com/builds/144041240. But that's probably just luck.

At any rate, this will make reading Travis CI logs easier, since we don't need to see the stack traces from 500s that are just caused by non-essential fixture files being missing.

r? @karlcow


karlcow:
r+ 

When I see these fixtures, I have the feeling we need to do something different. But I guess :) it is working for now. Balance in between work to do and perfect. 🚡

karlcow:
Merci @miketaylr

miketaylr:
Thanks!

> When I see these fixtures, I have the feeling we need to do something different.

Yep, lots of room for improvement here.

PR #1152, opened 2016-08-02
Fixes #1139. Add a loginDelay commandline argument.

### LINKED ISSUE #1139: update functional test config for login helper (shorten login timeout)
in /lib/helper.js, a long timeout was added to the login function to allow for local test manual entry of the github auth code. the timeout should be configurable.

> it should be possible to create some kind of param that people can pass into the config like LOGIN_TIMEOUT=10 (that defaults to 0`).

miketaylr:
r? @cch5ng

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,15 +77,15 @@ project.
 
 All code contributions should come in the form of a [pull request](https://help.github.com/articles/creating-a-pull-request), as a topic branch.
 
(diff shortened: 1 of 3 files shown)


cch5ng:
@miketaylr LGTM. I did not test it locally, should I?

miketaylr:
@cch5ng nah, no need -- I've tested it a number of times. :)

And looks like I need to fix some merge conflicts.

Issue #1145, opened 2016-07-28
Create method to backup issues.db


miketaylr:
This can be a command line argument to `run.py`: https://github.com/webcompat/webcompat.com/blob/master/run.py#L93

Something like `python run.py --backup`. This should:

Copy the existing issues.db to a folder that lives at `$DB_BACKUP_LOCATION` (or similar, somewhere configurable via secrets.py).

deepthivenkat:
@miketaylr  I have a question here. Before moving the issues.db to a backup location that $DB_BACKUP_LOCATION, I need to verify if issues.db exists.

Should I check if issues.db exists in webcompat server / just in the current directory(where ever the code is executed) ?

miketaylr:
> Should I check if issues.db exists in webcompat server / just in the current directory(where ever the code is executed) ?

just in the current directory is fine.

> Should I ssh to a host and check if issues.db exists there and also send the backup folder to the host?

Which host would this be?

miketaylr:
This is merged now!

PR #1157, opened 2016-08-07
Issue #1145 - Move issues.db to backup location configured in secrets.py

deepthivenkat:


.gitignore
```diff
@@ -12,6 +12,10 @@ uploads/
 .idea/
 config/secrets.py
 
+# backup folder contains the issues.db backup file with issues dump
+#for the previous version of schema
```
miketaylr: nit: add a space between `#` and `for`

config/secrets.py.example
```diff
@@ -24,6 +24,19 @@ elif LOCALHOST:
     UPLOADS_DEFAULT_DEST = BASE_DIR + '/uploads/'
     UPLOADS_DEFAULT_URL = 'http://localhost:5000/uploads/'
 
+
+# Database backup path. PRODUCTION and STAGING can be ignored for
+# local testing.
+if PRODUCTION:
+    BACKUP_DEFAULT_DEST = ''
+    BACKUP_DEFAULT_URL = ''
```
miketaylr: `BACKUP_DEFAULT_URL` isn't used, right? 
(And I imagine we don't want to expose this to the web)

miketaylr:
Note to self: it shouldn't be possible to download this backup over the web.

Double check on staging that this is the case (it should be, according to https://github.com/webcompat/webcompat.com/blob/master/CONTRIBUTING.md#production-server-setup)

config/secrets.py.example
```diff
@@ -24,6 +24,19 @@ elif LOCALHOST:
...
+elif STAGING:
+    BACKUP_DEFAULT_DEST = ''
+    BACKUP_DEFAULT_URL = ''
+elif LOCALHOST:
```
miketaylr: What about:

```
if LOCALHOST:
    BACKUP_DEFAULT_URL = 'http://localhost:5000/backup/'
else:
    BACKUP_DEFAULT_URL = ''
```

That way staging and prod are as close as possible (and to minimize the amount of configuration here).

run.py
```diff
@@ -90,13 +103,20 @@ def config_validator():
     if app.config['OAUTH_TOKEN'] == '':
         sys.exit(TOKEN_HELP)
 
+def execute_command(command):
+    #pattern, filename = 'test', 'tmp'
```
miketaylr: Is this just a debug comment?

run.py
```diff
@@ -106,6 +126,19 @@ def config_validator():
         app.config['TESTING'] = True
         print("Starting server in ~*TEST MODE*~")
         app.run(host='localhost')
+    elif args.backup:
+        try:
+            command = ['ls', '-l', 'issues.db']
```
miketaylr: Can we use https://docs.python.org/2/library/os.path.html#os.path.exists instead? That way it will work cross-platform.

run.py
```diff
@@ -106,6 +126,19 @@ def config_validator():
...
+            if execute_command(command)[1] == 0:
+                execute_command(['mkdir', '-p', app.config['BACKUP_DEFAULT_DEST']])
```
miketaylr: Same thing here, using the `os` module. 

https://docs.python.org/2/library/os.html#os.mkdir

(And for mv as well below)

run.py
```diff
@@ -106,6 +126,19 @@ def config_validator():
...
+                current_ts = time.time()
+                time_stamp = datetime.datetime.fromtimestamp(current_ts).strftime('%Y-%m-%d-%H:%M:%S')
+                issue_backup_db = 'issues_backup_' + time_stamp + '.db'
```
miketaylr: As is, this will keep all backups.

Maybe we could keep the previous 2 or something? (So we have 3 total)?

run.py
```diff
@@ -106,6 +126,19 @@ def config_validator():
...
+                execute_command(['mv', 'issues.db', app.config['BACKUP_DEFAULT_DEST'] + issue_backup_db])
+            else:
+                print FILE_NOT_FOUND_ERROR
+        except OSError, e:
+            raise OSError('{0}\n\n{1}'.format(e, FILE_NOT_FOUND_ERROR))
```
miketaylr: We can probably just `sys.exit(FILE_NOT_FOUND_ERROR)` if we didn't find the file using os.path.exists.

miketaylr:
Can we squash the "fixing merge conflicts" commit into the first one?

Thanks!

run.py
```diff
@@ -106,6 +121,39 @@ def config_validator():
...
+                                if idx < no_of_backup_files - 2:
+                                    os.remove(app.config['BACKUP_DEFAULT_DEST'] + val)
+                    current_ts = time.time()
+                    time_stamp = datetime.datetime.fromtimestamp(current_ts).strftime('%Y-%m-%d-%H:%M:%S')
+                    issue_backup_db = 'issues_backup_' + time_stamp + '.db'
+                    os.rename(os.getcwd() + '/issues.db', app.config['BACKUP_DEFAULT_DEST'] + issue_backup_db)
+                    try:
+                        os.remove(os.path.join(os.getcwd(), 'issues.db'))
+                    except OSError:
+                        print 'Backup issues.db complete'
```
deepthivenkat: @miketaylr  I get a weird issue here. Even though I try multiple times to verify that issues.db is not available in /Users/username/webcompat.com, I get true for line number 129
`if os.path.isfile(os.path.join(os.getcwd(), 'issues.db')):`
 when I rerun the script.

Even in try[line 131]:
` with open(os.path.join(os.getcwd(), 'issues.db')) as file:`
I have placed an additional check to open issues.db which passes. 

I am not able to figure out why this happens.

miketaylr: This sounds like a bug. `Both os.path.exists and os.path.isfile returns true even when file is not there in the path` is sort of a red flag.

```
$ touch file.txt
$ python
Python 2.7.12 (default, Jun 29 2016, 14:05:02) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.exists('file.txt')
True
>>> os.path.exists('file.lol')
False
>>> 
```

run.py
```diff
@@ -97,6 +110,8 @@ def config_validator():
     parser = argparse.ArgumentParser()
     parser.add_argument('-t', '--testmode', action='store_true',
                         help='Run server in "test mode".')
+    parser.add_argument('--backup', action='store_true',
+                        help='backup existing issues.db and update db schema and issues dump.')
```
miketaylr: One minor thing: `update db schema and issues dump` -- This should just be a command to create a backup, not do any schema updates.

config/secrets.py.example
```diff
@@ -24,6 +24,14 @@ elif LOCALHOST:
     UPLOADS_DEFAULT_DEST = BASE_DIR + '/uploads/'
     UPLOADS_DEFAULT_URL = 'http://localhost:5000/uploads/'
 
+
+# Database backup path. PRODUCTION and STAGING can be ignored for
```
miketaylr: There is no more PRODUCTION and STAGING as it relates to the backup dest, so this comment needs to be updated.

run.py
```diff
@@ -106,6 +121,39 @@ def config_validator():
...
         app.run(host='localhost')
+    elif args.backup:
+        # To verify if issues.db exist, simply calling os.path.exists is not enough.
+        # Both os.path.exists and os.path.isfile returns true even when file is not there in the path
+        # Every time the script is re run, os.path.exists returns true because of cache.
+        # To avoid this, we open the file in try catch - just to be sure
+        if os.path.isfile(os.path.join(os.getcwd(), 'issues.db')):
+            try:
+                with open(os.path.join(os.getcwd(), 'issues.db')) as file:
+                    if not os.path.exists(app.config['BACKUP_DEFAULT_DEST']):
```
karlcow: When the `if` passes, aka the file exists. The `try` becomes useless with regards to the `with open` because the file exists.

run.py
```diff
@@ -122,36 +125,32 @@ def config_validator():
...
+            issue_db = scoped_session(issue_session_maker)
+            # Take a backup if issues.db has data dump.
+            if issue_db().execute('select count(*) from webcompat_issues').fetchall()[0][0] > 0:
+                if not os.path.exists(app.config['BACKUP_DEFAULT_DEST']):
+                    print 'Backup db folder does not exist'
+                    os.makedirs(app.config['BACKUP_DEFAULT_DEST'])
+                else:
+                    print 'Backup db folder exists'
+                current_ts = time.time()
+                time_stamp = datetime.datetime.fromtimestamp(current_ts).strftime('%Y-%m-%d-%H:%M:%S')
```
karlcow: No need to import `time` and then to `datetime`. It is possible to get the now time from datetime directly

time stamps with colon for filename might not be the best idea.

```python
>>> import datetime
>>> datetime.datetime.now()
datetime.datetime(2016, 8, 18, 11, 39, 30, 422410)
>>> datetime.datetime.now().isoformat()
'2016-08-18T11:39:36.208304'
>>> datetime.datetime.now().strftime('%Y-%m-%d-%H:%M:%S')
'2016-08-18-11:40:47'
```

I would do instead with the `datetime` module

```python
>>> datetime.datetime.now().strftime('%Y-%m-%dT%H-%M-%S')
'2016-08-18T11-41-58'
```

Or if you really prefer the `time` module

```python
>>> time.strftime('%Y-%m-%dT%H-%M-%S', time.gmtime())
'2016-08-18T02-45-25'
```

The only difference is the TZ of the computer. Local or UTC time.

karlcow: also for local time

```python
>>> time.strftime('%Y-%m-%dT%H-%M-%S', time.localtime())
'2016-08-18T11-47-55'
```

run.py
```diff
@@ -122,36 +125,32 @@ def config_validator():
...
+                issue_backup_db = 'issues_backup_' + time_stamp + '.db'
+                os.rename(os.getcwd() + '/issues.db', app.config['BACKUP_DEFAULT_DEST'] + issue_backup_db)
+                # Retain last 3 recent backup files
+                no_of_backup_files = len(os.listdir(app.config['BACKUP_DEFAULT_DEST']))
+                print no_of_backup_files
+                if no_of_backup_files > 3:
+                    for idx, val in enumerate(os.listdir(app.config['BACKUP_DEFAULT_DEST'])):
+                        print idx
```
karlcow: this seems to be a remaining print of local testing.

run.py
```diff
@@ -122,36 +125,32 @@ def config_validator():
...
+                    print 'Backup db folder does not exist'
+                    os.makedirs(app.config['BACKUP_DEFAULT_DEST'])
+                else:
+                    print 'Backup db folder exists'
+                current_ts = time.time()
+                time_stamp = datetime.datetime.fromtimestamp(current_ts).strftime('%Y-%m-%d-%H:%M:%S')
+                issue_backup_db = 'issues_backup_' + time_stamp + '.db'
+                os.rename(os.getcwd() + '/issues.db', app.config['BACKUP_DEFAULT_DEST'] + issue_backup_db)
+                # Retain last 3 recent backup files
+                no_of_backup_files = len(os.listdir(app.config['BACKUP_DEFAULT_DEST']))
```
karlcow: from the documentation of `os.listdir`

> Return a list containing the names of the entries in the directory given by path. **The list is in arbitrary order.** It does not include the special entries '.' and '..' even if they are present in the directory.

There's no guarantee that the list you are building will be in time order or alphabetical order. It's very likely but not mandatory. It's also not necessary to get the length of the list. It's often a sign of a JavaScriptism :) List are enumerable properties.

run.py
```diff
@@ -122,36 +125,32 @@ def config_validator():
...
+                print no_of_backup_files
+                if no_of_backup_files > 3:
+                    for idx, val in enumerate(os.listdir(app.config['BACKUP_DEFAULT_DEST'])):
+                        print idx
+                        if idx < no_of_backup_files - 3:
+                            os.remove(app.config['BACKUP_DEFAULT_DEST'] + val)
```
karlcow: ok let's see a better way to do that. Let's say we have this list of files.

```
→ ls -1
db2016-08-17T01-03-20
db2016-08-17T05-02-20
db2016-08-18T01-03-20
db2016-08-18T04-05-06
db2016-08-18T14-03-01
db2016-08-18T14-03-20
```

And the list is: 
```python
>>> import os
>>> files_list = os.listdir('.')
>>> # And let's say the files_list comes into a different order
>>> files_list
['db2016-08-17T01-03-20', 'db2016-08-18T14-03-01', 'db2016-08-18T14-03-20', 'db2016-08-17T05-02-20', 'db2016-08-18T01-03-20', 'db2016-08-18T04-05-06']
>>> # Let's do a sort so the files will now be in alphabetical order
>>> files_list.sort()
>>> files_list
['db2016-08-17T01-03-20', 'db2016-08-17T05-02-20', 'db2016-08-18T01-03-20', 'db2016-08-18T04-05-06', 'db2016-08-18T14-03-01', 'db2016-08-18T14-03-20']
>>> # the oldest are at the beginning and the most recent ones at the end of the list
>>> # we want to keep the most recent one. So we need the list of files to remove.
>>> # This will cut the list by removing the 3 last items in the list
>>> files_list[:3]
['db2016-08-17T01-03-20', 'db2016-08-17T05-02-20', 'db2016-08-18T01-03-20']
>>> # so basically you can do 
>>> for old_file in files_list[:3]:
...     os.remove(old_file)
... 
```

Let's see in our directory:

```
→ ls -1
db2016-08-18T04-05-06
db2016-08-18T14-03-01
db2016-08-18T14-03-20
```

run.py
```diff
@@ -139,18 +138,13 @@ def config_validator():
...
                 os.rename(os.getcwd() + '/issues.db', app.config['BACKUP_DEFAULT_DEST'] + issue_backup_db)
                 # Retain last 3 recent backup files
-                no_of_backup_files = len(os.listdir(app.config['BACKUP_DEFAULT_DEST']))
-                print no_of_backup_files
-                if no_of_backup_files > 3:
-                    for idx, val in enumerate(os.listdir(app.config['BACKUP_DEFAULT_DEST'])):
-                        print idx
-                        if idx < no_of_backup_files - 3:
-                            os.remove(app.config['BACKUP_DEFAULT_DEST'] + val)
+                backup_files = os.listdir(app.config['BACKUP_DEFAULT_DEST'])
```
karlcow: @deepthivenkat you need to sort the list. See my earlier comment.

run.py
```diff
@@ -139,18 +138,13 @@ def config_validator():
...
+                for old_files in backup_files[:-3]:
```
karlcow: nit: `old_file` because the for loop is about `for one item of this list`.

run.py
```diff
@@ -27,6 +32,16 @@
...
 '''
 
+FILE_NOT_FOUND_ERROR = '''
+==============================================
+The issues.db file seems to be missing.
+
+Please create a issues.db file by running:
+
+python run.py
+==============================================
```
karlcow: How does it work? 
To access this message we need to run `python run.py` and the message is telling us the same thing.

run.py
karlcow: +1 message to modify to be simpler.

run.py
```diff
@@ -106,6 +123,30 @@ def config_validator():
...
+                                               autoflush=False,
+                                               bind=issue_engine)
+            issue_db = scoped_session(issue_session_maker)
+            # Take a backup if issues.db has data dump.
+            if issue_db().execute('select count(*) from webcompat_issues').fetchall()[0][0] > 0:
+                if not os.path.exists(app.config['BACKUP_DEFAULT_DEST']):
+                    print 'Backup db folder does not exist'
+                    os.makedirs(app.config['BACKUP_DEFAULT_DEST'])
+                else:
+                    print 'Backup db folder exists'
```
karlcow: The message `Backup db folder does not exist` is confusing here. Indeed the folder doesn't exist, but the message should say: "Creating backup directory" and the `else` is not necessary. No need to tell the users that the backup folder exists. This is supposed to run only once when the user install for the first time or has cleaned up a bit its repo.

run.py
```diff
@@ -106,6 +123,30 @@ def config_validator():
...
+    elif args.backup:
+        if os.path.isfile(os.path.join(os.getcwd(), 'issues.db')):
+            issue_engine = create_engine('sqlite:///' + os.path.join(
+                app.config['BASE_DIR'], 'issues.db'))
+            issue_session_maker = sessionmaker(autocommit=False,
+                                               autoflush=False,
+                                               bind=issue_engine)
+            issue_db = scoped_session(issue_session_maker)
+            # Take a backup if issues.db has data dump.
+            if issue_db().execute('select count(*) from webcompat_issues').fetchall()[0][0] > 0:
```
karlcow: I'm not sure why we need this here. It doesn't seem to be used.

run.py
```diff
@@ -106,6 +123,30 @@ def config_validator():
...
+                if not os.path.exists(app.config['BACKUP_DEFAULT_DEST']):
+                    print 'Backup db folder does not exist'
+                    os.makedirs(app.config['BACKUP_DEFAULT_DEST'])
+                else:
+                    print 'Backup db folder exists'
+                time_stamp = time.strftime('%Y-%m-%dT%H-%M-%S', time.localtime())
+                issue_backup_db = 'issues_' + time_stamp + '.db'
+                os.rename(os.getcwd() + '/issues.db', app.config['BACKUP_DEFAULT_DEST'] + issue_backup_db)
```
karlcow: It doesn't seem necessary to move the file to another place. After this operation issues.db will not exist. Which mean it would require to create it again, but that's not part of this script either. :) Or maybe I missed a step.

deepthivenkat: The [line #44](https://github.com/deepthivenkat/webcompat.com/blob/1e6f8c3ef4f590342ba36db38b622baa329c9f04/run.py#L44) gets executed irrespective of the arguments given to run.py.

This line recreates issues.db with old schema. Now I need to write code to read alter commands for webcompat_issues table from a configured location and execute all of them. Then we will have the updated schema. I will write for this code in a separate issue.

Since issues.db gets recreated every time the python script is run, I am removing the check if issues.db file exists. I am also removing the FILE_NOT_FOUND error since the case will never happen

run.py
deepthivenkat: Here we are checking for the issues data dump in the issues.db file. If there is a data dump, we are taking a backup of the issues.db file. Else the issues.db is a freshly generated file for the old schema. We are ready to execute alter statements and [change it to new schema](https://github.com/webcompat/webcompat.com/pull/1157#discussion_r75314190).

deepthivenkat: The code to create data dump is also to be implemented in a separate issue. I am adding that as a TODO for now

run.py
karlcow: > Then we will have the updated schema. I will write for this code in a separate issue.

We decided to not do that if I remember correctly. See my mail
https://groups.google.com/forum/#!topic/mozilla.compatibility/f7tJmi9IB40

run.py
```diff
@@ -106,6 +114,26 @@ def config_validator():
...
         print("Starting server in ~*TEST MODE*~")
         app.run(host='localhost')
+    elif args.backup:
+        issue_engine = create_engine('sqlite:///' + os.path.join(app.config['BASE_DIR'], 'issues.db'))
+        issue_session_maker = sessionmaker(autocommit=False,
+                                           autoflush=False,
+                                           bind=issue_engine)
+        issue_db = scoped_session(issue_session_maker)
+        # Take a backup if issues.db has data dump.
+        # TODO: write code for creating issues dump in the issues.db with updated schema
```
karlcow: See https://github.com/webcompat/webcompat.com/issues/1160#issuecomment-241318273

karlcow:
it's in a lot better shape. Thanks @deepthivenkat 
And could be almost finished except for a couple of things (TODO, messages, etc) if we focus on just doing the backup and not migration schema.

.gitignore
```diff
@@ -12,6 +12,10 @@ uploads/
 .idea/
 config/secrets.py
 
+# backup folder contains the issues.db backup file with issues dump
+# for the previous version of schema
```
miketaylr: We can remove "of schema"

config/secrets.py.example
miketaylr: Change to: `# Database backup path.`

run.py
miketaylr: change to something like: `Backup existing issues.db`.

run.py
miketaylr: I think @karlcow is saying to remove the # TODO comment, since this PR is about backup (and schema updates are in the air).

run.py
```diff
@@ -106,6 +114,26 @@ def config_validator():
...
+        if issue_db().execute('select count(*) from webcompat_issues').fetchall()[0][0] > 0:
+            if not os.path.exists(app.config['BACKUP_DEFAULT_DEST']):
+                print 'Creating backup directory'
```
miketaylr: minor nit: put parens `print('Creating backup directory)`'

miketaylr: Might be kinda cool to do something like "Creating backup directory at _____BACKUP_DEST___whatever" (or similar).

miketaylr:
LGTM, just need to update some comments and tweak the print stuff!

miketaylr:
Trying to run this locally with `python run.py --backup` and I don't get a backup folder created. Does it for you @deepthivenkat? I would assume there would be a /backup dir created in my project root.

config/secrets.py.example
miketaylr: Seems like you forgot this change.

deepthivenkat:
The backup folder will get created only if the issues.db has a data dump. 
I am creating the folder only when I have to move the db file to the backup folder. Otherwise, there will be only an empty folder with no content.

miketaylr:
> The backup folder will get created only if the issues.db has a data dump.
> I am creating the folder only when I have to move the db file to the backup folder.

Ah, OK!

I think we should add some print statements so it's less confusing like: "Not doing a backup because there's nothing to back up!" (or whatever). But let's file a follow-up issue for that.

miketaylr:
We can ignore travis failures here, seems good to merge. 👍

miketaylr:
Oops, forgot that we should have had a "Fixes #1145" in the PR title. Something to remember for the future @deepthivenkat :)

Issue #1002, opened 2016-04-08
Upgrade to ESLint 2


miketaylr:
http://eslint.org/docs/user-guide/migrating-to-2.0.0

miketaylr:
Or version or 3 or whatever http://eslint.org/blog/2016/08/eslint-v3.2.2-released

Issue #978, opened 2016-03-23
Use Node 0.12.7 to be compatible with postcss changes


miketaylr:
Travis is busted and I'm not 100% sure why.

https://travis-ci.org/webcompat/webcompat.com/builds/118036930

It works locally for me, but I don't want to leave master in a broken state until we understand it. So let's back it out.

miketaylr:
```
Loading "cssnext.js" tasks...ERROR

>> ReferenceError: Promise is not defined
```

I see https://github.com/webpack/css-loader/issues/144#issuecomment-142591201.

I'm on a super old Node (v0.12.7), maybe we need to upgrade and standardize on something more modern (v 4 or 5)

cc @magsout

miketaylr:
Hmm, https://github.com/webpack/css-loader/issues/145 suggests that v0.12.7 should be OK. Maybe the problem is that Travis is using an *even* older version. Let me test something..

miketaylr:
OK, "Build #1555 passed."

Phew. :fire: :fire_engine: :droplet:

miketaylr:
(FWIW, I think Travis uses Node 0.8 if you don't specify a newer version).

PR #1172, opened 2016-09-07
Issue #1002 - Upgrade eslint

denschub:
This updates eslint to the most recent 3.4.0:

* Moved the eslint config out of the `package.json` into `.eslintrc` for compatibility reasons.
* Replaced `space-after-keywords`,`space-before-blocks`, and `space-return-throw-case` with `keyword-spacing`
* Added `// eslint-disable-line no-use-before-define` to the default object assignments. eslint 3 is a bit smarter about variable hoisting and the default assignments are indeed a use before define. It's not an issue for us at the moment, but there is no way to tell eslint this is okay without the comments or completely disabling `no-use-before-define`.

r? @miketaylr

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,46 @@
+{
+  "env": {
+    "browser": true
+  },
+  "extends": "eslint:recommended",
+  "globals": {
+    "jQuery": true,
+    "$": true,
+    "_": true,
+    "Backbone": true,
+    "console": true,
+    "define": true,
+    "issueNumber": true,
+    "markdownitEmoji": true,
+    "markdownitSanitizer": true,
+    "md": true,
+    "module": true,
+    "moment": true,
+    "Mousetrap": true,
+    "PaginationMixin": true,
+    "Prism": true,
+    "qr": true,
+    "repoPath": true,
+    "require": true,
+    "wcEvents": true
+  },
+  "rules": {
+    "comma-dangle": 0,
+    "curly": [2, "all"],
+    "eqeqeq": [2, "smart"],
+    "indent": [2, 2],
+    "keyword-spacing": 2,
+    "linebreak-style": [2, "unix"],
+    "new-cap": 2,
+    "no-cond-assign": 0,
+    "no-spaced-func": 2,
+    "no-use-before-define": 2,
+    "one-var": [2, "never"],
+    "quotes": [2, "single"],
+    "semi": [2, "always"],
+    "space-before-blocks": 2,
+    "space-before-function-paren": [2, "never"],
+    "space-infix-ops": 2,
+    "space-unary-ops": 2
+  }
+}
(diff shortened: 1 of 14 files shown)


denschub:
Oh! eslint 3 dropped support for Node.js < 4, travis builds with 0.12.7. Is updating Node an option or should I downgrade to eslint 2, which is still supported?

.eslintrc
```diff
@@ -0,0 +1,46 @@
+{
```
miketaylr: For some reason we moved from .eslintrc to having the rules defined in package.json.... I can't remember why and I don't personally feel strongly one way or another. Just commenting to create a paper trail of moving back!

webcompat/static/js/lib/comments.js
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-var issues = issues || {};
+var issues = issues || {}; // eslint-disable-line no-use-before-define
```
miketaylr: One day we need to really move away from these globals and use some kind of module system... I blame myself. ;_;

miketaylr:
LGTM!

miketaylr:
Hm, oops. I just saw you used Gecko conventions here and I didn't let you change your commit message before merging.

> Upgrade eslint r?miketaylr 

If you feel strongly, we can back that out and you can update the commit message. I'm OK either way.

miketaylr:
Oops.

```
Loading "eslint.js" tasks...ERROR

>> SyntaxError: /home/travis/build/webcompat/webcompat.com/node_modules/grunt-eslint/tasks/eslint.js:2

>> const chalk = require('chalk');

>> ^^^^^

>> Use of const in strict mode.

Warning: Task "eslint" not found. Use --force to continue.
```

miketaylr:
Uh, I think this is caused by us pinning Node to ancient version:  

`TRAVIS_NODE_VERSION="0.12.7"`

miketaylr:
> Oh! eslint 3 dropped support for Node.js < 4, travis builds with 0.12.7. Is updating Node an option or should I downgrade to eslint 2, which is still supported?

Which is what you already commented.

miketaylr:
I'm trying to find why we pinned to 0.12... ideally we just upgrade Travis to use Node LTS.

miketaylr:
Here we go: https://github.com/webcompat/webcompat.com/issues/978

miketaylr:
Fixed it here: https://github.com/webcompat/webcompat.com/commit/1c43f46374454d1c6ba2e2eef1876bda8cb9f2e3

denschub:
> If you feel strongly, we can back that out and you can update the commit message. I'm OK either way.

Oh no, that's fine by me. After all, this looks like I've been pushing commits without review, which is a pretty badass thing to do, right? 💣 

Thanks!

Issue #1004, opened 2016-04-12
[bug] text cut off on contributors page


adamopenweb:
If you maximize the browser on the contributor page, not sure the exact resolution, the text is cut off in the 'How to Do Site Outreach" section.

<img width="1172" alt="cutoff" src="https://cloud.githubusercontent.com/assets/6587558/14476347/33d96c46-00d5-11e6-966f-4149a37fa977.png">

PR #1173, opened 2016-09-07
Issue #1004 - Do not cut off text on the contributors page

denschub:
The box has `box-sizing: border-box`, so better use a padding to not extend the element beyond its container...

Before:
screen shot 2016-09-07 at 17 01 38

After:
screen shot 2016-09-07 at 17 01 49

r? @miketaylr


miketaylr:
Much better, thanks!

miketaylr:
r+

PR #1174, opened 2016-09-11
Fixes #1153 - Adding new fields to issues.db and webhooks

### LINKED ISSUE #1153: Add status, creation and last changed time, domain and reported from to issue db
These new fields are necessary if we need use an offline database to display the list of issues.

deepthivenkat:


webcompat/db/__init__.py
```diff
@@ -37,13 +39,23 @@ class WCIssue(IssueBase):
...
     url = Column(String(1024))
+    domain = Column(String(1024))
+    status = Column(String(256))
     body = Column(String(2048))
+    reported_from = Column(String(256), default='null')
+    creation_time = Column(DateTime)
+    last_change_time = Column(DateTime)
 
-    def __init__(self, issue_id, summary, url, body):
+    def __init__(self, issue_id, summary, url, domain, body, status, reported_from, creation_time, last_change_time):
```
karlcow: line is too long

webcompat/db/__init__.py
```diff
@@ -37,13 +39,23 @@ class WCIssue(IssueBase):
...
         self.issue_id = issue_id
         self.summary = summary
         self.url = url
+        self.domain = domain
+        self.status = status
         self.body = body
+        self.reported_from = reported_from
+        self.creation_time = datetime.datetime.strptime(creation_time.split('Z')[0],"%Y-%m-%dT%H:%M:%S")
+        self.last_change_time = datetime.datetime.strptime(last_change_time.split('Z')[0],"%Y-%m-%dT%H:%M:%S")
```
karlcow: lines are too long

webcompat/form.py
```diff
@@ -147,6 +148,12 @@ def domain_name(url):
     # testing if it's an http URL
     if url.startswith(SCHEMES):
         domain = urlparse.urlparse(url).netloc
+        if domain.lower().startswith('www.'):
+            domain = domain.split('www.')[1]
+        port_number = re.search(r':\d+$', domain)
+        if port_number is not None:
```
karlcow: in Python it is not necessary to this construct. The `if` will execute only if the value is not None. see for example. so you can directly use the value.

```python
>>> import urlparse
>>> domain = 'http://www.org:5000'
>>> url = 'http://www.org:5000/foobar?blah'
>>> urlparse.urlparse(url)
ParseResult(scheme='http', netloc='www.org:5000', path='/foobar', params='', query='blah', fragment='')
>>> domain = urlparse.urlparse(url).netloc
>>> import re
>>> port_number = re.search(r':\d+$', domain)
>>> port_number
<_sre.SRE_Match object at 0x1097712a0>
>>> url2 = 'http://www.org/foobar?blah'
>>> domain2 = urlparse.urlparse(url2).netloc
>>> port_number2 = re.search(r':\d+$', domain2)
>>> port_number2
>>> if port_number:
...     print 'do something'
... 
do something
>>> if port_number2:
...     print 'do something'
... 
>>> 
```

webcompat/form.py
```diff
@@ -147,6 +148,12 @@ def domain_name(url):
     # testing if it's an http URL
     if url.startswith(SCHEMES):
         domain = urlparse.urlparse(url).netloc
+        if domain.lower().startswith('www.'):
+            domain = domain.split('www.')[1]
```
karlcow: It doesn't give us benefit to remove the `www.`, it's a domain on its own. Plus it leads to the potential mistake of cutting off `www.TLD` example: `www.org`.

webcompat/form.py
```diff
@@ -147,6 +148,12 @@ def domain_name(url):
...
+        port_number = re.search(r':\d+$', domain)
+        if port_number is not None:
+            if domain.endswith(port_number.group()):
```
karlcow: because of the previous if on `port_number`, this next `if` will always be true, so it's useless. You are basically testing this:

```
>>> domain
'www.org:5000'
>>> port_number.group()
':5000'
>>> domain.endswith(port_number.group())
True
>>> 'www.org:5000'.endswith(':5000')
True
```

webcompat/form.py
```diff
@@ -147,6 +148,12 @@ def domain_name(url):
...
+                domain = domain.split(port_number.group())[0]
     else:
```
karlcow: All of that said, I don't think we should normalize `www.` and `:port_number`. There are part of the useful information of a domain name and even clearly identify different web properties.

webcompat/webhooks/__init__.py
```diff
@@ -39,11 +39,19 @@ def hooklistener():
...
+                issue_reported_from = payload.get('issue')['reported_from']
+            else:
+                issue_reported_from = 'null'
             parse_and_set_label(issue_body, issue_number)
             # Setting "Needs Triage" label by default
             # to all the new issues raised
             set_label('status-needstriage', issue_number)
-            dump_to_db(issue_title, issue_body, issue_number)
+            dump_to_db(issue_title, issue_body, issue_number, issue_status, issue_reported_from,
+                       issue_creation_time, issue_last_change_time)
```
karlcow: line too long

webcompat/webhooks/helpers.py
```diff
@@ -49,7 +51,10 @@ def set_label(label, issue_number):
     api_post('labels', payload, issue_number)
 
 
-def dump_to_db(title, body, issue_number):
+def dump_to_db(title, body, issue_number, status, reported_from, creation_time, last_change_time):
```
karlcow: line too long.
You really need to install a Python linter for your code editor. These are very simple syntactic issues.

webcompat/webhooks/__init__.py
```diff
@@ -39,11 +39,19 @@ def hooklistener():
             issue_body = payload.get('issue')['body']
             issue_title = payload.get('issue')['title']
             issue_number = payload.get('issue')['number']
+            issue_status = payload.get('issue')['status']
+            issue_creation_time = payload.get('issue')['created_at']
+            issue_last_change_time = payload.get('issue')['updated_at']
+            if 'reported_from' in payload.get('issue'):
+                issue_reported_from = payload.get('issue')['reported_from']
+            else:
+                issue_reported_from = 'null'
```
karlcow: instead of if/else here. You could set issue_reported_from = 'null' at the beginning after line 37 (aka a default value).
Then just setting the value if it's found in the payload.

karlcow:
ok there is another issue in this pull request. It reuses the labeler webhook.
the labeler webhook is working only on `action: opened`, we probably want to change this to other [kind of events on issues](https://developer.github.com/v3/activity/events/types/#issuesevent) such as 'updated', 'deleted', etc. because in the current context the status will always be the one of the opening time. The last_change_time will never change etc.

Also there is a "security" issue with our current setup. But maybe this should go in private, I'll talk with @miketaylr about it. It is not dramatic, but still.

webcompat/db/__init__.py
karlcow: I have the feeling (testing locally) that you just can do:

```python
self.last_change_time = datetime.datetime.strptime(last_change_time,"%Y-%m-%dT%H:%M:%SZ") 
```
The `Z` will not be recorded in sqlite.

webcompat/db/__init__.py
```diff
@@ -37,13 +39,23 @@ class WCIssue(IssueBase):
     issue_id = Column(String(128), unique=True, primary_key=True)
```
karlcow: why not constraining this to an Integer type

karlcow:
* If we go the way of normalizing url. We need to add tests for it.
* Currently a lot of errors with the data fail with a Error 500 if the data are not correct, we need probably to make it more robust and catch the errors to log them and/or display a useful error message for the users.

miketaylr:
+1 to @karlcow's review so far. Will take a closer look when those comments are addressed. Thanks!

miketaylr:
I just pushed the security fixes from @karlcow -- please rebase!

deepthivenkat:
Okay!

miketaylr:
> If we really go with normalizing domains, then we need to add tests for it.
> Currently a lot of errors with the data fail with a Error 500 if the data are not correct, we need probably to make it more robust and catch the errors to log them and/or display a useful error message for the users.

@karlcow can you clarify this comment? I think @deepthivenkat wasn't 100% sure what it means. Thanks!

deepthivenkat:
Thanks @miketaylr.
@karlcow I wanted to know if by tests if you mean additional validations for the url in the form.py.
I wanted to clarify this before pushing the changes because, these comments will get erased(I am rebasing after mike's security push).

karlcow:
There are two things here:

* normalizing domains: refer to test ins the test suite. But we decided to not do that. So let's ignore.
* About `Error 500`, some parts of the code doesn't really constraint the input in a meaningful way. Like for example, the `String(128)` instead of `Integer`. So the game (if you accept it :D ) is basically to challenge the code. Each time there's a chance for the server to do a 500, it means we didn't handle the error gracefully. 

**BUT** first fix the code with the new version then we can chase down the `Error 500`.

karlcow:
@deepthivenkat Is there anything I can help with to move this forward? Do not hesitate to say if you feel you are too busy, we can finish it. Hope everything is fine with the new life.

deepthivenkat:
I have pushed some changes today. I had 2 problems that were holding me from pushing changes:

1) I wanted to know the difference between 'labeled' or 'unlabeled' actions in payload
2) I am updating a few fields in issue db for 'edited', 'closed', 'labeled', 'unlabeled' - I wanted to test it in the 'issue.db' file from my local webcompat folder.
I have a repository in github where the issues that I create in my localhost will be stored. I tried to add webhook to that repository and checked if it stored the fields in my local issue.db file. This attempt was not successful.
3) In [helpers.py](https://github.com/webcompat/webcompat.com/pull/1174/commits/c2e88d92dd607e9a0e8d1de199d7b42c5b795725#diff-53c7dc964488671110b629cf6885fb3eR72), the update command will be called once for every key, value pairs passed into **kwargs. However, I want it to be called once.

- These are the problems i faced. So now i thought I will push the code and seek your opinion.

karlcow:
@deepthivenkat There are a couple of issues with this last push. There are many commits from @miketaylr . Not only your modifications.

There are files which should not be there such as
https://github.com/webcompat/webcompat.com/pull/1174/commits/c2e88d92dd607e9a0e8d1de199d7b42c5b795725#diff-01b9f10c14726ede6655dcc207991901R1

I'm trying to get a better understanding of what has been pushed to the pull request.

karlcow:
@deepthivenkat 

> 1) I wanted to know the difference between 'labeled' or 'unlabeled' actions in payload
https://developer.github.com/webhooks/#events

On [issue event](https://developer.github.com/v3/activity/events/types/#issuesevent) gives you answer.

`"action": "labeled"` and `"action": "unlabeled"` means that a label has been added or removed to the issue.

karlcow:
@deepthivenkat 

> 2) I am updating a few fields in issue db for 'edited', 'closed', 'labeled', 'unlabeled' - I wanted to test it in the 'issue.db' file from my local webcompat folder. I have a repository in github where the issues that I create in my localhost will be stored. I tried to add webhook to that repository and checked if it stored the fields in my local issue.db file. This attempt was not successful.

To test webhooks, you can do either:

* Put a test server online on a personal space and creates a webhook.
* Impersonate github locally by doing a HTTP POST with a JSON file on your server running on localhost.

See for example https://github.com/miketaylr/web-bugs-issue-tagger#testing-locally

karlcow:
@deepthivenkat 

> 3) In helpers.py, the update command will be called once for every key, value pairs passed into **kwargs. However, I want it to be called once.

I didn't know the answer on the top of my head, but looking at the function:

```python
def update_issue_in_db(issue_number, **kwargs):
    for key, value in kwargs:
        if value:
            update(issue_db).where(issue_db.issue_id == issue_number).values(key=value)
```

Here you are iterating on `key, value` and makes the update for each of them. So I checked the [documentation of SQLAlchemy for `values`](http://docs.sqlalchemy.org/en/latest/core/dml.html#sqlalchemy.sql.expression.update.params.values).

In the documentation they say:

> Optional **dictionary** which specifies the SET conditions of the UPDATE. 

1. You can either pass a dictionary beforehand and uses it in the function call.
2. Or you iterate other the data in the function and creates a dictionary that you then pass to the update.

Personally I would go with option 1.

So instead of doing 

```python
                    update_issue_in_db(int(issue_number),
                                       issue_title=issue_title,
                                       issue_body=issue_body,
                                       issue_last_change_time=updated)
```

I would pass:

```python
                    update_issue_in_db(int(issue_number),
                                       {'issue_title'=issue_title,
                                       'issue_body'=issue_body,
                                       'issue_last_change_time'=updated})
```

and

```python
def update_issue_in_db(issue_number, new_values):
    '''Updates the database with new values.

    values is a dictionary.'''
    update(issue_db).where(issue_db.issue_id == issue_number).values(new_values)
```

miketaylr:
Is there a status update on this work @deepthivenkat @karlcow? Do we plan on landing this PR, or should we close it, iterate on a branch, and attempt to land it in a new PR?

deepthivenkat:
I discussed with Karl on IRC but forgot to update it here. 
@miketaylr we should close it and use a new PR.

miketaylr:
OK, let's do that. Thanks!

PR #1189, opened 2016-10-05
Fixing typos/grammar/wording

SemicolonExpected:
ln 6 and 36: Pluralized words for consistency. Note that the other headings that don't have 'a' or 'an' before it are all pluralized. eg closing bugs, issues, requests, etc. Questions or Problems also sounds slightly better.

ln 62: duplicate link

ln 127: themselves is gender neutral avoiding the him/herself so is a more succinct way of stating that they shouldn't merge their own pull request. I feel the themself should also be tacked on the end as "merging themselves the pull request" might sound slightly awkward. Similarly stating "or the admin" after "merge the branch" could also be awkward (and beg the question: or the admin what) and stating "the reviewer or the admin" flows better.

ln 143: "that much conventions" is grammatically incorrect, though not technically incorrect "we do not have yet" does not flow as well as "we do not yet have" I also added "When" before "in doubt" since on ln 170 you do use 'when in doubt' so I assume you just forgot "when"

ln 258: changed do to done -- tense agreement

ln 260: undefined reference: clarified what is needed, "installing pip" as what "if you need to" is referring to can be ambiguous

ln 287: changed 'user' to 'username' for consistency since in the next line 'username' is used

ln 376: I think this is what is meant, otherwise this is an odd sentence

ln 391: Windows

ln 403: ambiguous 'that'

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 You are welcome to contribute to this project. Here are the guidelines we try to stick to in this project.
 
 
-* [Question or Problem](#question-or-problem)
+* [Questions or Problems](#questions-or-problems)
 * [Filing an Issue](#filing-an-issue)
 * [Triaging Issues](#triaging-issues)
   * [Closing Bugs as Invalid](#closing-bugs-as-invalid)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -3,7 +3,7 @@
 You are welcome to contribute to this project. Here are the guidelines we try to stick to in this project.
 
 
-* [Question or Problem](#question-or-problem)
+* [Questions or Problems](#question-or-problem)
```
miketaylr: Can you update the link? GitHub automatically makes an anchor based on the heading, so it should be `(#questions-or-problems)`.

CONTRIBUTING.md
```diff
@@ -402,7 +400,7 @@ npm run watch
  npm run watch
  ```
 
-By default, a build will *not* optimize images (which is done before deploys). If you'd like to do that, you can run `grunt imagemin`.
+By default, a build will *not* optimize images (which is done before deploys). If you'd like to do optimize images, you can run `grunt imagemin`.
```
miketaylr: s/you'd like to do optimize images/you'd like to optimize images/

SemicolonExpected: Got it :)

miketaylr:
Great! I'm gonna rebase that second commit into the first and then merge. 👍

PR #1194, opened 2016-10-10
Small grammatical change + consistency/clarity fix

SemicolonExpected:
ln 2: changed the "Find a bug with a "needsdiagnosis" label or that is otherwise new" to be the same as the corresponding step in "How to Reproduce the Bug" to make that instruction easier to understand, and be consistent in terms of linking to the "Need Diagnosis" tag rather than sending users to a label. 

ln 6;8-9: added the between "if bug" to both be consistent with the other instructions that refer to the bug as "the bug" rather than just "bug" and for grammatical correctness. 

ln 11: It's more correct to say "if there are any errors" than "if there are errors"

[ci skip]

### DIFF
--- a/webcompat/templates/contributors/diagnose-bug.html
+++ b/webcompat/templates/contributors/diagnose-bug.html
@@ -1,11 +1,14 @@
 <ul class="contributors__list">
-  <li class="contributors__list__item"><span>Find a bug with a "needsdiagnosis" label or that is otherwise new</span></li>
-  <li class="contributors__list__item"><span>Check to see if bug has been reproduced already</span></li>
+  <li class="contributors__list__item"><span>Find a bug that "<a href="https://webcompat.com/issues?page=1&per_page=50&state=open&stage=needsdiagnosis&sort=created&direction=desc">Needs Diagnosis or is otherwise new</a>"</span>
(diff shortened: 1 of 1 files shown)


webcompat/templates/contributors/diagnose-bug.html
```diff
@@ -1,11 +1,14 @@
 <ul class="contributors__list">
-  <li class="contributors__list__item"><span>Find a bug with a "needsdiagnosis" label or that is otherwise new</span></li>
-  <li class="contributors__list__item"><span>Check to see if bug has been reproduced already</span></li>
+  <li class="contributors__list__item"><span>Find a bug that "<a href="https://webcompat.com/issues?page=1&per_page=50&state=open&stage=needsdiagnosis&sort=created&direction=desc">Needs Diagnosis or is otherwise new</a>"</span>
```
miketaylr: Hmm, I think this is sort of out of date. Before we had this notion of "new" bugs. But now we call them "needs triage". Can we restrict the hyperlink to just `Needs Diagnosis` and add a link for "Needs Triage" (instead of "otherwise new")?

miketaylr:
Thank you!

PR #1198, opened 2016-10-12
Fixes #1193. Create a simple /issues/new view for reporting bugs

### LINKED ISSUE #1193: Make /issues/new endpoint a reduced version of the homepage reporting form
If we do this correctly, this could fix #1163, #1164, #1165. This will also fix #317.

Currently we POST form data to that endpoint, this bug should create a response for GET, which is the form, a reduced header, and not much else (if anything).

miketaylr:
I need to write some tests, not ready for review yet.

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -155,8 +155,22 @@ def show_issues():
     return render_template('list-issue.html', categories=categories)
 
 
-@app.route('/issues/new', methods=['POST'])
+@app.route('/issues/new', methods=['GET', 'POST'])
 def create_issue():
+    """Creates a new issue.
+
+    GET will return an HTML response for reporting issues
+    POST will create a new issue
+    """
+    if request.method == 'GET':
+        bug_form = IssueForm()
+        # add browser and version to bug_form object data
+        ua_header = request.headers.get('User-Agent')
+        bug_form.browser.data = get_browser(ua_header)
+        bug_form.os.data = get_os(ua_header)
+        if g.user:
+            get_user_info()
+        return render_template('new-issue.html', form=bug_form)
     # copy the form so we can add the full UA string to it.
     form = request.form.copy()
     # see https://github.com/webcompat/webcompat.com/issues/1141
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -14,9 +14,10 @@ from flask import g
 from flask import redirect
 from flask import render_template
 from flask import request
-from flask import session
 from flask import send_from_directory
+from flask import session
 from flask import url_for
+
 from form import AUTH_REPORT
 from form import IssueForm
 from form import PROXY_REPORT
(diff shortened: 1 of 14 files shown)


webcompat/views.py
```diff
@@ -155,8 +155,22 @@ def show_issues():
...
-@app.route('/issues/new', methods=['POST'])
+@app.route('/issues/new', methods=['GET', 'POST'])
 def create_issue():
+    """Creates a new issue
+
+    GET will return an HTML response for reporting issues
+    POST will create a new issue
+    """
+    if request.method == 'GET':
+        bug_form = IssueForm()
```
miketaylr: This starts to feel a bit redundant w/ the index view, but I'm not sure how to make it more DRY. 🍤

miketaylr:
(bunch of random Travis failures right now... possibly GitHub issues?)

webcompat/views.py
karlcow: What about making the redundant code into a function that you call in both places?

miketaylr:
> (bunch of random Travis failures right now... possibly GitHub issues?)

Yeah, the issue was a class was changed and the logout helper method was failing. Fixed now.

r? @karlcow 
r? @magsout (for CSS stuff)

webcompat/views.py
```diff
@@ -155,8 +150,18 @@ def show_issues():
     return render_template('list-issue.html', categories=categories)
 
 
-@app.route('/issues/new', methods=['POST'])
+@app.route('/issues/new', methods=['GET', 'POST'])
 def create_issue():
+    """Creates a new issue
```
karlcow: nitpicking: Missing '.' at the end of the sentences.

webcompat/views.py
```diff
@@ -155,8 +150,18 @@ def show_issues():
...
+
+    GET will return an HTML response for reporting issues
+    POST will create a new issue
+    """
```
karlcow: I have the feeling we could improve the way we handle the code here. But probably for another issue, you might want to open.
Right now we have

```python
if GET:
    this_code()
    return blah_get
# here the http post code (with logging and spam)
return blah_post
```

Maybe something along

```python
# common_tasks
# ua_header, etc.
if GET:
    this_code()
    return blah_get
if POST:
   # making functions for different features
    spam_protect()
    log_this()
    return blah_post
```

BUT Let's not block the progress of this PR. so +1 with the issue created.

magsout:
r+ @miketaylr 

Good job

miketaylr:
Thanks @magsout!

miketaylr:
Also, thanks @karlcow! https://github.com/webcompat/webcompat.com/issues/1200 filed for refactoring of create_issue method.

miketaylr:
Fixed up the docstring nit. Will merge once tests are green and cut a new release/deploy.

PR #1208, opened 2016-10-19
Fixes #1054. Don't allow non-webby URLs to pass URL validation.

### LINKED ISSUE #1054: Disallow non-webby URIs?
For example, https://github.com/webcompat/web-bugs/issues/2591#issuecomment-220498160.

Filing bugs against file://, about:blank, about:config, chrome://settings, etc. -- not very useful? We could show an error saying how they're non-webby URLs.

miketaylr:
r? @adamopenweb


adamopenweb:
Looks good! r+
Thanks Mike.

PR #1213, opened 2016-10-26
Fixes #1206. Use dot instead of paren inside issue description.

### LINKED ISSUE #1206: Change Steps to Reproduce Ordered List to use dot instead of paren
```
1)
2)
```

Should become

```
1.
2.
```

haseebgit:
@miketaylr r?


miketaylr:
Thanks!

PR #1242, opened 2016-12-14
Fixes #1239. Remove QR code related code.

### LINKED ISSUE #1239: Remove QR-related code on issue page
We unintentionally didn't include it in the new issue page redesign. Do we re-include it? Or remove it for-reals? Right now we're including the JS on the page but it won't do anything, so we should make a decision one way or the other.

miketaylr:
r? @tagawa -- in case I missed something obvious (but I think I got it all!).


tagawa:
Thanks @miketaylr. There's also the file `static/js/vendor/qr.min.js`, then I think that's everything.

miketaylr:
Oops, great catch! Thanks for review, @tagawa.

tagawa:
Happy to help, but a bit sad too, but still happy.

miketaylr:
😞 

FWIW, I'd be happy for this to come back at some point in the future, even though I never used it. But since it's not working right now, seems good to reduce page load.

magsout:
@tagawa @miketaylr my fault :(

tagawa:
@magsout No, not at all! It's a good chance to get rid of clutter that's probably not being used.

Issue #1252, opened 2016-12-22
Only handle image blobs on front and backends


miketaylr:
Filing this as a "future work" issue.

In #1248 we allow the frontend to handle blobs OR base64 screenshot data, but send it all up to the server as base64 strings. It would simplify code to just handle blobs everywhere, since you can treat blobs the same as File objects.

In order for this to happen, we need to make changes in a few places: frontend, server code, add-ons, and the Fennec Nightly reporter code. 

Potentially there's perf wins here. Interesting read that's only partially relevant: https://github.com/Esri/offline-editor-js/issues/206

PR #1253, opened 2016-12-22
Fixes #1248. Handle blobs in addition to data URIs in bugform.js

### LINKED ISSUE #1248: Handle blobs being postMessage'd to bug form
Right now the add-on sends data URIs. Feedback from Firefox is to send blobs, so let's handle both.

miketaylr:
This allows the Firefox "Report Site Issue" feature to postMessage screenshot blobs, in addition to the base64 data URIs that Fennec and the rest of the addons send.

See #1252 for making everything send blobs.

(Still need to write a test for image-uploads-non-auth.js, so not ready for merge.)

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -9,7 +9,6 @@ function BugForm() {
   this.reportButton = $('#js-ReportBug');
   this.loaderImage = $('.js-Loader');
   this.uploadLoader = $('.js-Upload-Loader');
-  this.screenshotData = '';
   // by default, submission type is anonymous
   this.submitType = 'github-proxy-report';
   this.UPLOAD_LIMIT = 1024 * 1024 * 4;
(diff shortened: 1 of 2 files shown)


webcompat/static/js/lib/bugform.js
```diff
@@ -77,23 +76,34 @@ function BugForm() {
...
-        // so, bytes = (encoded_string.length - 814) / 1.37)
-        // see https://en.wikipedia.org/wiki/Base64#MIME
-        if ((String(this.screenshotData).length - 814 / 1.37) > this.UPLOAD_LIMIT) {
-          this.downsampleImageAndUpload(this.screenshotData);
+        // See https://github.com/webcompat/webcompat.com/issues/1252 to track
+        // the work of only accepting blobs, which should simplify things.
+        if (event.data instanceof Blob) {
+          // showUploadPreview will take care of converting from blob to
+          // dataURI, and will send the result to resampleIfNecessaryAndUpload.
+          this.showUploadPreview(event.data);
```
denschub: I am a bit confused that a function called `showUploadPreview` ultimately triggers the upload itself, since I would normally not expect that. But changing that in this pull request is probably not in scope, and probably not even needed.

miketaylr:
Thanks! I agree the naming of the methods feels a bit weird, and will add a test before merging.

denschub:
Even better then. r++!

miketaylr:
Tests are green, bombs away. 💣

PR #1255, opened 2016-12-24
Fixes #1249: support a src param to track "reported-with" 

### LINKED ISSUE #1249: Support GET args for reported-with values
In addition to `X-Reported-With`, let's support `&reported-with=foo`, and eventually phase out the HTTP header approach. It makes it much simpler to implement things on the Firefox side.

miketaylr:
r? @karlcow

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -142,7 +142,7 @@ function BugForm() {
     if (!location.search.search(/open=1/) && !location.search.search(/url=/)) {
       return;
     }
-    var urlParam = location.search.match(/url=(.+)/);
+    var urlParam = location.href.match(/url=([^&]*)/);
     if (urlParam != null) {
       // weird Gecko bug. See https://bugzilla.mozilla.org/show_bug.cgi?id=1098037
       urlParam = this.trimWyciwyg(urlParam[1]);
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -160,6 +160,8 @@ def create_issue():
         bug_form = get_form(request.headers.get('User-Agent'))
         if g.user:
             get_user_info()
+        if request.args.get('src'):
+            session['src'] = request.args.get('src')
         return render_template('new-issue.html', form=bug_form)
     # copy the form so we can add the full UA string to it.
     form = request.form.copy()
@@ -175,8 +177,14 @@ def create_issue():
             flash(msg, 'notimeout')
             return redirect(url_for('index'))
     form['ua_header'] = request.headers.get('User-Agent')
-    # Store where the report originated from
-    form['reported_with'] = request.headers.get('X-Reported-With', 'web')
+    # Currently we support either a src GET param, or X-Reported-With header
+    # to track where the report originated from.
+    # See https://github.com/webcompat/webcompat.com/issues/1254 to track
+    # supporting only the src param
+    if session.get('src'):
+        form['reported_with'] = session.pop('src')
+    else:
+        form['reported_with'] = request.headers.get('X-Reported-With', 'web')
     # Logging the ip and url for investigation
     log = app.logger
     log.setLevel(logging.INFO)


webcompat/static/js/lib/bugform.js
```diff
@@ -142,7 +142,7 @@ function BugForm() {
     if (!location.search.search(/open=1/) && !location.search.search(/url=/)) {
       return;
     }
-    var urlParam = location.search.match(/url=(.+)/);
+    var urlParam = location.search.match(/url=([^&]*)/);
```
miketaylr: (otherwise you ended up with stuff like URL: `http://example.com/&src=desktop-reporter`, etc.)

karlcow: Do you have a test for URLs like 
http://www.openstreetmap.org/search?query=paris#map=11/48.8591/2.2611&layers=C

If they are properly escaped, it will be ok, but if not:

```js
/* this is bogus already but let say */
url = 'http://www.example.com/?open=1&url=http://www.openstreetmap.org/search?query=paris#map=11/48.8591/2.2611&layers=C&src=FASTER'
url.match(/url=([^&]*)/)
/* will return 

Array [ "url=http://www.openstreetmap.org/search?query=paris#map=11/48.8591/2.2611", "http://www.openstreetmap.org/search?query=paris#map=11/48.8591/2.2611" ]
*/
```
which is not what we want.

webcompat/views.py
```diff
@@ -175,8 +177,14 @@ def create_issue():
...
             return redirect(url_for('index'))
     form['ua_header'] = request.headers.get('User-Agent')
-    # Store where the report originated from
-    form['reported_with'] = request.headers.get('X-Reported-With', 'web')
+    # Currently we support either a src GET param, or X-Reported-With header
+    # to track where the report originated from.
+    # See https://github.com/webcompat/webcompat.com/issues/1254 to track
+    # supporting only the src param
+    if session['src']:
+        form['reported_with'] = session.pop('src', None)
```
miketaylr: Ah, right. I can also do `if session.get('src')`, since that defaults to `None` if the key isn't there and doesn't throw `KeyError`. Thanks!

miketaylr: Yeah, I think we want to remove `src` from the session cookie -- it's not useful after we get its value.

webcompat/static/js/lib/bugform.js
miketaylr: Ah yeah. `location.search` is def not what we want to match against, because it doesn't even get the hash content for your example URL:

```
> location.search
"?open=1&url=http://www.openstreetmap.org/search?query=paris"
```

As for the extra layers param, let me test after escaping.

miketaylr: OK, so from the add-on, we `encodeURIComponent` before sticking in the `url` param:

`https://webcompat.com/issues/new?url=http%3A%2F%2Fwww.openstreetmap.org%2Fsearch%3Fquery%3Dparis%23map%3D11%2F48.8591%2F2.2611%26layers%3DC`

That means we wend up with the right thing in the form:
screen shot 2016-12-28 at 4 03 34 pm


But if you were to somehow manually type in the URL to the urlbar w/o escaping: `http://localhost:5000/issues/new?url=http://www.openstreetmap.org/search?query=paris#map=11/48.8591/2.2611&layers=C&src=FASTER` we end up with:

`http://www.openstreetmap.org/search?query=paris#map=11/48.8591/2.2611`

So obviously that isn't what we want.

We could either ignore it because it seems like an edgecase, or chop off everything that isn't `&src=foo` (if it's at the end). WDYT, @karlcow?

miketaylr:
>  A question for the url containing &

Answered here: https://github.com/webcompat/webcompat.com/pull/1255#discussion_r94080753 (and another question for you).

> A change for the python code which will not work as you expect.

Done, thank you!

webcompat/static/js/lib/bugform.js
karlcow: Maybe just add a comment for now on what is expected.

karlcow:
Thanks @miketaylr

PR #1266, opened 2017-01-05
Fixes #722 - Handle two sizes images

### LINKED ISSUE #722: Have two sizes for one image
When uploading an image. We can display a smaller version and keep the original which is linked from the smaller one. Depends on #710

karlcow:
This modifies the python upload code and the JavaScript libraries handling the markup.
Note that for the JS, the previous code is using two different methods for inserting the code at the right place even if they do exactly the same thing. This should be abstracted. 

r? @miketaylr

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -428,10 +428,10 @@ function BugForm() {
     copy over the URL of a newly uploaded image asset to the bug
     description textarea.
   */
-   this.addImageURL = function(url) {
-    var file_ext = url.split('.').pop();
-    var thumb_url = url.replace('.' + file_ext, '-thumb.' + file_ext);
-    var imageURL = ['[![Screenshot Description](', thumb_url, ')](', url, ')'].join('');
+  this.addImageURL = function(response) {
+    var img_url = response.url;
+    var thumb_url = response.thumb_url;
+    var imageURL = ['[![Screenshot Description](', thumb_url, ')](', img_url, ')'].join('');
     this.descField.val(function(idx, value) {
       return value + '\n\n' + imageURL;
     });
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -428,8 +428,10 @@ function BugForm() {
     copy over the URL of a newly uploaded image asset to the bug
     description textarea.
   */
-  this.addImageURL = function(url) {
-    var imageURL = ['![Screenshot Description](', url, ')'].join('');
+   this.addImageURL = function(url) {
+    var file_ext = url.split('.').pop();
+    var thumb_url = url.replace('.' + file_ext, '-thumb.' + file_ext);
+    var imageURL = ['[![Screenshot Description](', thumb_url, ')](', url, ')'].join('');
     this.descField.val(function(idx, value) {
       return value + '\n\n' + imageURL;
     });
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -425,8 +425,8 @@ function BugForm() {
     this.uploadField.val(this.uploadField.get(0).defaultValue);
   };
   /*
-    copy over the URL of a newly uploaded image asset to the bug
-    description textarea.
+    create the markdown with the URL of a newly uploaded image
+    and its thumbnail URL assets to the bug description textarea.
   */
   this.addImageURL = function(response) {
     var img_url = response.url;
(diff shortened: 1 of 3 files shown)


webcompat/static/js/lib/bugform.js
```diff
@@ -428,8 +428,10 @@ function BugForm() {
     copy over the URL of a newly uploaded image asset to the bug
     description textarea.
   */
-  this.addImageURL = function(url) {
-    var imageURL = ['![Screenshot Description](', url, ')'].join('');
+   this.addImageURL = function(url) {
+    var file_ext = url.split('.').pop();
+    var thumb_url = url.replace('.' + file_ext, '-thumb.' + file_ext);
+    var imageURL = ['[![Screenshot Description](', thumb_url, ')](', url, ')'].join('');
```
miketaylr: Here we're doing string concatenation manually, but in issues.js we're using Lodash's `_.template` method. I don't have a preference (which makes since I authored the original two different versions... :/), but can we make them consistent?

webcompat/static/js/lib/bugform.js
```diff
@@ -428,8 +428,10 @@ function BugForm() {
     copy over the URL of a newly uploaded image asset to the bug
     description textarea.
   */
-  this.addImageURL = function(url) {
-    var imageURL = ['![Screenshot Description](', url, ')'].join('');
+   this.addImageURL = function(url) {
+    var file_ext = url.split('.').pop();
+    var thumb_url = url.replace('.' + file_ext, '-thumb.' + file_ext);
```
miketaylr: Aren't we getting back the `thumb_url` path back from the response sent back by the `@uploads` `/` route?

Why not just make `addImageURL` take a response argument, rather than the url, that way you can grab `response.url` and `response.thumb_url` without the string manipulation.

miketaylr: nit: Can we also update the comment to mention the thumbnail?

webcompat/static/js/lib/issues.js
```diff
@@ -199,8 +199,11 @@ issues.ImageUploadView = Backbone.View.extend({
     var DELIMITER = '\n\n';
     var textarea = $('.js-Comment-text');
     var textareaVal = textarea.val();
-    var imageURL = _.template('![Screenshot of the site issue](<%= url %>)');
-    var compiledImageURL = imageURL({url: response.url});
+    var img_url = response.url;
+    var file_ext = img_url.split('.').pop();
+    var thumb_url = img_url.replace('.' + file_ext, '-thumb.' + file_ext);
```
miketaylr: Same as before, we should be able to just get `response.thumb_url` here, right?

webcompat/static/js/lib/bugform.js
karlcow: yes that was the purpose of my comment in the pull request. :) I tried to minimize my modification of the JS. And I remember you telling me a couple of months ago that to avoid to repeat the same code, we should probably make a function which is called in the two places. 

I have 0 preferences on the way to do it given my understanding of JS (performance impact, elegance, etc.) The one without Lodash's template seems more direct?

miketaylr: > yes that was the purpose of my comment in the pull request

Gotcha, yeah there's duplication (in spirit) now, but let's make it real duplication (until we fix that).

webcompat/static/js/lib/issues.js
karlcow: done :)

webcompat/static/js/lib/bugform.js
karlcow: done :)

webcompat/static/js/lib/bugform.js
karlcow: done.

karlcow:
ok addressed all the comments. And also pushed up a bit the size of the image and changed the algorithm for compression, so screenshots of texts are usable on desktop.

NEAREST at the top, BILINEAR at the bottom.
capture d ecran 2017-01-06 a 09 39 03

karlcow:
Thanks @miketaylr for the cool review. :)

miketaylr:
Looks good! r+

(Can you file an issue for adding some thumbnail tests?)

karlcow:
ahah. forgotten about that. Done with #1267

❤️

miketaylr:
Let's revert this for now, sorry @karlcow!

Another thing to fix when you figure out why the bug form uploads are broken: if you click "Remove image upload", it should remove the link in description.

See https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L383-L385

PR #1272, opened 2017-01-06
Fixes #1262. Display new bug report URI, rather than /issues/new.

### LINKED ISSUE #1262: Reporting new issues doesn't redirect to its permanent URI
Steps to reproduce.

1. Report a new issue.
2. Get the new flash message on top of the issue
3. Check the URI in the address bar `https://webcompat.com/issues/new` instead of the `https://webcompat.com/issues/[NNNN]`

Note that the flash knows the URI to go 
`https://twitter.com/intent/tweet?text=I%20just%20filed%20a%20bug%20on%20the%20internet%3A&url=https%3A%2F%2Fwebcompat.com%2Fissues%2F4078&via=webcompat`

So it should not be an issue to either redirect to the right place or change window.history

miketaylr:
Originally I wanted to solve this on the Python side... but there's an issue:

Flask's `url_for` can only pass in variables if they're defined in the route as "variables". So for the following route we have the variable `<int:number>`

```
@app.route('/issues/<int:number>')
def show_issue(number, show_thanks_flash=False):
```

if we do `return redirect(url_for('show_issue', number=123, show_thanks_flash=False)`,

we get redirected to `/issues/123?show_thanks_flash=False` (because stuff it doesn't recognize is treated as GET args).

We could work around this by defining a route like so:

`@app.route('/issues/<int:number>/<thanks>')`, but people might bookmark that.

I wonder if we could work around this with the session cookie, rather than doing it in JS? Or some other idea?

Not ready for review. Let me try the session thingy.

### DIFF
--- a/webcompat/static/js/lib/flash-message.js
+++ b/webcompat/static/js/lib/flash-message.js
@@ -57,6 +57,9 @@ var FlashMessageView = Backbone.View.extend({
     this.$el.addClass('is-active wc-FlashMessage--thanks');
     this.$el.html(buildTemplate({number: opts.message}))
             .insertBefore('.wc-Issue-information').show();
+
+    // show the correct URL now, rather than /issues/new
+    history.replaceState({}, '', '/issues/' + opts.message);
   },
   hide: function() {
     this.$el.fadeOut();


miketaylr:
Yeah, I don't like this. Closing in favor of https://github.com/webcompat/webcompat.com/pull/1273.

karlcow:
Just a note. I wonder if you could have done it with using the Referer.

PR #1273, opened 2017-01-06
Fixes #1262. Redirect to issue page after bug creation.

### LINKED ISSUE #1262: Reporting new issues doesn't redirect to its permanent URI

miketaylr:
Let's try this rather than https://github.com/webcompat/webcompat.com/pull/1272.

r? @karlcow

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -114,7 +114,8 @@ def file_issue():
     response = report_issue(session['form_data'])
     # Get rid of stashed form data
     session.pop('form_data', None)
-    return show_issue(response.get('number'), show_thanks_flash=True)
+    session['show_thanks'] = True
+    return redirect(url_for('show_issue', number=response.get('number')))
 
 
 @app.route('/', methods=['GET', 'POST'])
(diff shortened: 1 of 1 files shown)


miketaylr:
Thanks @karlcow  -- mind explaining how you might do this via Referer?

karlcow:
My naive thinking…

```
POST /issues/new
```

which creates a redirect to `/issues/[number]` could send along that the Referer is `/issues/new`. 
When this `Referer` is there you know you need to send the thanks flash.
when someone does a direct query to `issues/[number]` there's no referer with `/issues/new` value.

miketaylr:
Ah, that's kinda cool. Using `session` feels less elegant, I'm not sure if it is or isn't.

PR #1265, opened 2017-01-05
Fixes #1261 - Improve the comment form and upload image area

karlcow:
This pull request implements the proposal of #1261 following the recommendation of @magsout for the syntax

Issue #1007, opened 2016-04-14
Improving UX for bug reporting


karlcow:
Extracted from a thread on compat mailing-list

---

For example, when we say "desktop site instead of mobile site", it's not clearly obvious what it means. Often it means I guess: 
    text/pictures looks very tiny on my mobile screen

For mobile, we often lack description and steps to reproduce. It probably can be understood because for two reasons:
1. Typing on mobile keyboard is cumbersome.
2. It's a long form instead of a step by step form (series of screens). 

Maybe we should do stats 1st on what has been reported the ~2500 bugs and how. 

For Desktop, we have more real estate screen and a real keyboard. It is probably easier for people to report issues with more details.  I would discourage to hide some fields on desktop. For example, I usually report the bug for mobile on desktop, because it's just easier. I can have the mobile on the side and describe the steps instead of having to type on the mobile keyboard and/or memorize them. Also my github login is always here on desktop, but never on the mobile. 

I would instead  a multi-step, multi-screen reporting, something along
- Screen 1: Are you reporting an issue for a mobile, tablet, desktop, others [ ]
- Screen 2: We identified you are using this browser, screen size, etc.
         Is it the device you want to report the issue for.
- Screen 3: Type of issues + screenshot
- Screen 4: Steps to reproduce.

(Needs thinking, not necessary these screens, these wordings in this order, but you get the picture).

See for example https://www.virginamerica.com/book-travel
How the site propose different steps to accomplish the process.
Each screen being focused on 1 task, it becomes a lot clearer.

Or if we do a single form we need a way to make it more progressive, probably with a more vertical linear design with rewarding visual cues. We do visual cues for errors or missing info, but not that much rewarding. When the form has been completed something could appear on the page with love. 

---

Then I have also seen this form, quite well done.
http://codepen.io/epilande/pen/eZJGpP

magsout:
Totaly agree with you. When I reported mobile issue, I use my computer instead of my mobile, easier and faster. Steps like America is interesting. A page for a need, 
1)website
2)main issue
3) behavior but with a list ? (
to avoid to write but leaves anyway this possibility)
4) screenshot
5) reported

miketaylr:
> Maybe we should do stats 1st on what has been reported the ~2500 bugs and how.

This could be very cool.

karlcow:
@locketmm @zoepage to discuss about form redesign issues. This is the right place.

lockettm:
[...]

PR #1293, opened 2017-01-19
Fixes #1292 - Regression introduced by #1265

### LINKED ISSUE #1292: Upload image on bugform layout busted
This is on master, but not deployed (just noticed on staging....).

Seems like a regression from https://github.com/webcompat/webcompat.com/pull/1265. @karlcow want to fix?

screen shot 2017-01-18 at 10 24 05 am

karlcow:
1. reintroduces the overflow
2. handles the margin for the actions button
3. use flexbox for the main form.
4. introduces the margin scoped to comments for the upload zone.

capture d ecran 2017-01-19 a 12 54 52

capture d ecran 2017-01-19 a 12 55 31

### DIFF
--- a/webcompat/static/css/development/components/report-form.css
+++ b/webcompat/static/css/development/components/report-form.css
@@ -15,6 +15,7 @@
     text-align: center;
     justify-content: center;
     flex-wrap: wrap;
+    margin-top: 1em;
   }
 
     @media all and (min-width: 30.75em) {
--- a/webcompat/static/css/development/components/report-form.css
+++ b/webcompat/static/css/development/components/report-form.css
@@ -18,6 +18,18 @@
     margin-top: 1em;
   }
 
+  .wc-ReportForm .wc-UploadForm {
+    border: dashed 1px rgb(204, 204, 204);
+    border-radius: 20px;
+  }
+
+  .wc-ReportForm .r-Grid-cell {
+    margin-bottom: 1em;
+  }
+  .wc-ReportForm div.r-minM--1of2:nth-child(2) .wc-Form-group {
+    margin-bottom: 2em;
+  }
+
     @media all and (min-width: 30.75em) {
 
       /* space between button */
--- a/webcompat/static/css/development/components/comment.css
+++ b/webcompat/static/css/development/components/comment.css
@@ -13,6 +13,10 @@
   box-shadow: 0 1px 3px rgba(0,0,0, 0.10), 0 1px 2px rgba(0,0,0, 0.25);
 }
 
+  .wc-Comment .wc-UploadForm-wrapper {
+    margin: 1em;
+  }
+
   /* space */
   .wc-Comment + .wc-Comment {
     margin-top: 1em;
--- a/webcompat/static/css/development/components/comment.css
+++ b/webcompat/static/css/development/components/comment.css
@@ -15,6 +15,8 @@
 
   .wc-Comment .wc-UploadForm-wrapper {
     margin: 1em;
+    border: 2px dashed #CCCCCC;
+    border-radius: 15px;
   }
 
   /* space */
--- a/webcompat/static/css/development/components/upload-form.css
+++ b/webcompat/static/css/development/components/upload-form.css
@@ -5,12 +5,13 @@
 }
 
 .wc-UploadForm {
-  display: block;
+  display: flex;
   position: relative;
   text-align: center;
   min-height: 250px;
   cursor: pointer;
   z-index: 3;
+  overflow: hidden;
 }
 
   /* modifier new comment of existing bug */
@@ -23,13 +24,12 @@
   /* Form upload wrapper */
   .wc-UploadForm-wrapper {
     z-index: 2;
-    margin-top: 1rem;
     padding: 1rem;
     border: 2px dashed #CCCCCC;
     border-radius: 15px;
     align-content: center;
     align-items: center;
-    margin: 1rem auto;
+    flex: 1 0 0;
   }
     /* state */
     .wc-UploadForm-wrapper.is-hidden {
--- a/webcompat/static/css/development/components/upload-form.css
+++ b/webcompat/static/css/development/components/upload-form.css
@@ -25,12 +25,11 @@
   .wc-UploadForm-wrapper {
     z-index: 2;
     padding: 1rem;
-    border: 2px dashed #CCCCCC;
-    border-radius: 15px;
     align-content: center;
     align-items: center;
     flex: 1 0 0;
   }
+
     /* state */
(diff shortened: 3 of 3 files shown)


magsout:
heu @karlcow regression ? How is possible ? with some style ? O_O

karlcow:
heh @magsout with the previous CSS PR, I broke the home page form by removing some essential things in the css ;)

magsout:
oups, who is the reviewer ? 😝

miketaylr:
Are the changes in padding/spacing of the form elements intentional?

webcompat.com:
screen shot 2017-01-19 at 10 49 44 am

this PR branch:
screen shot 2017-01-19 at 10 49 52 am

Also, this border-radius is missing from the `label.wc-UploadForm` that was there before:

webcompat.com:
screen shot 2017-01-19 at 10 50 05 am

this PR branch:
screen shot 2017-01-19 at 10 50 16 am

magsout:
@karlcow sorry for my previous review

karlcow:
ok let's see if this one is working. :) 

capture d ecran 2017-01-20 a 15 51 50

r? @magsout @miketaylr 
and sorry for butchering the CSS.

miketaylr:
I'll leave @magsout to review CSS, but LGTM in terms of functionality after re-testing locally. Nice work @karlcow.

miketaylr:
One issue we might consider filing as a follow up bug: the layout of the report buttons seems a bit wonky. To be clear though, your branch makes it look *less* weird... but alignment feels a bit off.

This branch:
screen shot 2017-01-20 at 1 46 41 pm

Production:
screen shot 2017-01-20 at 1 47 38 pm

karlcow:
agreed. Honestly, I think the full form should be redone. I have advocated for this for a bit of time now. It should be on its own page and it should work more on a step by step process. :) 

BUT that's a longer work to design and think about.

zoepage:
@karlcow 
> I think the full form should be redone. [...] It should be on its own page and it should work more on a step by step process. :)

@lockettm what do you think? Should we re-do the form now or keep it like it is and work in it later?

karlcow:
Let's not reply/discuss about this in this pull request. My bad for hinting at it.
There is an already existing issue for this #1007

webcompat/static/css/development/components/report-form.css
```diff
@@ -15,6 +15,19 @@
...
     flex-wrap: wrap;
+    margin-top: 1em;
+  }
+
+  .wc-ReportForm .wc-UploadForm {
+    border: dashed 1px rgb(204, 204, 204);
+    border-radius: 20px;
+  }
+
+  .wc-ReportForm .r-Grid-cell {
```
magsout: I'm not a big fan to this kind of selector. Each component is independ to avoid side effect. We can add ne wrapper to handle it.

webcompat/static/css/development/components/report-form.css
```diff
@@ -15,6 +15,19 @@
...
+    margin-bottom: 1em;
+  }
+  .wc-ReportForm div.r-minM--1of2:nth-child(2) .wc-Form-group {
```
magsout: Same here .

webcompat/static/css/development/components/report-form.css
karlcow: ah. :) I would like to avoid another div/span. They drive me already crazy in the current templates. Can we just put a specific class attribute?

webcompat/static/css/development/components/report-form.css
karlcow: here definitely agree with you. But instead of another wrapper (aka div or span if I understood you), I would prefer a class attribute.

webcompat/static/css/development/components/report-form.css
karlcow: re-reading my comment 🤡 I feel we are entering a style/css philosophical discussion. So to avoid this, I would encourage @magsout to suggest a way to organize it as he's the owner and master of the templates. 👍

magsout: @karlcow heh, yes I'll do it.

webcompat/static/css/development/components/report-form.css
magsout: @karlcow two `class` in same div is too bad for me, same side effect than a selector.

```html
div class="wc-ReprortForm r-Grid-cell"></div>
```

For this example, a new class modifier for `r-Grid` is too specfic. So it's why I think it's better to add another wrapper.

But Maybe I can simplfiy the HTML.

magsout: @karlcow two `class` in same div is too bad for me, same side effect than a selector.

```html
div class="wc-ReprortForm r-Grid-cell"></div>
```

For this example, a new class modifier for `r-Grid` is too specfic. So it's why I think it's better to add another wrapper.

But Maybe I can simplfiy the HTML.

karlcow: when I was starting to touch to the form, I had the feeling to move everything to flexbox. with a very minimal markup. but it seems the `r-Grid-cell` is a css framework used for the forms? or site? Not sure. I didn't want to break everything given that I'm just occasionally editing it.

webcompat/static/css/development/components/report-form.css
karlcow: I'm happy to implement it too. :) if you tell me what I should do. As you prefer.

webcompat/static/css/development/components/report-form.css
magsout: Good idea for flexbox. 'Grid ´ is a small grid component  https://github.com/cssrecipes/grid

miketaylr:
@karlcow @magsout what's the status of this PR? Is someone working on the review issues? (hard to tell who is, based on comments).

If we're just blocked on adding or not adding a wrapper, I would suggest landing it with the wrapper and filing a bug to re-think or remove the wrapper div. That way we can move forward. 

We're currently blocked on release (https://github.com/webcompat/webcompat.com/issues?q=label%3A%22status%3A+add+to+changelog%22+is%3Aclosed) until this gets merged.

magsout:
@miketaylr we can merged, that's ok for me. 

@karlcow ?

karlcow:
@miketaylr that the current solution is working. Not perfect in terms of syntax. So we could release it. And then add a wrapper or a more overall solution such as a flexbox.

karlcow:
ah! reload baby ;) before replying. So yes like @magsout ship it!
and we will fix it in the next iteration. :)

miketaylr:
OK, let's do that. Ship, then file a bug to clean it up (w/ @magsout's guidance). I'll do both a bit later today/tomorrow. Thanks to both of you. ❤️

PR #1301, opened 2017-01-24
Fixes #1289. Remove user from Sessions.db User table when logging out

### LINKED ISSUE #1289: Submit issue "via Github" doesn't work when logging in the first time
1. I typed in a site issue while not logged in.
2. Submitted the issue via the Github button
3. Authorize flow app on Github

Expected:

Issue submitted

Actual:

Redirect to webcompat site, issue is forgotten.

miketaylr:
This branch is currently on staging if you'd like to test (but has been verified by bug reporter).

r? @karlcow

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -22,9 +22,11 @@ RATELIMIT_STORAGE_URL = 'memory://'
 RATELIMIT_STRATEGY = 'moving-window'
 
 DEBUG = False
+SESSION_COOKIE_SECURE = True
 
 if not PRODUCTION:
     DEBUG = True
+    SESSION_COOKIE_SECURE = False
 
 # Logging Capabilities
 # To benefit from the logging, you may want to add:
--- a/webcompat/db/__init__.py
+++ b/webcompat/db/__init__.py
@@ -13,15 +13,8 @@ from sqlalchemy import String
 from sqlalchemy.orm import scoped_session
 from sqlalchemy.orm import sessionmaker
 from hashlib import sha512
-from uuid import uuid4
 
 from webcompat import app
-
-session_engine = create_engine('sqlite:///' + os.path.join(
-    app.config['BASE_DIR'], 'session.db'))
-session_db = scoped_session(sessionmaker(autocommit=False,
-                                         autoflush=False,
-                                         bind=session_engine))
 
 issue_engine = create_engine('sqlite:///' + os.path.join(
     app.config['BASE_DIR'], 'issues.db'))
@@ -47,19 +40,3 @@ class WCIssue(IssueBase):
         self.body = body
 
 IssueBase.metadata.create_all(bind=issue_engine)
-
-UsersBase = declarative_base()
-UsersBase.query = session_db.query_property()
-
-
-class User(UsersBase):
-    __tablename__ = 'users'
-
-    user_id = Column(Integer, primary_key=True)
-    access_token = Column(String(128), unique=True)
-
-    def __init__(self, access_token):
-        self.access_token = access_token
-
-
-UsersBase.metadata.create_all(bind=session_engine)
--- a/webcompat/db/__init__.py
+++ b/webcompat/db/__init__.py
@@ -8,6 +8,7 @@ import os
 from sqlalchemy import create_engine
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import Column
+from sqlalchemy import Integer
 from sqlalchemy import String
 from sqlalchemy.orm import scoped_session
 from sqlalchemy.orm import sessionmaker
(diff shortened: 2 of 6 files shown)


karlcow:
@miketaylr 
Here a mechanism which is expiring sessions.
https://github.com/fengsp/flask-session/blob/master/flask_session/sessions.py#L516-L520

Could we ask someone from mozilla with a good security background to review it?

miketaylr:
> The thing which worries me a bit is now there is an incremented value which can be known and stays in the DB if the user is not logging out.

Can you clarify what you mean by "can be known"?

> The session will stay there for ever happily after and the DB will grow. I wonder if there should be an expiration date on the DB for the session, aka new params which are creation_date and last_access and determine if we need to clean up the old session tokens with a specific clean up function (activated when running python run.py or at regular time).

Yeah, good point. Note this is currently the case as well. 

Re-thinking this a bit... perhaps there is no need for a DB table at all, and we only use the `session` cookie.

miketaylr:
OK, @karlcow, re-thinking this. The Session.db, and its User table aren't really needed if we store everything in the [`SecureCookie`][1] session. We don't need `user_id`, because we were just checking if it existed as a proxy for `logged_in`. 

A user would be able to modify their own cookie, but it would fail silently unless they had the secret key used to decrypt the secure cookie. (If they have that, game over, they're inside the server.)

This is what your session cookie looks like on the client. 
screen shot 2017-01-25 at 5 27 55 pm

(`Secure: false` is because localhost and staging are not https)


So, this branch adds a `logged_in` session key. And the user token is stored when logged in, and deleted when logged out. See https://github.com/webcompat/webcompat.com/pull/1301/commits/abd3550f27d06b29c412bfdbef2d85f40c2a54f6.

(Note to self, get staging on https)

[1]: http://flask.pocoo.org/docs/0.12/api/#flask.sessions.SecureCookieSessionInterface



There's also the following: http://flask.pocoo.org/snippets/51/ if we wanted to feel more secure.

karlcow:
> There's also the following: http://flask.pocoo.org/snippets/51/ if we wanted to feel more secure.

The snippet has been published in 2011 for Flask 0.8.
In the current version of Flask, it is now the default behavior.

For example in 
https://github.com/pallets/flask/blob/0e79aba40d2497218736448ced708fcf4f8943b3/flask/sessions.py#L294

config/__init__.py
```diff
@@ -22,9 +22,11 @@
 RATELIMIT_STRATEGY = 'moving-window'
 
 DEBUG = False
+SESSION_COOKIE_SECURE = True
 
 if not PRODUCTION:
```
karlcow: Don't you want to extend that to staging too? or change it to `if LOCALHOST:` 
because we introduce a security risk for users.

webcompat/db/__init__.py
```diff
@@ -8,20 +8,14 @@
 from sqlalchemy import create_engine
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import Column
+from sqlalchemy import Integer
```
karlcow: to remove, not needed anymore.

webcompat/db/__init__.py
```diff
@@ -8,20 +8,14 @@
...
 from sqlalchemy import String
 from sqlalchemy.orm import scoped_session
 from sqlalchemy.orm import sessionmaker
 from hashlib import sha512
```
karlcow: to remove `from hashlib import sha512` not used anymore.

webcompat/db/__init__.py
```diff
@@ -46,23 +40,3 @@ def __init__(self, issue_id, summary, url, body):
         self.body = body
 
 IssueBase.metadata.create_all(bind=issue_engine)
```
karlcow: nit: two blank lines after a class (or above `IssueBase.metadata.…` in this case) ;)

karlcow: Maybe an opportunity to document also `class WCIssue(IssueBase)`

webcompat/views.py
```diff
@@ -29,35 +29,21 @@
 from webcompat import app
```
karlcow: not your code but `import urllib` is not used anymore. To scrap.

config/__init__.py
miketaylr: I think until https://github.com/webcompat/webcompat.com/issues/1303 is done, having a secure cookie on non-HTTPS site won't do what we want. AFAIU, the browser won't send the cookie unless it's talking to a TLS endpoint:

https://tools.ietf.org/html/rfc6265#section-4.1.2.5

webcompat/db/__init__.py
miketaylr: noted: thanks.

webcompat/db/__init__.py
miketaylr: Yeah, also maybe an opportunity to remove it altogether since nobody uses it. ^__^

But that can be a separate discussion (and we will want something like it in the future).

miketaylr:
> Now what I just did by having access to the cookie in the browser and copying it manually.
> 
> I edited out the values
> 
> >>> import base64
> >>> import zlib
> >>> zlib.decompress(base64.urlsafe_b64decode('here_the_cookie_string_in_between_the_two_periods==='))
> '{"access_token":"******************","avatar_url":"https://avatars.githubusercontent.com/u/505230?v=3","csrf_token":{" b":"******************=="},"logged_in":true,"username":"karlcow"}'
> 
> Courtesy of https://www.youtube.com/watch?v=mhcnBTDLxCI

Hmm. If somehow cookies are intercepted then yeah, it's possible to steal a token for a single user. That's bad. Let's give up on this approach. I'm gonna look at Flask-Session or go back to the old way and have some kind of automatic session expiry (1 month?).

miketaylr:
Let's close this PR, I'll send another branch (but I'll come back to the suggested comments). I should set up HTTP on staging as well so we can enable secure session cookies at the same time, even if it's only holding a URI to an avatar and a username.

PR #1331, opened 2017-02-07
Fixed #1155 - added padding to wc-Link

### LINKED ISSUE #1155: "GitHub view" text rendering problem on issues list page
On both the [issues list](https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc) and [issue](https://webcompat.com/issues/2938) pages there's a paragraph at the bottom:

> takes you to the GitHub view of this page.

On the list page it renders without proper spacing between the words, like:

> takes you to theGitHub viewof this page.

image

That's in Firefox 48 / Chrome 51 Win 7.

It's a result of the `display: flex` styling (looks like that was introduced in #901). One solution would be to apply margin to the GitHub View link. However, perhaps it's better to just remove the `display: flex`? The paragraph at the bottom of the issue template doesn't have it -- it's just the default `display: block`. It seems like the paragraph at the bottom of the issue list template with that and `text-align: right` would be fine, but is there some reason it's important for that to have flex layout?

zoepage:
r? @magsout

### DIFF
--- a/webcompat/static/css/development/components/search-issue.css
+++ b/webcompat/static/css/development/components/search-issue.css
@@ -68,3 +68,7 @@
     align-items: center;
     margin: 3em 0 0 0;
   }
+
+  .wc-SearchIssue-github .wc-Link { 
+    padding: 0 0.3em; 
+  }


webcompat/static/css/development/components/search-issue.css
```diff
@@ -68,3 +68,7 @@
     align-items: center;
     margin: 3em 0 0 0;
   }
+
+  .wc-SearchIssue-github .wc-Link {
```
magsout: IMHO this selector is dangerous because of side effect.

For me, a selector has unique Component (and 2/3 max selector), or can target a simple element.

I prefer `.wc-SearchIssue-github a {}`  or you can write another wrapper aroud the link ``

zoepage: Me personally don't like to add styles to tags. For me, this is a more dangerous side effect then adding a class. 

While `.wc-SearchIssue-github` seems like a closed module used just for the footer information, so adding an already existing class should make it independent?

magsout:
Seen together on IRC, we merge and change in the future as needed

magsout:
thanks @zoepage

PR #1333, opened 2017-02-07
Fixes #705. Update privacy policy around report logging and deletion.

### LINKED ISSUE #705: Update privacy policy with regards to logging for security and performance issues
We landed #245 which adds capabilities for logging, we need to update the privacy policy if we want to log things related to the user such as an ip address.

miketaylr:
Log rotation is set up on the server now (#717), so we're ready to update the privacy policy.

r? @zoepage

### DIFF
--- a/webcompat/templates/privacy.html
+++ b/webcompat/templates/privacy.html
@@ -4,7 +4,7 @@
   <main class="wc-UIContent" role="main">
     <section class="wc-UISection wc-UISection">
       <h1 class="wc-Title wc-Title--l">Webcompat.com Privacy Policy</h1>
-      <small>Last Updated: August 21, 2015</small>
+      <small>Last Updated: February 7, 2017</small>
       <h2 class="wc-Title wc-Title--s">About Us</h2>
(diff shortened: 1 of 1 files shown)


webcompat/templates/privacy.html
```diff
@@ -15,6 +15,8 @@ <h2 class="wc-Title wc-Title--s">Reporting Anonymously</h2>
       <p>If you choose the anonymous option, the only information that we’ll receive is the content of your report.</p>
       <h2 class="wc-Title wc-Title--s">Reporting via GitHub</h2>
       <p>If you choose the GitHub option, webcompat.com will receive your GitHub username and avatar.  You can revoke our access to this information at any time from the <a class="wc-Link" href="https://github.com/settings/applications">"Applications" section of your profile page</a>.</p>
+      <h2>Website Logs</h2>
+      <p>When a bug report is filed on webcompat.com we log your IP address, the time at which the bug was filed, and the URI of the bug report. This data may be used to investigate potential security issues and spam efforts. These logs are only kept for 14 days, after which they are deleted.</p>
```
zoepage: > after which they are deleted.

after which they are deleted automatically.

miketaylr: Good suggestion.

miketaylr:
Thanks for review!

Issue #1254, opened 2016-12-23
Remove support for X-Reported-With header


miketaylr:
And only support a `src=` param, to make things simple. Before we can do this we need to update all the add-ons and get them published.

karlcow:
How do we treat legacy? Aka add-on deployed in the wild but not updated.

karlcow:
Also I guess this is linked to #1249 or am I confused 
`X-Requested-With` here in #1254 and `X-Reported-With` in #1249

miketaylr:
> How do we treat legacy? Aka add-on deployed in the wild but not updated.

So, it turns out I never published the addons that sent `X-Reported-With` -- I dropped the ball on that. I did just publish versions 0.5 of Firefox Desktop and Chrome that send `&src=reporter-addon-chrome`, etc. yesterday.

The odds that someone built the addon from the repo for the few weeks that code existed are probably low enough to not worry about.

(Also, I just keep mixing it up, it's actually just `X-Reported-With`)

miketaylr:
Here's the relevant code to change: https://github.com/webcompat/webcompat.com/blob/274dc8c63aa506baceb235cfeb9ec26e1cdcb4b1/webcompat/views.py#L181

@zoepage lmk if you have any questions!

zoepage:
Closed after PR merge: https://github.com/webcompat/webcompat.com/pull/1334

PR #1334, opened 2017-02-07
Fixing #1254 - Remove support for X-Reported-With header

zoepage:
r? @miketaylr

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -179,14 +179,7 @@ def create_issue():
             flash(msg, 'notimeout')
             return redirect(url_for('index'))
     form['ua_header'] = request.headers.get('User-Agent')
-    # Currently we support either a src GET param, or X-Reported-With header
-    # to track where the report originated from.
-    # See https://github.com/webcompat/webcompat.com/issues/1254 to track
-    # supporting only the src param
-    if session.get('src'):
-        form['reported_with'] = session.pop('src')
-    else:
-        form['reported_with'] = request.headers.get('X-Reported-With', 'web')
+    form['reported_with'] = session.pop('src')
     # Logging the ip and url for investigation
     log = app.logger
     log.setLevel(logging.INFO)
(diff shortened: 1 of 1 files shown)


miketaylr:
In terms of fixing the issue, you nailed it. I have a small Python doubt (which means a bug I think I introduced... will comment, because you can probably fix it in this PR).

webcompat/views.py
```diff
@@ -179,14 +179,7 @@ def create_issue():
...
     form['ua_header'] = request.headers.get('User-Agent')
-    # Currently we support either a src GET param, or X-Reported-With header
-    # to track where the report originated from.
-    # See https://github.com/webcompat/webcompat.com/issues/1254 to track
-    # supporting only the src param
-    if session.get('src'):
-        form['reported_with'] = session.pop('src')
-    else:
-        form['reported_with'] = request.headers.get('X-Reported-With', 'web')
+    form['reported_with'] = session.pop('src')
```
miketaylr: So `src` will end up in the `session` if there was a `src` param:

https://github.com/webcompat/webcompat.com/blob/a5b5d84a06cbf1564247208ac8bf8e825134326d/webcompat/views.py#L166

(for all add-on and "report site issue "reported issues, that gets added for the user).

If this was a web-reported issue, `src` won't be in the session and a `KeyError` will be raised (which means the page will explode, as is).

[`dict#pop`](https://docs.python.org/2/library/stdtypes.html#dict.pop) does allow you to specify a default value if the key isn't there though, as the second arg. 

So I think we want `session.pop('src', 'web')` -- which should mean if there wasn't a `src` param added by an add-on, we should set `reported_with` to `web`.

Does that make sense, @zoepage? Anything I'm overlooking?

zoepage: @miketaylr 

so basically `form['reported_with'] = session.pop('src')` in line 182 would get the additional param, right? (Just to be sure and don't waste your time :) )

miketaylr:
r+, thanks!

webcompat/views.py
karlcow: @zoepage For a descriptive example of what @miketaylr is talking about
https://github.com/webcompat/webcompat.com/pull/1255#discussion_r93997278

PR #1340, opened 2017-02-13
Fixes #999 - Add domain name search

### LINKED ISSUE #999: [feature] domain name search for webcompat.com
Probably related to #865
But could work on his own.

if we had a keyword `domain` or `site`, we could interesting stuff in terms of search

`site:google.com` will pull out all the bugs about google.com. How to make it work right now?

using Github's `is:issue google.com in:title` 

Example: https://github.com/webcompat/web-bugs/issues?utf8=%E2%9C%93&q=is%3Aissue+google.com+in%3Atitle

MDTsai:
Replace site:DOMAIN_NAME to DOMAIN_NAME in:title query string
Utilize github title search function do domain name search.
@miketaylr could you give a review for this?

### DIFF
--- a/webcompat/static/js/lib/models/issue.js
+++ b/webcompat/static/js/lib/models/issue.js
@@ -213,6 +213,7 @@ issueList.IssueCollection = Backbone.Collection.extend({
       creator:   'author',
       mentioned: 'mentions'
     };
+    var sitesearchRegExp = /site:([\w-\.]+(:\d+)?)/g;
 
     if (_.isString(paramsArg)) {
       var paramsArray = _.uniq(paramsArg.split('&'));
@@ -229,6 +230,12 @@ issueList.IssueCollection = Backbone.Collection.extend({
     if ('direction' in params) {
       params.order = params.direction;
       delete params.direction;
+    }
+
+    // Support domain name search
+    // Replace all site:DOMAIN_NAME to DOMAIN_NAME in:title
+    if (params.q.match(sitesearchRegExp)) {
+      params.q = params.q.replace(sitesearchRegExp, "$1 in:title");
     }
 
     // The rest need to be added to the "q" param as substrings


webcompat/static/js/lib/models/issue.js
```diff
@@ -231,6 +231,12 @@ issueList.IssueCollection = Backbone.Collection.extend({
       delete params.direction;
     }
 
+    // Issue 999, support domain name search
```
miketaylr: nit: We don't typically leave bug numbers for features we're adding, just for bugs that still need to be fixed.

webcompat/static/js/lib/models/issue.js
```diff
@@ -231,6 +231,12 @@ issueList.IssueCollection = Backbone.Collection.extend({
...
+    // Replace all site:DOMAIN_NAME to DOMAIN_NAME in:title
+    if (!!params.q.match(/site:([\w-\.]+(:\d+)?)/i)) {
```
miketaylr: Any reason to use `!!` to coerce to boolean?

`String.prototype.match` will return an array (truthy) if it finds a match, and null if it doesn't (falsey).

I could be missing something though!

webcompat/static/js/lib/models/issue.js
```diff
@@ -231,6 +231,12 @@ issueList.IssueCollection = Backbone.Collection.extend({
...
+      params.q = params.q.replace(/site:([\w-\.]+(:\d+)?)/g, "$1 in:title");
```
miketaylr: Can we create a local variable for the `/site:([\w-\.]+(:\d+)?)` regex, since we use it twice? That way if we need to tweak if, there's less of a chance or forgetting to change it in both places.

MDTsai:
@miketaylr  thanks for the review! I've updated with your comment, please see if this works.

miketaylr:
Sorry for the delay here @MDTsai  -- looks good! I found a few bugs when testing, but I'll file some follow up issues (including an issue to add a test for this, so we don't break it in the future).

PR #1355, opened 2017-02-22
Fixes #1244 - align 'please login' message left

### LINKED ISSUE #1244: centered "please login" looks strange
screen shot 2016-12-16 at 3 37 01 pm

If you're logged out and look at an issues page, the please login message is centered -- and looks strange, IMO. We should maybe left-align it w/ the issue container?

(Another possible option is to move it to the right hand side up top in another column, w/ the G for GitHub thing, I imagine we'd want labels to go there as well.).

zoepage:
#1244 

r? @magsout && @miketaylr

### DIFF
--- a/webcompat/static/css/development/components/issue.css
+++ b/webcompat/static/css/development/components/issue.css
@@ -65,21 +65,21 @@
   .wc-Issue-title {
     font-size: 1.2em;
     line-height: 1em;
-    margin:0 0 .5em 0;
+    margin: 0 0 .5em 0;
     word-wrap: break-word;
   }
 
   @media all and (max-width: 28.12em) {
 
     .wc-Issue-title {
-      font-size:1.2em;
+      font-size: 1.2em;
     }
   }
 
   /* information issue : date / author / number of comment */
   .wc-Issue-create {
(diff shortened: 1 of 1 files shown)


webcompat/static/css/development/components/issue.css
```diff
@@ -156,7 +156,7 @@
 /* Link to log in */
 .wc-Issue-login {
   margin-top: 2.5em;
-  text-align:center;
+  text-align:left;
```
miketaylr: Can we add a space between the prop and val here? (I know it was missing before...)

magsout: 👍

magsout:
r+ 👍

webcompat/static/css/development/components/issue.css
zoepage: added space and spaces to all other values in the css that were missing. :)

miketaylr:
> added space and spaces to all other values in the css that were missing. :)

Cool, thanks! In the future a whitespace change like that can be its own commit (since the other whitespace changes have nothing to do with the commit message).

PR #1361, opened 2017-02-24
Fixes #1347 - single quotes to double quotes for strings

### LINKED ISSUE #1347: eslint: change to double quotes for strings in JS
Single quotes in JS feels gross, I think I chose that for consistency with our Python style, but I don't like it anymore.

Unless somebody objects, I'm going to do submit a PR to do this in the next few days.

zoepage:
r? @miketaylr 

Please also test the grunt tasks / new npm scripts. Just to be sure I didn't break anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "setup": "npm run module && npm run virtualenv && npm install && npm run config",
     "watch" : "grunt watch",
     "build": "grunt",
+    "lint": "grunt eslint",
     "module": "git submodule init && git submodule update",
     "start": "source env/bin/activate && python run.py",
     "virtualenv" : "pip install virtualenv && virtualenv env && source env/bin/activate && npm run pip",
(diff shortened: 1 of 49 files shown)


miketaylr:
Awesome -- thank you!

package.json
```diff
@@ -43,6 +43,8 @@
     "setup": "npm run module && npm run virtualenv && npm install && npm run config",
     "watch" : "grunt watch",
     "build": "grunt",
+    "lint": "grunt eslint",
```
miketaylr: Can you file a follow up bug: our grunt eslint task should be checking all .js files and currently doesn't (my bad, probably). It looks like what you use for the fix command is more what we want.

miketaylr: @zoepage ah, if we document eslint anywhere in CONTRIBUTING.md, it might be good to add these commands as alternatives.

magsout:
@zoepage ❤👍

PR #1365, opened 2017-02-27
Fixes #1364 - add specific paths to js lint

### LINKED ISSUE #1364: add paths to linter
https://github.com/webcompat/webcompat.com/pull/1361#discussion_r103008192

zoepage:
#1364

Adding path to npm script to lint just specific js files.
- Gruntfile.js 
- tests/ 
- grunt-tasks/ 
- webcompat/static/js/lib/

r? @miketaylr


zoepage:
Intention for removing the eslint part of grunt was to start being less dependent on grunt as we'll get rid of it soon-ish. So, I'll add the task to the travis.yml.

(Will do a test push, so make sure it works and fails.)

miketaylr:
> Intention for removing the eslint part of grunt was to start being less dependent on grunt as we'll get rid of it soon-ish.

So will you remove that Grunt task in this PR?

zoepage:
not the build as we'd need to move to webpack for that. Just the grunt task for the linter. small steps. 

or do you think about it otherwise?

miketaylr:
I guess I don't like the idea of having two eslint commands that are linting different sets of files, even temporarily (a few months?).

Removing the `grunt:eslint` task seems like a simple way to fix that in this PR that w/o requiring a full webpack re-write. 

Just remove https://github.com/webcompat/webcompat.com/blob/master/grunt-tasks/eslint.js, and remove the task from https://github.com/webcompat/webcompat.com/blob/master/Gruntfile.js#L26 and https://github.com/webcompat/webcompat.com/blob/master/Gruntfile.js#L31.

And https://github.com/webcompat/webcompat.com/blob/28ac4928f56dec681349fa977d619da3d2888eac/package.json#L29 I guess.

Issue #1328, opened 2017-02-03
Move security headers into app


miketaylr:
I've added the following security headers to our nginx config. But we should move them into the app, IMO. That way we can easily enable or tweak certain values for certain assets (and do it in python).

```
add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
add_header X-Content-Type-Options "nosniff";
add_header X-XSS-Protection "1; mode=block";
add_header X-Frame-Options "DENY";
```

Relates to #763

PR #1367, opened 2017-02-27
Fixes #1329, #1328. Add a CSP reporting endpoint and move security headers into app.

### LINKED ISSUE #1329: Add CSP report endpoint
A place that accepts a CSP report and logs the input to disk. Blocks #763

miketaylr:
This will report CSP violations to `/tmp/webcompat-csp-reports.log` (and DevTools also will dump them for the page you're on). Let's let this run wild for a week or so then figure out what needs to change. 

(I already noticed the following, which points to lodash):
```
[Report Only] Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' https://www.google-analytics.com".
```

r? @karlcow

### DIFF
--- a/tests/test_urls.py
+++ b/tests/test_urls.py
@@ -93,6 +93,19 @@ class TestURLs(unittest.TestCase):
         # A random post should 401, only requests from GitHub will 200
         self.assertEqual(rv.status_code, 401)
 
+    def test_csp_report_uri(self):
+        '''Test POST to /csp-report w/ correct content-type returns 204.'''
+        headers = {'Content-Type': 'application/csp-report'}
+        rv = self.app.post('/csp-report', headers=headers)
+        self.assertEqual(rv.status_code, 204)
+
+    def test_csp_report_uri_bad_content_type(self):
+        '''Test POST w/ wrong content-type to /csp-report returns 400.'''
+        headers = {'Content-Type': 'application/json'}
+        rv = self.app.post('/csp-report', headers=headers)
+        self.assertNotEqual(rv.status_code, 204)
+        self.assertEqual(rv.status_code, 400)
+
     def test_tools_cssfixme(self):
         '''Test that the /tools/cssfixme route gets 200.'''
         rv = self.app.get('/tools/cssfixme')
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -315,3 +315,13 @@ def contributors():
 def cssfixme():
     '''Route for CSS Fix me tool'''
     return render_template('cssfixme.html')
+
+
+@app.route('/csp-report', methods=['POST'])
+def log_csp_report():
+    '''Route to record CSP header violations.'''
+    if 'application/csp-report' not in request.headers.get('content-type', ''):
+        return ('Wrong Content-Type.', 400)
+    with open(app.config['CSP_REPORTS_LOG'], 'a') as r:
+        r.write(request.data + '\n')
+    return ('', 204)
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -21,6 +21,7 @@ from flask import url_for
 
 from form import AUTH_REPORT
 from form import PROXY_REPORT
+from helpers import add_csp
 from helpers import add_sec_headers
 from helpers import cache_policy
 from helpers import get_browser_name
@@ -54,6 +55,7 @@ def before_request():
 def after_request(response):
     session_db.remove()
     add_sec_headers(response)
+    add_csp(response)
     return response
 
 
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -21,6 +21,7 @@ from flask import url_for
 
 from form import AUTH_REPORT
 from form import PROXY_REPORT
+from helpers import add_sec_headers
 from helpers import cache_policy
 from helpers import get_browser_name
 from helpers import get_form
@@ -52,6 +53,7 @@ def before_request():
 @app.after_request
 def after_request(response):
     session_db.remove()
+    add_sec_headers(response)
     return response
 
 
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -323,9 +323,18 @@ def cssfixme():
 
 @app.route('/csp-report', methods=['POST'])
 def log_csp_report():
-    '''Route to record CSP header violations.'''
-    if 'application/csp-report' not in request.headers.get('content-type', ''):
-        return ('Wrong Content-Type.', 400)
-    with open(app.config['CSP_REPORTS_LOG'], 'a') as r:
-        r.write(request.data + '\n')
-    return ('', 204)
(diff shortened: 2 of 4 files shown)


webcompat/views.py
```diff
@@ -306,3 +321,13 @@ def contributors():
...
+
+
+@app.route('/csp-report', methods=['POST'])
+def log_report():
+    '''Route to record CSP header violations.'''
+    if 'application/csp-report' not in request.headers.get('content-type'):
+        return
+    with open(app.config['CSP_REPORTS_LOG'], 'a') as r:
+        r.write(request.data + '\n')
+    return ('', 204)
```
miketaylr: I don't know if `204` is the best here. ❓

miketaylr:
(just pushed 2 simple tests and a return value for the non-correct content-type case)

webcompat/views.py
```diff
@@ -306,3 +321,13 @@ def contributors():
 def cssfixme():
     '''Route for CSS Fix me tool'''
     return render_template('cssfixme.html')
+
+
+@app.route('/csp-report', methods=['POST'])
+def log_report():
```
karlcow: nit: maybe log_csp_report()

webcompat/views.py
```diff
@@ -306,3 +321,13 @@ def contributors():
...
+    '''Route to record CSP header violations.'''
+    if 'application/csp-report' not in request.headers.get('content-type'):
```
karlcow:
```python
request.headers.get('content-type', '')
```

to avoid people trying to hit the server with wrong request and generating 500.

webcompat/views.py
```diff
@@ -306,3 +321,13 @@ def contributors():
...
+        return ('Wrong Content-Type.', 400)
+    with open(app.config['CSP_REPORTS_LOG'], 'a') as r:
+        r.write(request.data + '\n')
```
karlcow: nothing weird but you could have used the logging module too which we instantiate.

webcompat/views.py
```diff
@@ -51,6 +51,21 @@ def before_request():
 @app.after_request
 def after_request(response):
     session_db.remove()
+    # Security headers
```
karlcow: I would put the full csp thing in a specific function: `add_csp(response)` which returns `response` with the new headers, that I would call here.

webcompat/views.py
```diff
@@ -306,3 +321,13 @@ def contributors():
...
+
+
+@app.route('/csp-report', methods=['POST'])
+def log_report():
+    '''Route to record CSP header violations.'''
+    if 'application/csp-report' not in request.headers.get('content-type'):
+        return ('Wrong Content-Type.', 400)
+    with open(app.config['CSP_REPORTS_LOG'], 'a') as r:
+        r.write(request.data + '\n')
+    return ('', 204)
```
karlcow: [204](https://tools.ietf.org/html/rfc7231#section-6.3.5) means 'No content' and is indeed the appropriate response.

webcompat/views.py
karlcow: Another reason to do that is with the combination of a logging file configuration to allow for activating and deactivating the feature when needed and have timestamp when the log is happening. 

```python
CSP_LOG = True # or False
if CSP_LOG:
    csp_log.info(request.data)
```

karlcow: when CSP_LOG is False. You can just send a [403](https://tools.ietf.org/html/rfc7231#section-6.5.3).

tests/test_urls.py
```diff
@@ -93,6 +93,17 @@ def test_labeler_webhook(self):
         # A random post should 401, only requests from GitHub will 200
         self.assertEqual(rv.status_code, 401)
 
+    def test_csp_report_uri(self):
```
karlcow: Missing comment, explaining the test.

tests/test_urls.py
```diff
@@ -93,6 +93,17 @@ def test_labeler_webhook(self):
...
+        headers = {'Content-Type': 'application/csp-report'}
+        rv = self.app.post('/csp-report', headers=headers)
+        self.assertEqual(rv.status_code, 204)
+
+    def test_csp_report_uri_bad_content_type(self):
```
karlcow: missing comment explaining the test.

karlcow:
nosetests -v 

```
…
test_csp_report_uri (tests.test_urls.TestURLs) ... ok
test_csp_report_uri_bad_content_type (tests.test_urls.TestURLs) ... ok
…
```

tests/test_urls.py
miketaylr: oops!

miketaylr: done.

tests/test_urls.py
miketaylr: done.

webcompat/views.py
miketaylr: Good point, let's do this.

webcompat/views.py
miketaylr: done!

webcompat/views.py
miketaylr: good point, done.

webcompat/views.py
miketaylr: Thanks, agree that's less messy. I'll define add_csp in helpers.py.

miketaylr: ...and add_sec_headers.

miketaylr:
Have addressed all feedback except the CSP_LOG part -- will get to that soon.

> This branch has conflicts that must be resolved

Looks like I need to rebase and fix conflicts as well.

webcompat/views.py
miketaylr: @karlcow one question:

Currently the default `file_handler` is set to `logging.INFO` level, which means anything `INFO` and above will get logged. For CSP, probably something like `logging.WARNING` is appropriate... but that means CSP logs will pollute the regular `webcompat.log`. Which I think we don't want.

The current dumb approach writes to a its own log file, so it avoids this problem.

Is there something obvious I'm missing? Admittedly, the logging module is confusing. 😕

karlcow: So my suggestion was a bad idea, given the context of flask which is exploding the normal logging python module. The [logging module](https://docs.python.org/2/library/logging.html) is ["simple"](https://pymotw.com/3/logging/index.html), but flask makes it more difficult. Cf. https://github.com/pallets/flask/issues/2023

so let's go with your initial suggestion.

miketaylr:
@karlcow ok, let's go back to the original log file.

r? for https://github.com/webcompat/webcompat.com/pull/1367/commits/c8065f7b0594e84ed7aa795b5437638c378d1359

karlcow:
ok r+ @miketaylr

miketaylr:
Thanks @karlcow.

PR #1379, opened 2017-03-02
Fixes #1358: Add a details param to put info in description textarea.

### LINKED ISSUE #1358: Add a GET param to pre-populate extra details inside issue description
(See #1356 for discussion of Media team's needs.)

miketaylr:
r? @zoepage

### DIFF
--- a/tests/functional/reporting-non-auth.js
+++ b/tests/functional/reporting-non-auth.js
@@ -14,6 +14,7 @@ define([
   var cwd = intern.config.basePath;
   var VALID_IMAGE_PATH = path.join(cwd, "tests/fixtures", "green_square.png");
   var BAD_IMAGE_PATH = path.join(cwd, "tests/fixtures", "evil.py");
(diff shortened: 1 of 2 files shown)


zoepage:
@miketaylr 
The tests related to the changes pass. Code matches the requests in the issue.
Just one question: Why did you remove the check in bugform.js? 
Otherwise it looks good to me.  👍 



General:
Locally 28 tests fail. In the Travis build 7.
But that's probably a different bug.
https://gist.github.com/zoepage/f4e07f5bcb544d124fe4d626c0de6e3c

miketaylr:
> Just one question: Why did you remove the check in bugform.js?

If you mean https://github.com/webcompat/webcompat.com/pull/1379/files#diff-ce241059c28a13e03bcd0829da8d4eb7L142, we don't use the `?open=1` param in the addons/browser button anymore (we send users to `/issues/new` instead) so the code in `checkParams` would never run.

> Locally 28 tests fail. In the Travis build 7.

Yeah, different bugs. That's the point of the next Outreachy round -- figure out what's up with intermittents (they all pass on my machine).

miketaylr:
> Otherwise it looks good to me. 👍

Based on that comment, let's merge!

miketaylr:
Thanks for review @zoepage!

PR #1392, opened 2017-03-06
Fixes #1370 - Add docstring to describe tests

### LINKED ISSUE #1370: Add comments describing tests in test_uploads.py
To fix this, the following methods need a docstring:

`testBase64ScreenshotUploads`: https://github.com/webcompat/webcompat.com/blob/master/tests/test_uploads.py#L118
`testRegularUploads`: https://github.com/webcompat/webcompat.com/blob/master/tests/test_uploads.py#L82

To know that this is fixed, the output from the following command won't show the test method names, but the docstring:

```
🐓 nosetests -v tests/test_uploads.py
testBase64ScreenshotUploads (tests.test_uploads.TestUploads) ... ok
testRegularUploads (tests.test_uploads.TestUploads) ... ok

Ran 3 tests in 0.125s
```

sergiorojasa:


tests/test_uploads.py
```diff
@@ -80,6 +80,7 @@ def testGet(self):
         self.assertEqual(rv.status_code, 404)
 
     def testRegularUploads(self):
+        '''Test that regular file should not be possible to upload.'''
```
miketaylr: This isn't quite accurate, if you look at what the code is doing.

What it does (in a loop), is try to upload some *bad* files  and some *good* files, and then asserts that it got the expected HTTP response code.

Unfortunately, there is a comment 2 lines down that says the wrong thing, so this is my fault.

tests/test_uploads.py
```diff
@@ -80,6 +80,7 @@ def testGet(self):
...
         # Loop over some files and the status codes that we are expecting
         # Basically it should never be possible to upload a "regular" file.
```
miketaylr: Can you remove this comment? It's just wrong. Possibly leftover from an earlier version of this test.

tests/test_uploads.py
```diff
@@ -116,6 +117,7 @@ def files(self):
             self.assertEqual(rv.status_code, status_code)
 
     def testBase64ScreenshotUploads(self):
+        '''Test should check screenshot is Base64.'''
```
miketaylr: This isn't quite right as well. 

`self.assertEqual(rv.status_code, status_code)` is checking that we get the expected return status for some bad base64 images, and a good one.

tests/test_uploads.py
miketaylr: (to be more clear, since it's hard to know which comment I mean...

`Basically it should never be possible to upload a "regular" file` is the bad comment)

tests/test_uploads.py
```diff
@@ -80,8 +80,8 @@ def testGet(self):
         self.assertEqual(rv.status_code, 404)
 
     def testRegularUploads(self):
+        '''Test that uploaded file return the expected status code.'''
```
miketaylr: Looks better! One small grammar nit: `file return` should probably be `files return`.

tests/test_uploads.py
```diff
@@ -116,6 +116,7 @@ def files(self):
             self.assertEqual(rv.status_code, status_code)
 
     def testBase64ScreenshotUploads(self):
+        '''Test that uploaded Base64 screenshot return the expected status code.'''
```
miketaylr: Same grammar nit: `screenshots return`.

miketaylr:
Great, thank you!

miketaylr:
I rebased the commits (and tweaked the commit message to be a bit more descriptive):

`git rebase -i HEAD~4`

From the rebase dialog, I did the following:

screen shot 2017-03-07 at 1 11 49 pm

Then pushed that back up to this branch:

`git push -f`

miketaylr:
Oops, my bad. I should have checked on the Pep8 errors.

miketaylr:
Pushed a follow up commit @ 3266a7cfafccc18e72ebcf2af9855c7dfb77643a.

PR #1395, opened 2017-03-06
Fixes #1378: Add alternative shell command to activate virtualenv

### LINKED ISSUE #1378: Dev env setup: "source: not found" error with "npm start"
Hi! I am trying to set up the dev environment and got as far as 
[x] git clone webcompat.com
[x] npm run module

Then `npm run setup` didn't work so I ran the commands individually:  
[x] virtualenv env
[x] . env/bin/activate
[x] pip install -r config/requirements.txt
[x] sudo npm install -g grunt-cli
[x] npm install
*edit: I forgot to run grunt here! That's why the server couldn't find the css. -_- Oops.*
[x] cp config/secrets.py.example config/secrets.py

Then made a test repo, filled in OAUTH_TOKEN and github client secret & id.

Then I ran `python run.py` and while it serves the site to localhost:5000, the CSS didn't seem to be applying - I found in the browser console that it couldn't locate some files in css and api dirs:

```
http://localhost:5000/css/webcompat.dev.css
http://localhost:5000/api/issues/labels?per_page=100
http://localhost:5000/api/issues/category/needstriage
```
*edit: running `grunt` fixed the css, but the api dir is still a 404.*

I tried leaving the virtualenv and running `npm start`, but it had the same problem as when I tried to `npm run setup` before - 
```bash
sh: 1: source: not found
```
I guess it is using sh shell instead of bash (or zsh, which I normally use). I modified package.json to not use source:
```json
"start": ". env/bin/activate && python run.py",
```
and then running `npm start` worked equally as well as manually entering the virtualenv and doing  `python run.py`.

Running the server also tells me 
```
home/webcompat/env/local/lib/python2.7/site-packages/flask_limiter/extension.py:124: UserWarning: Use of the default `get_ipaddr` function is discouraged. Please refer to https://flask-limiter.readthedocs.org/#rate-limit-domain for the recommended configuration
```
but if I am reading [this comment](https://github.com/webcompat/webcompat.com/issues/1304#issuecomment-275597739) in #1304 correctly, that is a non-issue right now??

Anyway. I'm running Linux Mint 18 Sarah and running the commands from zshell, if that makes a difference (*edit*: apparently not, I tried running it from bash and same outcome.). I would love to know how to tackle this! :)

heatherbooker:
Fixes #1378 @miketaylr

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,8 +47,8 @@
     "fix": "eslint --fix ./Gruntfile.js ./tests ./grunt-tasks ./webcompat/static/js/lib",
     "imagemin":"grunt imagemin",
     "module": "git submodule init && git submodule update",
-    "start": "source env/bin/activate && python run.py",
-    "virtualenv": "pip install virtualenv && virtualenv env && source env/bin/activate && npm run pip",
+    "start": "source env/bin/activate || . env/bin/activate && python run.py",
+    "virtualenv": "pip install virtualenv && virtualenv env && source env/bin/activate || . env/bin/activate && npm run pip",
     "pip": "pip install -r config/requirements.txt",
     "config": "cp config/secrets.py.example config/secrets.py",
     "project-update": "pip install --upgrade pip && npm run backend-install && npm update"


miketaylr:
r+, thanks!

Maxim-Mazurok:
Based on this info, I think we can safely use dot `.` all the time, instead of `source`: https://stackoverflow.com/questions/20094271/using-dot-or-source-while-calling-another-script-what-is-the-difference

PR #1399, opened 2017-03-07
Fixes #1398 - changed twitter link to include replies

### LINKED ISSUE #1398: Twitter account link should include replies
This is a followup issue on "change twitter link in footer" #1372

In  her blog post @lockettm mentioned the comment of a UX test participant:

> “Last post was 6:35 am Jan 30—so no, I would say they would not respond quickly. And then I’d have to take my chances with email—I would probably not even try to ask them on Twitter.”

We fixed it with #1372 That's cool it was an improvement. I guess we can improve this a bit more, by using https://twitter.com/webcompat/with_replies

Here the screenshots. Left with replies, and right without replies (aka the current one)

capture d ecran 2017-03-07 a 08 17 34

zoepage:
r? @karlcow 

#1398

### DIFF
--- a/webcompat/templates/shared/footer.html
+++ b/webcompat/templates/shared/footer.html
@@ -25,7 +25,7 @@
       </ul>
       <ul class="wc-Footer-list r-ResetList">
         <li class="wc-Footer-item wc-Footer-item--follow">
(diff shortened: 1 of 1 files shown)


karlcow:
Thanks a lot @zoepage

PR #1410, opened 2017-03-07
Fixes #1409. Tweak CSP policy based on incoming reports.

### LINKED ISSUE #1409: Make updates to Content-Security-Policy policy
Based on observations from the Report-Only logs in https://github.com/webcompat/webcompat.com/issues/763#issuecomment-284854940.

miketaylr:
I'd like to deploy this and look at Report-Only logs after another few days.

r? @karlcow

### DIFF
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -484,8 +484,8 @@ def add_csp(response):
         "default-src 'none'; " +
         "connect-src 'self'; " +
         "font-src 'self'; " +
-        "img-src 'self'; " +
-        "script-src 'self' https://www.google-analytics.com; " +
-        "style-src 'self'; " +
+        "img-src 'self' https://www.google-analytics.com https://*.githubusercontent.com; " +  # nopep8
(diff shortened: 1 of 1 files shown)


webcompat/helpers.py
```diff
@@ -484,8 +484,8 @@ def add_csp(response):
         "default-src 'none'; " +
         "connect-src 'self'; " +
         "font-src 'self'; " +
-        "img-src 'self'; " +
-        "script-src 'self' https://www.google-analytics.com; " +
-        "style-src 'self'; " +
+        "img-src 'self' https://www.google-analytics.com https://*.githubusercontent.com; " +  # nopep8
+        "script-src 'self' 'unsafe-eval' 'unsafe-inline' https://www.google-analytics.com; " +  # nopep8
+        "style-src 'self' 'unsafe-inline; " +
```
karlcow: Is there a missing single quote at the end of unsafe-inline?

miketaylr: There is! Good catch. 👼

miketaylr:
> Apart of my comment I guess it's a good to go.

Thanks, I fixed the string. So let's merge.

PR #1411, opened 2017-03-07
Fixes #1047 -  Functional tests for file upload screenshots

### LINKED ISSUE #1047: Functional tests for file upload screenshots
The other half to #1043.

brizental:
r? @miketaylr


tests/functional-all.js
```diff
@@ -16,6 +16,7 @@ define([
   "./functional/comments-non-auth.js",
   "./functional/contributors-non-auth.js",
   "./functional/comments-auth.js",
+  "./functional/image-uploads-auth.js",
```
miketaylr: Can we change the name of this file to something like `comments-image-uploads-auth.js`? Initially I was a bit confused why we needed authed upload tests (but also, I'm very slow 🐢...)

brizental: I decided on this name at first just to follow the pattern that I found on the naming of the files, but this is more semantic I think.

brizental: Maybe, instead of renaming the file I just put this test inside of comments-auth.js. What do you think?

miketaylr: @brizental yeah, that would also work. 👍

miketaylr:
(FYI, I'm heading on holiday today for the next 10 days or so -- but I'll come back and merge this tomorrow at some point!) @karlcow and @zoepage are around to help with any other questions related to other issues. 

And thanks for the PR! 👏

tests/functional-all.js
brizental: Done.

brizental:
Ok, thanks for letting me know :)

No problem, its been really fun contributing to the project. You know I started doing this for the Outreachy application, so if you think that any of the issues here are more interesting for me, given the project, let me know.

Have a nice holiday!

Issue #609, opened 2015-04-13
HTML resources HTTP Caching is broken on webcompat.com


karlcow:
We need to have a better story at caching HTML resources on webcompat.com.
See #590 and comments on #608

miketaylr:
Yeah, especially since all the dynamic stuff is happening on the client side there's no good reason to not heavily cache the HTML.

karlcow:
Working on this.

zoepage:
@karlcow are you adding AppCache or what is your plan?
(Didn't see anything related to that question in the other linked issues)

karlcow:
@zoepage the issue is about HTTP caching for HTML resources.

Currently if we request an issue
```
→ http --print h GET https://webcompat.com/issues/100
```

The HTTP response has no HTTP caching headers

```http
HTTP/1.1 200 OK
Connection: keep-alive
Content-Encoding: gzip
Content-Type: text/html; charset=utf-8
Date: Thu, 19 Jan 2017 12:48:55 GMT
Server: nginx/1.1.19
Strict-Transport-Security: max-age=31536000
Transfer-Encoding: chunked
Vary: Accept-Encoding
```

I'm adding `Etag` and `Cache-Control`. See the commits already done in that issue, or the tests.
https://github.com/karlcow/webcompat.com/commit/f6a65b7098eb307ef2190296d55f92259e89a1e2

Does it help to understand?

karlcow:
And in case, someone else reading this issue wants to know how HTTP caching is working, there is a very good resource: [Caching Tutorial for Web Authors and Webmasters](http://www.web-caching.com/mnot_tutorial/)

karlcow:
let me restart a branch for this. So I do it against the latest master.

miketaylr:
I'm gonna back this out in https://github.com/webcompat/webcompat.com/issues/1386, because it caused #1384. When we re-land this code, we should verify that #1384 is fixed.

(see also #1385 for writing tests to make sure we don't break logging in/out again)

miketaylr:
Let's also make sure we don't break the test in https://github.com/webcompat/webcompat.com/issues/1371 when we fix it too, I think this changeset caused the permafail (which makes sense, it was a test on an issue page that involved logging in).

miketaylr:
I wonder if the reason it broke is because we serve different HTML if the user is authed:

https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/shared/shared-nav-links.html#L7-L12

[...]

PR #1415, opened 2017-03-09
Fixes 609 - Implements a caching policy for HTML resources

karlcow:
Hopefully I didn't break anything this time. 
A couple of things have changed. So review when you have time 
r? @miketaylr

### DIFF
--- a/tests/test_http_caching.py
+++ b/tests/test_http_caching.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+'''Tests for HTTP Caching on webcompat resources.'''
+
+import os.path
+import sys
+import unittest
+
+# Add webcompat module to import path
+sys.path.append(os.path.realpath(os.pardir))
+import webcompat  # nopep8
+
+# Any request that depends on parsing HTTP Headers (basically anything
+# on the index route, will need to include the following: environ_base=headers
+html_headers = {
+    'HTTP_USER_AGENT': ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; '
+                        'rv:53.0) Gecko/20100101 Firefox/53.0'),
+    'HTTP_ACCEPT': 'text/html'}
+
+
+class TestHTTPCaching(unittest.TestCase):
+    def setUp(self):
+        webcompat.app.config['TESTING'] = True
+        self.app = webcompat.app.test_client()
+
+    def tearDown(self):
+        pass
+
+    def test_issue_has_etag(self):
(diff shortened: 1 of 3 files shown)


tests/test_http_caching.py
```diff
@@ -0,0 +1,68 @@
...
+import os.path
+import sys
+import unittest
+
+# Add webcompat module to import path
+sys.path.append(os.path.realpath(os.pardir))
+import webcompat  # nopep8
+
+# Any request that depends on parsing HTTP Headers (basically anything
+# on the index route, will need to include the following: environ_base=headers
```
miketaylr: I notice you updated the dict to be `html_headers`, maybe we should update this comment as well.

webcompat/views.py
```diff
@@ -147,6 +147,7 @@ def index():
 
 
 @app.route('/issues')
+@cache_policy(private=True, uri_max_age=0, must_revalidate=True)
```
miketaylr: Just so I understand: here and for the other endpoints with this decorator, by setting `uri_max_age=0`, you're telling the user agent that it should revalidate the resource always?

Do we plan on using a longer (i.e., the day default) max age for other resources, in a follow up bug perhaps?

karlcow: Tell me if what I explained here, answer your question.
https://github.com/webcompat/webcompat.com/issues/609#issuecomment-284322939

We specifically want a revalidation at each request, because if we do not do it, we will get the issue with the cached resource (on login/logout state). So no. max-age must always be 0 for revalidation.

tests/test_http_caching.py
karlcow: dang! Thanks for catching this.

webcompat/views.py
karlcow: ping @miketaylr

miketaylr: Ah, OK. That's helpful, thanks @karlcow.

LGTM to merge w/ the comment fix.

tests/test_http_caching.py
karlcow: Done

``` bash
git commit -m '#609 - fixes a comment (minor)' tests/test_http_caching.py
git rebase -i HEAD~2
git push -f karlcow 609/3
git log --oneline
# 3cf04f4 Issue #609 - Implements cache policies on HTML views
# 8e20552 Issue #609 - Creates a cache policy wrapper for views
# 3e55497 Issue #609 - Cleans up the modules import
# 57c63ee Issue #609 - Adds tests for HTTP Caching Policy
```

webcompat/helpers.py
```diff
@@ -489,3 +491,36 @@ def add_csp(response):
         "style-src 'self' 'unsafe-inline'; " +
         "report-uri /csp-report"
     )
+
+
+def cache_policy(private=True, uri_max_age=86400, must_revalidate=False):
```
davidbgk: @karlcow you might want to take a look at https://github.com/opendatateam/udata/pull/802/files#diff-74fa3771bbb3507e52e7f20e6ed64351R22 ;)

Issue #899, opened 2016-01-07
inline-svg instead of icon-font


magsout:
http://www.sitepoint.com/icon-fonts-vs-svg-debate/

### Header
- [ ] Search
- [ ] Contribute
- [ ]  Issues
- [ ] Log in
- [ ] Home
- [ ] Caret down

### Submit form

- [ ] Cloud upload

### Threesteps

- [ ] Bug reporting
- [ ]  Bug diagnosis  
- [ ]  Site outreach

###  Thanks 

- [ ]  Browser stack

### Footer 

- [ ] Twitter
- [ ] Github 

### Search

- [ ] Caret down
- [ ] Search
- [ ]  Github

### Issue
- [ ] cog

miketaylr:
Yeah, seems like icon fonts aren't cool anymore. +1 for inline svg at some point.

magsout:
Just re up this issue.

What do you think @zoepage  ?

zoepage:
@magsout I agree on SVG. Could be a great issue for our all-hands list? :)

magsout:
@zoepage added on my all hands SF todolist ;)

zoepage:
This will be fully solved by the refactor in 2 weeks. Already implemented in http://webcompat.github.io/design/

PR #1416, opened 2017-03-09
Fixes #1404: Add contribute link with svg to top navbar

### LINKED ISSUE #1404: Add Contributor link to top navigation
### Todo's
- [ ] create anchor we can link in top navigation to https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/home-page/join-us.html#L3
- [ ] Add contributors link to top navigation with the label `Contribute` https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/shared/nav.html
- [ ] link to anchor in `join the team` section title
- [ ] add contributor SVG icon https://thenounproject.com/search/?similar=898865&i=346424 to link
- [ ] hide link description for responsive layout as it is right now & seen in following
(label should be hidden by default with same semantics, https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/css/development/components/navbar.css#L41)
 (the icon should still be visible)

---
### Context
Follow up issue on https://github.com/webcompat/webcompat.com/issues/1281#issue-200691564.

heatherbooker:
Fixes #1404 . Looks like:

screenshot of navbar

@zoepage , what do you think? :)

Also, I may be unclear on whether this fulfills the "inline SVG" requirement. Does that mean the SVG code literally needs to be dropped into the HTML, and not used the way I've done it as a background-image using a url?


webcompat/static/css/development/components/navbar.css
```diff
@@ -53,7 +53,23 @@
...
+      display: inline-block;
+      background-repeat: no-repeat;
+      background-position: center center;
+      position: relative;
     }
+
+  /* "Contribute" link icon */
+  .wc-Icon--contribute {
+    background-image: url(/img/contribute.svg);
+    height: 1.2em;
```
heatherbooker: These height and width styles should move to line 55, `.wc-Navbar-link-icon` once the other icons are [changed to svg](https://github.com/webcompat/webcompat.com/issues/899).

zoepage:
@heatherbooker could you please make the icon bigger so it fits the proportions of the other ones?
(+2-3px) It doesn't align well with the others.
slice 1

Plus the icon is full black. Could you change that to #404040? 
https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/css/development/components/style-variable.css#L12

Thank you!

heatherbooker:
Sure! How's this:

zoepage:
@heatherbooker the aligning and color look great, but maybe that was the wrong icon. :/
Do you want to take a look at different ones? :)

heatherbooker:
Haha, ah okay.

This one? 


Or this one?


Or keep trying? :)

zoepage:
@lockettm what do you think? :)

zoepage:
Btw, I am ok with both. It's an improvement to not having the link at all. :)
And if we find something that fits better at some point, cool! We still able to change it then!

miketaylr:
Slight preference for screen shot 2017-03-09 at 9 34 13 am (maybe that's the pointy one?) 😄

heatherbooker:
haha, I actually thought the top one was the pointy one! anyway I'll wait for a couple hrs and see if there's more opinions I guess.

adamopenweb:
I like both, like @zoepage said, this is a nice improvement. 😄

karlcow:
The round one looks like a Russian orthodox church, the pointy one looks like startrek logo. 
Or at least it's what my weird poetic brain tells me. 

I don't have strong preferences. Choose the one you want. :)

lockettm:
Both look good to me ☺

heatherbooker:
Thank you everyone for the input!

One last concern is the placement of the icon - I'm having a hard time making it work for both FF and Safari:

`Top` value | FF/Chrome/Opera | Safari on iPhone | Safari on Desktop
--- | --- | --- | ---
(not set) | too high :( | perfecto :) | ---
2px or 0.1em | perfecto :) | too low :( | apparently fine

Screenshot on mobile safari of it being ever so slightly too low:

image

I'm not sure what to do about this, and also don't have the ability to test on other mobile platforms. Thoughts? Is it subtle enough to get away with, esp. considering all the other icons should soon change to svg as well and all be in weird alignment together? ;P

zoepage:
> The round one looks like a Russian orthodox church, the pointy one looks like startrek logo.

Wow, right. STAR TREEEEEK! :D


> Screenshot on mobile safari of it being ever so slightly too low.

Me personally, I am actually okay with it as it looks like. It doesn't need to be pixel perfect. What do the others think?

zoepage:
@heatherbooker I've checked the PR and it looks good to me (also on mobile as far as I could simulate it) :)))

It's so cool and fits so well, I was thinking... Could you also add the icon to the index page navigation between search and all issues?

heatherbooker:
@zoepage definitely! I just pushed that. :)

zoepage:
@heatherbooker awesome, thanks!

### View port issue
There seems to appear a breaking point issue with the longer content, so we should hide the labels earlier. (Works again starting from device width: 681px )

screen shot 2017-03-10 at 10 57 03 pm

screen shot 2017-03-10 at 10 56 55 pm



### double icon
 Same issue with the other navigation, plus the contribute icon appears twice now :)
screen shot 2017-03-10 at 11 00 07 pm
screen shot 2017-03-10 at 11 00 13 pm

heatherbooker:
@zoepage oops my bad! Sorry about that! Looks good to me now on ff, chrome and opera at various widths. And I haven't seen the contribute icon doubling up at all, that's strange! :S Is it still doing that?

zoepage:
@heatherbooker 
the double icon is gone. maybe it was just a build or caching issue :)

One last point and we should be good! :)

The labels hide on different break points. From the UX side, I personally prefer steady interactions. 
I'm not sure, if this is an issue in general. What you you think @heatherbooker ?

screen shot 2017-03-13 at 10 14 02 am

screen shot 2017-03-13 at 10 14 42 am

heatherbooker:
@zoepage That's fair! I was thinking about maximizing the amount of time the labels are visible, but keeping it steady is also a good point. The other thing then is that only on the home page, the icons' margins decrease on small viewports - shall I also extend that to the other pages?

Smaller margin on home page:


Unchanged margin:
image

zoepage:
> The other thing then is that only on the home page, the icons' margins decrease on small viewports - shall I also extend that to the other pages?

👍 :)

heatherbooker:
@zoepage done!

zoepage:
@heatherbooker great job! Thank you for all your effort! :)))

PR #1388, opened 2017-03-03
Issues/980/15 [not ready for merging]

cch5ng:
hi @miketaylr I'm just sharing what I have

browserstack now offers a way to run the app server locally and run the tests on browserstack (I guess this would replace the old running tests locally if you want)... anyhow I got the tests running for FF51 and Chrome 56 (MacOS Sierra).

what is the best way to share the app key?

prerequisites: need to install browserstack-local
`npm i -D browserstack-local`

add local env var
`export $BROWSERSTACK_USER='carolchung2'`
`export $BROWSERSTACK_KEY='***'` ?how do you communicate stuff like this? lastpass?

command to run tests (this can change)
`node_modules/.bin/intern-runner config=tests/browserstack BROWSERSTACK_USER=$BROWSERSTACK_USER BROWSERSTACK_KEY=$BROWSERSTACK_KEY`

testing from travis is another todo...

PR #1430, opened 2017-03-17
[test #980 not ready for merge]

cch5ng:
try to trigger travis tests with browserstack
(branch based on clean upstream master)
fix: #980; test config for browserstack tests on travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,6 @@ cache:
 git:
   depth: 5
 
-addons:
-  firefox: "50.1.0"
-
 env:
   global:
     - ISSUES_REPO_URI=webcompat/webcompat-tests/issues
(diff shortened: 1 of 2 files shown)


cch5ng:
@miketaylr closing this PR but think this is the closest to the correct configs so far...
 it is possible I might have made mistakes in creating the travis encrypted vars for browserstack (username, access_key)... but the larger issue is the travis security restriction where pr's from forked repos do not have access to encrypted environ vars. so I think I'm blocked from testing this and even if the configs would work, the browserstack tests would only get run for the project owner (?)...

one possible workaround so contributors working from forks get the benefit of browserstack testing is to do something like here https://github.com/webcompat/webcompat.com/pull/1388
(the code/PR is not ready to merge but the concept is valid)... I was trying out a config to trigger tests locally but get them running on browserstack. 

one way is to keep the username/access_key info in lastpass and give access to regular contributors. then they can run the tests and put in the info as cli parameters. this can work (though it might not be totally secure).

let me know what you think and if #1388 is useful I can clean up those configs and make a PR for that.

cch5ng:
a couple additional thoughts, where I first saw open source projects using browserstack were these projects

* (originally) https://github.com/react-dnd/react-dnd
* (more recently) https://github.com/ReactTraining/react-router

so I kind of doubt they are using intern but maybe looking through their testing configs would give some insight into whether they are using some travis-browserstack testing combo, and if that fails, maybe ask around on their slack/gitter type channels - how to work around trick of keeping access_key secure (?)

miketaylr:
(Note: this is the place to start for travis browserstack integration)

PR #1431, opened 2017-03-19
Fixes #1229  Added support for ctrl+enter for posting new comment

### LINKED ISSUE #1229: Support cmd/ctrl+enter for posting new comment
Just like on GitHub and elsewhere.

nj4710:
Fixes #1229

### DIFF
--- a/webcompat/static/js/lib/issues.js
+++ b/webcompat/static/js/lib/issues.js
@@ -347,6 +347,7 @@ issues.MainView = Backbone.View.extend({
     this.comments = new issues.CommentsCollection({pageNumber: 1});
     this.initSubViews();
     this.fetchModels();
+    this.handleKeyShortcuts();
   },
   closeLabelEditor: function(e) {
     var target = $(e.target);
@@ -484,6 +485,10 @@ issues.MainView = Backbone.View.extend({
   },
   render: function() {
     this.$el.fadeIn();
+  },
+
+  handleKeyShortcuts: function() {
+    Mousetrap.bind("mod+enter", _.bind(this.addNewComment, this));
   }
 });
 
--- a/webcompat/templates/issue/issue-comment-submit.html
+++ b/webcompat/templates/issue/issue-comment-submit.html
@@ -11,7 +11,7 @@
   </div>
   <div class="wc-Comment-body">
   <label for="wc-Comment-text" class="wc-Accessibility">Comment Text</label>
-    <textarea class="wc-Comment-submit js-Comment-text" placeholder="Leave a comment"></textarea>
+    <textarea class="wc-Comment-submit js-Comment-text mousetrap" placeholder="Leave a comment"></textarea>
     <div class="wc-Form-group js-ImageUploadView">
       <script type="text/template" id="upload-input-tmpl">
         <form>


miketaylr:
Sorry for the delay @nj4710, just coming back from vacation. 

(Also, normally we request reviewer in the pull request, not in the issue (https://github.com/webcompat/webcompat.com/issues/1229#issuecomment-287635267).

webcompat/static/js/lib/issues.js
```diff
@@ -325,15 +325,16 @@ issues.StateButtonView = Backbone.View.extend({
     } else {
       this.model.toggleState();
     }
-  }
+  }
```
miketaylr: nit: can we remove this extra whitespace

webcompat/static/js/lib/issues.js
```diff
@@ -484,6 +485,12 @@ issues.MainView = Backbone.View.extend({
   },
   render: function() {
     this.$el.fadeIn();
+  },
+
+  keyAction: function(e) {
+    if (e.keyCode === 13 && e.ctrlKey ) {
```
miketaylr: Could we add cmd + e support for Mac users as well?

Since we're already shipping [Mousetrap](https://github.com/webcompat/webcompat.com/blob/2ca6d388907ec1f721e651df518857ff851d99e6/webcompat/templates/layout.html#L57), maybe we can use that: https://craig.is/killing/mice

miketaylr: Since we're already shipping [Mousetrap](https://github.com/webcompat/webcompat.com/blob/2ca6d388907ec1f721e651df518857ff851d99e6/webcompat/templates/layout.html#L57), maybe we can use that: https://craig.is/killing/mice

webcompat/static/js/lib/issues.js
```diff
@@ -325,15 +325,16 @@ issues.StateButtonView = Backbone.View.extend({
...
 });
 
 issues.MainView = Backbone.View.extend({
   el: $(".js-Issue"),
   events: {
     "click .js-Issue-comment-button": "addNewComment",
     "click": "closeLabelEditor",
-    "click .wc-Comment-content-nsfw": "toggleNSFW"
+    "click .wc-Comment-content-nsfw": "toggleNSFW",
+    "keydown": "keyAction"
```
miketaylr: Can we give this a more descriptive name, like `handleKeyShortcuts` (or something better)?

nj4710:
Hi @miketaylr   I am having some problems using Mousetrap. According to me, this should be the solution but this does not work. Could you please help me to call the addNewComment function from handleKeyShortcuts?

webcompat/static/js/lib/issues.js
```diff
@@ -484,6 +485,10 @@ issues.MainView = Backbone.View.extend({
   },
   render: function() {
     this.$el.fadeIn();
+  },
+
+  handleKeyShortcuts: function(e) {
+     Mousetrap.bind('mod+enter', this.addNewComment() ,'keydown');
```
miketaylr: This is probably a scope/`this` binding issue -- so possible something like:

`Mousetrap.bind('mod+enter', _.bind(this.addNewComment, this), 'keydown');` will work (or similar -- untested!)

Is there any reason we need the 3rd `keydown` arg? AFAIK, Mousetrap will just do the right thing™ without it.

However, thinking more about this, with this approach aren't we re-binding a new listener on every keydown event? That doesn't seem to be what we want. 

We should probably remove the "keydown: handleKeyShortcuts" key/value pair from the `events` object and just call `this.handleKeyShortcuts()` from inside the `initialize` method.

miketaylr: (this has the advantage of making an obvious place for any new complex keyboard shortcuts to go)

nj4710:
Yeah that was a scope issue. Thanks a lot @miketaylr. Also you are right, my approach will probably create more bugs. Please let me know if there are any improvements in this?

webcompat/static/js/lib/issues.js
```diff
@@ -484,6 +485,10 @@ issues.MainView = Backbone.View.extend({
   },
   render: function() {
     this.$el.fadeIn();
+  },
+
+  handleKeyShortcuts: function(e) {
+    Mousetrap.bind('mod + enter', _.bind(this.addNewComment, this));
```
miketaylr: A few more things, then this is ready!

eslint will complain about single quotes, so can you change to use double quotes? Also, there's an `e` argument getting passed in that isn't being used.

Also, for me it's not working as `mod + enter`, but does work without the spaces `mod+enter'. Does the spaces variant work for you?

nj4710:
I have changed it to `mod+enter` but it was working for me with the spaces too.

r? @miketaylr

miketaylr:
> I have changed it to mod+enter but it was working for me with the spaces too.

Interesting!

r+, let's merge. Thanks for fixing this @nj4710!

nj4710:
Thanks @miketaylr

PR #1435, opened 2017-03-22
Fixes #1199 - Simple bugform.js cleanup

### LINKED ISSUE #1199: Simple bugform.js cleanup
🚨 Wait until https://github.com/webcompat/webcompat.com/pull/1186 is merged, to avoid merge conflicts!

`BugForm.submitType` is unused (it was needed when we did a JS `FormData` POST to the relevant endpoint, but now we do regular form submission, so the server gets that from the multipart/form-data junk)

https://github.com/webcompat/webcompat.com/blob/f8c3b75122ee4ec6d918833b01f71c457a3bc9f5/webcompat/static/js/lib/bugform.js#L18

https://github.com/webcompat/webcompat.com/blob/f8c3b75122ee4ec6d918833b01f71c457a3bc9f5/webcompat/static/js/lib/bugform.js#L57

`loadingIndicator` and `loaderImage` can be consolidated into one (in usage as well):

https://github.com/webcompat/webcompat.com/blob/f8c3b75122ee4ec6d918833b01f71c457a3bc9f5/webcompat/static/js/lib/bugform.js#L12

https://github.com/webcompat/webcompat.com/blob/f8c3b75122ee4ec6d918833b01f71c457a3bc9f5/webcompat/static/js/lib/bugform.js#L14

laghee:
r? @miketaylr

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -7,10 +7,8 @@ function BugForm() {
   this.submitButtons = $("#js-ReportForm .js-Button");
   this.loadingIndicator = $(".js-Loader");
   this.reportButton = $("#js-ReportBug");
-  this.loaderImage = $(".js-Loader");
   this.uploadLoader = $(".js-Upload-Loader");
   // by default, submission type is anonymous
-  this.submitType = "github-proxy-report";
   this.UPLOAD_LIMIT = 1024 * 1024 * 4;
 
   this.inputs = {
@@ -60,14 +58,7 @@ function BugForm() {
     this.uploadField.on("change",  _.bind(this.checkImageTypeValidity, this));
     this.osField.add(this.browserField)
                 .on("blur input", _.bind(this.checkOptionalNonEmpty, this));
-    this.submitButtons.on("click", _.bind(function(e) {
-      if (e.target && e.target.value) {
-        // store a reference to what report button was clicked
-        this.submitType = e.target.value;
-      }
-
-      this.loadingIndicator.show();
-    }, this));
+    this.submitButtons.on("click", _.bind(this.loadingIndicator.show, this));
 
     // See if the user already has a valid form
     // (after a page refresh, back button, etc.)
@@ -434,7 +425,7 @@ function BugForm() {
           msg = "The image is too big! Please choose something smaller than 4MB.";
           wcEvents.trigger("flash:error", {message: msg, timeout: 5000});
         }
-        this.loaderImage.hide();
+        this.loadingIndicator.hide();
       }, this)
     });
 


miketaylr:
(let me know if that's confusing, i can also do it for you if you'd prefer!)

laghee:
I _think_ I did it successfully. It was confusing at first because the difference between `~` and `-` was not apparent when looking at the email update rather than the actual thread on GitHub!

miketaylr:
> I think I did it successfully. It was confusing at first because the difference between ~ and - was not apparent when looking at the email update rather than the actual thread on GitHub!

Ah, yeah -- depending on the font I bet that could be confusing! I still see 2 commits. Are you sure you pushed the single commit with `git push -f`?

You can also check by logging out the last 2 commits and making sure yours is only the first one via something like `git log -n 2`.

laghee:
I have a feeling I missed something with the push. It looks like I squashed it locally, but maybe I'm missing something fundamental in pushing? the `git push` defaults to origin, right? So maybe I should have specifically done `git push -f origin issues/1199/2`?

screenshot 2017-03-22 22 01 23

miketaylr:
> So maybe I should have specifically done git push -f origin issues/1199/2?

That looks about right. For me, my first push typically looks like `git push --set-upstream origin issues/NNN/N/N`, so `git push -f` will default to what i set the upstream to.

laghee:
> For me, my first push typically looks like git push --set-upstream origin issues/NNN/N/N, so git push -f will default to what i set the upstream to.

That sounds eminently sensible. I wish I had done that. Is there an elegant-ish fix I can run now? I found some suggestions on stackoverflow, but they're not identical situations, and I'm leery of making things worse...

laghee:
Hi, @miketaylr, when you get a chance could you let me know if I managed to send the squashed commits correctly?

denschub:
Looks great, there is only one little nag left: The commit message contains the same message ... twice! That's no big deal and we probably could edit that when merging (maybe Mike is doing that), but if you want, you can call `git commit --amend` and edit the commit message so it does not appear twice. After that, force-push again and you should be done!

Thanks for your patience. I know it can be confusing at first... :)

(jumping in since MIke is probably busy with some work traveling)

miketaylr:
@laghee yep -- it did work! \o/

Like @denschub said, could you tweak the commit message so it's not duplicated on different lines?

`git commit --amend`
(delete second line of commit message, save file)
`git push -f` (or whatever magic git command you just used to push the new commit!)

thx!

laghee:
Thanks, @denschub and @miketaylr! I amended and pushed again. I appreciate all the hand-holding! 🤓

miketaylr:
Awesome! Thanks @laghee

PR #1440, opened 2017-03-26
Fixes #1211: Added auto-expanding for bug reporting/comments area

### LINKED ISSUE #1211: auto-expand the bug reporting/comments area
Sometimes I feel the form is a bit tiny when editing a long bug or comment. I resize by dragging the corner of the form, but we could also helps users with an auto-expanding feature.

See for example http://codepen.io/vsync/pen/frudD

(not a priority)

nj4710:
r? @miketaylr

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -73,6 +73,9 @@ function BugForm() {
     // (after a page refresh, back button, etc.)
     this.checkForm();
 
+    // Auto Expanding feature for form
+    this.autoExpand();
+
     // Set up listener for message events from screenshot-enabled add-ons
     window.addEventListener("message", _.bind(function(event) {
       // Make sure the data is coming from ~*inside the house*~!
@@ -465,6 +468,15 @@ function BugForm() {
       }
       return value.replace(firstLine, prefix + this.urlField.val() + "\n");
     }, this));
+  };
+
+  // See function autoExpand in issues.js
+  this.autoExpand = function() {
+    var initialHeight = $("textarea.js-autoExpand").height();
+    $("textarea.js-autoExpand").on("input", function() {
+      $(this).css("height", initialHeight);
+      $(this).css({ "overflow":"hidden", "height": this.scrollHeight + "px"});
+    });
   };
 
   return this.init();
--- a/webcompat/static/js/lib/issues.js
+++ b/webcompat/static/js/lib/issues.js
@@ -348,6 +348,7 @@ issues.MainView = Backbone.View.extend({
     this.initSubViews();
     this.fetchModels();
     this.handleKeyShortcuts();
+    this.autoExpand();
   },
   closeLabelEditor: function(e) {
     var target = $(e.target);
@@ -489,6 +490,15 @@ issues.MainView = Backbone.View.extend({
 
   handleKeyShortcuts: function() {
     Mousetrap.bind("mod+enter", _.bind(this.addNewComment, this));
+  },
+
+  // See function autoExpand in bugform.js 
+  autoExpand: function() {
+    var initialHeight = $("textarea.js-autoExpand").height();
+    $("textarea.js-autoExpand").on("input", function() {
+      $(this).css("height", initialHeight);
+      $(this).css({"overflow":"hidden", "height": this.scrollHeight + "px"});
+    });
   }
 });
 
--- a/webcompat/templates/issue/issue-comment-submit.html
+++ b/webcompat/templates/issue/issue-comment-submit.html
@@ -11,7 +11,7 @@
   </div>
   <div class="wc-Comment-body">
   <label for="wc-Comment-text" class="wc-Accessibility">Comment Text</label>
-    <textarea class="wc-Comment-submit js-Comment-text mousetrap" placeholder="Leave a comment"></textarea>
+    <textarea class="wc-Comment-submit js-Comment-text mousetrap js-autoExpand" placeholder="Leave a comment"></textarea>
     <div class="wc-Form-group js-ImageUploadView">
       <script type="text/template" id="upload-input-tmpl">
         <form>
--- a/webcompat/templates/home-page/form.html
+++ b/webcompat/templates/home-page/form.html
@@ -63,7 +63,7 @@
       <div class="r-Grid r-Grid--withGutter ">
         <div class="r-Grid-cell r-Grid-cell--alignBottom r-minS--1of2">
             <div class="wc-Form-group js-Form-group">
-              {{ form.description.label(class_='wc-Form-label') }} {{ form.description(class_='wc-Form-item wc-Form-item--textarea', rows=7) }}
+              {{ form.description.label(class_='wc-Form-label') }} {{ form.description(class_='wc-Form-item wc-Form-item--textarea js-autoExpand', rows=7) }}
             </div>
         </div>
         <div class="r-Grid-cell r-Grid-cell--alignBottom r-minS--1of2">


miketaylr:
(In the future, PR titles should begin with "Fixes #NNNN: blah", thanks!)

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,21 @@ function BugForm() {
     }, this));
   };
 
+  this.autoExpand = function() {
+    $(document).one("focus.js-autoExpand", "textarea.js-autoExpand", function() {
+      var savedValue = this.value;
+      this.value = "";
+      this.baseScrollHeight = this.scrollHeight;
+      this.value = savedValue;
```
miketaylr: Why are we clearing the `value` then restoring it?

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,21 @@ function BugForm() {
...
+    })
+    .on("input.js-autoExpand", "textarea.js-autoExpand", function() {
+      var minRows = this.getAttribute("data-min-rows") | 0, rows;
```
miketaylr: Bitwise OR here feels a bit magical... what are you trying to accomplish?

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,21 @@ function BugForm() {
...
+      this.rows = minRows;
+      rows = Math.ceil((this.scrollHeight - this.baseScrollHeight) / 17);
```
miketaylr: Is 17 meaningful here? Or just something that looks good?

webcompat/static/js/lib/issues.js
```diff
@@ -489,6 +490,23 @@ issues.MainView = Backbone.View.extend({
 
   handleKeyShortcuts: function() {
     Mousetrap.bind("mod+enter", _.bind(this.addNewComment, this));
+  },
+
+  // Auto expanding feature for comments
+  autoExpand: function() {
+    $(document).one("focus.js-autoExpand", "textarea.js-autoExpand", function() {
```
miketaylr: My jQuery is a bit rusty, can you explain the event scoping here?

webcompat/static/js/lib/issues.js
```diff
@@ -489,6 +490,23 @@ issues.MainView = Backbone.View.extend({
 
   handleKeyShortcuts: function() {
     Mousetrap.bind("mod+enter", _.bind(this.addNewComment, this));
+  },
+
+  // Auto expanding feature for comments
```
miketaylr: In general, duplicating code is not great... but I think we have this problem in a few other places. Something that #909 (or similar) should be able to fix. Can we at least leave a comment pointing to the other dupe'd code so they stay in sync?

webcompat/static/js/lib/bugform.js
nj4710: I was worried we might lose the comment data but it turns out the `savedValue` variable is not even needed.

webcompat/static/js/lib/bugform.js
nj4710: I have removed the bitwise OR and added different values for `minRows` for comment section and form section.

webcompat/static/js/lib/bugform.js
nj4710: 17 might be for font size(I maybe wrong). But I tested different values on my machine. It works fine with all the values that I used.

webcompat/static/js/lib/issues.js
nj4710: Yes, I understand the value of code reusability. Can we make a global function that works for all the elements that has js-autoExpand class? Although, I have added a comment there too.

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,18 @@ function BugForm() {
     }, this));
   };
 
+  this.autoExpand = function() {
+    $(document).one("focus.js-autoExpand", "textarea.js-autoExpand", function() {
+      this.baseScrollHeight = this.scrollHeight;
+    })
+    .on("input.js-autoExpand", "textarea.js-autoExpand", function() {
+      var minRows = 2;
```
miketaylr: Can you help me understand `minRows = 2`? Currently the textarea on this form has a `rows=7` attribute.

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,18 @@ function BugForm() {
...
+      this.rows = minRows;
+      var rows = Math.ceil((this.scrollHeight - this.baseScrollHeight) / 17);
```
miketaylr: Can you leave a comment explaining where the number 17 comes from? It will be confusing to come back to later. :)

webcompat/static/js/lib/issues.js
miketaylr: We could make a global function, but currently there isn't a "global.js" file -- ideally we're moving to a world of modules where we just pull in what we need. I think for now having it duped for now is fine, but we should add comments to both methods.

webcompat/static/js/lib/issues.js
```diff
@@ -489,6 +490,21 @@ issues.MainView = Backbone.View.extend({
...
+  // See Line 473 in bugform.js
```
miketaylr: Just as a heads up, line numbers will change very quickly and become stale. A comment like "see autoExpand in bugform.js and keep it in sync with this" or something is good enough.

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,18 @@ function BugForm() {
     }, this));
   };
 
+  this.autoExpand = function() {
+    $(document).one("focus.js-autoExpand", "textarea.js-autoExpand", function() {
+      this.baseScrollHeight = this.scrollHeight;
+    })
+    .on("input.js-autoExpand", "textarea.js-autoExpand", function() {
+      var minRows = 2;
+      this.rows = minRows;
```
miketaylr: Can you talk me through your logic of getting this calculation? I'm still trying to understand a few things.

For example, we set `textarea.rows` to 2 here but the other textarea gets set to 3. Why the difference?

webcompat/static/js/lib/issues.js
```diff
@@ -489,6 +490,21 @@ issues.MainView = Backbone.View.extend({
...
+  },
+
+  // Auto expanding feature for comments
+  // See Line 473 in bugform.js 
+  autoExpand: function() {
+    $(document).one("focus.js-autoExpand", "textarea.js-autoExpand", function() {
+      this.baseScrollHeight = this.scrollHeight;
+    })
+    .on("input.js-autoExpand", "textarea.js-autoExpand", function() {
+      var minRows = this.getAttribute("data-min-rows");
```
miketaylr: FYI, since we're using jQuery, you can just do $(this).data('min-rows').

webcompat/static/js/lib/issues.js
```diff
@@ -489,6 +490,21 @@ issues.MainView = Backbone.View.extend({
...
+      this.rows = minRows;
+      var rows = Math.ceil((this.scrollHeight - this.baseScrollHeight) / 10);
```
miketaylr: We're dividing by 10 here, but 17 in the other place. Why the difference?

webcompat/static/js/lib/bugform.js
nj4710: minRows is the number of rows that I would want to add in the total number of rows.

webcompat/static/js/lib/bugform.js
nj4710: I will leave comments on both the files.

webcompat/static/js/lib/bugform.js
nj4710: It was just my thinking that for bugform 2 would be enough.

webcompat/static/js/lib/issues.js
nj4710: It was to demonstrate the use of that number but I think that might get confusing later. We should use the same number.

nj4710:
If you feel that this approach is way too confusing, we can also do this
`document.getElementByClassName('.js-autoExpand').addEventListener('keyup', function() {
        this.style.overflow = 'hidden';
        this.style.height = 0;
        this.style.height = this.scrollHeight + 'px';
      },false);`

This is probably simpler.

nj4710:
I tried a less confusing approach today. How about this?
r? @miketaylr

nj4710:
Ping @miketaylr

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,17 @@ function BugForm() {
     }, this));
   };
 
+  // See function autoExpand in issues.js
+  this.autoExpand = function() {
+    $(document).one("focus.js-autoExpand", "textarea.js-autoExpand",function(){})
```
miketaylr: This event handler is an empty function, so is it needed?

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,17 @@ function BugForm() {
...
+    .on("input.js-autoExpand", "textarea.js-autoExpand", function() {
+      this.style.overflow = 'hidden';
+        this.style.height = 0;
```
miketaylr: Why are we setting height to 0, then resetting it to `scrollHeight` right after? It's the `scrollHeight` we're really interested in.

webcompat/templates/issue/issue-comment-submit.html
```diff
@@ -11,7 +11,7 @@
   </div>
   <div class="wc-Comment-body">
   <label for="wc-Comment-text" class="wc-Accessibility">Comment Text</label>
-    <textarea class="wc-Comment-submit js-Comment-text mousetrap" placeholder="Leave a comment"></textarea>
+    <textarea class="wc-Comment-submit js-Comment-text mousetrap js-autoExpand" data-min-rows='3' placeholder="Leave a comment"></textarea>
```
miketaylr: `data-min-rows` is now unused, should be removed.

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,17 @@ function BugForm() {
     }, this));
   };
 
+  // See function autoExpand in issues.js
+  this.autoExpand = function() {
+    $(document).one("focus.js-autoExpand", "textarea.js-autoExpand",function(){})
+    .on("input.js-autoExpand", "textarea.js-autoExpand", function() {
+      this.style.overflow = 'hidden';
```
miketaylr: If you run eslint on this code, you'll see a few errors that should be fixed up (double quotes, indentation, etc.)

There's more details on the travis results: https://travis-ci.org/webcompat/webcompat.com/builds/216025269

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,17 @@ function BugForm() {
...
+        this.style.height = 0;
+        this.style.height = this.scrollHeight + 'px';
+
```
miketaylr: nit: extra whitespace

webcompat/static/js/lib/issues.js
```diff
@@ -489,6 +490,19 @@ issues.MainView = Backbone.View.extend({
 
   handleKeyShortcuts: function() {
     Mousetrap.bind("mod+enter", _.bind(this.addNewComment, this));
+  },
+
+  // Auto expanding feature for comments
+  // See function autoExpand in bugform.js
```
miketaylr: Same comments apply here as in bugform.js

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,17 @@ function BugForm() {
     }, this));
   };
 
+  // See function autoExpand in issues.js
+  this.autoExpand = function() {
+    $(document).one("focus.js-autoExpand", "textarea.js-autoExpand",function(){})
+    .on("input.js-autoExpand", "textarea.js-autoExpand", function() {
```
miketaylr: I understand the benefits of jQuery event namespaces (`.js-autoExpand` in this case), but we're not using them here. Maybe we can just listen for `input` to make the code easier to understand?

webcompat/static/js/lib/bugform.js
miketaylr: Also, since we're using jQuery, you can just use the $.fn.css API for consistency. That is `$(this).css({'height': foo, 'overflow': bar'});`

See http://api.jquery.com/css/ for more docs.

webcompat/static/js/lib/bugform.js
nj4710: Once we delete the long comment, the height should decrease too. right? Probably, for that we will need to set height =0

miketaylr: Do you mean we need to adjust the height if the user delete some stuff from their textarea comment? `scrollHeight` will account for that. Have you tested without setting height to 0?

nj4710: Yes, I tested it without setting height to 0. The height did not adjust if the user wanted to delete some stuff from the comment. Could please test the latest commit on your machine?

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,15 @@ function BugForm() {
     }, this));
   };
 
+  // See function autoExpand in issues.js
+  this.autoExpand = function() {
+    $(document).one("focus.js-autoExpand", "textarea.js-autoExpand")
+    .on("input", function() {
+      $(this).css({"overflow":"hidden", "height": "scrollHeight+px"});
```
miketaylr: The string `"scrollHeight+px"` won't do what you want, that's just an invalid string as far as CSS is concerned. Did you try to run this code?

Try: `"height": this.scrollHeight + "px"`

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,15 @@ function BugForm() {
...
+
```
miketaylr: nit: Empty line

webcompat/static/js/lib/bugform.js
nj4710: First, I tried running the code with `"height": this.scrollHeight + "px"`. It did not change the height. Then I tried `"scrollHeight+px"`. The height increased fine but did not adjust when we delete a few lines. Could you please explain me what am I doing wrong here?

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,15 @@ function BugForm() {
     }, this));
   };
 
+  // See function autoExpand in issues.js
+  this.autoExpand = function() {
+    $(document).one("focus.js-autoExpand", "textarea.js-autoExpand")
```
miketaylr: `.one("focus.js-autoExpand", "textarea.js-autoExpand")` isn't doing anything. It can be removed.

webcompat/static/js/lib/bugform.js
miketaylr: That's odd, I also tested without setting height to 0 and it worked.

nj4710:
Tested this code on my machine, the height adjusted for bugform but not for comments. Does this work for you? @miketaylr

webcompat/static/js/lib/bugform.js
nj4710: Then I will set height to 0 too? Although I did not understand why it did not work without it.

miketaylr:
> Tested this code on my machine, the height adjusted for bugform but not for comments. Does this work for you? @miketaylr

I think we should probably pause here. 😄 

It seems like a little bit of learning about how jQuery and the DOM works would be useful to you so we don't have so much of a back and forth (which unfortunately I don't have that much time to do... :(). 

The nature of the changes in this PR suggest that maybe it's a bit above your current knowledge (which means it's a great issue to learn from!). I've been asking some questions along the way to try to get you to learn a little bit more, but many of them have gone unanswered.

You might want to start with the following docs:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style
http://api.jquery.com/on/
http://api.jquery.com/css/
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight

Give those a read and try to figure out why `$(document).on("input","textarea", function() {...}` won't do what you want it to. 

Then, let's continue with the PR. 👍

webcompat/static/js/lib/bugform.js
miketaylr: So, looking at this again, I'm totally wrong. Sorry!

> Once we delete the long comment, the height should decrease too. right? Probably, for that we will need to set height =0

So, we do want to set height to 0 (or some value like the original offsetHeight of the textarea element). The reason why is we need to re-caculate the height since we've deleted some stuff now. 

So sorry for sending you down a confusing rabbit hole there!

nj4710: Oh, Thank you so much for the explanation and sorry for taking so much of your time. I will read more about jquery and get back to this PR as soon as possible.

webcompat/static/js/lib/bugform.js
```diff
@@ -467,6 +470,14 @@ function BugForm() {
     }, this));
   };
 
+  // See function autoExpand in issues.js
+  this.autoExpand = function() {
+    $("textarea.js-autoExpand").on("input", function() {
+     $(this).css({"height":0});
```
miketaylr: This is looking better!

(Small tip, when using jQuery, if you're just setting a single property via `$.fn.css`, convention is to just do `$(foo).css("prop", val)`, rather than use the object syntax.)

That said, I think we want to do something a bit nicer than using 0. What if we set a local variable to the initial height of the textarea, then use that to "reset" the height. That way it will never get shorter than it initially was:


```
this.autoExpand = function() {
  var initialHeight = $("textarea.js-autoExpand").height();
  $("textarea.js-autoExpand").on("input", function() {
  $(this).css({"height": intialHeight});
  ...

miketaylr:
This looks good -- in my testing, the textarea on the bug reporting form doesn't auto-grow in Chrome, but does for the issue comment textarea. Both work in Firefox.

Let's go ahead and merge, but can you file a follow up bug for the Chrome issue @nj4710?

nj4710:
Sure thing!

miketaylr:
Let's close this PR, I'll need to create a new one (because your branch wasn't up to date with latest master, and didn't have the prettier changes, so I've just cherry picked it) and merge from there.

nj4710:
I tested this on chromium. Both works fine in Chromium

miketaylr:
> I tested this on chromium. Both works fine in Chromium

Interesting... I'll test again once we deploy to the production site. Thanks for checking!

Issue #1425, opened 2017-03-14
[Graphics]  SVG icons need to be improved


karlcow:
So following discussions in #1421 PR.

As we can see for the three SVG icons, the one on the left, the report bug, is a bit thinner and smaller. 
![Icons for three steps](https://cloud.githubusercontent.com/assets/505230/23858535/078735f4-0801-11e7-8cec-ed63a54826d7.png)

* [ ] SVG icons need to be re-centered and  in their own view box
* [ ] Report bug icon needs to be thicker for matching the two others
* [x] The included text inside the SVG can be cleaned up too

It might be possible to just improve the tracing. Or just remake them. It's not really complicated but long and a fun project. It requires a bit of expertise with SVG too.

zoepage:
I know, @bastianalbers is working on this and will send a PR soon.

zoepage:
The placement of SVG's was improved and the message was taken out. The other two todo's are still open, if anyone wants to tackle them.

zoepage:
In the new design, there won't be icons as you can see here http://webcompat.github.io/design/ "Module: Hero Tiles". So this issue can be closed.

PR #1443, opened 2017-03-27
#1425 - join team icon display

bastianalbers:
This addresses some of the issues in https://github.com/webcompat/webcompat.com/issues/1425

- removed the text from the svg icons
- fixed issues with flex display and incosistent sizes of icon box by adding an extra wc-Threesteps-item-icon class

Probably still needs to be adjusted with design eye: icon/box sizes.


karlcow:
The modified 3 icons.
capture d ecran 2017-03-28 a 07 40 45

The current PR is trying to adjust the icons through CSS, but not by remaking the SVG (to make them regular). I wonder if it will bite us in the future. We need to check how it looks too.

bastianalbers:
Yes, the SVGs themselves need some work. I tried to Provide a fix for now as the boxes were varying heavily, but i lack the actual SVG editing skills (and tools) to properly modify those paths. I'll try to have another look on those though.

zoepage:
I feel like this is already an improvement to the actual state. 
As long as we keep this is mind and an issue open for this, we should be good, I guess?

@bastianalbers 
screen shot 2017-03-28 at 1 36 42 pm

Did you check Safari? Seems something is not right there. :/

miketaylr:
> As long as we keep this is mind and an issue open for this, we should be good, I guess?

I think that sounds like a good way forward.

zoepage:
As this is not solving #1425, but improving the actual state and we agreed on merging, I did edit the title to not close the bug. This still needs to be improved. 

Thanks @bastianalbers for your help :)

PR #1445, opened 2017-03-28
 Fixes #1276 Added a referer header when new issues are created

### LINKED ISSUE #1276: Add a referer header when new issues are created and redirected to issue page
See https://github.com/webcompat/webcompat.com/pull/1273#issuecomment-271752559

We can use that as a signal to show the "Thanks!" flash message, rather than the current (hacky, imo) solution of using `session`.

See https://github.com/webcompat/webcompat.com/blob/master/webcompat/views.py#L214 and all other instances of `session['show_thanks'] = True`

nj4710:
I am fairly new to python but the docs are really helpful. I tried something, feel free to discard this approach if it is not what you expected.
I am having trouble figuring out what the URL should be in 
`if get_referer(request) == url`
for testing on my machine I gave it the url for localhost.

r? @karlcow

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -119,7 +119,7 @@ def file_issue():
     response = report_issue(session['form_data'])
     # Get rid of stashed form data
     session.pop('form_data', None)
-    session['show_thanks'] = True
+    set_referer(request)
     return redirect(url_for('show_issue', number=response.get('number')))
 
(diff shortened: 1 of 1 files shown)


karlcow:
@nj4710 almost going to bed. I'll check that tomorrow. Thanks.

PR #1451, opened 2017-03-28
Fixes #1450. Updates to CSP policy based on reading logs.

### LINKED ISSUE #1450: More CSP tweakings - Round 2
We've learned a few things, need to update the policy. See https://github.com/webcompat/webcompat.com/issues/763#issuecomment-289847715.

miketaylr:
After we let this one bake for a week or so, I think we'll be close to turning this on "for realsies", then begin the work to remove certain things one by one (i.e., inline scripts, etc.)

r? @karlcow

### DIFF
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -487,8 +487,8 @@ def add_csp(response):
         "connect-src 'self'; " +
         "font-src 'self'; " +
         "img-src 'self' https://www.google-analytics.com https://*.githubusercontent.com; " +  # nopep8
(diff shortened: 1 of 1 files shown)


miketaylr:
> That sounds a bit scary. Does it open the door to inject a script through HTTPS with a bug in a third party script?

It's no more scary than what we currently have (which is nothing). Right now the only 3rd party script we host is Google analytics. The plan is to allow it so we can turn on CSP, then work on a solution where we can remove it (either by removing GA, or moving everything to script tags).

PR #1453, opened 2017-03-28
Fixes #1360. Provide a param to assign a label to a new issue.

### LINKED ISSUE #1360: Add a GET param to add a label by default
(See #1356 for discussion of Media team's needs.)

It would be nice to be able to apply a `type-video` label on issue creation. Probably want to make sure this only works for pre-existing labels to reduce abuse.

miketaylr:
Not ready for review, just opening as a place for discussion right now.

### DIFF
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -211,6 +211,7 @@ def build_formdata(form_object):
     # Append "from webcompat.com" message to bottom (for GitHub issue viewers)
     body += u'\n\n{0}'.format(GITHUB_HELP)
     rv = {'title': summary, 'body': body}
-    if 'label' in form_object:
-        rv.update({'labels': [form_object.get('label')]})
+    extra_label = form_object.get('label', None)
+    if extra_label:
+        rv.update({'labels': [extra_label]})
     return rv
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -23,6 +23,7 @@ from wtforms.validators import InputRequired
 from wtforms.validators import Length
 from wtforms.validators import Optional
 
+from webcompat import app
 from webcompat.api.uploads import Upload
 
 AUTH_REPORT = 'github-auth-report'
@@ -224,6 +225,6 @@ def build_formdata(form_object):
     body += u'\n\n{0}'.format(GITHUB_HELP)
     rv = {'title': summary, 'body': body}
     extra_label = form_object.get('label', None)
-    if extra_label:
+    if extra_label and extra_label in app.config['EXTRA_LABELS']:
         rv.update({'labels': [extra_label]})
     return rv
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -10,6 +10,9 @@ power the issue reporting form on webcompat.com.'''
 import random
 import urlparse
 
+from helpers import get_browser
+from helpers import get_os
+
 from flask_wtf import FlaskForm
 from flask_wtf.file import FileAllowed
 from flask_wtf.file import FileField
@@ -76,6 +79,15 @@ class IssueForm(FlaskForm):
     image = FileField(u'Attach a screenshot image',
                       [Optional(),
                        FileAllowed(Upload.ALLOWED_FORMATS, image_message)])
+
+
+def get_form(ua_header):
+    """Return an instance of flask_wtf.FlaskForm with browser and os info."""
+    bug_form = IssueForm()
+    # add browser and version to bug_form object data
+    bug_form.browser.data = get_browser(ua_header)
+    bug_form.os.data = get_os(ua_header)
+    return bug_form
 
 
 def get_problem(category):
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -210,4 +210,7 @@ def build_formdata(form_object):
             image_url=form_object.get('image_upload').get('url'))
     # Append "from webcompat.com" message to bottom (for GitHub issue viewers)
     body += u'\n\n{0}'.format(GITHUB_HELP)
-    return {'title': summary, 'body': body}
+    rv = {'title': summary, 'body': body}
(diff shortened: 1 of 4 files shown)


webcompat/form.py
```diff
@@ -215,4 +215,7 @@ def build_formdata(form_object):
             image_url=form_object.get('image_upload').get('url'))
     # Append "from webcompat.com" message to bottom (for GitHub issue viewers)
     body += u'\n\n{0}'.format(GITHUB_HELP)
-    return {'title': summary, 'body': body}
+    rv = {'title': summary, 'body': body}
+    if 'label' in form_object:
+        rv.update({'labels': [form_object.get('label')]})
```
miketaylr: Hmm, looking again at the Issues API docs,

> Labels to associate with this issue. NOTE: Only users with push access can set labels for new issues. Labels are silently dropped otherwise.

This probably means that if someone creates an issue authenticated as themselves, it will fail. webcompat-bot proxied issues would be OK. So maybe this is the wrong layer, and I should do this in the labeler webhook instead.

miketaylr:
@karlcow any ideas here? the original plan is here: https://github.com/webcompat/webcompat.com/issues/1360#issuecomment-289904680, but done this way, anybody could add any arbitrary label to our repo because @webcompat-bot has elevated privs.

miketaylr:
The other approach I was thinking about was to fetch the current labels via the API when the app is booted, and keep those in memory. Or we could have a yaml file or similar in the project config, and parse that on app boot-up. 

It really just depends on where we want to manage what labels exist for the web-bugs repo.

If there's a match for an already existing label, we add it via the labeler webhooks. If there isn't a match, we ignore it.

karlcow:
I don't think the creation of labels will be intense. Or at least it should not. We are not in the free tagging business. So I would make a list of accepted labels and use the webhooks.

Benefits it gives us an overview of all labels. And we [have already code](https://github.com/webcompat/webcompat.com/blob/c8065f7b0594e84ed7aa795b5437638c378d1359/config/__init__.py#L46-L61), we can export in a new module dedicated to labels.

miketaylr:
Cool, thanks @karlcow . I'll cook something up in the next few days.

> We are not in the free tagging business.

Speak for yourself... 🎆

karlcow:
@miketaylr The code in my previous comment relates to [namedtuples](https://docs.python.org/3/library/collections.html#collections.namedtuple), a cheap way to create structured objects. 

```python
>>> categories = CATEGORIES
>>> categories
[Category(name='needstriage', dataAttribute='needstriage', label='Needs Triage'), Category(name='needsDiagnosis', dataAttribute='needsdiagnosis', label='Needs Diagnosis'), Category(name='needsContact', dataAttribute='needscontact', label='Needs Contact'), Category(name='ready', dataAttribute='contactready', label='Ready for Outreach'), Category(name='sitewait', dataAttribute='sitewait', label='Site Contacted'), Category(name='close', dataAttribute='closed', label='Closed')]
>>> categories[0].name
'needstriage'
>>> categories[0].dataAttribute
'needstriage'
>>> categories[0].label
'Needs Triage'
```

but we can change that to another model if we prefer or if this one is too confusing. No issue for me. 
 YAML, JSON, n3, rdf, cvs, turtle, name your own evil.

miketaylr:
> I don't think the creation of labels will be intense. Or at least it should not. We are not in the free tagging business. So I would make a list of accepted labels and use the webhooks.

One concern I have about this is browsers that we dont currently know about wont get their automatic labels. But maybe thats not a big issue.... If the triagers note it, they can file a bug and we can just add that new label to the config.

miketaylr:
Oh wait, I'm totally overthinking this. We can just have a small set of labels that whitelist what can be set via the `extra_labels` param, which for now will be 1 or 2 (type-media, type-webvr). That won't affect the labeler webhook.

karlcow:
@miketaylr yup. The benefit of having `category-label` is that we can adjust our policy depending on the `category-`. We can still have limited vocab for some of them.

miketaylr:
(update: i wrote patches for this 2 nights ago, hope to get them up for review today or over the weekend)

webcompat/views.py
```diff
@@ -165,8 +165,9 @@ def create_issue():
         bug_form = get_form(request.headers.get('User-Agent'))
         if g.user:
             get_user_info()
-        if request.args.get('src'):
-            session['src'] = request.args.get('src')
+        for param in ['src', 'label']:
```
miketaylr: I should probably leave a comment here, explaining why `src` and `label` are special, and might need to end up in the session cookie.

miketaylr:
So, I guess `type-media` reports are coming in already, so landing this sooner than later will make less of a headache for triagers (cc @MDTsai @karlcow @adamopenweb).

Not 100% sure about tests, since we don't directly test issue creation.

r? @karlcow

webcompat/views.py
karlcow: yes probably, at least to remember.

webcompat/form.py
```diff
@@ -210,4 +223,8 @@ def build_formdata(form_object):
             image_url=form_object.get('image_upload').get('url'))
     # Append "from webcompat.com" message to bottom (for GitHub issue viewers)
     body += u'\n\n{0}'.format(GITHUB_HELP)
-    return {'title': summary, 'body': body}
+    rv = {'title': summary, 'body': body}
+    extra_label = form_object.get('label', None)
+    if extra_label and extra_label in app.config['EXTRA_LABELS']:
+        rv.update({'labels': [extra_label]})
+    return rv
```
karlcow: Just a note for myself. 
`build_formdata` starts to be a long piece of code. Maybe we want in the future to break it in smaller components. Not sure.

miketaylr: Agree.

PR #1459, opened 2017-03-29
Fixes #1455 - prevent shortcut to open when typing comment

### LINKED ISSUE #1455: Click "e" on webcompat.com comment area opens label menu
Expect behavior: nothing happen
Actual behavior: click "e" always open label menu, which interrupt comment.

zoepage:
r? @miketaylr 

As described in #1455, the label editor opens even when you are typing a comment.
With the fix, it just opens when it doesn't have focus on the textarea. The shortkey was changed from `e` to `l` and info in the footer was added.

### DIFF
--- a/webcompat/static/js/lib/labels.js
+++ b/webcompat/static/js/lib/labels.js
@@ -24,7 +24,16 @@ issues.LabelsView = Backbone.View.extend({
     "click .js-LabelEditorLauncher.is-active": "closeEditor"
   },
   keyboardEvents: {
-    "e": "editLabels"
+    "e": "labelWrap"
+  },
+  labelWrap: function(e) {
+    // make sure we're not typing in the search input.
+    if (e.target.nodeName === "TEXTAREA") {
+      return;
+    } else {
+      e.preventDefault();
+      this.editLabels();
+    }
   },
   template: _.template($("#issue-labels-tmpl").html()),
   // this subTemplate will need to be kept in sync with
--- a/webcompat/static/js/lib/labels.js
+++ b/webcompat/static/js/lib/labels.js
@@ -24,7 +24,7 @@ issues.LabelsView = Backbone.View.extend({
     "click .js-LabelEditorLauncher.is-active": "closeEditor"
   },
   keyboardEvents: {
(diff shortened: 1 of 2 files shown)


webcompat/static/js/lib/labels.js
```diff
@@ -24,7 +24,16 @@ issues.LabelsView = Backbone.View.extend({
     "click .js-LabelEditorLauncher.is-active": "closeEditor"
   },
   keyboardEvents: {
-    "e": "editLabels"
+    "l": "labelWrap"
+  },
+  labelWrap: function(e){
```
miketaylr: nit: can you add a space between `) {` here? jslint is unhappy:

`29:25  error  Missing space before opening brace  space-before-blocks`

webcompat/static/js/lib/labels.js
```diff
@@ -24,7 +24,16 @@ issues.LabelsView = Backbone.View.extend({
     "click .js-LabelEditorLauncher.is-active": "closeEditor"
   },
   keyboardEvents: {
-    "e": "editLabels"
+    "l": "labelWrap"
```
miketaylr: (I don't follow the method name, `labelWrap`, but maybe that's just me)

miketaylr:
Actually, I'll push the fix myself, since you're probably away right now and this is urgent.

miketaylr:
(In a perfect world, we'd have a test for this, so it doesn't regress but I'll file a follow up)

karlcow:
@miketaylr @zoepage 
And this is breaking "g" which now is failing when typing in textarea.

It would be good to find which previous deploy broke the initial e. Maybe that would help a bit. I searched but didn't find. Probably it happened among all the deploy made by @miketaylr after the 3 weeks off (travel + business).

miketaylr:
> It would be good to find which previous deploy broke the initial e.

See here for how this broke: https://github.com/webcompat/webcompat.com/pull/1464#issuecomment-290451911 (basically, a contribution that wasn't well reviewed by me, and a lack of tests for these features).

#1462 was filed to cover the testing scenario so this doesn't break in the future.

PR #1461, opened 2017-03-29
Fixes #1448 - added grunt build as prestart script

### LINKED ISSUE #1448: add npm build to npm start
As a lot of new contributors forget the `npm run build` step and file issues or ping on the channel, because the CSS files are missing.

To help and make the setup steps simpler, it seems like a good idea to put `npm run build` in one of the steps. As running the build before starting the server makes sense, I'd suggest to put it [here](https://github.com/webcompat/webcompat.com/blob/master/package.json#L51).

What do you think?

zoepage:
Added build as prestart to the npm scripts. So it will be called automated before start.


miketaylr:
Oh cool, I didn't know about `prestart` (RTFM to self...):

https://docs.npmjs.com/misc/scripts

PR #1463, opened 2017-03-30
Fixes #1437. makes /rate_limit route send 410 Gone

### LINKED ISSUE #1437: /rate-limit shows webcompat-bot when non authentified
Currently, we are displaying the rate limit when requesting https://webcompat.com/rate_limit and being logged out.

```
Current user: webcompat-bot

Core resources rate limit: 4998 / 5000 requests.
Reset in 45 minutes.

Search rate limit: 30 / 30 requests.
Reset in 1 minute.
```

I don't have the feeling it is useful for anyone but us. Maybe we should just return a 403, with a simple text message saying that rate limit is available only to authenticated users. 

It also removes the need for 
https://github.com/webcompat/webcompat.com/blob/db4e879a167604debfa04d3e208be68cb4f64c6d/webcompat/views.py#L253-L256

webcompat-bot is a person like us.

karlcow:
I can change the quote to something just plain boring.

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -4,7 +4,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import json
 import logging
 import os
 
@@ -31,7 +30,6 @@ from helpers import set_referer
 from issues import report_issue
 from webcompat import app
 from webcompat import github
-from webcompat.api.endpoints import get_rate_limit
 from webcompat.db import User
 from webcompat.db import session_db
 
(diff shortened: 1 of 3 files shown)


webcompat/views.py
```diff
@@ -250,19 +248,19 @@ def show_user_page(username):
...
+    '''Retired route. 410 Gone.
+
+    Decision made on March 2017. See
+    https://github.com/webcompat/webcompat.com/issues/1437
+    '''
+    msg = """
+    All those moments will be lost in time…
+    like tears in rain…
+    Time to die.
+    – Blade Runner
```
miketaylr: ❤️

miketaylr:
r+

PR #1464, opened 2017-03-30
Fixes #1455 - prevent shortkeys to fire in comment textarea

### LINKED ISSUE #1455: Click "e" on webcompat.com comment area opens label menu

zoepage:


webcompat/static/js/lib/labels.js
```diff
@@ -24,9 +24,9 @@ issues.LabelsView = Backbone.View.extend({
     "click .js-LabelEditorLauncher.is-active": "closeEditor"
   },
   keyboardEvents: {
-    "l": "labelWrap"
+    "l": "labelWarp"
```
miketaylr: Ah, now I get `labelWrap`, it was a typo for `labelWarp`. :)))

So, the name `githubWarp` was a silly name playing off the mario brothers concept of using a pipe to warp to another level:

 (hence the dumb `warpPipe` variable name...)

So I'm not sure `labelWarp` is meaningful. What we're doing is opening the label editor.

miketaylr:
(I think this needs to be rebased against latest master...)

So, it turns out this regression was introduced by https://github.com/webcompat/webcompat.com/pull/1431... and it was my suggestion to use mousetrap to handle ctrl+enter... which meant the two existing shortcuts would work. Oops. 🙈 

This looks good, but we should rename the `labelWarp` to something else like `openEditor` or `editLabels`, IMO.

miketaylr:
I'll file a followup issue for the renaming, that's a very good "good first bug". :)

webcompat/static/js/lib/labels.js
zoepage: I got the reference and liked the idea, but I also get the benefits of having a meaningful name. 
+1 on "good first bug"

zoepage:
@miketaylr you merged, so no rebase? :)

miketaylr:
> @miketaylr you merged, so no rebase? :)

Yeah, sorry -- the merge applied cleanly. All good!

PR #1468, opened 2017-03-31
Fixes #1434 - adding Prettier

### LINKED ISSUE #1434: Add prettier instead of eslint
I like the idea behind prettier. 

you can write your code however you like and instantly prettier format it correctly.

http://jlongster.com/A-Prettier-Formatter

We can add a script in the file package.json to automate prettier to start before a commit

magsout:
Closes #1434 

r? @miketaylr you write all JS of webcompat.

Every commit, this recipe https://github.com/webcompat/webcompat.com/pull/1468/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R56 Is triggered.

if any JS is modified, it is automatically added after a `git commit` https://github.com/webcompat/webcompat.com/pull/1468/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R62 ? I'm not sure if it's a good idea.

The configuration is simple, I have modified only this option --trailing-com. (added a comma at the end of object)


miketaylr:
Would it be possible to split out the commit with all the changes to the code? Ideally there would be one commit for removing eslint, one for adding prettier, and one for transforming code.

Also, we'll need to update the .travis.yml config... right now it's exploding on `npm run lint`.

miketaylr:
I'd be interested to have @denschub take a look at this PR as well.

miketaylr:
There's still some eslint specific comments in the code, for example: https://github.com/webcompat/webcompat.com/blob/ea87d2f8278b696f1c602d03cb00bb818c8100f4/webcompat/static/js/lib/comments.js#L5

So if we decide there's no more usage for eslint, we should remove those.

I wonder if we should document how this affects workflow in CONTRIBUTING.md as well.

miketaylr:
I wonder how people are using this w/ eslint though... with this change we lose out on all the recommended linting rules: http://eslint.org/docs/rules/ (and the few that we have added).

magsout:
@miketaylr hum yes, I may have been a bit quick on this PR.

tests/functional/comments-auth.js
```diff
@@ -2,131 +2,158 @@
...
-        .then(function(val) {
-          assert.include(val, "![Screenshot of the site issue](http://localhost:5000/uploads/", "The image was correctly uploaded and its URL was copied to the comment text.");
-        })
-        .end();
-    }
-
-  });
-});
+      "Add a screenshot to a comment": function() {
+        return FunctionalHelpers.openPage(
```
miketaylr: This style is kinda weird, but if I never have to think about authoring it myself, I can live with it. :)

magsout:
>I wonder how people are using this w/ eslint though... with this change we lose out on all the recommended linting rules: http://eslint.org/docs/rules/ (and the few that we have added).

@miketaylr Yes, it's Prettier's role, but maybe it's a bit too gross? You find that you lose what compared to eslint?

tests/functional/comments-auth.js
magsout: or you can change every times you code in this file and let Prettier re-format 💃

miketaylr:
> Yes, it's Prettier's role, but maybe it's a bit too gross? You find that you lose what compared to eslint?

We lose the "safety" stuff that eslint was checking. eslint was also checking boring things like, do you have a space before a paren -- in my mind, this is what Prettier is good at. Don't think about style, let the machine do style. But it doesn't tell us if we accidentally introduced a new global variable, for example: https://github.com/webcompat/webcompat.com/blob/master/.eslintrc#L6

It should be possible to have both eslint and prettier -- just get rid of the code style rules from eslint, and run prettier after that?

I didn't study these yet, but I see:

https://github.com/prettier/eslint-config-prettier
https://github.com/prettier/prettier-eslint

magsout:
>But it doesn't tell us if we accidentally introduced a new global variable, for example: https://github.com/webcompat/webcompat.com/blob/master/.eslintrc#L6

@miketaylr right now, this recipe https://github.com/webcompat/webcompat.com/blob/master/package.json#L48 handle your comment above ?

magsout:
@miketaylr 

hum, let me rethink completely my PR.
Thanks for your feedback and your review.

miketaylr:
> @miketaylr right now, this recipe https://github.com/webcompat/webcompat.com/blob/master/package.json#L48 handle your comment above ?

Yeah, I guess. (But I never use that...)

To be clear, I'm +1 on prettier, it does way more than eslint --fix. I just don't want us to lose linting non-style stuff.

magsout:
>To be clear, I'm +1 on prettier, it does way more than eslint --fix. I just don't want us to lose linting non-style stuff.

Yeah, and I agree with you.

magsout:
@miketaylr another approch. Combined prettier with eslint. I removed rules of eslint and prettier instead. We keep our lint/fix, with the Formatted Code by Prettier.

What do you think ?

Gruntfile.js
```diff
@@ -9,7 +9,7 @@ module.exports = function(grunt) {
     cssPath: "webcompat/static/css",
     imgPath: "webcompat/static/img",
     banner: "/*! <%= pkg.title %>\n" +
-      " *  Copyright (c) <%= grunt.template.today(\"yyyy\") %> <%= pkg.author.name %>\n" +
+      ' *  Copyright (c) <%= grunt.template.today("yyyy") %> <%= pkg.author.name %>\n' +
```
miketaylr: Seems a little funky to use a single quote here, when the lines above and below use double quotes. Any reason why?

miketaylr: (or did prettier do this...)

miketaylr: Looks like prettier did, so i'll stop caring. :D

miketaylr:
Interesting, while trying this locally I got the following error:

```
🐓 npm run lint

> webcompat@ lint /Users/miket/dev/compat/webcompat.com
> eslint ./Gruntfile.js ./tests ./grunt-tasks ./webcompat/static/js/lib


/Users/miket/dev/compat/webcompat.com/webcompat/static/js/lib/issues.js
  118:19  error  Follow `prettier` formatting (expected '\n' but found '"')  prettier/prettier

✖ 1 problem (1 error, 0 warnings)
```

Wonder why that didn't get picked up... (I can fix it up).

miketaylr:
> Interesting, while trying this locally I got the following error:

(ah that's because that code got merged in before @magsout made this PR. forgive me, I'm v slow sometimes...)

miketaylr:
@magsout this looks good, we should probably update CONTRIBUTING.md as well. One question about workflow... for some imaginary scenario:

1. write some code
2. run `npm run lint`
3. run `npm run fix` (if there were simple style errors...)
4. commit, make PR, etc.

Does that seem like the best way?

miketaylr:
Actually, I think we can merge this and I'll file a follow up issue to document recommended workflow.

magsout:
@miketaylr yes, it does 👍

Issue #1470, opened 2017-03-31
Intermittent errors in tests


brizental:
I have played these tests four times, one after the other and each time I got a different result.

The first time I got these errors:
```bash
× firefox on any platform - Issue-list - Pressing g goes to github issues
NoSuchElement: [POST http://localhost:4444/wd/hub/session/47c56f18-2777-4e86-bbf9-0caf33a82ecc/element / {"using":"css selector","value":".issues-listing"}] Unable to locate element: .issues-listing
For documentation on this error, please visit: http://seleniumhq.org/exceptions/no_such_element.html
Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:48:19 -0700'
System info: host: 'bmachado-Lenovo-Z40-70', ip: '127.0.1.1', os.name: 'Linux', os.arch: 'amd64', os.version: '4.4.0-21-generic', java.version: '1.8.0_111'
Driver info: org.openqa.selenium.firefox.FirefoxDriver
[...]

PR #1481, opened 2017-04-05
Fixes #1462 - Add functional tests for when typing 'g' or 'l' in an issue or issue comment

### LINKED ISSUE #1462: Add functional regression test for preventing shortcut to open label editor when typing comment
Fix was landed in https://github.com/webcompat/webcompat.com/pull/1459, but we should write an Intern test for this.

brizental:
r? @miketaylr 

The tests are failing sometimes, but it's the same thing as I explained in #1470

### DIFF
--- a/tests/functional/issues-auth.js
+++ b/tests/functional/issues-auth.js
@@ -53,7 +53,28 @@ define(
           .then(function(text) {
             assert.equal(text, "Close Issue", "Button says Close not Reopen");
           });
+      },
+
+      "Pressing 'l' opens the label editor box": function () {
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues/70"),
+          ".wc-Issue-commentSubmit",
+          true
+        )
+          .findByCssSelector(".wc-Issue-labels")
+          .click()
+          .type("l")
+          .end()
+          .setFindTimeout(0)
+          .findByCssSelector(".js-LabelEditorLauncher")
+          .getAttribute("class")
+          .then(function(className) {
+            assert.include(className, "is-active");
+          })
+          .end();
       }
+
     });
   }
 );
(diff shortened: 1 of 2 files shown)


tests/functional/issues-auth.js
```diff
@@ -53,7 +53,28 @@ define(
...
+          this,
+          url("/issues/70"),
+          ".wc-Issue-commentSubmit",
+          true
+        )
+          .findByCssSelector(".wc-Issue-labels")
+          .click()
+          .type("l")
+          .end()
+          .setFindTimeout(0)
```
miketaylr: I think this has the potential to be racy. Can we remove it?  

(It should fall back to the default that `FunctionalHelpers.openPage` uses (`config.wc.pageLoadTimeout`).

tests/functional/issues-auth.js
```diff
@@ -53,7 +53,28 @@ define(
...
             assert.equal(text, "Close Issue", "Button says Close not Reopen");
           });
+      },
+
+      "Pressing 'l' opens the label editor box": function () {
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues/70"),
+          ".wc-Issue-commentSubmit",
+          true
```
miketaylr: any reason to request a longer timeout here?

brizental: oh sorry, i forgot to remove that.
i was just testing to see if the timeout was the problem with the tests that are intermittently not working.
i'll change it back.

tests/functional/issues-auth.js
brizental: these were also tests that i forgot to remove, my bad

miketaylr:
Thanks!

PR #1494, opened 2017-04-08
Fixes #1485 - removed missing backend-install script call

### LINKED ISSUE #1485: [package.json] `npm run backend-install` missing
In https://github.com/webcompat/webcompat.com/blob/master/package.json#L60 we're trying to call a script called `backend-install`, but after checking the history, I saw it was never defined.

What was the purpose and do we need it?

zoepage:
r? @miketaylr


miketaylr:
Looks good, thanks!

PR #1504, opened 2017-04-12
Fixes #1035 - Unified code for version string

### LINKED ISSUE #1035: Simplifiy/unify/clean up UA parsing code
See @karlcow's suggestion in https://github.com/webcompat/webcompat.com/pull/1032/files#r61705335.

moraveyo:
r? @miketaylr

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -146,10 +146,10 @@ class TestHelpers(unittest.TestCase):
         self.assertEqual(get_browser_name(None), 'unknown')
         self.assertEqual(get_browser_name(), 'unknown')
         self.assertEqual(get_browser_name(u'💀'), 'unknown')
-        self.assertEqual(get_browser('<script>lol()</script>'), 'Unknown')
-        self.assertEqual(get_browser(True), 'Unknown')
-        self.assertEqual(get_browser(False), 'Unknown')
-        self.assertEqual(get_browser(None), 'Unknown')
+        self.assertEqual(get_browser_name('<script>lol()</script>'), 'unknown')
+        self.assertEqual(get_browser_name(True), 'unknown')
+        self.assertEqual(get_browser_name(False), 'unknown')
+        self.assertEqual(get_browser_name(None), 'unknown')
 
     def test_get_browser(self):
         '''Test browser parsing via get_browser helper method.'''
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -17,7 +17,9 @@ import webcompat
 from webcompat.helpers import format_link_header
 from webcompat.helpers import get_browser_name
 from webcompat.helpers import get_browser
+from webcompat.helpers import get_name
 from webcompat.helpers import get_os
+from webcompat.helpers import get_version_string
 from webcompat.helpers import normalize_api_params
 from webcompat.helpers import parse_link_header
 from webcompat.helpers import rewrite_and_sanitize_link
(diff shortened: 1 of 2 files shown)


webcompat/helpers.py
```diff
@@ -100,6 +100,18 @@ def get_user_info():
         session['avatar_url'] = gh_user.get('avatar_url')
 
 
+def version_string(json_section):
```
miketaylr: Can we rename the method to `get_version_string`?

Also, the argument being called `json_section` is a bit hard to understand what exactly it is. It should be a Python Dict, not a json object, right? Maybe we can rename to something like `user_agent_dict` (or just `user_agent`, with a comment that the arg is a dict, in the docstring)?

webcompat/helpers.py
```diff
@@ -100,6 +100,18 @@ def get_user_info():
...
+    '''Gives version string out of section with `major`, `minor`, and `patch`.'''
```
miketaylr: (section here is confusing as well, so this should be updated to reflect what we rename the argument to)

The other `get_foo` methods in this file generally start like `Return blah from`. Can we follow that convention as well?

Also, the comment should ideally indicate that it will only return minor and patch if they exist.

webcompat/helpers.py
moraveyo: About Dict - I'm pretty new to Python, sorry :)
It's used not only for `user_agent`, but for `os` too.

moraveyo: Maybe just `dict` with arg comment about `major`, `minor`, and `patch`?

I first didn't add `get_` to function name because I thought it can be confused with other `get_` functions, that "get" as "extract" some info from protocol objects. While our function is just an ancillary one.

miketaylr: Ah yeah, so maybe something vague like `version_dict` would be better...

miketaylr: `dict` is a built-in type/class, so we don't want to use that name by itself: https://docs.python.org/2/library/stdtypes.html#dict

tests/test_helpers.py
```diff
@@ -144,10 +145,10 @@ def test_get_browser_name(self):
         self.assertEqual(get_browser_name(None), 'unknown')
         self.assertEqual(get_browser_name(), 'unknown')
         self.assertEqual(get_browser_name(u'💀'), 'unknown')
-        self.assertEqual(get_browser('<script>lol()</script>'), 'Unknown')
-        self.assertEqual(get_browser(True), 'Unknown')
-        self.assertEqual(get_browser(False), 'Unknown')
-        self.assertEqual(get_browser(None), 'Unknown')
+        self.assertEqual(get_browser_name('<script>lol()</script>'), 'unknown')
```
miketaylr: Ah, nice catch. 👍

miketaylr:
@moraveyo I'm not sure if this is ready for re-review yet, can you ping when it is?

Thanks!

moraveyo:
r? @miketaylr
What "bizarre UA strings" can I add to tests, to test the name treatment? (Just can't make it up)

miketaylr:
> What "bizarre UA strings" can I add to tests, to test the name treatment? (Just can't make it up)

You could use some of the bogus stuff in these tests: https://github.com/moraveyo/webcompat.com/blob/eadb82be50b3e6d8e7095c57959725b467b0114e/tests/test_helpers.py#L145-L151, or you could invent something wacky like 

"Nozilla/1.0 (TI-83 Graphing Calculator;) CoolBrowser/1.0"

miketaylr:
@moraveyo can you run [pep8](https://pypi.python.org/pypi/pep8) on your changes? There's a few style nits here -- thanks!

```
pep8 webcompat/helpers.py
webcompat/helpers.py:104:1: W293 blank line contains whitespace
webcompat/helpers.py:105:80: E501 line too long (85 > 79 characters)
webcompat/helpers.py:109:19: E701 multiple statements on one line (colon)
webcompat/helpers.py:111:17: E701 multiple statements on one line (colon)
webcompat/helpers.py:113:17: E701 multiple statements on one line (colon)
webcompat/helpers.py:114:19: E221 multiple spaces before operator
webcompat/helpers.py:119:1: W293 blank line contains whitespace
```

moraveyo:
Oops... I first tried `pep8 --show-source --show-pep8 .` from [CONTRIBUTING.md](https://github.com/webcompat/webcompat.com/blob/master/CONTRIBUTING.md#python), and it does million of messages, so I thought it's broken.

moraveyo:
r? @miketaylr
As to "bizarre UA strings", I thought about some string resulting in 'family':'Other' with any version number. After some research I see it's not possible.
So I guess tests already cover things, and it's done.

miketaylr:
LGTM! We just need to clean up the git history. Maybe 3 commits: 1 for the unified code refactor, one for the new tests, and one for fixing the existing tests? LMK if you get stuck on that!

moraveyo:
r? @miketaylr
I'm putting the code with new tests in one commit: there already was a commit with both, I don't know how to split it.
Hope I didn't do wrong with Rebase.

moraveyo:
Looks like I did the mess with rebase

moraveyo:
r? @miketaylr
Now I hope it's OK.

miketaylr:
Awesome, thanks! And yeah, splitting git commits is not fun (or easy...). This looks great, thanks for all your work here.

miketaylr:
(there were some pep8 errors, but I pushed a fixup commit here: https://github.com/webcompat/webcompat.com/commit/e02eccba317414ba3da0926b209dc04eece6885b)

moraveyo:
Strange, I've already removed some spaces so that `pep8 tests/test_helpers.py` made no messages for remaining ones.
All those spaces were for visual clearness...

moraveyo:
I first even tried `  # nopep8` at the end of each line - didn't work.

miketaylr:
Hmm, odd that we saw different things. In general, python style doesn't like whitespace for alignment. See https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements (look for "More than one space around an assignment (or other) operator to align it with another. ")

PR #1507, opened 2017-04-14
Fixes #1506: Upgrade Pretter to 1.1.0

### LINKED ISSUE #1506: Upgrade Prettier to 1.0
1.0 was just released: http://jlongster.com/prettier-1.0

miketaylr:
I know we don't want too many changes like this (because it makes reading history harder), but prettier is considered to be stable now, and changes this introduces are more readable, IMO.

r? @magsout

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -71,25 +71,22 @@ function BugForm() {
     // Set up listener for message events from screenshot-enabled add-ons
     window.addEventListener(
       "message",
-      _.bind(
-        function(event) {
-          // Make sure the data is coming from ~*inside the house*~!
-          // (i.e., our add-on or some other priviledged code sent it)
-          if (location.origin === event.origin) {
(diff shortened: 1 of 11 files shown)


webcompat/static/js/lib/bugform.js
```diff
@@ -111,32 +108,29 @@ function BugForm() {
...
+      }
 
-        this.addPreviewBackgroundAndUpload(screenshotData);
-        (img = null), (canvas = null);
-      },
-      this
-    );
+      this.addPreviewBackgroundAndUpload(screenshotData);
+      (img = null), (canvas = null);
+    }, this);
```
miketaylr: This change makes me very happy, before it was too many line breaks/whitespace. 👍

magsout:
@miketaylr so coool, this version is awesome, 

r+ 👍 

Great work @vjeux and @jlongster

webcompat/static/js/lib/bugform.js
vjeux: 👍

PR #1513, opened 2017-04-15
Fixes #1512. Upgrade Grunt-related dependencies.

### LINKED ISSUE #1512: Update Grunt deps
(I know we want to move away from Grunt, but our grunt deps are currently super out of date)

Let's see if we can upgrade without breaking anything.

miketaylr:
Everything seems to "just work™" when running grunt locally.

Let's see if any Travis tests explode (which would be surprising...)

r? @magsout


package.json
```diff
@@ -22,18 +22,18 @@
...
     "grunt-contrib-imagemin": "~1.0.0",
-    "grunt-contrib-uglify": "~0.2.7",
-    "grunt-contrib-watch": "~0.5.3",
+    "grunt-contrib-uglify": "^2.3.0",
+    "grunt-contrib-watch": "^1.0.0",
     "grunt-postcss": "^0.8.0",
     "grunt-travis-lint": "0.0.3",
     "intern": "3.4.0",
-    "load-grunt-tasks": "~0.6.0",
+    "load-grunt-tasks": "^3.5.2",
```
magsout: 3 major version ^^

magsout:
looks good r+ 👍

package.json
miketaylr: Yeah, I  was sure it would explode. 💣

Issue #1023, opened 2016-04-28
Search is broken


adamopenweb:
Looks like search is broken.
STR:
- Navigate to webcompat.com
- Click and type 'video'
- Click View All Open Issues and switch to View All Issues

Expected Results:
Display related bugs

Actual Results:
Error: "Not Found. Lost in Punk Cat Space"

Also navigating direct to this link doesn't work: https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=%22country%3A+jp%22

hallvors:
IMHO this is a side effect of #795 - the site does not allow you filter by various separate options right now. You can't both give a search word and a bug state, for example. The bugs #795 and #863 are about fixing that but my attempts so far were not solid enough.

(While I think it could be closed as a dup, I guess it will take time to fix those issues so we should perhaps add a workaround here)

miketaylr:
I don't think this is a dupe, this used to work.

miketaylr:
I think this bug is related, if not the cause:

0) from /issues, search for "country: jp". Note that it works and the URL is `https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=country%3A+jp`.
1) Try to navigate to `https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=country%3A+jp` (or just refresh).
2) Notice the URL is actually `https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=country%253A%2Bjp`

I think we're double encoding the URL here and the GitHub API is like wtf r u doing.

miketaylr:
I suspect that's happening here: https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/issue-list.js#L517
`jQuery.param` does `encodeURIComponent`: https://github.com/jquery/jquery/blob/df2051cf59d054d7ee91bdb554b583471942c6b4/src/serialize.js#L57-L90

So, original stuff works, but we are encoding already encoded junk and then it breaks.

miketaylr:
> I think this bug is related, if not the cause:

Nah, not the cause, just related to crappy search code. But might as well fix it in this branch.

magsout:
@miketaylr I writed a `SearchForm` Component with React, Redux, History, Form Serialized, Fetch ... Maybe it can be interesting you for webcompat

[...]

Issue #1474, opened 2017-04-03
Document recommended eslint/prettier workflow


miketaylr:
Support for prettier landed in #1468.

We should determine a simple workflow, and document it so we're all on the same page.

Strawman proposal:

 1.  write some code
 2.   run npm run lint
 3.   run npm run fix (if there were simple style errors...)
 4.   commit, make PR, etc.

magsout:
@miketaylr looks good 👍 it's I think the best way.

zoepage:
+1

Do we want to add a hook to lint automatically before committing? 
https://github.com/prettier/prettier#pre-commit-hook-for-changed-files

magsout:
@zoepage 

>Do we want to add a hook to lint automatically before committing?

It was my first thought and I made it in my initial PR. But after I did not know if it was good or practical to do that especially after the comment of @karlcow  https://github.com/webcompat/webcompat.com/issues/1434#issuecomment-288255993

zoepage:
@magsout Just to clarify, I was just talking about linting. Not fixing :)

Not a fan of automated fixing either, so I agree with @karlcow. Linting is something people like to forget (as same as running the build and wondering, why the styles are missing... Me. Always. :grin:)

miketaylr:
Decision: we'd like to figure out a pre-commit style workflow. But possibly good to document manual way as well?

magsout:
@zoepage 
>@magsout Just to clarify, I was just talking about linting. Not fixing :)

ah yes, you meant `npm run lint` ? not `npm run fix`  ?

magsout:
So something like:

```js
"scripts": {
    "precommit": "lint-staged",
 },
"lint-staged": {
    "{webcompat/static/js/lib,tests}__}/**/*.js": [
      "npm run lint"
    ]
  },
```

miketaylr:
So with a pre-commit lint, I guess once you try to commit w/ style errors, it will fail and then it's up to you to run `npm run fix`? @zoepage is that what you would expect?

zoepage:
@miketaylr 
The pre-commit should lint after adding the files through `git add, but before the commit itself. 
Then you could run `npm run fix` or fix manually. This is what I would expect. :)

miketaylr:
Cool, thanks! I can live with that. :)

@magsout want to implement what you wrote in https://github.com/webcompat/webcompat.com/issues/1474#issuecomment-291502599?

miketaylr:
[...]

PR #1518, opened 2017-04-18
Fixes #1515 - fixed encoding param

### LINKED ISSUE #1515: Typing "label:type-media" doesn't work in the homepage search form
Noticed by @magsout 

Steps to reproduce.
1. Go the homepage
2. click/tap on the search icon
3. Enter `label:type-media`

Expected:
Get the list of bugs with `type-media` label

Actual:
Empty list with a URL being 
`https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=label%253Atype-media`

Note: `&q=label%253Atype-media`

Going to the issues list page and typing `label:type-media` in the search form is working and gives a URL
`https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=label%3Atype-media`

but if I take this URI and I cut and paste it in the browser, I get back the gibberish URI from the home page
`https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=label%253Atype-media` 
where the browser is trying to convert the percent into `%25`. So the progression goes indeed like this.

1. `label%3Atype-media`
2. `label%253Atype-media`
3. `label%25253Atype-media`

etc.

magsout:
r? @miketaylr 

I have fixed the issue #1515  and more.

- url: param q is encoded with `encodeURIComponent` to avoid `%25`
- input search is set when we refresh page if there is a `q` param
- `_currentSearch` is init directly in `this.issueList = new issueList.IssueView();`  to avoid double search at first page init.
- param `q` is removed of XHR rquest when search input is empty

~~There is still a bug. When you delete the search field. It is necessary to send a request (ok) and delete the param `q` (NOK, the test is here https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/issue-list.js#L464)~~

### Edit

Seems to fixe this issue #1023 

~~Last bug: Delete the request parameter https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/issue-list.js#L464~~

### DIFF
--- a/webcompat/static/js/lib/issue-list.js
+++ b/webcompat/static/js/lib/issue-list.js
@@ -489,7 +489,7 @@ issueList.IssueView = Backbone.View.extend(
           var kvArray = param.split("=");
           var key = kvArray[0];
           var value = kvArray[1];
-          this.issues.params[key] = value;
+          this.issues.params[key] = decodeURIComponent(value);
         }, this)
       );
     },
--- a/webcompat/static/js/lib/issue-list.js
+++ b/webcompat/static/js/lib/issue-list.js
@@ -232,9 +232,10 @@ issueList.SearchView = Backbone.View.extend({
     // if it's empty and the user searches, show the default results
     // (but only once)
     if (!searchValue.length && this._currentSearch !== "") {
+      this.doSearch("");
       this.currentSearch("");
+      issueList.events.trigger("filter:clear", { removeQ: true });
       this._isEmpty = true;
-      issueList.events.trigger("issues:update");
     }
   }, 350),
   doSearch: _.throttle(function(value) {
@@ -485,12 +486,21 @@ issueList.IssueView = Backbone.View.extend(
           this.issues.setURLState("/api/issues/search", paramsCopy);
         }
       } else if (_.contains(issuesAPICategories, category)) {
-        this.issues.setURLState("/api/issues/category/" + category, params);
+        this.issues.setURLState(
+          "/api/issues/category/" + category,
+          this.removeParamQuery(params)
+        );
       } else {
-        this.issues.setURLState("/api/issues", params);
+        this.issues.setURLState("/api/issues", this.removeParamQuery(params));
       }
 
       this.fetchAndRenderIssues();
+    },
+    removeParamQuery: function(params) {
+      if (params && params.q) {
+        delete params.q;
+      }
+      return params;
     },
     addParamsToModel: function(paramsArray) {
       // this method just puts the params in the model's params property.
--- a/webcompat/static/js/lib/issue-list.js
+++ b/webcompat/static/js/lib/issue-list.js
@@ -336,14 +336,18 @@ issueList.IssueView = Backbone.View.extend(
       // There are some params in the URL
       if (urlParams.length !== 0) {
         queryMatch = urlParams.match(this._searchRegex);
+        if (queryMatch) {
+          _.delay(function() {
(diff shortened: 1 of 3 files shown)


miketaylr:
Whoa, awesome! Will have time to review this afternoon.

miketaylr:
@magsout can you push a prettier error fix commit so tests can run?

magsout:
>@magsout can you push a prettier error fix commit so tests can run?


Rah, need to add this issue https://github.com/webcompat/webcompat.com/issues/1474

magsout:
Seems to fixe this issue https://github.com/webcompat/webcompat.com/issues/1023

magsout:
@miketaylr hum, hum where is this commit https://github.com/webcompat/webcompat.com/commit/5998f12efa47be0dbb49b97ee85f5851414a1c82 because it fixed the bug like I did.. But I cannot find it..

magsout:
ok, I get it. I find how to handle my last case, let me ush something

miketaylr:
> hum, hum where is this commit 5998f12 because it fixed the bug like I did.. But I cannot find it..

Hmmm...... 👻

magsout:
@miketaylr I add this https://github.com/webcompat/webcompat.com/pull/1518/files#diff-4ba6be3180115f78e14125480dcf4d7dR495

I don't know if it's clean or not.

magsout:
Maybe it's better to make another PR, I fixed lot of things, what do you think @miketaylr  ?

miketaylr:
> Maybe it's better to make another PR, I fixed lot of things, what do you think @miketaylr ?

There's not that much code changing in these 4 commits, I think it should be fine here.

webcompat/static/js/lib/issue-list.js
```diff
@@ -336,14 +336,18 @@ issueList.IssueView = Backbone.View.extend(
       // There are some params in the URL
       if (urlParams.length !== 0) {
         queryMatch = urlParams.match(this._searchRegex);
+        if (queryMatch) {
```
miketaylr: I wonder if there's a difference in behavior for logged-in or not logged-in users...

miketaylr:
@magsout this looks good, but I'd like to write some tests as well. Can you help me think of the scenarios we should test?

1. have a `type-media` label in the test repo, make sure that `label:type-media` search query returns a result
2. have a q param with `foo`, and the search input should load with `foo`

> param q is removed of XHR rquest when search input is empty

How can we test this one?

magsout:
@miketaylr is it enought if:

3) have a q param with something With 0 result
4) remove q param and have result again ?

magsout:
>r+, but let's write some tests before merging. I'm happy to write the tests, unless you want to

@miketaylr  unless it can be wait one or two weeks, be my guest ;)

webcompat/static/js/lib/issue-list.js
magsout: Hum, I don't think so..

miketaylr:
:))))) I'll work on these over the next few days.

miketaylr:
> @miketaylr is it enought if:

Another way would just be to load a page with `q=thing`, then search for `""` and check that `q=thing` isn't in the URL.

miketaylr:
@magsout just pushed some tests, will likely merge tomorrow morning (don't want to break the universe at the end of the day, in case something goes wrong). ❤️

PR #1524, opened 2017-04-24
Fix #1523 - Adds a list of files to ignore for rsync

### LINKED ISSUE #1523: Create a deploy.ignore file for fabfile
After discussing with @miketaylr 
Currently, we rely on .cvsignore file. It doesn't have to be system dependent and clutter the space of every person deploying the site.
We can create a deploy.ignore file in our own repo. 

rsync --exclude-from=/path/to/deploy.ignore

karlcow:
As discussed in Berlin to ease a common process for deployment.

r? @miketaylr 

Do not forget to modify your fabfile.


miketaylr:
r+

PR #1555, opened 2017-05-11
Fixes #1037 - Removes dead code and method.

### LINKED ISSUE #1037: HTTP POST tests
This is good.`http POST http://localhost:5000/issues`

``` http
HTTP/1.0 405 METHOD NOT ALLOWED
Allow: HEAD, GET, OPTIONS
Content-Length: 178
Content-Type: text/html
Date: Tue, 03 May 2016 04:41:55 GMT
Server: Werkzeug/0.10.4 Python/2.7.10

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>405 Method Not Allowed</title>
<h1>Method Not Allowed</h1>
<p>The method is not allowed for the requested URL.</p>

```

But this is "not good". (I know it is a leftover of the preview design)

`http POST http://localhost:5000/`

``` http
HTTP/1.0 200 OK
Content-Length: 21155
Date: Tue, 03 May 2016 04:43:06 GMT
Server: Werkzeug/0.10.4 Python/2.7.10
Set-Cookie: session=eyJjc3JmX3Rva2VuIjp7IiBiIjoiWXprMlpUTXdOemRqTlRZME5XRTNOemMwTURNMVlUTXpZVFZqTVRrNFptTTVNalkwTURsak5BPT0ifX0.Cgm92g.LNDGffBQG-ZCkaEcA0gnzhdPZ9M; HttpOnly; Path=/

```

When the POST is successful the user get a `302` with redirection to a thanks page. 

This could be fixed at the same time than #646 
with:
- returns 405 for any URI which are not about issues reporting so everything but `/issues/new`
- returns 201 Created to the successful POST with Location: being the issue page.

karlcow:
This removes dead code on the home page. Basically we never use this code anymore since we completely transitioned to /issues/new for creating a new issue. 

To check if add-on is hitting "/" instead of "/issues/new" @miketaylr will know

r? @miketaylr

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -121,7 +121,7 @@ def file_issue():
     return redirect(url_for('show_issue', number=response.get('number')))
 
 
-@app.route('/', methods=['GET', 'POST'])
+@app.route('/', methods=['GET'])
 def index():
     '''Main view where people come to report issues.'''
     ua_header = request.headers.get('User-Agent')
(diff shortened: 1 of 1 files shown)


miketaylr:
> To check if add-on is hitting "/" instead of "/issues/new" @miketaylr will know

If people have older versions installed they will.

miketaylr:
r+ thanks @karlcow!

Issue #1560, opened 2017-05-12
Cannot close or reopen issues


miketaylr:
Fallout from #1559.

STR
1. go to any open issue
2. try to close issue

expected: 200 PATCH to edit endpoint
actual: 403

I have a quick fix I'll push then ask @karlcow to review after (and probably improve).

zoepage:
@miketaylr as the push seemed to fix the issue and you pushed to master, shall we close the issue?

miketaylr:
Ah yeah, this should have been closed. Thanks!

zoepage:
Thanks!

PR #1561, opened 2017-05-12
Issue #1560. Fix regression that prevents closing or re-opening an issue.

miketaylr:
r=miketaylr so we can deploy and unbreak master.

But, I'd like @karlcow to review after the fact and file any bugs for improvement.


miketaylr:
OK, push build is at least green. Let's merge.

Issue #1553, opened 2017-05-10
The status-needstriage label is added and removed after filing an issue


softvision-sergiulogigan:
Browser / Version: Chrome 57.0.2987.133 (Desktop)
Operating System: Windows 10

**Steps to Reproduce:**
1. Navigate to: https://github.com/webcompat/web-bugs/issues/6591.
2. Observe the labels assigned.

**Expected Behavior:**
The status-needstriage label is applied.

**Actual Behavior:**
The status-needstriage label is applied and then removed.

**Notes:**
1. The issue is also reproducible for https://github.com/webcompat/web-bugs/issues/6580

sv;
![screenshot_1](https://cloud.githubusercontent.com/assets/16573771/25903572/78ee2780-35a5-11e7-963c-cb0d2a3f18fa.png)

softvision-sergiulogigan:
Could this be related to https://github.com/webcompat/webcompat.com/issues/1228 ?

karlcow:
@softvision-sergiulogigan I'm inclined to think that's the case. I will come up with something in #1228 
As for now I will close this one as a duplicate of the other and we will see if my modifications are working.

PR #1566, opened 2017-05-19
Fix #1228 - Improves webhooks to send labels at once

### LINKED ISSUE #1228: Investigate why issue didn't get "needstriage" label
@adamopenweb noticed the following issue didn't get a "needstriage" label:

https://github.com/webcompat/web-bugs/issues/3646

It was reported w/ `Mozilla/5.0 (Android 4.4.4; Tablet; rv:51.0) Gecko/51.0 Firefox/51.0 ` -- so possibly the UA string parsing for the labeler webhook has some bugs.

karlcow:
r? @miketaylr 

Instead of sending two requests which might have created the issue #1553, we are now sending one request only.
We also created a bunch of additional tests in their own file.
Comments in each commit have additional details explaining the context. 

You can test locally the Webhook by doing.

```
# start webcompat.com on localhost
python run.py
# In another shell
http POST http://localhost:5000/webhooks/labeler 'Content-Type:application/json' 'X-GitHub-Event: issues' 'X-Hub-Signature:sha1=43da93fa444d40c2bbec6c3b055ef29859870c79' < /path/to/webcompat.com/tests/fixtures/webhooks/new_event_valid.json
```

Testing

```
nosetests --verbosity=2 --nocapture 
API access to comments greater than 30 returns pagination in Link ... ok
API issue search with bad keywords returns JSON 404. ... ok
Check ETAG for issues. ... ok
Test issues and integer for: ... ok
Test that the /login route 302s to GitHub. ... ok
Checks 500 is not accepted for /issues/new POST. ... ok
POST with action different of opened fails. ... ok

Ran 55 tests in 6.471s

OK
```

or for just testing the webhook.

```
nosetests --verbosity=2 --nocapture tests/test_webhook.py
```

### DIFF
--- a/tests/test_webhook.py
+++ b/tests/test_webhook.py
@@ -18,11 +18,7 @@ from webcompat.webhooks import helpers
 sys.path.append(os.path.realpath(os.pardir))
 import webcompat  # nopep8
 
-
-# Any request that depends on parsing HTTP Headers (basically anything
-# on the index route, will need to include the following: environ_base=headers
-headers = {'HTTP_USER_AGENT': ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.9; '
-                               'rv:31.0) Gecko/20100101 Firefox/31.0')}
+# The key is being used for testing and computing the signature.
 key = app.config['HOOK_SECRET_KEY']
 
 
--- a/tests/test_webhook.py
+++ b/tests/test_webhook.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Tests for our webhooks."""
+
+import json
+import os
+import sys
+import unittest
+
+from webcompat import app
+from webcompat.webhooks import helpers
+
+# Add webcompat module to import path
+sys.path.append(os.path.realpath(os.pardir))
+import webcompat  # nopep8
+
+
+# Any request that depends on parsing HTTP Headers (basically anything
+# on the index route, will need to include the following: environ_base=headers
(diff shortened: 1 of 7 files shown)


webcompat/webhooks/helpers.py
```diff
@@ -40,21 +27,21 @@ def parse_and_set_label(body, issue_number):
...
-def dump_to_db(title, body, issue_number):
-    url = extract_url(body)
-    issue_db.add(WCIssue(issue_number, title, url, body))
-    issue_db.commit()
+    headers = {
+        'Authorization': 'token {0}'.format(app.config['OAUTH_TOKEN'])
+    }
+    uri = 'https://api.github.com/repos/{0}/{1}/labels'.format(
+        app.config['ISSUES_REPO_URI'], issue_number)
+    return requests.post(uri, data=json.dumps(payload), headers=headers)
```
miketaylr: Any reason we're not using `api_post` here?

karlcow: Yeah. That was the essence of my comment on https://github.com/webcompat/webcompat.com/issues/1228#issuecomment-300952788
with 

> So Basically we are doing the same thing with two different functions. We probably would gain rationalizing a bit. :)

I can certainly modify a bit more. so we use the only one. I was hesitating and I was wondering if there was an initial reason for insulation.

tests/test_webhook.py
```diff
@@ -0,0 +1,126 @@
...
+from webcompat.webhooks import helpers
+
+# Add webcompat module to import path
+sys.path.append(os.path.realpath(os.pardir))
+import webcompat  # nopep8
+
+
+# Any request that depends on parsing HTTP Headers (basically anything
+# on the index route, will need to include the following: environ_base=headers
+headers = {'HTTP_USER_AGENT': ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.9; '
```
karlcow: Let me test that. 
indeed mistake. We should remove it. There's no dependency on it. I didn't add any tests going to GitHub live. And there should not be anyway.

karlcow:
@miketaylr Modifications done.

miketaylr:
Thanks Karl. Let's merge.

Issue #1547, opened 2017-05-02
Refactor bug submit form


zoepage:
## Context
-> https://webcompat.com/issues/new & https://webcompat.com/?open=1

More Context: https://github.com/webcompat/webcompat.com/issues/1007#issuecomment-297720355 (please see the work week notes in the details)

**NOTE:** *There will NOT be a redesign yet as seen in the mockups.* We aim to change the form structure first as well as the wording of the labels. :)

## Todos
**Text**
- [x] Finalize wording of labels (https://github.com/webcompat/webcompat.com/issues/1007#issuecomment-297736870)
- [x] Figure out which values are optional / required

**HTML/CSS**
- [x] Switch upper grid cells in the form (as seen in mockups)
- [x] Split textarea in two different input fields. Add `Describe what was wrong?` and `How did you get there?` as replacement for `Expected behavior` and `Actual behavior`
- [x] Group inputs for browser / OS version
- [x] Change label of browser / OS version to `Is this information correct?`
- [x] Add `Did you test in another browser?` label and two radio buttons with [yes/no] options.
- [ ] Make new form information readable in issue template (https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/issue.html)

**TESTS**
- [ ] Update tests
- [ ] After this gets merged, follow up to check that the fix applied for #1552 does not get broken

**API/PYTHON/JS** 
- [x] Add fields for `Did you test in another browser?` radio buttons
- [x] Change textarea -> two input fields and store information
- [ ] Massaging the data before sending them to the issue template. The structure being slightly different. (https://github.com/webcompat/webcompat.com/issues/1547#issuecomment-298896321)
- [ ] Update form validation in bugform.js 

## Mockups

**Desktop**
![desktop hd](https://cloud.githubusercontent.com/assets/562354/25616704/9fdd5914-2f3e-11e7-9a07-f62755f29f67.png)

**Mobile**
![mobile portrait](https://cloud.githubusercontent.com/assets/562354/25616705/9fdf4652-2f3e-11e7-8496-bfa9516b5fe4.png)

zoepage:
@adamopenweb @miketaylr @karlcow anything I forgot in the todos?

miketaylr:
I added a "Update form validation in bugform.js" to the JS section, maybe there should be a "figure out what is optional/required" task as well?

zoepage:
@miketaylr thanks. Also added the task.

karlcow:
[...]

PR #1577, opened 2017-06-01
fix: #1547; bug report form content updates;

cch5ng:
also validation updates and tests

hi @zoepage I tried to leave the old validation in place and keep its logic for now. not sure when and whether to expect feedback on new validation suggestions.

I'm not clear about this bugform.js > checkParams() function, around 
https://github.com/cch5ng/fork-webcompat.com/blob/issues/1547/5/webcompat/static/js/lib/bugform.js#L188

not sure if this logic requires an update since the old descField got split into 2 fields?

I added just a couple of tests but let me know if additional tests should be added

### DIFF
--- a/tests/functional/reporting-non-auth.js
+++ b/tests/functional/reporting-non-auth.js
@@ -103,6 +103,36 @@ define(
           .end();
       },
 
+      "description validation": function() {
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues/new"),
+          ".wc-ReportForm-actions-button"
+        )
+          .findByCssSelector("#description")
+          .click()
+          .end()
+          .execute(function() {
+            var elm = document.querySelector("#description");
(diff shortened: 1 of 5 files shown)


zoepage:
Hey @cch5ng!
You pushed this branch again to your repository. Pushed it to the webcompat.com repo as in (https://github.com/webcompat/webcompat.com/tree/issues/1547/5). There are again the issues with the GH API.

Could you open a PR with this one and add all your changes to it from now on? :)
No need to create a new version of the branch. We do this after merging a PR and a new issue related to the issue number comes up (which almost never happens). Thanks!

> not sure if this logic requires an update since the old descField got split into 2 fields?

Yes, it needs to be split up. LMK, if you have specific questions. Thanks!

cch5ng:
closing this PR as it should be made in a branch on webcompat/webcompat.com

Issue #1552, opened 2017-05-10
“Report” button becomes active after adding a screenshot and no other details are typed/set for the issue


softvision-oana-arbuzov:
**URL:** www.webcompat.com  
**Browser / Version:** Firefox 52.0.2, Chrome 56.0.2924 
**Operating System:** Windows 10 Pro

**Steps to Reproduce:**
1. Navigate to: www.webcompat.com.   
2. Tap “Report bug” button.
3. Without typing any details or selecting an option, attach a screenshot.
4. Observe behavior.

**Expected Behavior:**
Error messages are displayed (“Problem type required”, “A valid URL is required”, “Need more Details”) and “Report” button becomes active only after the mandatory fields/options are set.

**Actual Behavior:**
No error messages are displayed and “Report” button becomes active.

**Notes:**
1. Issues is logged if tapping the “Report” button.
2. Check example #[6596](https://github.com/webcompat/web-bugs/issues/6596).
3. Screenshot attached.

**Watchers:**
@softvision-sergiulogigan
@softvision-oana-arbuzov

sv; 

![reportissue_nodetails](https://cloud.githubusercontent.com/assets/12184325/25891276/1db3d88c-3579-11e7-90f5-a1e6d59e54a0.png)

zoepage:
We should tackle this one in #1547.

cch5ng:
This seems like a relatively small fix, removing this line:
https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L457

I am just wondering if there was a reason why it was included originally. This change will be made in #1547

cch5ng:
![screen shot 2017-06-02 at 6 29 58 pm](https://cloud.githubusercontent.com/assets/4755432/26749517/bf063dec-47c1-11e7-92d4-e3c97fab3481.png)

there is a pointer to issue #1049 in the blame for line 415 but I still don't understand why that line was needed

karlcow:
@miketaylr see question above. I don't remember either.

miketaylr:
> I am just wondering if there was a reason why it was included originally.

Hm, it's possible I meant to call `checkForm` there... but I don't remember. If removing it doesn't fail tests, that sounds like a good fix. :)

cch5ng:
revisiting this because @zoepage found a bug with my pr for it

[...]

PR #1578, opened 2017-06-01
Fixes #1547 - new bug form content refactor

### LINKED ISSUE #1547: Refactor bug submit form

cch5ng:
latest:
* text area placeholder update
* text area style/height update

### DIFF
--- a/tests/functional/reporting-non-auth.js
+++ b/tests/functional/reporting-non-auth.js
@@ -103,61 +103,6 @@ define(
           .end();
       },
 
-      "(optional) browser + os validation": function() {
-        return (
-          FunctionalHelpers.openPage(
-            this,
-            url("/issues/new"),
-            ".wc-ReportForm-actions-button"
-          )
-            // make sure we can see the valid checkbox (i.e. it's background image is non-empty)
-            .execute(function() {
-              return window
-                .getComputedStyle(
-                  document.querySelector(
-                    "div.wc-Form-group:nth-child(2) > span:nth-child(2)"
-                  ),
-                  ":after"
-                )
-                .getPropertyValue("background-image");
-            })
-            .then(function(bgImage) {
-              assert.include(
-                bgImage,
-                "valid.svg",
-                "The valid checkbox psuedo is visible"
-              );
-            })
-            .end()
-            .execute(function() {
-              var elm = document.querySelector("#os");
-              elm.value = "";
-              WindowHelpers.sendEvent(elm, "input");
-            })
-            .end()
-            .sleep(500)
-            // make sure we can't see the valid checkbox (i.e. it's background image is empty)
-            .execute(function() {
-              return window
-                .getComputedStyle(
-                  document.querySelector(
-                    "div.wc-Form-group:nth-child(3) > span:nth-child(2)"
-                  ),
-                  ":after"
-                )
-                .getPropertyValue("background-image");
-            })
-            .then(function(bgImage) {
-              assert.notInclude(
-                bgImage,
-                "valid.svg",
-                "The valid checkbox psuedo is not visible"
-              );
-            })
-            .end()
-        );
-      },
-
       "Problem type validation": function() {
         return (
           FunctionalHelpers.openPage(
@@ -186,6 +131,57 @@ define(
             .end()
             // validation message should be gone
             .waitForDeletedByCssSelector(".wc-Form-helpMessage")
+            .end()
+        );
+      },
+
+      "(optional) browser + os validation": function() {
+        return (
+          FunctionalHelpers.openPage(
+            this,
+            url("/issues/new"),
+            ".wc-ReportForm-actions-button"
+          )
+            // make sure we can see the valid checkbox (i.e. it's background image is non-empty)
+            .execute(function() {
+              return window
+                .getComputedStyle(
+                  document.querySelector(".js-bug-form-os"),
+                  ":after"
+                )
+                .getPropertyValue("background-image");
+            })
+            .then(function(bgImage) {
+              assert.include(
+                bgImage,
+                "valid.svg",
+                "The valid checkbox pseudo is visible"
+              );
+            })
+            .end()
+            .execute(function() {
+              var elm = document.querySelector("#os");
+              elm.value = "";
+              WindowHelpers.sendEvent(elm, "input");
+            })
+            .end()
+            .sleep(500)
+            // make sure we can't see the valid checkbox (i.e. it's background image is empty)
+            .execute(function() {
+              return window
+                .getComputedStyle(
+                  document.querySelector(".js-bug-form-os"),
+                  ":after"
+                )
+                .getPropertyValue("background-image");
+            })
+            .then(function(bgImage) {
+              assert.notInclude(
+                bgImage,
+                "valid.svg",
+                "The valid checkbox pseudo is not visible"
+              );
+            })
             .end()
         );
       },
@@ -239,7 +235,8 @@ define(
             .click()
             .end()
             .findByCssSelector("#description")
-            .click()
+            .type("test for desktop site")
+            //.click()
             .end()
             // wait a bit
             .sleep(250)
@@ -255,6 +252,7 @@ define(
         );
       },
 
+      // tests a scenario where bug reported from firefox nightly media type tool
       "problem_type param selects problem type": function() {
         return FunctionalHelpers.openPage(
           this,
@@ -275,7 +273,7 @@ define(
           url("/issues/new?details=" + DETAILS_STRING),
           "#description"
         )
-          .findByCssSelector("#description")
+          .findByCssSelector("#steps_reproduce")
           .getProperty("value")
           .then(function(text) {
             assert.include(
@@ -292,7 +290,7 @@ define(
           url("/issues/new?details=" + DETAILS_STRING2),
           "#description"
         )
-          .findByCssSelector("#description")
+          .findByCssSelector("#steps_reproduce")
           .getProperty("value")
           .then(function(text) {
             assert.notInclude(
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -54,3 +54,22 @@ class TestForm(unittest.TestCase):
 
         r = form.get_metadata(('cool', 'wow'), TEST_DICT)
         self.assertEqual(r, EXPECTED_MULTIPLE)
+
+    def test_radio_button_label(self):
+        '''Checks that appropriate radio button label is returned.'''
+        TEST_LABELS_LIST = [
+            (u'detection_bug', u'Desktop site instead of mobile site'),
+            (u'unknown_bug', u'Something else')
+        ]
+
+        r = form.get_radio_button_label('unknown_bug', TEST_LABELS_LIST)
+        self.assertEqual(r, u'Something else')
+
+        r = form.get_radio_button_label(u'detection_bug', TEST_LABELS_LIST)
+        self.assertEqual(r, u'Desktop site instead of mobile site')
+
+        r = form.get_radio_button_label(None, TEST_LABELS_LIST)
+        self.assertEqual(r, u'Unknown')
+
+        r = form.get_radio_button_label('failme', TEST_LABELS_LIST)
+        self.assertEqual(r, u'Unknown')
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -34,30 +34,37 @@ GITHUB_HELP = u'_From [webcompat.com](https://webcompat.com/) with ❤️_'
 
 problem_choices = [
     (u'detection_bug', u'Desktop site instead of mobile site'),
-    (u'mobile_site_bug', u'Mobile site is not usable'),
-    (u'video_bug', u'Video doesn\'t play'),
-    (u'layout_bug', u'Layout is messed up'),
-    (u'text_bug', u'Text is not visible'),
-    (u'unknown_bug', u'Something else - I\'ll add details below')
+    (u'site_bug', u'Site is not usable'),
+    (u'layout_bug', u'Design is broken'),
+    (u'video_bug', u'Video or audio doesn\'t play'),
+    (u'unknown_bug', u'Something else')
+]
+
+tested_elsewhere = [
+    (u'yes', u'Yes'),
+    (u'no', u'No')
 ]
 
 url_message = u'A valid URL is required.'
 image_message = (u'Please select an image of the following type:'
-                 ' jpg, png, gif, or bmp.')
+                 u' jpg, png, gif, or bmp.')
 radio_message = u'Problem type required.'
 username_message = u'A valid username must be {0} characters long'.format(
     random.randrange(0, 99))
 
-problem_label = (u'What seems to be the trouble?',
-                 '<span class="wc-Form-required">*</span>')
+desc_label = (u'Please describe what was wrong'
+              u' <span class="wc-Form-required">*</span>')
+desc_message = u'An issue description is required.'
+
 url_label = u'Site URL <span class="wc-Form-required">*</span>'
+browser_test_label = u'Did you test in another browser?'
 
-desc_default = u'''1. Navigate to: Site URL
-2. …
+steps_default = u'''For example,
+1. I've tried to log in.
+2. I've filled out the form details.
+3. I clicked on the submit button.
+4. Nothing happened.
 
-Expected Behavior:
-
-Actual Behavior:
 '''
 
 
@@ -65,15 +72,19 @@ class IssueForm(FlaskForm):
     '''Define form fields and validation for our bug reporting form.'''
     url = StringField(url_label,
                       [InputRequired(message=url_message)])
-    browser = StringField(u'Browser / Version', [Optional()])
+    browser = StringField(u'Is this information correct?', [Optional()])
     os = StringField(u'Operating System', [Optional()])
     username = StringField(u'Username',
                            [Length(max=0, message=username_message)])
-    description = TextAreaField(u'Give more details', [Optional()],
-                                default=desc_default)
-    problem_category = RadioField(problem_label,
-                                  [InputRequired(message=radio_message)],
+    description = StringField(desc_label,
+                              [InputRequired(message=desc_message)])
+
+    steps_reproduce = TextAreaField(u'How did you get there?', [Optional()],
+                                    default=steps_default)
+    problem_category = RadioField([InputRequired(message=radio_message)],
                                   choices=problem_choices)
+    browser_test = RadioField(browser_test_label, [Optional()],
+                              choices=tested_elsewhere)
     # we filter allowed type in uploads.py
     # Note, we don't use the label programtically for this input[type=file],
     # any changes here need to be updated in form.html.
@@ -91,11 +102,11 @@ def get_form(ua_header):
     return bug_form
 
 
-def get_problem(category):
+def get_radio_button_label(field_value, label_list):
     '''Return human-readable label for problem choices form value.'''
-    for choice in problem_choices:
-        if choice[0] == category:
-            return choice[1]
+    for value, text in label_list:
+        if value == field_value:
+            return text
     # Something probably went wrong. Return something safe.
     return u'Unknown'
 
@@ -105,7 +116,7 @@ def get_problem_summary(category):
     if category == 'unknown_bug':
         return u'see bug description'
     else:
-        return get_problem(category).lower()
+        return get_radio_button_label(category, problem_choices).lower()
 
 
 def wrap_metadata(metadata):
@@ -162,12 +173,15 @@ def build_formdata(form_object):
     '''Convert HTML form data to GitHub API data.
 
     Summary -> title
-    Browser -> part of body, labels
     Version -> part of body
     URL -> part of body
-    Description -> part of body
-    Image Upload -> part of body
     Category -> labels
+    Detail -> part of body
+    Description -> part of body
+    Browser -> part of body, labels
+    OS -> part of body, labels
+    Tested Elsewhere -> labels
+    Image Upload -> part of body
 
     We'll try to parse the Browser and come up with a browser label, as well
     as labels like mobile, desktop, tablet.
@@ -202,20 +216,23 @@ def build_formdata(form_object):
         'url': form_object.get('url'),
         'browser': form_object.get('browser'),
         'os': form_object.get('os'),
-        'problem_type': get_problem(form_object.get('problem_category')),
-        'description': form_object.get('description')
+        'problem_type': get_radio_button_label(
+            form_object.get('problem_category'), problem_choices),
+        'browser_test_type': get_radio_button_label(form_object.get(
+            'browser_test'), tested_elsewhere),
+        'description': form_object.get('description'),
+        'steps_reproduce': form_object.get('steps_reproduce')
     }
 
     # Preparing the body
     body = u'''{metadata}
 **URL**: {url}
+**Problem type**: {problem_type}
+**Description**: {description}
+**Steps to Reproduce** {steps_reproduce}
 **Browser / Version**: {browser}
 **Operating System**: {os}
-**Problem type**: {problem_type}
-
-**Steps to Reproduce**
-{description}
-
+**Tested Another Browser**: {browser_test_type}
 '''.format(**formdata)
     # Add the image, if there was one.
     if form_object.get('image_upload') is not None:
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -22,11 +22,21 @@ function BugForm() {
       valid: null,
       helpText: "Problem type required."
     },
+    description: {
+      el: $("#description"),
+      valid: null,
+      helpText: "Description required."
+    },
+    steps_reproduce: {
+      el: $("#steps_reproduce"),
+      valid: true,
+      helpText: null
+    },
     image: {
       el: $("#image"),
       // image should be valid by default because it's optional
       valid: true,
-      helpText: "Image must be one of the following: jpg, png, gif, or bmp.",
+      helpText: "Image must be one of the following: jpe, jpg, jpeg, png, gif, or bmp.",
       altHelpText: "Please choose a smaller image (< 4MB)"
     },
     browser: {
@@ -38,6 +48,11 @@ function BugForm() {
       el: $("#os"),
       valid: true,
       helpText: null
+    },
+    browser_test_type: {
+      el: $("[name=browser_test]"),
+      valid: true,
+      helpText: null
     }
   };
 
@@ -46,14 +61,19 @@ function BugForm() {
   this.problemType = this.inputs.problem_type.el;
   this.uploadField = this.inputs.image.el;
   this.urlField = this.inputs.url.el;
-  this.descField = $("#description");
+  this.descField = this.inputs.description.el;
+  this.browserTestField = this.inputs.browser_test_type.el;
+  this.stepsToReproduceField = this.inputs.steps_reproduce.el;
 
   this.init = function() {
     this.checkParams();
     this.disableSubmits();
-    this.urlField.on("input", _.bind(this.copyURL, this));
     this.urlField.on("blur input", _.bind(this.checkURLValidity, this));
     this.descField.on("focus", _.bind(this.checkProblemTypeValidity, this));
+    this.descField.on(
+      "blur input",
+      _.bind(this.checkDescriptionValidity, this)
+    );
     this.problemType.on("change", _.bind(this.checkProblemTypeValidity, this));
     this.uploadField.on("change", _.bind(this.checkImageTypeValidity, this));
     this.osField
@@ -147,7 +167,6 @@ function BugForm() {
       // weird Gecko bug. See https://bugzilla.mozilla.org/show_bug.cgi?id=1098037
       urlParam = this.trimWyciwyg(urlParam[1]);
       this.urlField.val(decodeURIComponent(urlParam));
-      this.copyURL();
       this.makeValid("url");
     }
 
@@ -159,10 +178,10 @@ function BugForm() {
       $("[value=" + problemType[1] + "]").click();
     }
 
-    // If we got a details param, add that to the end of the issue description.
+    // If we got a details param, add that to the end of the steps to reproduce field
     var details = location.href.match(/details=([^&]*)/);
     if (details !== null) {
-      this.descField.val(function(idx, value) {
+      this.stepsToReproduceField.val(function(idx, value) {
         return (
           value +
           "\n" +
@@ -246,6 +265,16 @@ function BugForm() {
     }
   };
 
+  /* Check to see that the description input element is not empty. */
+  this.checkDescriptionValidity = function() {
+    var val = this.descField.val();
+    if ($.trim(val) === "") {
+      this.makeInvalid("description");
+    } else {
+      this.makeValid("description");
+    }
+  };
+
   /* Check if Browser and OS are empty or not, only
      so we can set them to valid (there is no invalid state) */
   this.checkOptionalNonEmpty = function() {
@@ -268,11 +297,13 @@ function BugForm() {
     var inputs = [
       this.problemType.filter(":checked").length,
       this.urlField.val(),
+      this.descField.val(),
       this.uploadField.val()
     ];
     if (_.some(inputs, Boolean)) {
       // then, check validity
       this.checkURLValidity();
+      this.checkDescriptionValidity();
       this.checkProblemTypeValidity();
       this.checkImageTypeValidity();
       // and open the form, if it's not already open
@@ -316,10 +347,9 @@ function BugForm() {
           .removeClass("is-error js-form-error");
         break;
       case "url":
-        inlineHelp.insertAfter("label[for=" + id + "]");
-        break;
+      case "description":
       case "problem_type":
-        inlineHelp.appendTo("fieldset .wc-Form-information");
+        inlineHelp.insertAfter("label[for=" + id + "]");
         break;
       case "image":
         // hide the error in case we already saw one
@@ -357,7 +387,8 @@ function BugForm() {
     if (
       this.inputs["url"].valid &&
       this.inputs["problem_type"].valid &&
-      this.inputs["image"].valid
+      this.inputs["image"].valid &&
+      this.inputs["description"].valid
     ) {
       this.enableSubmits();
     }
@@ -491,26 +522,9 @@ function BugForm() {
       img_url,
       ")"
     ].join("");
-    this.descField.val(function(idx, value) {
-      return value + "\n\n" + imageURL;
+    this.stepsToReproduceField.val(function(idx, value) {
+      return value + "\n" + imageURL;
     });
-  };
-  /*
-     copy URL from urlField into the first line of the
-     description field. early return if the user has deleted
-     the first so we don't make them sad.
-  */
-  this.copyURL = function() {
-    var firstLine = /^1\.\sNavigate.*\n/;
-    this.descField.val(
-      _.bind(function(idx, value) {
-        var prefix = "1. Navigate to: ";
-        if (!firstLine.test(value)) {
-          return value;
-        }
-        return value.replace(firstLine, prefix + this.urlField.val() + "\n");
-      }, this)
-    );
   };
 
   // See function autoExpand in issues.js
--- a/webcompat/templates/home-page/form.html
+++ b/webcompat/templates/home-page/form.html
@@ -15,9 +15,25 @@
       <div class="r-Grid r-Grid--withGutter">
         <div class="r-Grid-cell r-minM--1of2">
           <div class="wc-Form-group js-Form-group">
+            <div class="wc-Form-information">
+              {{ form.url.label(class_='wc-Form-label') }}
+              {% if form.url.errors %}
+                {% for error in form.url.errors %}
+                  <span class="wc-Form-helpMessage">{{ error }}</span>
+                {% endfor %}
+              {% endif %}
+            </div>
+            <span class="wc-Form-validation">
+            {{ form.url(class_='wc-Form-item required',
+               placeholder='e.g. https://example.com',
+               required=True, type='url', pattern='') }}
+            </span>
+          </div>
+
+          <div class="wc-Form-group js-Form-group">
             <fieldset class="{% if form.problem_category.errors %}is-error{% endif %}">
               <div class="wc-Form-information">
-                <label class="wc-Form-label">What seems to be the trouble? <span class="wc-Form-required">*</span></label>
+                <label class="wc-Form-label" for="problem_type">Type of bug <span class="wc-Form-required">*</span></label>
                 {% if form.problem_category.errors %}{% for error in form.problem_category.errors %}
                 <span class="wc-Form-helpMessage">{{ error }}</span>
                 {% endfor %}{% endif %}
@@ -29,44 +45,54 @@
                 {% endfor %}
             </fieldset>
           </div>
-        </div>
-        <div class="r-Grid-cell r-minM--1of2">
-          <div class="wc-Form-group wc-Form-group--biggerOffset js-Form-group {% if form.url.errors %}is-error{% endif %}">
+
+          <div class="wc-Form-group js-Form-group">
             <div class="wc-Form-information">
-              {{ form.url.label(class_='wc-Form-label') }}
-              {% if form.url.errors %}
-                {% for error in form.url.errors %}
+              {{ form.description.label(class_='wc-Form-label') }}
+              {% if form.description.errors %}
+                {% for error in form.description.errors %}
                   <span class="wc-Form-helpMessage">{{ error }}</span>
                 {% endfor %}
               {% endif %}
             </div>
             <span class="wc-Form-validation">
-            {{ form.url(class_='wc-Form-item required',
-               placeholder='e.g., http://www.example.com',
-               required=True, type='url', pattern='') }}
+            {{ form.description(class_='wc-Form-item required',
+               placeholder='e.g. The video should have started after 3 sec.',
+               required=True, type='text', pattern='') }}
             </span>
           </div>
-          <div class="wc-Form-group wc-Form-group--biggerOffset js-Form-group">
+          <div class="wc-Form-group js-Form-group">
+            {{ form.steps_reproduce.label(class_='wc-Form-label') }} {{ form.steps_reproduce(class_='wc-Form-item wc-Form-item--textarea js-autoExpand', rows=4) }}
+          </div>
+        </div>
+        <div class="r-Grid-cell r-minM--1of2">
+
+          <div class="wc-Form-group wc-Form-group--biggerOffset js-Form-group {% if form.url.errors %}is-error{% endif %}">
             {{ form.browser.label(class_='wc-Form-label') }}
             <span class="wc-Form-validation">
             {{ form.browser(class_='wc-Form-item') }}
             </span>
-          </div>
-          <div class="wc-Form-group wc-Form-group--biggerOffset js-Form-group">
-            {{ form.os.label(class_='wc-Form-label') }}
-            <span class="wc-Form-validation">
+            <span class="wc-Form-validation js-bug-form-os">
             {{ form.os(class_='wc-Form-item') }}
             </span>
           </div>
-        </div>
-      </div>
-      <div class="r-Grid r-Grid--withGutter ">
-        <div class="r-Grid-cell r-Grid-cell--alignBottom r-minS--1of2">
-            <div class="wc-Form-group js-Form-group">
-              {{ form.description.label(class_='wc-Form-label') }} {{ form.description(class_='wc-Form-item wc-Form-item--textarea js-autoExpand', rows=7) }}
-            </div>
-        </div>
-        <div class="r-Grid-cell r-Grid-cell--alignBottom r-minS--1of2">
+
+          <div class="wc-Form-group js-Form-group">
+            <fieldset class="{% if form.browser_test.errors %}is-error{% endif %}">
+              <div class="wc-Form-information">
+                {{ form.browser_test.label(class_='wc-Form-label') }}
+                {% if form.browser_test.errors %}{% for error in form.browser_test.errors %}
+                <span class="wc-Form-helpMessage">{{ error }}</span>
+                {% endfor %}{% endif %}
+              </div>
+                <div class="wc-Form-radio">
+                {% for subradio in form.browser_test %}
+                  {{ subradio }} {{ subradio.label }}
+                {% endfor %}
+                </div>
+            </fieldset>
+          </div>
+
             <div class="js-image-upload wc-Form-group js-Form-group {% if form.image.errors %}is-error{% endif %}">
               <label
                 class="wc-UploadForm wc-UploadForm--homePage js-image-upload-label"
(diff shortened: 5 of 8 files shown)


cch5ng:
@zoepage I found some additional issues with validation today. if you could hold off the review for a bit, I will work further on fixes. thanks!

zoepage:
Sure, thanks for letting me know :) just ping when you need feedback or have questions!

cch5ng:
@zoepage I think the travis process is failing due to prettier issues. am I expected to run the files through the editor at 

https://prettier.github.io/prettier/#%7B%22content%22%3A%22hello%20(%20%5C%22world%5C%22%5Cn)%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22none%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%7D%7D

to fix these issues? when I look at the source, it looks ok.

but otherwise, I think the content (labels, placeholders) should be ok. I just reverted the validation (select description and a warning pops up for bug categories) to the original logic b/c otherwise my changes broke the functionality. validation logic should be revisited I think.

#1552 was fixed

I think for the form, Guillame's grid was a little nicer in terms of vertical spacing. but if I followed his grid, this would break the spec's field order when the view shifted between desktop vs mobile view (b/c it had 2 rows defined). now there is only one row defined.

for functional tests I tried to fix mostly reporting and issues tests but there are a number of failing tests regardless. where should I communicate those? thanks...

cch5ng:
@zoepage hi, I made a mistake adding a file to .gitignore and will remove that change on Mon. I think also I need to improve tests for the new form fields. is it better for me to keep working or wait and get a review first? thanks.

tests/functional/reporting-non-auth.js
```diff
@@ -155,14 +151,14 @@ define(
...
+            // .execute(function() {
+            //   var elm = document.querySelector("#steps_reproduce");
+            //   WindowHelpers.sendEvent(elm, "input");
+            // })
             .sleep(800)
             // using xpath because 2 errors display here
-            .findByXpath(
-              "/html/body/div/main/section/div[1]/form/div/div[1]/div[3]/div/span"
+            .findByCssSelector(
+              "div.is-error:nth-child(3) > div:nth-child(1) > span:nth-child(2)"
```
cch5ng: @zoepage I am stuck with this selector. I grabbed it using firebug and it passes when I run the test locally but keeps failing on travis. any suggestions? thanks

zoepage:
The PR looks good so far. 

I found one thing related to #1552. When you upload an image and the required info is set, you remove the image the submit buttons get inactive. 

If you again edit the required fields, it becomes active again. 

@cch5ng can you reproduce it?

cch5ng:
@zoepage 

> I found one thing related to #1552. When you upload an image and the required info is set, you remove the image the submit buttons get inactive.

> If you again edit the required fields, it becomes active again.

yes, I can reproduce this. I will look into it.

zoepage:
@cch5ng & I have decided to take #1552 out of this PR and to open a separate one, so @karlcow can take over and finish the last bits (as we discussed in the April work week).

cch5ng:
@zoepage the last pr tests are still running but it includes removing #1552, and there were a few other changes in my branch I had not pushed. thanks.

karlcow:
> @cch5ng & I have decided to take #1552 out of this PR and to open a separate one, so @karlcow can take over and finish the last bits (as we discussed in the April work week).

refresh my memory. which last bits? :)

karlcow:
I need to better understand here. What is being asked of me :)

karlcow:
@cch5ng Ping me when we can start a review of this pull request on the python side.

webcompat/form.py
```diff
@@ -48,32 +52,42 @@
...
+                              [InputRequired(message=desc_message)])
+
+    steps_reproduce = TextAreaField(u'How did you get there?', [Optional()],
+                                    default=steps_default)
     problem_category = RadioField(problem_label,
                                   [InputRequired(message=radio_message)],
                                   choices=problem_choices)
+    browser_test = RadioField(problem_label,
+                              [Optional()],
+                              choices=browser_test_choices)
```
karlcow: So if I understand/remember, this field is a radio button asking if we tested in another browser. 
https://cloud.githubusercontent.com/assets/562354/25616704/9fdd5914-2f3e-11e7-9a07-f62755f29f67.png

For the variable name I would pick up another name. It is confusing right now. Something like `choices=tested_elsewhere`

@zoepage Currently it is a radio field button Yes/No. I wonder if it should be just one checkbox for a "yes" (true). And with a no for (false) default value? No strong idea. Just wondering.

webcompat/form.py
```diff
@@ -100,6 +114,15 @@ def get_problem(category):
     return u'Unknown'
 
 
+def get_browser_test(category):
```
karlcow: the name is probably not good too. Something like `get_tested_elsewhere()` and it should probably return True/False. but as said in other comments, this function has not been written. yet.

It probably would be good to add tests for most of the functions in this code. I realized we have never written tests for all the functions. And it's probably why we don't get to know if it's legit or not. 

Proposal: Add tests for the new functions you created. Testing input/output. And open a new issue for adding tests to all the functions you have not touched in this PR and assign it to me if you wish. 
https://github.com/webcompat/webcompat.com/blob/master/tests/test_form.py

webcompat/form.py
```diff
@@ -48,32 +52,42 @@
...
-                                default=desc_default)
+    description = StringField(desc_label,
+                              [InputRequired(message=desc_message)])
+
+    steps_reproduce = TextAreaField(u'How did you get there?', [Optional()],
+                                    default=steps_default)
     problem_category = RadioField(problem_label,
                                   [InputRequired(message=radio_message)],
                                   choices=problem_choices)
+    browser_test = RadioField(problem_label,
```
cch5ng: hi, you are right. that is the wrong label. I got a little confused because in the html, sometimes the label is defined in that file. for ex

https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/home-page/form.html#L20

should the label value always be coming from the form.py definition?

webcompat/form.py
cch5ng: > What this function is supposed to do?
it seems to be the copy of get_problem but without any changes on it. Not finished?
What parameters does it accept, and what is supposed to return?

I did copy it from the existing code for the problem categories. It appeared to be used like an iterator to get the labels for the Yes/No radio buttons. But that whole section, with the formdata variable definition and build_formdata() function seemed a bit abstract. I tried looking up docs for WTForms but even then that section was not clear for me.

Do you mean that I should have updated the comment and argument names?

karlcow: It's better to have different names for different things. And probably starting with the test will help clarify what we really need to achieve with the function. Try as much as possible to write your test first that will fail. Then write your function to fulfill the tests. It helps a lot to have a clearer idea. 

if two functions are really doing the same thing, then it's only one function which should be versatile enough to handle the two cases to avoid repeating code. In the current form of the code, it looks like only one function. But probably we need to wait for @zoepage answer on the design Radio field or checkbox.

webcompat/form.py
```diff
@@ -203,19 +226,20 @@ def build_formdata(form_object):
         'browser': form_object.get('browser'),
         'os': form_object.get('os'),
         'problem_type': get_problem(form_object.get('problem_category')),
-        'description': form_object.get('description')
+        'browser_test_type': get_browser_test(form_object.get('browser_test')),
```
cch5ng: for buildFormData() comments, I'm not sure if these items need to stay in:

    Summary -> title
    Version -> part of body

webcompat/form.py
cch5ng: ok, I agree it should probably be one function. but will wait for further decision.

webcompat/form.py
karlcow: Oh! check that. We are not using the problem_label in the form.html. It's written there instead of passing the variable. This is dead code. 
https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/home-page/form.html#L17-L31

This was added by @miketaylr two years ago.
https://github.com/webcompat/webcompat.com/search?q=problem_label&type=Commits&utf8=%E2%9C%93

but was never used.
https://github.com/webcompat/webcompat.com/blob/ec418f78ba0209a9c06f04fdd089aa4af30ebb8a/webcompat/templates/form.html#L7


No wonder you are confused.

So we can remove `problem_label` safely from the source code of `form.py`

a good way to test a library is to use the console.

```python
>>> from wtforms import Form, RadioField
>>> from wtforms.validators import Optional
>>> browser_test_choices = [(u'yes', u'Yes'), (u'no', u'No')]
>>> problem_label = (u'What seems to be the trouble?', '<span class="wc-Form-required">*</span>')
>>> class MyForm(Form):
...     browser_test = RadioField(problem_label, [Optional()], choices=browser_test_choices)
... 
>>> form = MyForm()
>>> str(form.browser_test)
'<ul id="browser_test"><li><input id="browser_test-0" name="browser_test" type="radio" value="yes"> <label for="browser_test-0">Yes</label></li><li><input id="browser_test-1" name="browser_test" type="radio" value="no"> <label for="browser_test-1">No</label></li></ul>'
```

zoepage:
@karlcow 
Here you can find the todo checklist, we've created after our work week: https://github.com/webcompat/webcompat.com/issues/1547#issue-225677195

Would you be able to check, if anything is missing in the PR or we forgot a todo? 
Thank you!

webcompat/form.py
cch5ng: @karlcow not clear about this function, form_object.get('problem_category') from 
https://github.com/webcompat/webcompat.com/blob/master/webcompat/form.py#L192

is it getting the current value selected on the webpage form?

webcompat/form.py
miketaylr: > So we can remove problem_label safely from the source code of form.py

Won't it throw at that point? `RadioField` needs a label argument. Another option is to use `form.problem_category.label(class_='wc-Form-label')` in the template, rather than the bare HTML.

miketaylr: Oh wait, I think I know why we didn't do that. It's because that will generate the `<label></label>` and we need content to go inside. Since we're not using the value of `problem_label` in form.py, maybe we should just leave a comment saying it's unused, but `RadioField` does need a label argument.

miketaylr:
I haven't reviewed this yet -- I never got any r? @miketaylr notifications, sorry about that.

(my email is set up to put those in a special bucket... the new style review requests slip through the cracks when things get busy...)

miketaylr:
@cch5ng can you ping me for review when things are ready again? @karlcow has the python bits squared away, but I'd like to check over bugform.js since I wrote 99% of that. Thanks!

cch5ng:
@miketaylr sure. sorry about that. I'd been making prs for a couple weeks and didn't want to bug you during the conference. but that would be great. thanks.

> @cch5ng can you ping me for review when things are ready again? @karlcow has the python bits squared away, but I'd like to check over bugform.js since I wrote 99% of that. Thanks!

zoepage:
@karlcow 

> @zoepage Currently it is a radio field button Yes/No. I wonder if it should be just one checkbox for a "yes" (true). And with a no for (false) default value? No strong idea. Just wondering.

I feel like the Yes / No is the better option as it can be tricky to best on mobile (and in a hurry) on another browser, but with just `yes` it feels mandatory.

cch5ng:
@karlcow R?  the refactor for the functions for getting radio button labels was done. it can be adjusted later if necessary. one unit test was added and it appeared to pass on travis (do python tests only run on travis?).

Mike had a couple later comments about adding back the problem_label (var/argument) which are todos.

karlcow:
@zoepage 

> I feel like the Yes / No is the better option as it can be tricky to best on mobile (and in a hurry) on another browser, but with just yes it feels mandatory.

It makes sense. We should probably put "No" as the checked value. Most people have not tested elsewhere.

zoepage:
@karlcow 
> It makes sense. We should probably put "No" as the checked value. Most people have not tested elsewhere.

Adam had a very good point there when we discussed this during the work week. When we set `no` as a default, the reporter will assume that it is not important to test in another browser. But this is actually something we want to teach them. This is why we decided on not having a default value here.

karlcow:
ah! indeed inducing/encouraging a good behavior. I wonder how we could measure that if people really see the radio button or will just ignore with the default answer. Also how does it help us or not to have false positives or is it business as usual. And 42 too.

webcompat/form.py
karlcow: in the deployed template, we have:

```html
<label class="wc-Form-label" for="problem_type">Type of bug <span class="wc-Form-required">*</span></label>
```

So to check I commented it and replace it by:

```python
<!-- <label class="wc-Form-label" for="problem_type">Type of bug <span class="wc-Form-required">*</span></label> -->
                {{ form.problem_category.label(class_='wc-Form-label') }}
```

```python
problem_label = ''.join((u'What seems to be the trouble?',
                         u'<span class="wc-Form-required">*</span>'))
```

which gives me:
capture d ecran 2017-06-09 a 08 37 25

The markup of this is:

```html
<div class="wc-Form-information">
                <!-- <label class="wc-Form-label" for="problem_type">Type of bug <span class="wc-Form-required">*</span></label> -->
                <label class="wc-Form-label" for="problem_category">What seems to be the trouble?<span class="wc-Form-required">*</span></label>
                
              </div>
```

Maybe the trouble @miketaylr had is 

```python
problem_label = (u'What seems to be the trouble?',
                             u'<span class="wc-Form-required">*</span>')
```

which gives:

capture d ecran 2017-06-09 a 08 34 13

Did I miss anything?

tests/test_form.py
```diff
@@ -54,3 +54,16 @@ def test_metadata_wrapping(self):
...
+        TEST_LABELS_LIST = [
+            (u'detection_bug', u'Desktop site instead of mobile site'),
+            (u'unknown_bug', u'Something else')
+        ]
+
+        r = form.get_radio_button_label('unknown_bug', TEST_LABELS_LIST)
+        self.assertEqual(r, u'Something else')
+
+        r = form.get_radio_button_label(u'detection_bug', TEST_LABELS_LIST)
+        self.assertEqual(r, u'Desktop site instead of mobile site')
```
karlcow: adding two more tests.

```python

        r = form.get_radio_button_label(None, TEST_LABELS_LIST)
        self.assertEqual(r, u'Unknown')

        r = form.get_radio_button_label('failme', TEST_LABELS_LIST)
        self.assertEqual(r, u'Unknown')
```

This makes sure we get the right thing when the key is unknown.

webcompat/templates/home-page/form.html
```diff
@@ -15,9 +15,25 @@ <h3 class="wc-Title--s">Report Site Issue</h3>
...
+               placeholder='e.g. https://example.com',
+               required=True, type='url', pattern='') }}
+            </span>
+          </div>
+
+          <div class="wc-Form-group js-Form-group">
             <fieldset class="{% if form.problem_category.errors %}is-error{% endif %}">
               <div class="wc-Form-information">
-                <label class="wc-Form-label">What seems to be the trouble? <span class="wc-Form-required">*</span></label>
+                <label class="wc-Form-label" for="problem_type">Type of bug <span class="wc-Form-required">*</span></label>
```
karlcow: See my comment in https://github.com/webcompat/webcompat.com/pull/1578#discussion_r121022903 We can use and pass the variables.

webcompat/form.py
cch5ng: @karlcow I think he is refering to the WTForm RadioField. see https://wtforms.readthedocs.io/en/latest/fields.html#field-definitions
(scroll down a couple paragraphs to the base class where the constructor takes a label parameter). but to me it seems label is optional since there's a default None value (?).

ex RadioField: https://github.com/webcompat/webcompat.com/blob/master/webcompat/form.py#L74

in the current PR, I also removed the problem_label parameter and I didn't see any issue in the form display or ability to save the issue (and see saved data). but I don't know if there were backend impacts.

karlcow: yes it's what he is referring to. 
What I'm saying is that we can use it. it's working as shown above.

cch5ng: @karlcow the main difference that I can see is the resulting `for="..."` value. 

screen shot 2017-06-09 at 10 47 53 am

my best guess is that the label is hand coded in the html to be able to hard code the `for` value. 
the `for="problem_type"` seems to be referenced by `problem_type` in form.py. I do not know exactly how it gets used end to end but seems to be significant and I felt it safer to preserve the existing way.

ex https://github.com/webcompat/webcompat.com/blob/master/webcompat/form.py#L205

cch5ng: @miketaylr do you know why for radio buttons the top most label has been defined in the html whereas for other fields, the label is just defined from the python?

ex https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/home-page/form.html#L20

vs https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/home-page/form.html#L36

is there some significance? thanks.

webcompat/templates/home-page/form.html
cch5ng: @karlcow in the latest pr version, I made all requested changes except this one. I wanted to get Mike's reasoning why the original code used label for radio buttons differently. if too much time passes without any response, I will change to your suggestion though.

cch5ng:
@miketaylr R? I think you were just going to look at bugform.js but if you have time to skim form.html and form.py (just the parts you authored), that would be great.

webcompat/form.py
```diff
@@ -91,11 +100,11 @@ def get_form(ua_header):
...
 
-def get_problem(category):
+def get_radio_button_label(field_value, label_list):
     '''Return human-readable label for problem choices form value.'''
-    for choice in problem_choices:
-        if choice[0] == category:
-            return choice[1]
+    for label in label_list:
+        if label[0] == field_value:
+            return label[1]
```
cch5ng: @karlcow thanks

cch5ng:
@miketaylr please hold on the review. I have a couple additional changes thanks

webcompat/templates/home-page/form.html
cch5ng: @karlcow Mon I will make this change (for both radio button field sets). my reluctance is that the old way seemed to be working ok. I kind of assume when a developer does something weird or breaks the pattern, it is a workaround for some unwanted behavior. but I am only guessing at best what was the original developer's intent.

please let me know afterward if you see issues from the backend side though. thanks

karlcow: @miketaylr pingy ping.

cch5ng: @karlcow now I remember, I broke the pattern (and forgot). while running the related functional tests, changing the label to use only the python defined label breaks this test: reporting-non-auth.js > Problem type validation

https://github.com/webcompat/webcompat.com/blob/master/tests/functional/reporting-non-auth.js#L161

(side note, I don't understand why the validation is designed to work this way but that is a different topic; I am trying to preserve existing logic)

the dependent JS logic is in bugform.js, https://github.com/webcompat/webcompat.com/pull/1578/files#diff-ce241059c28a13e03bcd0829da8d4eb7R355

instead of using the old css selector (which broke because of the field reordering), it seemed simpler to use a different selector using the `for` attribute as a marker for placing the field validation error message.

so in the latest pr, the problem type radio buttons use the hard-coded html label. but the latter (tested another browser) radio buttons use the python defined label. there is a merge conflict note I have to resolve before another review.

cch5ng:
@karlcow R?

webcompat/form.py
cch5ng: (no reply needed) not sure if expected to provide a rationale but chose option 1 because it seemed more accessible for all levels of future contributors. think that options 2 and 3 are better for efficiency, however since the length of radio button options is relatively short, the first option should not be negative.

cch5ng:
@karlcow when working on assigned issues am I expected to try to clean up existing code (as you did)? so far I have been trying to make minimal changes necessary to fix bugs. thanks

karlcow:
@cch5ng I will look at it tomorrow! :) Sorry for the delay.

webcompat/templates/home-page/form.html
karlcow: @cch5ng understood. 

Ah simple. It's because two different names are being used in the logic in the code `problem_category` and `problem_type` if the same name is used everywhere. There should be no issue. 

As an experiment I did this in form.html

```
                {{ form.problem_category.label(class_='wc-Form-label') }}
                <label class="wc-Form-label" for="problem_type">Type of bug <span class="wc-Form-required">*</span></label>
```

with the following in form.py

```python
problem_category_label = (u'Type of bug'
                          u'<span class="wc-Form-required">*</span>')

# …

    problem_category = RadioField(problem_category_label,
                                  [InputRequired(message=radio_message)],
                                  choices=problem_choices)

```

which resulted in the html as:

```html
<div class="wc-Form-information">
                <label class="wc-Form-label" for="problem_category">Type of bug<span class="wc-Form-required">*</span></label>
                <label class="wc-Form-label" for="problem_type">Type of bug <span class="wc-Form-required">*</span></label>
                
              </div>
```

The only difference here in between the hard coded value and the programmatically generated is indeed the `for` attribute value, but this comes from the chosen variable name in form.py. This can be changed. 

So either we switch everything to `problem_category` or everything to `problem_type` but mixing both seems not a good idea as we noticed. :)

Searching for `problem type`.

```
→ git rg problem_type
CHANGELOG.md
124:* Add support for a problem_type param to preselect type in form [Pull #1362](https://github.com/webcompat/webcompat.com/pull/1362)

tests/functional/reporting-non-auth.js
293:      "problem_type param selects problem type": function() {
296:          url("/issues/new?problem_type=video_bug"),

webcompat/form.py
218:        'problem_type': get_radio_button_label(
229:**Problem type**: {problem_type}

webcompat/static/js/lib/bugform.js
20:    problem_type: {
61:  this.problemType = this.inputs.problem_type.el;
175:    // If we have a problem_type param, and it matches the value, select it for
178:    var problemType = location.href.match(/problem_type=([^&]*)/);
222:      this.makeInvalid("problem_type");
224:      this.makeValid("problem_type");
354:      case "problem_type":
392:      this.inputs["problem_type"].valid &&

webcompat/templates/home-page/form.html
36:                <label class="wc-Form-label" for="problem_type">Type of bug <span class="wc-Form-required">*</span></label>
```

Searching for `problem_category`

```
→ git rg problem_category
tests/functional/reporting-non-auth.js
132:            .findByCssSelector("#problem_category-0")
272:            .findByCssSelector("#problem_category-0")

tests/test_urls.py
128:        data = {'problem_category': u'mobile_site_bug',

webcompat/form.py
83:    problem_category = RadioField([InputRequired(message=radio_message)],
205:    problem_summary = get_problem_summary(form_object.get('problem_category'))
219:            form_object.get('problem_category'), problem_choices),

webcompat/static/js/lib/bugform.js
21:      el: $("[name=problem_category]"),
221:    if (!$("[name=problem_category]:checked").length) {

webcompat/templates/home-page/form.html
34:            <fieldset class="{% if form.problem_category.errors %}is-error{% endif %}">
37:                {% if form.problem_category.errors %}{% for error in form.problem_category.errors %}
41:                {% for subradio in form.problem_category %}

webcompat/views.py
172:        must_parameters = set(['url', 'problem_category', 'description',
```

I think we have a hard dependency on `problem_type` because of the `type-media` people sending us a URI with `problem_type` in parameters.

So I proposed we switch to from `problem_category` to `problem_type`. @miketaylr opinion?

cch5ng:
@karlcow I was reflecting on the required field validation error messages and think there is a bit of redundancy between form.py and bugform.js. examples:

https://github.com/webcompat/webcompat.com/blob/master/webcompat/form.py#L44

https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L18

I do not know what happens for each required field case but there are at least a couple cases where it seems the validation error gets displayed from the js code and not the python.

side note: for these types of errors:
https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/home-page/form.html#L38

should I assume they are backend validation errors and not related to required field validation?

cch5ng:
@karlcow re
> As an experiment I did this in form.html

                {{ form.problem_category.label(class_='wc-Form-label') }}
                <label class="wc-Form-label" for="problem_type">Type of bug <span class="wc-Form-required">*</span></label>

yes, I also tried to use the python label for product_category attribute. so your point is exactly right about the sticking point of where to use product_type vs product_category. thanks for the info. I will wait I guess for Mike's comment.

cch5ng:
@miketaylr there might be a little bit of pending changes to the js for https://github.com/webcompat/webcompat.com/pull/1578#discussion_r122099902

but otherwise I think the bugform.js could be reviewed?

miketaylr:
OK! sorry to be blocking everyone here. Looking at this now...

tests/functional/reporting-non-auth.js
```diff
@@ -73,88 +73,36 @@ define(
...
-                  ":after"
-                )
-                .getPropertyValue("background-image");
-            })
-            .then(function(bgImage) {
-              assert.notInclude(
-                bgImage,
-                "valid.svg",
-                "The valid checkbox psuedo is not visible"
+            // using xpath as workaround for timeout issue when error clears
```
miketaylr: nit: this comment should probably be removed, since it isn't using findbyxpath (or whatever the method is called)

tests/functional/reporting-non-auth.js
```diff
@@ -190,6 +138,92 @@ define(
...
+            url("/issues/new"),
+            ".wc-ReportForm-actions-button"
+          )
+            .findByCssSelector("#description")
+            .click()
+            .end()
+            .findByCssSelector("#steps_reproduce")
+            .click()
+            .end()
+            // .execute(function() {
```
miketaylr: If we don't need this, can we remove it, rather than comment out?

tests/functional/reporting-non-auth.js
```diff
@@ -190,6 +138,92 @@ define(
...
+            //   var elm = document.querySelector("#steps_reproduce");
+            //   WindowHelpers.sendEvent(elm, "input");
+            // })
+            .sleep(800)
+            // using xpath because 2 errors display here
```
miketaylr: nit: this comment mentions xpath, but we're using css selectors.

tests/functional/reporting-non-auth.js
```diff
@@ -190,6 +138,92 @@ define(
...
+            .execute(function() {
+              return window
+                .getComputedStyle(document.getElementById("span-os"), ":after")
+                .getPropertyValue("background-image");
+            })
+            .then(function(bgImage) {
+              assert.include(
+                bgImage,
+                "valid.svg",
+                "The valid checkbox psuedo is visible"
```
miketaylr: nit: s/psuedo/pseudo/

tests/functional/reporting-non-auth.js
```diff
@@ -190,6 +138,92 @@ define(
...
+            .execute(function() {
+              return window
+                .getComputedStyle(document.getElementById("span-os"), ":after")
+                .getPropertyValue("background-image");
+            })
+            .then(function(bgImage) {
+              assert.notInclude(
+                bgImage,
+                "valid.svg",
+                "The valid checkbox psuedo is not visible"
```
miketaylr: nit: s/psuedo/pseudo/

(I realize you inherited this typo, but might as well fix it while we're here. :))

webcompat/form.py
```diff
@@ -34,11 +34,15 @@
 
 problem_choices = [
     (u'detection_bug', u'Desktop site instead of mobile site'),
-    (u'mobile_site_bug', u'Mobile site is not usable'),
-    (u'video_bug', u'Video doesn\'t play'),
-    (u'layout_bug', u'Layout is messed up'),
-    (u'text_bug', u'Text is not visible'),
-    (u'unknown_bug', u'Something else - I\'ll add details below')
+    (u'mobile_site_bug', u'Site is not usable'),
```
miketaylr: I don't want to bikeshed `problem_choices` naming, but "Site is not usable" will get used for desktop sites as well. I'm not sure it matters much though?

webcompat/form.py
```diff
@@ -34,11 +34,15 @@
...
+    (u'layout_bug', u'Design is broken'),
+    (u'video_bug', u'Video or audio doesn\'t play'),
```
miketaylr: 🚲 🏠: maybe we can use `media_bug` instead of `video` since this now covers audio?

(yes I realize I said I didn't want to bikeshed... oops)

webcompat/form.py
```diff
@@ -48,32 +52,38 @@
 username_message = u'A valid username must be {0} characters long'.format(
     random.randrange(0, 99))
 
-problem_label = (u'What seems to be the trouble?',
-                 '<span class="wc-Form-required">*</span>')
-url_label = u'Site URL <span class="wc-Form-required">*</span>'
+desc_label = 'Please describe what was wrong <span class="wc-Form-required">' \
+    '*</span>'
```
miketaylr: +1 to @karlcow's feedback.

webcompat/form.py
```diff
@@ -48,32 +52,38 @@
...
+desc_message = u'An issue description is required.'
 
-desc_default = u'''1. Navigate to: Site URL
-2. …
+url_label = u'Site URL <span class="wc-Form-required">*</span>'
+browser_test_label = u'Did you test in another browser?'
 
-Expected Behavior:
+steps_default = u'''1. I've tried to log in.
```
miketaylr: I feel like this might be confusing without some kind of "For example," text before this.

webcompat/form.py
```diff
@@ -202,20 +215,23 @@ def build_formdata(form_object):
...
+        'description': form_object.get('description'),
+        'steps_reproduce': form_object.get('steps_reproduce')
     }
 
     # Preparing the body
     body = u'''{metadata}
 **URL**: {url}
+**Problem type**: {problem_type}
+**Description**: {description}
+**Steps to Reproduce** {steps_reproduce}
```
miketaylr: I know it doesn't follow the format of the form, but I feel like STR shoul come last. It could contain a large image or content, which would mean you have to scroll down/up to see browser version, etc.

To be clear, I don't think we should change the flow of the reporting form, just the order once the report is made.

@karlcow @denschub @zoepage  thoughts?

webcompat/static/js/lib/bugform.js
```diff
@@ -159,10 +180,11 @@ function BugForm() {
       $("[value=" + problemType[1] + "]").click();
     }
 
-    // If we got a details param, add that to the end of the issue description.
+    // If we got a details param, add that to the end of the steps to reproduce field
+    // github.com/webcompat/webcompat.com/issues/1547#issuecomment-305649762
```
miketaylr: I don't understand what the point of this comment is (specifically, the link to the GitHub issue comment). Can we be more descriptive? Or remove it if it's unintentional.

webcompat/static/js/lib/bugform.js
```diff
@@ -46,14 +61,20 @@ function BugForm() {
   this.problemType = this.inputs.problem_type.el;
   this.uploadField = this.inputs.image.el;
   this.urlField = this.inputs.url.el;
-  this.descField = $("#description");
+  this.descField = this.inputs.description.el;
+  this.browserTestField = this.inputs.browser_test_type.el;
+  this.stepsReproduceField = this.inputs.steps_reproduce.el;
```
miketaylr: Can we go ahead and rename this to `stepsToReproduceField`. Seems less awkward.

webcompat/templates/home-page/form.html
miketaylr: Yes, we need to keep `problem_type` around forever. Apart from that, no preference from me.

miketaylr: > I kind of assume when a developer does something weird or breaks the pattern, it is a workaround for some unwanted behavior. but I am only guessing at best what was the original developer's intent.

I'm not sure what this refers to (or if I need to at this point :p). What was the weird pattern?

webcompat/templates/home-page/form.html
```diff
@@ -29,44 +45,55 @@ <h3 class="wc-Title--s">Report Site Issue</h3>
...
-              {{ form.description.label(class_='wc-Form-label') }} {{ form.description(class_='wc-Form-item wc-Form-item--textarea js-autoExpand', rows=7) }}
-            </div>
-        </div>
-        <div class="r-Grid-cell r-Grid-cell--alignBottom r-minS--1of2">
+
+          <div class="wc-Form-group js-Form-group">
+            <fieldset class="{% if form.browser_test.errors %}is-error{% endif %}">
+              <div class="wc-Form-information">
+                {{ form.browser_test.label(class_='wc-Form-label') }}
+                <!--<label class="wc-Form-label" for="browser_test_type">Did you test in another browser? </label>-->
```
miketaylr: Is this intentionally commented out?

webcompat/templates/home-page/form.html
```diff
@@ -29,44 +45,55 @@ <h3 class="wc-Title--s">Report Site Issue</h3>
...
+          <div class="wc-Form-group js-Form-group">
+            {{ form.steps_reproduce.label(class_='wc-Form-label') }} {{ form.steps_reproduce(class_='wc-Form-item wc-Form-item--textarea js-autoExpand', rows=4) }}
+          </div>
+        </div>
+        <div class="r-Grid-cell r-minM--1of2">
+
+          <div class="wc-Form-group wc-Form-group--biggerOffset js-Form-group {% if form.url.errors %}is-error{% endif %}">
             {{ form.browser.label(class_='wc-Form-label') }}
-            <span class="wc-Form-validation">
+            <span class="wc-Form-validation" id="span-browser">
```
miketaylr: What's the purpose of the `#span-browser` and `#span-os` ids?

webcompat/templates/home-page/form.html
```diff
@@ -76,7 +103,7 @@ <h3 class="wc-Title--s">Report Site Issue</h3>
                   <div class="wc-UploadForm-icon" aria-hidden="true">
                     <span class="wc-Icon wc-Icon--cloud-upload"></span>
                   </div>
-                  <label class="wc-UploadForm-label">Attach a screenshot image</label>
+                  <label class="wc-UploadForm-label">Screenshot upload</label>
```
miketaylr: Upload screenshot, maybe? Sounds more natural.

miketaylr:
@cch5ng I assume you've looked at test failures? The image upload stuff, for example, seems to all be busted. (Could just be that tests need updating)

webcompat/form.py
miketaylr: Actually, nevermind. If we want to do this, let's do it in a follow up bug. Let's ship first then see what it feels like. :)

zoepage: @miketaylr as long as it is helpful for the work on the bugs, I'd +1 for changing the flow in the report!

webcompat/form.py
cch5ng: @miketaylr ok, I will make this update:
(u'mobile_site_bug', u'Site is not usable'), =>
(u'site_bug', u'Site is not usable'),

webcompat/form.py
cch5ng: @miketaylr ok, I will make this update:
(u'video_bug', u'Video or audio doesn\'t play'), =>
(u'media_bug', u'Video or audio doesn\'t play'),

webcompat/static/js/lib/bugform.js
cch5ng: @miketaylr replying here to your comment below:
>I kind of assume when a developer does something weird or breaks the pattern, it is a workaround for some unwanted behavior. but I am only guessing at best what was the original developer's intent.

>I'm not sure what this refers to (or if I need to at this point :p). What was the weird pattern?

I think you already responded to the conversation. we were debating why that outer label for problem_type seemed to be defined in form.py (and not used) and then hard-coded in form.html. and whether that was significant to preserve.

webcompat/form.py
cch5ng: @miketaylr @zoepage I am adding the "For example," text to the placeholder.

webcompat/static/js/lib/bugform.js
miketaylr: Ah, yeah. I don't remember. As long as it does what we need it to do, I have no preference. :)

webcompat/templates/home-page/form.html
cch5ng: @miketaylr one of the functional tests was repeatedly failing for me. It appeared to be this selector

https://github.com/webcompat/webcompat.com/blob/master/tests/functional/reporting-non-auth.js#L118

I added the id to simplify the selector. I could remove the id for the browser input.

webcompat/form.py
cch5ng: @miketaylr if I change that input value to 'media_bug', then one of the current functional tests breaks

https://github.com/webcompat/webcompat.com/blob/master/tests/functional/reporting-non-auth.js#L258

is this going to cause some domino effect where the media type reporting tool will also need to be changed? ex this is the resulting url when I use the media type reporting tool for new issue:

https://webcompat.com/issues/new?url=http%3A%2F%2Fsafeschool.mon.bg%2FVideos%2F201705210928251574.mp4%3FAspxAutoDetectCookieSupport%3D1&label=type-media&problem_type=video_bug&src=media-decode-error&details=Technical+Information%3A%0AError+Code%3A+NS_ERROR_DOM_MEDIA_METADATA_ERR+%280x806e0006%29%0AResource%3A+http%3A%2F%2Fsafeschool.mon.bg%2FVideos%2F201705210928251574.mp4

cch5ng:
@miketaylr 
> I assume you've looked at test failures? The image upload stuff, for example, seems to all be busted. (Could just be that tests need updating)

sorry, I missed that. outdated selectors were fixed. tests are currently running on travis. I think I made about 98% of requested changes (except where you had an open question and I made very recent comments)

webcompat/templates/home-page/form.html
miketaylr: @cch5ng ah, makes sense. (but I didn't see any other usage of `span-browser` in this PR...)

We have a convention to use `js-thing-for-whatever` classes as general hooks for JS or tests. For example,

https://github.com/webcompat/webcompat.com/blob/master/tests/functional/comments-auth.js#L34
https://github.com/webcompat/webcompat.com/blob/master/tests/functional/comments-auth.js#L51
etc.

Can you create a `js-form-browser-wrapper` (terrible name, pick something that makes sense :p) and use that instead of an ID?

webcompat/form.py
miketaylr: Oh... oops. >_< You're right:

http://searchfox.org/mozilla-central/source/browser/base/content/browser-media.js#350

We have a couple of options here... 

1. Write a patch against Gecko and update to `media_bug` (and update the tests as well), and ensure we support both `video_bug` as a legacy alias to`media_bug`. We probably wouldn't have to maintain the alias for very long -- this is Nightly-only code and that population updates pretty quickly.

2. Just use `video_bug` and add a comment saying we use `video_bug` for legacy reasons, even though it might be used for audio bugs.

3. Do 2, but file a bug to come back and do 1 after the form rewrite lands.

3 is probably my preference. I can help you write your first Firefox patches if that's something that interests you.

webcompat/form.py
cch5ng: @zoepage @miketaylr the current placeholder format for the description details text area is breaking one functional test and seems to render one existing function obsolete.

https://github.com/webcompat/webcompat.com/blob/master/tests/functional/index-non-auth.js#L32

https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L503

is the copyURL() function now obsolete and I should remove it?

webcompat/form.py
miketaylr: Cool, we can continue discussion in #1607.

webcompat/form.py
cch5ng: I think in order to keep the existing copyURL() intact and keep the existing functional test working, it would make most sense to revert the first line of the description detail (text area) placeholder back to:

'1. Navigate to: ' ... otherwise copyURL() breaks

miketaylr: I think it's fine to keep the existing behavior. But we've also got a URL: field in the created report, so we wouldn't lose information if you got rid of the copyURL functionality (and test). I trust whatever you think is best.

cch5ng:
@miketaylr I think the latest push includes your changes and fixes broken tests.

I was having trouble with selectors for a new required field validation test (I added) and removed that test.

screen shot 2017-06-21 at 9 10 40 am

there's a new required field validation error for the description text input field. I wanted the new test to verify if you click the field and click away, that error should display. I could get the error selector to pass locally but would always fail on travis. in that flow, it is the second error that comes up and I had trouble selecting the 2nd error.

one version of the test

      "Description validation": function() {
        return (
          FunctionalHelpers.openPage(
            this,
            url("/issues/new"),
            ".wc-ReportForm-actions-button"
          )
            .findByCssSelector("#description")
            .click()
            .end()
            .click()
            .end()
            // .execute(function() {
            //   var elm = document.querySelector("#steps_reproduce");
            //   WindowHelpers.sendEvent(elm, "input");
            // })
            .sleep(800)
            // using xpath because 2 errors display here
            .findByCssSelector(
              "div.is-error:nth-child(3) > div:nth-child(1) > span:nth-child(2)"
            )
            .getVisibleText()
            .then(function(text) {
              assert.include(
                text,
                "Description required.",
                "Description validation message is shown"
              );
            })
            .end()
            .findByCssSelector("#description")
            .type("bug description")
            .end()
            .waitForDeletedByCssSelector(
              "div.is-error:nth-child(3) > div:nth-child(1) > span:nth-child(2)"
            )
            .end()
        );
      },

cch5ng:
@miketaylr just chatted with @zoepage  and she wanted me to revert the text area placeholder to her original text. I'll re-add the 'For example,' too. it will take me some time to remove the references to copyUrl() and the related test. I'll message again afterwards. thanks.

cch5ng:
@miketaylr the latest changes were pushed. the travis tests all passed (surprise!)

I should follow up with one intern test (noted above), let me know if you prefer that now or later. if I'm having difficulty getting a css selector to work, do you think using xpath is a good alternative?

miketaylr:
> @miketaylr the latest changes were pushed. the travis tests all passed (surprise!)

Awesome!

> I should follow up with one intern test (noted above), let me know if you prefer that now or later. if I'm having difficulty getting a css selector to work, do you think using xpath is a good alternative?

xpath works, sometimes I'll cheat and use devtools to create a selector for me. There's a copy CSS selector and CSS path option from the inspector. Chrome also has a copy xpath option:

screen shot 2017-06-21 at 5 05 09 pm

If that doesn't work, and if it's something you can follow up with reasonably quickly (say in the next week, or before the end of SF) you might just file a follow up bug and stick it in the SF milestone.

cch5ng:
@miketaylr I had tried to get the css selector from firebug inspector in the past and for some reason the selector errors on travis only. just got the same result for xpath on travis. I've opened another issue for that test, #1609

cch5ng:
@miketaylr this should be ready for another check (travis tests running now). I will follow up with the missing test in another issue. thanks.

cch5ng:
@miketaylr I think I found the fix for the missing test selector. I'll just add the test back into this pr. sorry about going back and forth.

miketaylr:
> I think I found the fix for the missing test selector. I'll just add the test back into this pr. sorry about going back and forth.

No worries, I hadn't started yet. :)

cch5ng:
@miketaylr hmmm, the test is failing on travis although I'm pretty sure the selector is right this time. do the intern tests run in a concurrent way at all? this particular test has some timing dependencies because of triggering error messages. what if I tried to move the test file to the top of the list of tests? or maybe if I put this test at the top of its own file (?)

the test passes locally. I might have to pull it out again.

miketaylr:
Timing can be pretty flaky on Travis... :/ Another option is to pull it out, put the body of the test in an issue and @brizental can investigate (since the focus of her Outreachy is making tests suck less).

cch5ng:
@miketaylr this latest version has the latest test mentioned removed

the test was documented in #1609 with the latest source. @brizental I'm stumped with why the test only fails on travis. if you have suggestions or if you want to take on the bug, let me know what you think is best.

miketaylr:
OK! will review this in the morning (or later tonight). Nice work!

miketaylr:
@cch5ng this looks good! i have one small piece of feedback, at mobile viewports, the placeholder text for the bug description will be cut off. 

but i'll file a follow up bug for that -- i don't think it should block merge.

screen shot 2017-06-27 at 1 50 02 pm

miketaylr:
(filed https://github.com/webcompat/webcompat.com/issues/1617)

@karlcow @zoepage did you have any pending feedback? Looks good to merge from my POV. Nice work @cch5ng!

zoepage:
Very small changes @cch5ng:

This bug appears and could be tackled too https://github.com/webcompat/webcompat.com/issues/1611 so we could deploy soon. :)
https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/css/development/components/form.css#L95 the value needs to be changed from `.75` to `0.0.75`

https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/css/development/components/form.css#L84 please add `margin-bottom: 5px` so the inputs don't look like "glued together".

https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/css/development/components/form.css#L118 `min-height:250px` should be `min-height: 125px` so the text area is closing with the submit buttons. 

Then we totally can merge. Great work Carol! :)
Thanks!

cch5ng:
planning to merge changes for #1611 into this issue per latest change request

zoepage:
@cch5ng just add `Issue #1661 - fixing textshadow value` to the commit and it will be connected :)

cch5ng:
@zoepage I've made your change requests but if the text area becomes too short, then right around the breakpoint (806px ish), there is a weird alignment of the bottom of the text area and the buttons (because the button labels are a bit long). thanks.

screen shot 2017-06-27 at 3 27 38 pm

zoepage:
@cch5ng let's create a follow up issue for that. We'll refactor the design of the form soon anyway. :)
Thanks!

karlcow:
let's move on. :) no issue. Thanks @cch5ng

miketaylr:
OK, the only thing left to do is to fixup the commit message before merge. I can help with that if you'd like @cch5ng. Grab me if so.

cch5ng:
@miketaylr 
> the only thing left to do is to fixup the commit message before merge

the top commit message is fixed now. travis tests running currently.

miketaylr:
Awesome. Thanks for all your work here @cch5ng. :)

Issue #1569, opened 2017-05-21
Functional tests fail starting with a certain test


magsout:
The tests fail on all branches. It would appear that the tests fail from: `Going back in history updates issue list and URL state`

```
× firefox on any platform - Issue-list - Going back in history updates issue list and URL state
```

Fx ANY:     [✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓×××××✓✓✓✓×××✓\      ] 65/78, 15 fail

zoepage:
@miketaylr @brizental 
Would one of you be able to take a look here? Thanks!

miketaylr:
I don't think this reproduces for me. @brizental does it for you?

brizental:
I just changed from Ubuntu to MacOS an this has started happening to me too... I'll have a look, now that I can reproduce it.

miketaylr:
whoops, this reproduces for me now (using v54 stable):

<img width="751" alt="screen shot 2017-06-15 at 4 24 59 pm" src="https://user-images.githubusercontent.com/67283/27185989-56128090-51e7-11e7-9c8b-915e702b7f93.png">

miketaylr:
Failing test starts here: https://github.com/webcompat/webcompat.com/blob/master/tests/functional/issue-list-non-auth.js#L290

brizental:
The problem is that the `goBack()` function (https://github.com/webcompat/webcompat.com/blob/master/tests/functional/issue-list-non-auth.js#L320) should return a Promise but it never returns anything.

I tried to look and see if the problem is in our code, but I didn't find anything wrong, so I think that this might be a problem in Leadfoot, it only happens for the newest Firefox versions, like 54 and forward. I  also tried to mess around with Leadfoot's code, but didn't have success.

Apparently everything is working fine, the URL changes right as it should and everything, it just gets stuck in some Promise limbo and never gets to the next line (https://github.com/webcompat/webcompat.com/blob/master/tests/functional/issue-list-non-auth.js#L321). Also if you manually refresh the page when it gets stuck it gets going again.

I'll keep on investigating, but just wanted to see if anybody has any ideia of what might be happening :)

miketaylr:
> so I think that this might be a problem in Leadfoot, it only happens for the newest Firefox versions, like 54 and forward.

It's also possible it's a Geckodriver bug or regression: https://github.com/mozilla/geckodriver

PR #1588, opened 2017-06-09
Fixes #1582 - Gets mock login to work

### LINKED ISSUE #1582: Mock Github login for testing
I have been working on #980 and Github requests have been really slowing the process down. Not only because it's a request to a remote server, but also because Github requires re-authorization from time to time.

We have a work around (https://github.com/webcompat/webcompat.com/blob/master/tests/functional/lib/helpers.js#L88) for when Github asks the user for re-authorization, but that does not seem to be working :/

brizental:
So, I got mock login to work. I don't know if this is the best way, but it works. Let me know what you guys think.

The login is working, but the user activity page is not, it gets a "Bad credentials" error. I'll work on that tomorrow. I think it should be something close to what we already do with the issues fixtures.

Also, I noticed that even though now the login doesn't require remote server requests, there are some other tests that still do. Maybe it would be best to change this issue's name to "Get all tests to work offline" or something. What do you think?

Anyways, r? @miketaylr


brizental:
I just saw that along with this commit there is a package-lock.json file. I never saw that, but from a quick Google search I saw that it should be ignored... Should we add it to .gitignore? I can open a new issue for that.

package-lock.json
```diff
@@ -0,0 +1,5268 @@
+{
+  "name": "webcompat",
```
miketaylr: Is this `package-lock.json` file doing anything? Seems unneeded?

webcompat/views.py
```diff
@@ -73,10 +73,15 @@ def format_date(datestring):
 @app.route('/login')
 def login():
     if session.get('user_id', None) is None:
-        # manually set the referer so we know where to come back to
-        # when we return from GitHub
-        set_referer(request)
-        return github.authorize('public_repo')
+        if app.config['TESTING']:
+            session['username'] = 'testuser'
+            session['avatar_url'] = '/img/avatar.png?'
```
miketaylr: is the `?` intentional here?

miketaylr: Also, can we put this image in the `/tests/fixtures` folder? It feels a bit odd to expose this image to the web. 

If you use the `test-files` route (in views.py) `http://localhost:5000/test-files/fixtures/avatar.png` should work.

miketaylr:
> I just saw that along with this commit there is a package-lock.json file. I never saw that, but from a quick Google search I saw that it should be ignored... Should we add it to .gitignore? I can open a new issue for that.

oops, I should have read this question before commenting as much in review. Yeah, let's remove it and add a commit to .gitignore it.

miketaylr:
> Maybe it would be best to change this issue's name to "Get all tests to work offline" or something. What do you think?

Let's open a new issue for that, and list which tests require network access. 

> The login is working, but the user activity page is not, it gets a "Bad credentials" error. I'll work on that tomorrow. I think it should be something close to what we already do with the issues fixtures.

Yeah, there should be a way to detect that we're in test mode, and just return some JSON fixture file that has the right content.

webcompat/views.py
brizental: Yeah, it was intentional, but anyways it's bet to put it in the fixtures folder I'll change that.

I put it there because for some reason there was a string being added to the end of the URL and it was returning a 404.

miketaylr: Ah OK, makes sense. I'm guessing the github package prepends a GET param (and expects to be able to do something useful with it). Seems fine to leave in.

brizental:
r? @miketaylr 

Let me know if you find anything else I need to change. I tested and everything seems to be working fine, but since I started using the notebook that Mozilla sent I have been getting the #1569 error, so I wasn't able to run all tests...

brizental:
r? @miketaylr 

Let me know if you find anything else I need to change. I tested and everything seems to be working fine, but since I started using the notebook that Mozilla sent I have been getting the #1569 error, so I wasn't able to run all tests...

miketaylr:
@brizental can you rebase and push your branch back up before we merge this in?

`git pull --rebase upstream master` (assuming you have an upstream remote pointing to webcompat repo)
`git push -f`

miketaylr:
ok, looks good! i rebased everything down to a single commit. let's talk about rebasing commits in our next 1:1. it's possible to do some cool stuff. :)

(but sometimes it can get crazy....)

webcompat/views.py
```diff
@@ -73,10 +73,15 @@ def format_date(datestring):
...
-        set_referer(request)
-        return github.authorize('public_repo')
+        if app.config['TESTING']:
+            session['username'] = 'testuser'
+            session['avatar_url'] = '/test-files/fixtures/avatar.png?'
+            return authorized()
+        else:
+            # manually set the referer so we know where to come back to
+            # when we return from GitHub
+            set_referer(request)
```
miketaylr: I wonder if we'll want to do this for the mocked case, but let's merge and see if anything blows up. :)

PR #1591, opened 2017-06-12
Fix #1579 - Alexa Top Site to DB

### LINKED ISSUE #1579: Implement a python script to dump alexa top sites to db
This is a part of #1533 , I decide to cache alexa top sites in a sqlite db, not query when new issue opened.
Assume the ranking doesn't change that frequently, we manually launch the script to create a new db, then deploy to production server.
The plan is to have a top site table with 4 columns: url (primary key), priority (from 1 to 3), country_code and ranking. We use url and priority for label. Country_code and ranking are used to left a comment when a new issue created, let contributor know it's important because it's top N in which country.
I use the alexa top slite to query:

1. global top 10000 sites
2. tier 1 top 1000 sites (11 regions, total 11000 sites)

Priority is defined in https://github.com/webcompat/webcompat.com/issues/1533#issuecomment-299429187 . If the url doesn't exist, then add a record. If the url exists but priority is higher (like global 9000 but top 52 in Taiwan), update the priority.

MDTsai:
Implement a python script which could dump alexa top site to local DB
4 columns: url(primary key), priority, country_code, ranking

r? @zoepage could you help to arrange reviewer?


zoepage:
@MDTsai I'd say, @karlcow would be the perfect person here?

topsites.py
```diff
@@ -0,0 +1,166 @@
...
+class TopSites:
+    def __init__(self, access_key, secret_key):
+        self.access_key = access_key
+        self.secret_key = secret_key
+
+    def query(self, begin, begin_priority):
+        query_string = self.build_query(begin, 100)
+        toSign = "GET\n" + ATS_SERVICE_HOST + "\n/\n" + query_string
+        signature = self.gen_sign(toSign, self.secret_key)
+        uri = ATS_AWS_BASE_URL + query_string + "&" + urlencode({"Signature": signature}, "UTF-8")
```
karlcow: Use the same string format style along the program. So as in the project we used `format()`

This could be rewritten as 

```python
        uri = '{base}{query}&{signature}'.format(
            base=ATS_AWS_BASE_URL,
            query=query_string,
            signature=urlencode({"Signature": signature}, "utf-8"))
```

topsites.py
```diff
@@ -0,0 +1,166 @@
...
+
+        response = requests.get(uri)
+
+        if response.status_code == requests.codes.ok:
+            # See http://docs.aws.amazon.com/AlexaTopSites/latest/index.html?QUERY_QueryRequests.html
```
karlcow: This probably should be 

```python
            # See http://docs.aws.amazon.com/AlexaTopSites/latest/index.html?QUERY_QueryRequests.html  # nopep8
```

so the style linter does not complain about it.

topsites.py
```diff
@@ -0,0 +1,166 @@
...
+            dom = parseString(response.content)
+            sites = dom.getElementsByTagName("aws:Site")
+            for site in sites:
+                url = self.get_text(site.getElementsByTagName("aws:DataUrl")[0].childNodes)
+                rank = int(self.get_text(site.getElementsByTagName("aws:Rank")[0].childNodes))
```
karlcow: style nit: lines too long. It should be

```python
            for site in sites:
                url = self.get_text(
                    site.getElementsByTagName("aws:DataUrl")[0].childNodes)
                rank = int(self.get_text(
                    site.getElementsByTagName("aws:Rank")[0].childNodes))
```

topsites.py
```diff
@@ -0,0 +1,166 @@
...
+                    session.add(site_row)
+
+            session.commit()
+
+    # Build query string for alexa top site
+    # with start ranking and count
+    def build_query(self, start, count):
+        now = time.time()
+        mlsec = repr(now).split('.')[1][:3]
+        timestamp = datetime.datetime.utcnow().strftime(ATS_DATEFORMAT).format(mlsec)
```
karlcow: style nit: line too long.

```python
        timestamp = datetime.datetime.utcnow().strftime(
            ATS_DATEFORMAT).format(mlsec)
```

Though I would probably separate in two lines myself. Too many chained functions are not very pleasant on the eye.

```python
        now_date = datetime.datetime.utcnow()
        timestamp = now_date.strftime(ATS_DATEFORMAT).format(mlsec)
```

topsites.py
```diff
@@ -0,0 +1,166 @@
...
+ATS_RESPONSE_GROUP_NAME = "Country"
+ATS_SERVICE_HOST = "ats.amazonaws.com"
+ATS_AWS_BASE_URL = "http://" + ATS_SERVICE_HOST + "/?"
+
+ATS_DATEFORMAT = "%Y-%m-%dT%H:%M:%S.{}Z"
+ATS_HASH_ALGORITHM = "HmacSHA256"
+
+# Regions/Countries to dump to topsites.db
+regions = {
+    "": (10000, 1),  # Global
```
MDTsai: Empty string represents global in alexa top site API, so I put it here. Replace with "world" means we need to replace when use.

topsites.py
```diff
@@ -0,0 +1,166 @@
...
+    "IN": (1000, 2),
+    "DE": (1000, 2),
+    "TW": (1000, 2),
+    "ID": (1000, 2),
+    "HK": (1000, 2),
+    "SG": (1000, 2),
+    "PL": (1000, 2),
+    "GB": (1000, 2),
+    "RU": (1000, 2),
+}
```
MDTsai: I store country_code, count and begin priority here, so just what we need now. If the way gives priority become more complex, maybe it's not a good to use dictionary here.

topsites.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __init__(self, url, priority, country_code, ranking):
+        self.url = url
+        self.priority = priority
+        self.country_code = country_code
+        self.ranking = ranking
+
+    def __repr__(self):
+        return "<Site('%d','%d' '%s', '%s')>" % (self.priority, self.ranking, self.url, self.country_code)
+
```
MDTsai: I'll remove __repr__ since it's just for my development and debugging, not used in real world now.

MDTsai:
Thanks @karlcow, I use ''Python PEP8 Autoformat'' in my sublime now. Seems there is no PEP8 warning now. I re-push style fixed.

topsites.py
karlcow: Understood. From the point of view of the python program, this is a dictionary. And the first part is called the key. Having an empty key doesn't seem to be a good choice. I think it's better to special case it in the call. with something like 

if regions['world']:
    parameter = ""

topsites.py
```diff
@@ -0,0 +1,180 @@
...
+            "CountryCode": self.country_code,
+            "Count": count,
+            "Start": start,
+            "SignatureVersion": "2",
+            "SignatureMethod": ATS_HASH_ALGORITHM
+        }
+
+        # alexa top site only accept request with ordered query parameters
+        return urlencode(collections.OrderedDict(sorted(query_params.items())))
+
```
karlcow: Here I'm trying to understand.
```python
>>> query_params = {
...     "Action": ATS_ACTION_NAME,
...     "ResponseGroup": ATS_RESPONSE_GROUP_NAME,
...     "AWSAccessKeyId": "blah",
...     "Timestamp": "2017-06-14T05:50:09.480Z",
...     "CountryCode": "US",
...     "Count": 100,
...     "Start": 234,
...     "SignatureVersion": "2",
...     "SignatureMethod": ATS_HASH_ALGORITHM
... }
>>> 
>>> # This will give you a list of tuples.
>>> query_params.items()
[('Count', 100), ('Start', 234), ('SignatureVersion', '2'), ('AWSAccessKeyId', 'blah'), ('CountryCode', 'US'), ('Action', 'TopSites'), ('Timestamp', '2017-06-14T05:50:09.480Z'), ('SignatureMethod', 'HmacSHA256'), ('ResponseGroup', 'Country')]
>>> # This is sorting the list of tuple.
>>> sorted(query_params.items())
[('AWSAccessKeyId', 'blah'), ('Action', 'TopSites'), ('Count', 100), ('CountryCode', 'US'), ('ResponseGroup', 'Country'), ('SignatureMethod', 'HmacSHA256'), ('SignatureVersion', '2'), ('Start', 234), ('Timestamp', '2017-06-14T05:50:09.480Z')]
>>> import collections
>>> # This converts it to an ordered dictionary
>>> collections.OrderedDict(sorted(query_params.items()))
OrderedDict([('AWSAccessKeyId', 'blah'), ('Action', 'TopSites'), ('Count', 100), ('CountryCode', 'US'), ('ResponseGroup', 'Country'), ('SignatureMethod', 'HmacSHA256'), ('SignatureVersion', '2'), ('Start', 234), ('Timestamp', '2017-06-14T05:50:09.480Z')])
>>> from urllib import urlencode
>>> # This encodes it as a parameter URL.
>>> urlencode(collections.OrderedDict(sorted(query_params.items())))
'AWSAccessKeyId=blah&Action=TopSites&Count=100&CountryCode=US&ResponseGroup=Country&SignatureMethod=HmacSHA256&SignatureVersion=2&Start=234&Timestamp=2017-06-14T05%3A50%3A09.480Z'
>>> # But this would work too.
>>> urlencode(sorted(query_params.items()))
'AWSAccessKeyId=blah&Action=TopSites&Count=100&CountryCode=US&ResponseGroup=Country&SignatureMethod=HmacSHA256&SignatureVersion=2&Start=234&Timestamp=2017-06-14T05%3A50%3A09.480Z'
```

But why not make it even simpler. You are not using the dictionary for its benefit, aka deferencing with keys a value. We could use simply a list of tuples. List in python are ordered objects.

```python
>>> access_key = 'foo'
>>> count = 100
>>> country_code = 'US'
>>> start = '333'
>>> start = 333
>>> timestamp = '2017-06-14T05:50:09.480Z'
>>> query_params = [
...     ("Action", ATS_ACTION_NAME),
...     ("AWSAccessKeyId", access_key),
...     ("Count", count),
...     ("CountryCode", country_code),
...     ("ResponseGroup", ATS_RESPONSE_GROUP_NAME),
...     ("SignatureMethod", ATS_HASH_ALGORITHM),
...     ("SignatureVersion", "2"),
...     ("Start", start),
...     ("Timestamp", timestamp)]
>>> urlencode(query_params)
'Action=TopSites&AWSAccessKeyId=foo&Count=100&CountryCode=US&ResponseGroup=Country&SignatureMethod=HmacSHA256&SignatureVersion=2&Start=333&Timestamp=2017-06-14T05%3A50%3A09.480Z'
```

The full function could become.
```python
    def build_query(self, start, count):
        ATS_DATEFORMAT = '%Y-%m-%dT%H:%M:%S.%f'
        now_time = datetime.datetime.utcnow()
        timestamp = '{}Z'.format(nowtime.strftime('ATS_DATEFORMAT')[:-3])
        # Alexa top site only accept request with ordered query parameters
        # Keep the order!
        query_params = [
            ("Action", ATS_ACTION_NAME),
            ("AWSAccessKeyId", access_key),
            ("Count", count),
            ("CountryCode", country_code),
            ("ResponseGroup", ATS_RESPONSE_GROUP_NAME),
            ("SignatureMethod", ATS_HASH_ALGORITHM),
            ("SignatureVersion", "2"),
            ("Start", start),
            ("Timestamp", timestamp)]
        return urlencode(query_params)
```

topsites.py
```diff
@@ -0,0 +1,180 @@
...
+    def gen_sign(self, data, secret_key):
+        '''Computes RFC 2104-compliant HMAC signature'''
+        dig = hmac.new(secret_key, data, hashlib.sha256).digest()
+        return base64.b64encode(dig)
+
+    def get_text(self, node_list):
```
karlcow: This is not necessary. See above.

topsites.py
```diff
@@ -0,0 +1,180 @@
...
+
+    access_key = sys.argv[1]
+    secret_key = sys.argv[2]
+
+    site = TopSites(access_key, secret_key)
+
+    for country_code in REGIONS:
+        site.country_code = country_code
+        end_ranking = REGIONS[country_code][0]
+        begin_priority = REGIONS[country_code][1]
```
karlcow:
```python
end_ranking, begin_priority = REGIONS[country_code]
```
will work by assigning the two values of the tuple directly to each of the variable. One line instead of two without using indexes.

topsites.py
```diff
@@ -0,0 +1,180 @@
...
+
+engine = create_engine("sqlite:///" + os.path.join(
+    app.config["BASE_DIR"], "topsites.db"))
+Base = declarative_base()
+Session = sessionmaker(bind=engine)
+session = Session()
+
+
+class Site(Base):
+    '''Define a sqlalchemy base object for a alexa top site.'''
```
MDTsai: I think SQLAchemy base object is necessary. Even if it looks like do nothing, but SQLAchemy helps to update and commit value to DB. I just need to update the variables.
For the TopSites class, yes, maybe it could be move to functions instead.

karlcow: Ah yes! indeed for the initialization of SQLAlchemy. My bad. Thanks.

topsites.py
```diff
@@ -0,0 +1,180 @@
...
+                url = self.get_text(
+                    site.getElementsByTagName("aws:DataUrl")[0].childNodes)
+                rank = int(self.get_text(
+                    site.getElementsByTagName("aws:Rank")[0].childNodes))
+
+                priority = begin_priority
+                if rank > 100 and rank <= 1000:
+                    priority = priority + 1
+                elif rank > 1000 and rank <= 10000:
+                    priority = priority + 2
```
MDTsai: Basic idea is we have priority 1, 2 and 3. With the site info,
global top 1-100 go to P1, 101-1000 to P2 and 1001-10000 to P2
tier 1 top 1-100 to P2, 101-1000 to P3.
So I assign a initial priority to global and tier 1 (1 and 2 in REGIONS), and update priority in the same range.
If we need to provide tier 2 countries, I can extend REGIONS with initial priority 3.

topsites.py
```diff
@@ -0,0 +1,180 @@
...
+
+                if url in topsites:
+                    site_row = topsites[url]
+                    # Update priority is new is lower
+                    if site_row.priority > priority:
+                        site_row.priority = priority
+                        site_row.country_code = self.country_code
```
MDTsai: It's a question from Sergiu, if we have a site is low priority in global, but high priority in tier 1 country, finally the priority should be higher.
I put url and Site object in topsites dict. If url already exists but with higher priority (from P3 to P2), then update the priority and country code, we will know why it's important to us.

topsites.py
```diff
@@ -0,0 +1,180 @@
...
+                    session.add(site_row)
+
+            session.commit()
+
+    def build_query(self, start, count):
+        '''Build query string for request with start ranking and count'''
+        now = time.time()
+        mlsec = repr(now).split('.')[1][:3]
+        nowtime = datetime.datetime.utcnow()
+        timestamp = nowtime.strftime(ATS_DATEFORMAT).format(mlsec)
```
MDTsai: The sample code and sample query from amazon gives miliseconds. But seems work without that.
```
&Timestamp=2011-05-06T17%3A58%3A49.463Z
```

topsites.py
MDTsai: List is a better way for this +1

topsites.py
karlcow: I'm still confused. 

So you take the Alexa list for `world`, make a first pass and assign priorities following this scheme:

* `priority-critical` (1) if url is in rank 1 to rank 100
* `priority-important` (2) if url is in rank 101 to rank 1000
* `priority-normal` (3) if url is in rank 1001 to rank 10000

Then you go onto the countries list (aka `REGIONS`)

* `priority-important` (2) if url is in rank 1 to rank 100 AND 
* `priority-normal` (3) if url is in rank 101 to rank 1000

With this extra conditions:

* if `url` is classified as `priority-critical` (== 1) in global (already in db) and if in `country_code` it is `priority-important` (==2 aka rank <= 100), it will stay as `priority-critical`.
* if `url` is classified as `priority-normal` (== 3) in global (already in db) and if in `country_code` it is `priority-important` (==2 aka rank <= 100), it will become  `priority-important`.

Is that it?

karlcow:
@MDTsai I think we would benefit in flexibility if indeed we were dropping `class TopSites:` and reorganizing the code into small functions. The `def query():` should be probably break down into smaller units. 

That would also make it easier for you to create tests for it. Currently there are no unit tests at all. 
The unit tests are to be put in `tests/test_topsites.py`

Also would you mind create a `tools/` directory? Or do you want me to do that?
I think `topsites.py` needs to be there along `labels.py`

topsites.py
karlcow: yup and because we drop the milliseconds. Your code becomes a lot easier in one shot.

```python

    ATS_DATEFORMAT = '%Y-%m-%dT%H:%M:%SZ'

    def build_query(self, start, count):
        timestamp = datetime.datetime.utcnow().strftime('ATS_DATEFORMAT')
        # Alexa top site only accept request with ordered query parameters
        # Keep the order!
        query_params = [
            ("Action", ATS_ACTION_NAME),
            ("AWSAccessKeyId", access_key),
            ("Count", count),
            ("CountryCode", country_code),
            ("ResponseGroup", ATS_RESPONSE_GROUP_NAME),
            ("SignatureMethod", ATS_HASH_ALGORITHM),
            ("SignatureVersion", "2"),
            ("Start", start),
            ("Timestamp", timestamp)]
        return urlencode(query_params)
```

topsites.py
MDTsai: So I keep REGIONS, if it's "GLOBAL" as special case, I give different country_code and priority.

topsites.py
MDTsai: Yes, this is what I want to do.

MDTsai:
OK, I will move the code to tools, drop class TopSites and write test case :) Could be done tomorrow.

miketaylr:
@MDTsai could you go ahead and make a commit that moves this file to the tools directory? https://github.com/webcompat/webcompat.com/blob/master/labels.py (since I guess it's relevant to a tools directory being created)?

MDTsai:
r? = @karlcow Drop TopSite class to functions, add unittest. Not sure if I need to move labels.py in this PR, but now it's included.

karlcow:
@MDTsai thanks. I will make a pass at it. Tomorrow.

tools/topsites.py
```diff
@@ -0,0 +1,182 @@
...
+Base.metadata.create_all(engine)
+
+
+def query_topsites(country_code, count=1000):
+    """Query top sites with given country code and count."""
+    for index in range(1, count, 100):
+        uri = build_uri(country_code, index)
+        response = requests.get(uri)
+
+        if response.status_code == 200:
```
karlcow: You get a `if` which is only testing success, but not the other cases. 
What should happen if there's something different than `response.status_code` 

* put the URI on retry queue?
* send a warning so that we know it's incomplete
* Something else?

Cf https://github.com/webcompat/webcompat.com/pull/1591#discussion_r121859159

tools/topsites.py
```diff
@@ -0,0 +1,182 @@
...
+    url = node_text(site, 'aws:DataUrl')
+    rank = int(node_text(site, 'aws:Rank'))
+
+    # Global website, initial priority is 1
+    # Tier 1 regions, initial priority is 2
+    priority = 1 if country_code == "" else 2
+
+    # Ranking between 100 and 1000, add priority 1
+    # Ranking between 1000 and 10000, add priority 2
+    if rank > 100 and rank <= 1000:
```
karlcow: I missed that in the first review. But python is so nice that it allows you to do this in one shot.

```python
>>> rank = 150
>>> if 100 < rank <= 1000:
...     print 'DOUBLE RAINBOW'
... 
DOUBLE RAINBOW
```

same for the next one.

tools/topsites.py
```diff
@@ -0,0 +1,182 @@
...
+        priority = priority + 1
```
karlcow: You can also use this syntax:

```python
>>> priority = 1
>>> priority += 1
>>> priority
2
```

tools/topsites.py
```diff
@@ -0,0 +1,182 @@
...
+        base=ATS_AWS_BASE_URL,
+        query=query_string,
+        signature=urlencode({"Signature": signature}, "UTF-8"))
+
+    return uri
+
+
+def build_query_string(country_code, start_ranking, timestamp=None):
+    """Build query string for request with start ranking and count."""
+    if None == timestamp:
```
karlcow: This is the opposite.

```python
if timestamp == None:
```

but more to the point. Are there anytime where you do not need timestamp to build the query string. 
If it has always to be there. drop it from the parameter and just call the timestamp.

```python
def build_query_string(country_code, start_ranking):
    """Build query string for request with start ranking and count."""
    nowtime = datetime.datetime.utcnow()
    timestamp = '{}Z'.format(nowtime.strftime('ATS_DATEFORMAT')[:-3])
    # Alexa top site only accept request with ordered query parameters
    # … etc.
```

tools/topsites.py
MDTsai: Normally only authentication error message is available: http://docs.aws.amazon.com/AlexaTopSites/latest/QUERY_QueryAuthenticationErrors.html
I think show the error message in console will be nice.

tools/topsites.py
MDTsai: Add timestamp in parameter make it possible to run the unit test with given timestamp. If no timestamp, I may remove the test_build_query_string test case.

karlcow: You will need to install mock into the project. It is available in python 3.5 by default, but not in python 2.7

This is a way to solve the issue with your test.

```python
import datetime
import urllib
def build_query_string(foo):
    """Build query string for request with start ranking and count."""
    nowtime = datetime.datetime.utcnow()
    timestamp = '{}Z'.format(nowtime.strftime('%Y-%m-%dT%H:%M:%S'))
    query_params = [
        ("Action", 'blah'),
        ("Start", foo),
        ("Timestamp", timestamp)]
    return urllib.urlencode(query_params)


import unittest
import datetime
import mock

class TestTopsites(unittest.TestCase):

    def test_build_query_string(self):
    testdt = datetime.datetime(2020, 01, 02, 01, 02, 03, 0)
    with mock.patch('datetime.datetime') as dt_mock:
        dt_mock.utcnow.return_value = testdt
        actual = build_query_string('hello')
        expected = "Action=blah&Start=hello&Timestamp=2020-01-02T01%3A02%3A03Z"
```

Or on the CLI with the IDE.

```python
>>> import mock
>>> import datetime
>>> import urllib
>>> def build_query_string(foo):
...     """Build query string for request with start ranking and count."""
...     nowtime = datetime.datetime.utcnow()
...     timestamp = '{}Z'.format(nowtime.strftime('%Y-%m-%dT%H:%M:%S'))
...     query_params = [
...         ("Action", 'blah'),
...         ("Start", foo),
...         ("Timestamp", timestamp)]
...     return urllib.urlencode(query_params)
... 
>>> testdt = datetime.datetime(2020, 01, 02, 01, 02, 03, 0)
>>> with mock.patch('datetime.datetime') as dt_mock:
...     dt_mock.utcnow.return_value = testdt
...     actual = build_query_string('hello')
...     expected = "Action=blah&Start=hello&Timestamp=2020-01-02T01%3A02%3A03Z"
... 
>>> expected
'Action=blah&Start=hello&Timestamp=2020-01-02T01%3A02%3A03Z'
>>> actual
'Action=blah&Start=hello&Timestamp=2020-01-02T01%3A02%3A03Z'
```

tools/topsites.py
```diff
@@ -0,0 +1,182 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import base64
+import datetime
+import hashlib
+import hmac
+import os
+import requests
```
karlcow: requests is not part of python lib. It goes to the second block of import.

MDTsai: fixed

MDTsai:
r? = @karlcow Use Mock to replace datetime.datetime. Print error code and message from ATS result XML now. Change variable quote to single quote.

tests/test_topsites.py
```diff
@@ -0,0 +1,52 @@
...
+import datetime
+import os
+import sys
+import unittest
+from xml.dom.minidom import parseString
+
+from mock import patch
+
+# Add tools module to import path
+sys.path.append(os.path.realpath(os.pardir))
```
karlcow: the comment and the `sys.path…` are not necessary here.
And it means that you can get rid of `import os` and `import sys`

tests/test_topsites.py
```diff
@@ -0,0 +1,52 @@
...
+        with patch('datetime.datetime') as dt_mock:
+            dt_mock.utcnow.return_value = testdt
+            self.assertEqual(topsites.build_query_string('CN', 1),
+                             TEST_QUERY_STRING)
+
+    def test_node_text(self):
+        site = self.dom.getElementsByTagName('aws:Site')[0]
+        self.assertEqual(topsites.node_text(site, 'aws:DataUrl'), 'baidu.com')
+        self.assertEqual(topsites.node_text(site, 'aws:Rank'), '1')
+
```
karlcow: two blank lines here instead of one.

tools/topsites.py
```diff
@@ -0,0 +1,196 @@
...
+    country_code = Column(String)
+    ranking = Column(Integer)
+
+    def __init__(self, url, priority, country_code, ranking):
+        self.url = url
+        self.priority = priority
+        self.country_code = country_code
+        self.ranking = ranking
+
+Base.metadata.create_all(engine)
```
karlcow: two blank lines after a Class definition.

```python 
        self.ranking = ranking


Base.metadata.create_all(engine)
```

tools/topsites.py
```diff
@@ -0,0 +1,196 @@
...
+                code = node_text(dom, 'Code')
+                message = node_text(dom, 'Message')
+                print('Send request to {} get error: {}.\nMessage: {}'.format(
+                      uri, code, message))
+        except ConnectionError:
+            print('Unable send request to {}'.format(uri))
+
+
+
+def parse_site(site):
```
karlcow: two blank lines before a function definition, not 3.

tools/topsites.py
```diff
@@ -0,0 +1,196 @@
...
+    to_sign = 'GET\n{}\n/\n{}'.format(ATS_SERVICE_HOST, query_string)
+    signature = gen_sign(to_sign)
+
+    # URI with signature
+    uri = '{base}{query}&{signature}'.format(
+        base=ATS_AWS_BASE_URL,
+        query=query_string,
+        signature=urlencode({"Signature": signature}, "UTF-8"))
+
+    return uri
```
karlcow: not a mistake but I would personally remove the blank line before `return uri`

tools/topsites.py
```diff
@@ -0,0 +1,196 @@
...
+    """Computes RFC 2104-compliant HMAC signature."""
+    dig = hmac.new(ats_secret_key, data, hashlib.sha256).digest()
+    return base64.b64encode(dig)
+
+
+def node_text(tree, tag_name):
+    """Extract the text node in a tree for a specific tag_name.
+    It will only work with unique text nodes.
+    """
+    node = tree.getElementsByTagName(tag_name)[0]
```
karlcow: We usually use the format:

```python 
def blah():
    """"Short sentence.

    Longer explanations or more details.
   """

There is a space in between the two sections.

File: tests/test_topsites.py (hunk 21 above)
Comment (MDTsai) [inline tests/test_topsites.py]: 2 blank line between test_node_text and main?

Comment (karlcow) [inline tests/test_topsites.py]: yes. :)

**Comment by MDTsai** [comment]
@karlcow: as our discussion in SF, I need to:
[ ] Provide AWS app/secret key
[ ] Modify the script to read app/secret key from env set by fabric. 
Is there any special requirement for the second one?

**Comment by karlcow** [comment]
so we have a devops special repository for the fabric file.  I'm not here for the next two weeks, so I guess @miketaylr will take the relay on this one.

File: tools/topsites.py [hunk 26]
```diff
@@ -0,0 +1,196 @@
...
+# Alexa top site access and secret key
+ats_access_key = None
+ats_secret_key = None
+
+
+# Cache parsed sites, change priority if raised
+topsites = {}
+
+engine = create_engine('sqlite:///' + os.path.join(
+    app.config['BASE_DIR'], 'topsites.db'))
```
karlcow: @MDTsai Following our discussion in San Francisco. Maybe we should rename this `topsites-new.db`. So that when we run it, we don't crush the previous one and the fabric file or a script will be in charge of rotating the DB once the db has been populated. 

Once this is done. Feel free to merge if I'm already off.

MDTsai:
@karlcow archive implemented, ready to merge

karlcow:
Thanks a lot @MDTsai

miketaylr:
> I propose we deploy this on staging first.
> Probably mike will do it given that I'm off for the next two weeks.

Ah, missed this... should we let this run on staging before merging to master? That's a bit hard now that it's on master...

miketaylr:
@MDTsai @karlcow (realizing it's not a great time to be pinging people in Asia....)

miketaylr:
Actually, nothing is calling into this code right now. Nevermind. :)

MDTsai:
@miketaylr  we can use crontab or other similar tool to run the script, is there any existing service on stage server?

miketaylr:
> @miketaylr we can use crontab or other similar tool to run the script, is there any existing service on stage server?

Yep, we currently use crontab to delete logs when they are older than 14 days.

PR #1593, opened 2017-06-13
Fixes #1571 - image dnd

### LINKED ISSUE #1571: Attaching another photo to an issue using “Drag and drop” should remove previous one
**Steps to Reproduce:**
1. Navigate to: www.webcompat.com.   
2. Tap “Report bug” button.
3. Type in the required information and attach a screenshot.
4. After screenshot is attached, drag and drop another screenshot. 
5. Observe behavior.

**Expected Behavior:**
Screenshot is attached and previous one is removed.

**Actual Behavior:**
Screenshot is displayed and previous one remains attached.

**Watchers:**
@softvision-sergiulogigan
@softvision-oana-arbuzov

sv;

cch5ng:
placeholder: this code is not ready to merge; it has only been pushed as a point of reference for conversation about how to approach the issue solution; explanation why the requested change is non trivial and breaks some existing functionality and design (which already works pretty well)

ref:
[note to self] possible tech to update drag and drop response:
* https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop
  * add a drop file handler to outermost label
* https://developer.mozilla.org/en-US/docs/Web/API/File
* https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer

@zoepage I'm sharing a partial fix for #1571. I'm unclear how to move forward because making the requested changes (upload a second image to replace existing) breaks existing functionality and design.

previously the developer designed the file input control to have 2 specific states:
* empty: no image uploaded; the user can upload via dnd or open file dialog
* filled: image is uploaded; the control is not interactive (the whole section is hidden) until the image is removed by the Remove image upload button

in the pr code, I have the fix so you can now upload a second image (replacing the first) but this change now causes the input to have mixed state (some aspects of empty state from above are true and some aspects of filled state from above are true too). I think the original design to have 2 distinct states is better.

I'm not sure how to move forward. should the original developer have input as to how to fix this? do I need to go through an approval process to redesign how the input works? I do not know who to ask questions for this to get unstuck. thanks

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -410,13 +410,13 @@ function BugForm() {
 
     // hide upload image errors (this will no-op if the user never saw one)
     $(".wc-Form-helpMessage--imageUpload").remove();
-// TODO do not understand need
-    //$(".wc-UploadForm-label").show();
+    $(".wc-UploadForm-label").show();
 
     removeBanner.removeClass("is-hidden");
     removeBanner.attr("tabIndex", "0");
     uploadIcon.addClass("is-hidden");
     uploadLabel.addClass("is-hidden");
+    uploadLabel.hide();
     //uploadWrapper.addClass("is-hidden");
     removeBanner.on(
       "click",
@@ -427,6 +427,7 @@ function BugForm() {
         removeBanner.attr("tabIndex", "-1");
         uploadIcon.removeClass("is-hidden");
         uploadLabel.removeClass("is-hidden");
+        uploadLabel.show();
         //uploadWrapper.removeClass("is-hidden");
         removeBanner.off("click");
 
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -405,14 +405,19 @@ function BugForm() {
   this.showRemoveUpload = function(label) {
     var removeBanner = $(".wc-UploadForm-button");
     var uploadWrapper = $(".wc-UploadForm-wrapper");
+    var uploadIcon = $(".wc-UploadForm-icon");
+    var uploadLabel = $(".wc-UploadForm-label");
 
     // hide upload image errors (this will no-op if the user never saw one)
     $(".wc-Form-helpMessage--imageUpload").remove();
-    $(".wc-UploadForm-label").show();
+// TODO do not understand need
+    //$(".wc-UploadForm-label").show();
 
     removeBanner.removeClass("is-hidden");
     removeBanner.attr("tabIndex", "0");
-    uploadWrapper.addClass("is-hidden");
+    uploadIcon.addClass("is-hidden");
+    uploadLabel.addClass("is-hidden");
+    //uploadWrapper.addClass("is-hidden");
     removeBanner.on(
       "click",
       _.bind(function() {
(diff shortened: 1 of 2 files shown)


zoepage:
/home/travis/build/webcompat/webcompat.com/webcompat/static/js/lib/bugform.js

  407:9  error  'uploadWrapper' is assigned a value but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)

cch5ng:
@zoepage thanks. a few follow up questions:

do you have a style preference for the flow diagram?

I am not sure where to find the current dnd logic in the code.

I think the validation logic was fixed (or will be fixed) in #1552

zoepage:
@cch5ng 
> do you have a style preference for the flow diagram?

Big fan of pen & paper. Anything else is your choice. :)

> I am not sure where to find the current dnd logic in the code.

That is okay as you will not need it right now. Let's go through everything in our call today and document the progress here.

> I think the validation logic was fixed (or will be fixed) in #1552

True. Let's step back from that for a bit and think about "how it would make the most sense for you" or what you'd expect where it would be.

cch5ng:
@zoepage I forget, did you want me to write a flow diagram for how things should work or how they work today?

I started to do the latter and think if you want the former, there is a significant amount of detail that I would not know to include. thanks

zoepage:
@cch5ng 
The first option, yes. 

> I started to do the latter and think if you want the former, there is a significant amount of detail that I would not know to include.

Which kind of detail?

cch5ng:
@zoepage this is a flow chart for 2a (starting state is the file input has an image currently). the chart describes what happens in the existing code.

1571_flow_img_filled

cch5ng:
@zoepage this is the flow diagram for 2b (file input is filled and user does drag and drop with a 2nd image)

the general concept is there but some details in the event handlers probably need to be changed. ie even though there is state change overlap between diagrams 2a and 2b, I believe maybe a custom handler should be written for 2b b/c it does not seem necessary to do something like
* enable upload icon
* hide remove image btn
* hide upload icon
* enable remove image btn

1571_flow_img_filled_dnd

miketaylr:
> I am not sure where to find the current dnd logic in the code.

There is no current DND logic in the code. We rely on the fact that browsers allow you to drop an image onto a `input[type=file]` associated `<label>` element, and it will do the right thing. The current shipping behavior is for the "drop area" to be a `<label>` area sized to cover 100% of the thing. Hope that helps.

webcompat/static/js/lib/bugform.js
```diff
@@ -491,8 +499,13 @@ function BugForm() {
       img_url,
       ")"
     ].join("");
+    var delimeter = "[![Screenshot Description](";
     this.descField.val(function(idx, value) {
-      return value + "\n\n" + imageURL;
+      if (value.includes(delimeter)) {
```
miketaylr: Just FYI `String.prototype.includes` is a super new JS feature -- I wouldn't recommend using it. You could either use `indexOf` or lodash (since we include that....): https://lodash.com/docs/3.10.1#includes

cch5ng: @miketaylr will do. just to let you know that this PR is not ready to review. I only pushed the branch to have a point of reference to discuss with Ola about how to begin to approach the problem.

miketaylr: (yeah, was just going through all my github notifications... not really reviewing. 😄 )

cch5ng:
closing this to work on a slightly updated branch (based off current master)

Issue #1565, opened 2017-05-18
Update Github queries to filter out type-media bugs


adamopenweb:
In the webcompat.com meeting on 5/18/2017 there was a discussion about filtering out type-media bugs from appearing on webcompat.com.

Recently a contributor was triaging a type-media bug that we normally don't focus on. We think we can prevent this from happening again by updating the Github queries to filter out type-media bugs.

Using `-label:type-media` seems to work on Github.

@karlcow @miketaylr do you see any issues with doing this?

This is related to #1356.

karlcow:
https://wiki.mozilla.org/Compatibility/Meetings/2017-05-18#type-media_.28miketaylr.29

```
type-media (miketaylr)

Where are we at, and what should our next steps be? Hide from UI?

    Adam: Karl has probably some feedback here. I'd think, we should hide it as contributors already jumped on them and wanted to triage them. Would it be easy to hide it?
    Tom: It should be when we hide them through webcompat.com.
    Adam: Is everyone okay with that?
    Tom: I think so.
    Adam: I don't think its the perfect solution, but a step forward.
    Tom: Probably also a lot of dupes. Shall we close all of them?
    Adam: The media team will probably want to take a look but yes.
    Guillaume: Is it possible to hide them from GitHub?
    Adam: https://github.com/webcompat/web-bugs/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20-label%3Atype-media%20label%3Astatus-needstriage%20
    Tom: We just add a minus sign in front of the label and it should be filtered out.
    Guillaume: I think, its easy to hide that label. --> Adam will file a follow up bug.
```

karlcow:
@wisniewskit About duplicates. See #1551 

@magsout we can't hide them from GitHub apart of yourself filtering them out doing `-label:type-media`

For the UI, we can indeed exclude them from the webcompat.com UI. That means we have to not forget about handling them on the github side. What I usually do

1. Go to needstriage and filter all issues related to `type-media`
2. Find the first one
3. Close the rest as duplicate of the first one. Repeat and Rinse.

Sometimes it takes one hour to do that. Sometimes less.

karlcow:
The ultimate question is when the media team will start handling those.

karlcow:
For example

* web.whatsapp.com https://github.com/webcompat/web-bugs/issues/5659
* (music|radio).yandex.com https://github.com/webcompat/web-bugs/issues/5628

[...]

PR #1599, opened 2017-06-19
Fixes #1505 remove unused query-params.js

### LINKED ISSUE #1505: Remove query-params.js file
This code has never been actually used. So lets remove it.

(FTR, it has some great ideas in it, possibly something we want to revisit during a refactor).

https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/models/query-params.js

cch5ng:
r? @miketaylr


cch5ng:
@miketaylr I'm also currently assigned to #1565 and am still searching through the code to understand what part of the code to update (change the api request parameters or filter the api response). but I was wondering do you think I should keep any of this removed file for filtering out media-type labelled issues from the webcompat.com issues list?

I did not look closely into the contents of query-params.js for this pr but can look again.

cch5ng:
@miketaylr nvm, I think I found in github api docs info about the question I had earlier. I think it should be ok to just review this file removal.

cch5ng:
@miketaylr done

> Can you update the commit message to mention the query-params.js file

miketaylr:
Thanks!

PR #1604, opened 2017-06-20
Fixes #1603 - Updates CONTRIBUTING.md tests section for mock login.

### LINKED ISSUE #1603: Update CONTRIBUTING.md for mock login
Just remove the part about adding `testusername` and `testpassword` because they are no longer needed for functional tests since #1588

brizental:
r? @miketaylr

### DIFF
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -57,11 +57,12 @@ If you are a member of webcompat organization in GitHub, edit `config/secrets.py
 
 Change the value to : `ISSUES_REPO_URI = 'webcompat/webcompat-tests/issues'`.
 
-Start the application server:
+Start the application server in test mode:
 
 ```bash
-source env/bin/activate && python run.py
+source env/bin/activate && python run.py -t
 ```
+> We start the server in test mode to mock the communications with GitHub API servers using local fixture data. The files in `/tests/fixtures/` directory will be served as responses.
 
 In a separate terminal window or tab, run the tests:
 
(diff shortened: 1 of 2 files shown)


miketaylr:
Looks good! I think we should also just recommend running functional tests always in test mode:

So we would want to update the following command a bit before your edits:

```
source env/bin/activate && python run.py -t
```

I'm not really sure why you would run them w/o fixtures or test mode.... wondering if we should document that it's possible. @karlcow?

karlcow:
@miketaylr 
> I'm not really sure why you would run them w/o fixtures or test mode.... wondering if we should document that it's possible. @karlcow?

We should definitely document any requirements. 
if we look at the [current doc](https://github.com/webcompat/webcompat.com/blob/master/docs/tests.md), it says:

> * Running Tests
>
>    * Functional Tests
>    * Functional Tests using Fixture Data

Is there a reason why we ask to run the tests without the fixture data in the first part?

> Start the application server:
>     `source env/bin/activate && python run.py`
> In a separate terminal window or tab, run the tests:
>     `node_modules/.bin/intern-runner config=tests/intern`

I even wonder if `node_modules/.bin/intern-runner config=tests/intern` could simply becomes:

`grunt functests` which will start the server with the appropriate config.

karlcow:
> Intern also comes with Grunt tasks so it can be quickly added to existing Grunt-based workflows, and is designed to work out-of-the-box with many popular continuous integration services like Jenkins and Travis CI.
— https://theintern.github.io/intern/#what-is-intern

> Grunt support is built into Intern. Install Intern and load the Grunt task into your Gruntfile using grunt.loadNpmTasks('intern'):
— https://theintern.github.io/intern/#grunt

Issue #1482, opened 2017-04-05
Fix missing docstrings on the python files.


karlcow:
On #1473 there was a discussion about the project having missing docstrings. We didn't create docstrings for all the functions.

The output on view was. 

```
`pep257 webcompat/views.py`
`webcompat/views.py:1 at module level:
        D100: Missing docstring in public module
webcompat/views.py:40 in public function `shutdown_session`:
        D103: Missing docstring in public function
webcompat/views.py:45 in public function `before_request`:
        D103: Missing docstring in public function
webcompat/views.py:54 in public function `after_request`:
        D103: Missing docstring in public function
webcompat/views.py:62 in public function `token_getter`:
        D103: Missing docstring in public function
webcompat/views.py:69 in public function `format_date`:
        D300: Use """triple double quotes""" (found '''-quotes)
webcompat/views.py:76 in public function `login`:
        D103: Missing docstring in public function
webcompat/views.py:87 in public function `logout`:
        D103: Missing docstring in public function
webcompat/views.py:97 in public function `authorized`:
        D103: Missing docstring in public function
webcompat/views.py:117 in public function `file_issue`:
        D300: Use """triple double quotes""" (found '''-quotes)
webcompat/views.py:127 in public function `index`:
        D300: Use """triple double quotes""" (found '''-quotes)
webcompat/views.py:150 in public function `show_issues`:
        D300: Use """triple double quotes""" (found '''-quotes)
webcompat/views.py:159 in public function `create_issue`:
        D401: First line should be in imperative mood ('Create', not 'Creates')
webcompat/views.py:208 in public function `show_issue`:
        D300: Use """triple double quotes""" (found '''-quotes)
webcompat/views.py:219 in public function `me_redirect`:
        D401: First line should be in imperative mood ('Thi', not 'This')
webcompat/views.py:219 in public function `me_redirect`:
        D300: Use """triple double quotes""" (found '''-quotes)
webcompat/views.py:228 in public function `show_user_page`:
        D400: First line should end with a period (not ':')
webcompat/views.py:228 in public function `show_user_page`:
        D300: Use """triple double quotes""" (found '''-quotes)
webcompat/views.py:250 in public function `show_rate_limit`:
        D103: Missing docstring in public function
webcompat/views.py:268 in public function `download_file`:
        D300: Use """triple double quotes""" (found '''-quotes)
webcompat/views.py:278 in public function `get_test_helper`:
[...]

PR #1610, opened 2017-06-22
Fixes #1587 - Adds tests for all form.py functions

### LINKED ISSUE #1587: add python unit tests for form.py
some existing functions in form.py are missing unit tests.

should affect https://github.com/webcompat/webcompat.com/blob/master/tests/test_form.py

karlcow:
- testing all the functions we have in form.py
- fixes the syntax of comments related to the discussion in #1473 and #1482
- fixes URL normalization function with empty URL. The tests were breaking if url was None. So we fixed that weakness that would result in a 500. It should not happen upstream, but better play safe than say sorry.

first review: r? @cch5ng
then second review: r? @miketaylr 

@cch5ng reviews are a good way to learn more and you have worked a lot on this code. If things are breaking because of your recent modifications. I'll fix the tests and eventual merge conflicts.

Thanks.

### DIFF
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -10,7 +10,7 @@ from webcompat import form
 class TestForm(unittest.TestCase):
 
     def test_normalize_url(self):
-        '''Checks that URL is normalized.'''
+        """Checks that URL is normalized."""
         r = form.normalize_url('http://example.com')
         self.assertEqual(r, 'http://example.com')
 
@@ -36,7 +36,7 @@ class TestForm(unittest.TestCase):
         self.assertEqual(r, 'http://example.com')
 
     def test_domain_name(self):
-        '''Checks that domain name is extracted.'''
+        """Checks that domain name is extracted."""
         r = form.domain_name('http://example.com')
         self.assertEqual(r, 'example.com')
 
@@ -44,7 +44,7 @@ class TestForm(unittest.TestCase):
         self.assertEqual(r, 'example.com')
 
     def test_metadata_wrapping(self):
-        '''Checks that metadata is processed and wrapped.'''
+        """Checks that metadata is processed and wrapped."""
         TEST_DICT = {'cool': 'dude', 'wow': 'ok'}
         EXPECTED_SINGLE = '<!-- @cool: dude -->\n'
         EXPECTED_MULTIPLE = '<!-- @cool: dude -->\n<!-- @wow: ok -->\n'
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -4,10 +4,21 @@
 '''Tests for form validation.'''
 
 import unittest
+
+import webcompat
 from webcompat import form
+
+FIREFOX_UA = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:48.0) Gecko/20100101 Firefox/48.0'  # nopep8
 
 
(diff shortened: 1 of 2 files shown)


tests/test_form.py
```diff
@@ -54,3 +71,49 @@ def test_metadata_wrapping(self):
...
+    def test_get_metadata(self):
+        """HTML comments need the right values depending on the keys."""
+        metadata_keys = ('sky', 'earth')
+        form_object = {'blah': 'goo', 'hello': 'moshi', 'sky': 'blue'}
+        actual = form.get_metadata(metadata_keys, form_object)
+        expected = u'<!-- @sky: blue -->\n<!-- @earth: None -->\n'
+        self.assertEqual(actual, expected)
+
+    def test_build_formdata(self):
+        """The data body sent to GitHup API."""
```
miketaylr: nit: s/GitHup/GitHub/

miketaylr:
> first review: r? @cch5ng

Uh, oops. >_< Please review as well @cch5ng, I review as well as I read, probably. :)

tests/test_form.py
```diff
@@ -54,3 +71,49 @@ def test_metadata_wrapping(self):
 
         r = form.get_metadata(('cool', 'wow'), TEST_DICT)
```
cch5ng: @karlcow maybe this line and the next are a little redundant with the test below for test_get_metadata( )? ~line 103

webcompat/form.py
```diff
@@ -216,7 +223,7 @@ def build_formdata(form_object):
 **Steps to Reproduce**
 {description}
 
-'''.format(**formdata)
+""".format(**formdata)
```
cch5ng: @karlcow ? I had to look up format but still am not clear. is this converting the formdata dictionary into a string? (pardon me if I ever mix js and python terms b/c it has been a while since using python daily). my high level gist of the build_formdata() function is that this is where you are building the string that stores the bug form field values.

on a side note. is it considered unimportant whether a user would want to edit the issue field values in the future? I may have missed something but it seemed like you could only edit an existing issue by adding comments(?)

cch5ng:
@karlcow general ? ... do the images ever get removed from wherever they are stored?

webcompat/form.py
karlcow: @cch5ng For now we are only planning to allow some people to edit the summary of an issue. 

----

@cch5ng I'm not sure if you are asking about `str.format()` or `**formdata` syntax. Let's assume it's about [`**kwargs`](https://docs.python.org/dev/tutorial/controlflow.html#unpacking-argument-lists)

`**` is an operator to unpack keyword arguments. An example:

```python
>>> # a dictionary of data
>>> formdata = {'state': 'California', 'city': 'Los Angeles'}
>>> # no keys in the string
>>> 'No luck'.format(**formdata)
'No luck'
>>> # Some parameters in the string matching the keys.
>>> '{city} is a city in {state}'.format(**formdata)
'Los Angeles is a city in California'
```

karlcow:
@cch5ng Never. Except if we erase them by hand. It has happened in the past. Privacy, legal issues.

miketaylr:
I'm gonna fix conflicts and merge this. Thanks @karlcow!

miketaylr:
Will merge in https://github.com/webcompat/webcompat.com/pull/1644.

PR #1614, opened 2017-06-22
Fixes #1613. Check form validation in image upload success handler.

### LINKED ISSUE #1613: Do not call enableSubmits() in success handler of image upload
If you do, you make it possible to file bad requests... for example if you didn't select a problem type, we'll end up hitting a 400 due to not having that key on the form object (which feels like another kind of bug...).

miketaylr:
Blindly enabling them is bad. >_<

r? @cch5ng (this shouldn't create conflicts for you, AFAICT!)

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -454,7 +454,7 @@ function BugForm() {
       success: _.bind(function(response) {
         this.addImageURL(response);
         this.uploadLoader.removeClass("is-active");
-        this.enableSubmits();
+        this.checkForm();
       }, this),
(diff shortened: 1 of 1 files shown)


miketaylr:
Just verified manually by deploying this branch to staging and changing `extensions.webcompat-reporter.newIssueEndpoint` (from about:config in Dev Editino) to `https://staging.webcompat.com/issues/new`.

miketaylr:
Oh indeed, it's a dupe. Why don't we merge and deploy this now, to prevent 400s, and you can further improve it if you think it's needed after that. I'd like to prevent more busted bug reports.

PR #1616, opened 2017-06-24
image dnd change [not ready to merge]

cch5ng:
not ready to merge, the branch is used to discuss approaches to solve #1571
(also a continuation of thoughts from #1593); branch version increased b/c it is based on a more recent master...

(update 062517) high level paths to approach #1571 
* plan A: use html drag and drop api to put drop event listener on the label wrapper (around the image input + input label); hopefully get the file info and be able to replace an existing input image file
  * plan A first choice route:
    * use combo of html drag and drop + file reader api
    * for 2nd image dnd event, want to get the new file uri (there is an existing function for this) => (update 062817, unfortunately it appears impossible to both accept the 2nd drag/drop action and get the 2nd file uri)
    * for 2nd image dnd event, want to prevent replacing the browser url with the new image file url/uri
    * want to replace the existing image with the newly dropped image programmatically (ie not via the normal drag/drop response)
  * plan A, second choice route (local branch issues/1571/3): there is a quick partial fix available that disables dragging a 2nd image altogether; it doesn't replace the first image but it also doesn't replace the app view with just a new image url in the browser
* plan B: support 2nd image drag and drop (prefer to avoid this because requires breaking a lot of existing functionality/design)

@zoepage in this latest version, there is a proof of concept for partial fix. html dnd api could be used to prevent the behavior where the second image drag and drop event replaces the entire site. instead, the 2nd image dnd will do nothing (the first image will remain in place). in this scenario the user would need to remove the first image using the remove button and then could drag another image over.

note that preventDefault() had to be applied to both target label's drop and dragover events in order for this partial fix to work. but on initial view html dnd is still not simple for the requested behavior to replace an existing image with 2nd image dnd (it is still hard to get the dropped file info but I haven't looked much into referenced file api too).

coverage for modern desktop browsers looks pretty good. mobile browsers not so much (but I assume that is ok?)

to enable the behavior of replacing an existing image with 2nd image dnd event requires more work and breaks existing design/functionality to do so.

this code requires refactoring and I left in commented console statements (commented due to the linter errors).

reference:
https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop
http://caniuse.com/dragndrop/embed/eras=-4,2

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -47,6 +47,7 @@ function BugForm() {
   this.uploadField = this.inputs.image.el;
   this.urlField = this.inputs.url.el;
   this.descField = $("#description");
+  this.uploadLabel = $(".wc-UploadForm");
 
   this.init = function() {
     this.checkParams();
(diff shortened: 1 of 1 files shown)


cch5ng:
closing as replaced by #1634

PR #1628, opened 2017-06-30
Fixes #1627 - Adds coco.fr to the abuse list

### LINKED ISSUE #1627: Adds coco.fr to the abuse list
We currently had a series of report for coco.fr which looks like spam. 
Let's block them in the meantime.

karlcow:
r? @cch5ng 

We had abuse on the project for the domain coco.fr.

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -186,7 +186,8 @@ def create_issue():
         abort(400)
     # see https://github.com/webcompat/webcompat.com/issues/1141
     # see https://github.com/webcompat/webcompat.com/issues/1237
-    spamlist = ['qiangpiaoruanjian', 'cityweb.de']
+    # see https://github.com/webcompat/webcompat.com/issues/1627
+    spamlist = ['qiangpiaoruanjian', 'cityweb.de', 'coco.fr']
     for spam in spamlist:
         if spam in form.get('url'):
             msg = (u'Anonymous reporting for domain {0} '


cch5ng:
@karlcow sorry for the delay, I will look at this today

karlcow:
@cch5ng it's good you took time. The storm seems to have passed. I will close the pull request and we can reopen it, in case this is coming back.

Thanks a lot.

cch5ng:
@karlcow this is the start to my questions about views.py (for /issues/new route) but eventually I'll cover the entire views.py. I might include some questions I had when trying to address a different issue later.

for https://github.com/webcompat/webcompat.com/blob/master/webcompat/views.py#L179
should set() be updated to frozenset()? I had to look up issubset() and saw a deprecation notice in the docs. https://docs.python.org/2/library/sets.html ... but am not totally clear about it

https://github.com/webcompat/webcompat.com/blob/master/webcompat/views.py#L192
very nit picky but when I tested this it appeared that any type of reporting was blocked (anonymous or logged in) for the spamlist elements. is this the expected behavior?

https://github.com/webcompat/webcompat.com/blob/master/webcompat/views.py#L196
for the resulting flash message, should there be an easier way to close it? I felt like you have to refresh the browser

high level question about spam? it seems like the spam would be generated programmatically so I'm not clear how it is effective to update the spamlist (https://github.com/webcompat/webcompat.com/blob/master/webcompat/views.py#L189). does this mean generally spam is being generated using the bug form UI?

https://github.com/webcompat/webcompat.com/blob/master/config/__init__.py#L49
(this is something I looked at for a different issue) why do you use a namedtuple here? is it just to be more descriptive?

for the homepage view (/), I'm trying to understand the flow for how the triage issues list gets generated. in the source it seems to be coming from a backbone model but I'm not clear where in the code the request to the github api gets made. could I get a pointer to which source files I should be looking at?

in the future is it OK for me to include both you and MikeT as JS reviewers? I know both of you are very busy but your review concerns are a bit different and I think I learn from both forms.

Issue #1306, opened 2017-01-27
Don't start uploading image until user clicks submit


miketaylr:
Right now we auto-upload, the motivation being your upload is done while you're filling out details.

We should probably wait until the user hits submit before uploading to a) prevent a ton of unneeded images (people just clicking the button, checking it out) and b) not upload unneeded images if you decide to change it.

karlcow:
yes. And we need to be careful on the way we deal with the URI.

miketaylr:
> And we need to be careful on the way we deal with the URI.

Can you clarify? "be careful" is crucial for many things that we do around public user data.

karlcow:
so currently.

* User selects an image
* we upload it on the server
* we create a hash for the URI in a dated space
* we send back this URI to the form to create the markup in markdown

It's indeed an issue for the cruft it creates on the server. 

If we change the way we do it by just staging in the form and sending the image to the server, we can't have anymore the URI computed on the server side and added to the form.

What we could do though, either compute the URI on the JS side (so it is visible in the markdown textarea) and send this by the form, but it could be abused I guess. Not comfortable with this solution. 

Or maybe we could show a placeholder for the image in the textarea without putting the full URI, I even wonder if we could actually put a thumbnail of the image in this textarea. Then the user presses submit it really generates the real URI on the server side and the right markdown format.
If the user doesn't, everything vanishes in between the stars 🌃

karlcow:
Another way:

1. Select image
2. Buffer the image locally
3. Request a hash URI for the image on the server
4. Use this hash URI for the markdown
5. Once the user finally pushes submit, send the `HTTP POST` for the issue to GitHub and an `HTTP PUT` for the image at the location given by the hash URI.

miketaylr:
Good thoughts. I was just thinking something simple like:

1) select image, or add-on already put image [fill out form like normal], do nothing to image
2) when user clicks upload, do the XHR POST and request the image path, stick that in the markdown and submit the form immediately. 

We already show a spinner, so there will be visual feedback while the upload is happening.

But I like your "Another way" option too.

miketaylr:
I'm afraid this might create a ton of pain for @cch5ng if I work on this before #1547 lands. Maybe we can get it done by the end of next week, or at least coordinate in-person.

cch5ng:
@miketaylr thanks. I'm also assigned to #1571, should I assume that this issue (#1306) should be completed before I start making changes on #1571?

[...]

PR #1634, opened 2017-07-04
Fixes #1571 - handle multiple image drag and drop events

### LINKED ISSUE #1571: Attaching another photo to an issue using “Drag and drop” should remove previous one

cch5ng:
fix: #1571; add handler for multiple drag and drop events; update some styling

some history in #1616 also #1593 

for the record, I'm not too crazy about this new functionality, after the first image is added to the input field, the user can do one of the following:
* drag and drop another image
* click the top to select another image
* click the bottom to remove the image

my preference would be to limit the user options to either add the image (one state) or remove the image (another state) and not mix it all up. but it is what it is...

---
update (070617) ps

fwiw this is the list of test case scenarios I came up with; some may be redundant with functional tests. could not figure out at this time how to automate tests for file drag and drop.

X-pass

1 click to add image
2 click to remove image
=> image should be removed

---

X-pass

1 click to add image

http://localhost:5000/uploads/2017/6/060c17f0-75a9-4bf9-9025-6a5bb04c0908-thumb.jpg

2 click to replace image

http://localhost:5000/uploads/2017/6/db48d0d3-837d-4dd0-8d7b-8d77c0144da5-thumb.jpg

=> image should be updated
=> text area should be updated
3 click to remove image
=> image should be removed

---

X-pass

1 drag to add image

http://localhost:5000/uploads/2017/6/7c40b638-90cf-4191-87e6-5d7fff288beb-thumb.jpg

=> text area should be updated
2 drag to replace image

http://localhost:5000/uploads/2017/6/fa0d37d9-57ca-41f0-8075-351aac4e2da6-thumb.jpg

=> file input should be updated
=> text area should be updated
3 click to remove image
=> image should be removed

img1
http://localhost:5000/uploads/2017/6/513cb759-c08a-406f-9046-155a2f0e9b28-thumb.jpg

img2
http://localhost:5000/uploads/2017/6/b884ff32-a5d9-4852-a126-3d3d86c1a23b-thumb.jpg

---

X-pass

1 click to add image

http://localhost:5000/uploads/2017/6/5c1daecb-84ac-46d0-8b22-8cdb787720a8-thumb.jpg

=> text area should be updated
2 drag to replace image

http://localhost:5000/uploads/2017/6/1e28e2d9-b446-45d2-8efd-00fc3089d409-thumb.jpg

=> file input should be updated
=> text area should be updated
3 click to remove image
=> image should be removed

img1
http://localhost:5000/uploads/2017/6/707d338d-8015-48de-8f1e-a47bfea8baec-thumb.jpg

img2
http://localhost:5000/uploads/2017/6/3290d9c8-df3b-4a40-82d9-f5c2327305d1-thumb.jpg

---

X-pass

1 drag to add image

http://localhost:5000/uploads/2017/6/c5e1098e-aaff-4e22-80b2-ab6dd093df4c-thumb.jpg

=> text area should be updated
2 click to replace image

http://localhost:5000/uploads/2017/6/35d21095-77b1-45a2-acd7-3948263fc67c-thumb.jpg

=> file input should be updated
=> text area should be updated
3 click to remove image
=> image should be removed

img1
http://localhost:5000/uploads/2017/6/7b804914-badf-43ae-a38f-6ce8cf8e945e-thumb.jpg

img2
http://localhost:5000/uploads/2017/6/788fef15-317b-43ab-8e21-01cbe934a55b-thumb.jpg

### DIFF
--- a/webcompat/static/css/development/components/upload-form.css
+++ b/webcompat/static/css/development/components/upload-form.css
@@ -39,7 +39,9 @@
 }
 
 /* state */
-.wc-UploadForm-wrapper.is-hidden {
+.wc-UploadForm-wrapper.is-hidden,
+.wc-UploadForm-icon.is-hidden,
+.wc-UploadForm-label.is-hidden {
   display: none;
 }
 
@@ -92,6 +94,11 @@
   text-decoration: underline;
   transform: translateY(0);
   transition: transform .2s linear 0s;
+
+  /* workaround for mult dnd - allows selection of image remove btn when an
+     image is displayed otherwise change for #1571 would make the default click
+     result in an upload file dialog */
+  z-index: 10;
 }
 
 /* state */
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -442,6 +442,8 @@ function BugForm() {
   this.showRemoveUpload = function(label) {
     var removeBanner = $(".wc-UploadForm-button");
     var uploadWrapper = $(".wc-UploadForm-wrapper");
+    var uploadIcon = $(".wc-UploadForm-icon");
+    var uploadLabel = $(".wc-UploadForm-label");
 
     // hide upload image errors (this will no-op if the user never saw one)
     $(".wc-Form-helpMessage--imageUpload").remove();
@@ -449,7 +451,9 @@ function BugForm() {
 
     removeBanner.removeClass("is-hidden");
     removeBanner.attr("tabIndex", "0");
-    uploadWrapper.addClass("is-hidden");
+    uploadIcon.addClass("is-hidden");
+    uploadLabel.addClass("is-hidden");
+    uploadLabel.hide();
     removeBanner.on(
       "click",
       _.bind(function() {
@@ -457,15 +461,11 @@ function BugForm() {
         label.css("background", "none");
         removeBanner.addClass("is-hidden");
         removeBanner.attr("tabIndex", "-1");
+        uploadIcon.removeClass("is-hidden");
+        uploadLabel.removeClass("is-hidden");
+        uploadLabel.show();
         uploadWrapper.removeClass("is-hidden");
         removeBanner.off("click");
-
-        // remove the last embedded image URL
-        // Note: this could fail in weird ways depending on how
-        // the user has edited the descField.
-        this.stepsToReproduceField.val(function(idx, value) {
-          return value.replace(/\[!\[[^\]]+\]\([^\)]+\)\]\([^\)]+\)$/, "");
-        });
 
         this.hasImage = false;
       }, this)


cch5ng:
@miketaylr R? (should I continue to get JS reviews from you or should I sometimes ask Karl?)

miketaylr:
> @miketaylr R? (should I continue to get JS reviews from you or should I sometimes ask Karl?)

I'm happy to review, but @zoepage and @karlcow are also good candidates (dunno if Karl will self-identify as a JS reviewer :p)

miketaylr:
> (update) checking notes, forgot there was a dependency on #1306. should I close this until then? also reminder that #1306 is now unblocked.

Nah, let's not close right now. I'll work on #1306 today/tomorrow, then let's come back to this. Hopefully I don't create too many merge conflicts. :)

cch5ng:
in another issue I've worked on, there was some regex to remove the old image text url from the text area. 

https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L460

not sure if it's better to use that or to keep the way I'm doing the url text replacement.

miketaylr:
OK, #1306 has landed. Can you re-base and push to this branch again @cch5ng? There's no conflicts, but it will make future reviews/iterations a bit nicer.

webcompat/static/css/development/components/upload-form.css
```diff
@@ -92,6 +94,9 @@
   text-decoration: underline;
   transform: translateY(0);
   transition: transform .2s linear 0s;
+
+  /* workaround for mult dnd */
```
miketaylr: What does this achieve?

webcompat/static/js/lib/bugform.js
```diff
@@ -155,6 +156,18 @@ function BugForm() {
     img.src = dataURI;
   };
 
+  // Event handler for image file drop; purpose to detect multiple drops
+  this.handleImageDrop = function() {
+    var imageFilled = false;
```
miketaylr: We probably don't need this now, since we have the `hasImage` property on `BugForm` now.

webcompat/static/js/lib/bugform.js
```diff
@@ -522,8 +542,13 @@ function BugForm() {
       img_url,
       ")"
     ].join("");
+    var delimeter = "[![Screenshot Description](";
```
miketaylr: Do we need anything more than what's on master now? 

https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L560-L562

The way things work currently:

1. user clicks submit
2. if they have an image, POST to `/upload`, get a URL back
3. stick URL in textarea, call `<form>.submit()`

If the image changes before step 1, we don't need to handle anything differently.

webcompat/static/js/lib/bugform.js
```diff
@@ -155,6 +156,18 @@ function BugForm() {
...
 
+  // Event handler for image file drop; purpose to detect multiple drops
+  this.handleImageDrop = function() {
+    var imageFilled = false;
+    if ($(".wc-UploadForm-label.is-hidden").length) {
+      imageFilled = true;
+    }
+    if (imageFilled) {
+      // if 2nd drag and drop make wrapper hidden to make remove button available
+      $(".wc-UploadForm-wrapper").addClass("is-hidden");
```
miketaylr: I don't understand how this is supposed to work, can you explain it to me? Why does adding a `is-hidden` class to this wrapper div solve the problem?

It seems like in the `drop` handler we want to just do the same as we do for the change event on the file input -- call `checkImageTypeValidity`, right (we need to make sure we send it the right data from the event, or `dataTransfer` object or something, it's been a while since I've messed with DnD....)?

(we also want to prevent the default browser action, otherwise we end up navigating to the image. try to drop a new image 3 times.... why 3 and not 2 or 1... not sure. 🤔

miketaylr:
> in another issue I've worked on, there was some regex to remove the old image text url from the text area.

https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L466-L468 is actually dead code now, I forgot to remove it in #1306. If you can make a commit to fix that in this PR that's cool, or otherwise we can open a new issue.

webcompat/static/js/lib/bugform.js
miketaylr: Ideally we fix `checkImageTypeValidity` here to not know about form inputs, and to get its data from the `event` object instead.

For the drop event, we can get the file from here: https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/files.

cch5ng:
@miketaylr this is just a high level response to your questions. I will respond inline to your comments afterward. basically I feel this is a bit of a hack solution to what seems more like a new feature than a bug.

this is the cause of why currently the 2nd drag and drop event replaces the browser url with the 2nd file url:
https://github.com/webcompat/webcompat.com/issues/1571#issuecomment-306624657

so the solution I turned in was to prevent ever hiding the input field for image file. but this causes a bunch of breakage of existing logic for that input's state. documented:
https://github.com/webcompat/webcompat.com/pull/1593#issue-235617483
https://github.com/webcompat/webcompat.com/pull/1634#issue-240294228

looking at bugform.js > showRemoveUpload() around current lines
https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L450
(and about 10 lines down), it looks like the previous author only intended that input field to have 2 states 
* filled and you can remove the file with the button/link or 
* empty and you can click/drag 'n drop to add a file

that is why my changes are a hack and make the resulting code difficult to understand. I still believe the cleaner solution would have been more along the lines of Dennis' suggestion of only preventing the 2nd drag and drop attempt from updating the browser url (requiring the user to click Remove before another drag 'n drop can occur): https://github.com/webcompat/webcompat.com/issues/1571#issuecomment-306491631
(which I investigated and needed a little more logic but requires little code to resolve and is much clearer to maintain; this is available on a separate local branch)

webcompat/static/css/development/components/upload-form.css
cch5ng: @miketaylr this is part of the hack solution described at the bottom. originally the UI was designed to either:
* only have the file input field active or
* only have the remove button/link active

in order to make the 2nd drag/drop work, the design had to be updated to allow both fields to be interactive at the same time. otherwise the default behavior would be:
* user drags 2nd file over file input field
* user tries to click the Remove button/link => the dialog to upload a new file pops up
  * workaround was to put the Remove button on top

webcompat/static/js/lib/bugform.js
cch5ng: @miketaylr yes, I agree, if the text area never gets updated until the form gets submitted, I think that logic is obsolete but will test and verify.

cch5ng:
@miketaylr 
> https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L466-L468 is actually dead code now, I forgot to remove it in #1306. If you can make a commit to fix that in this PR that's cool, or otherwise we can open a new issue.

ok, I will add that as a todo here

cch5ng:
@miketaylr I think I made all of your change requests and fixed a couple issues after the rebase. let me know if I missed anything. travis tests running currently (will check if they pass a little later). thanks.

webcompat/static/js/lib/bugform.js
cch5ng: I ended up removing the event handler. previously during testing it was working I think. but currently it was causing the 2nd (+) drag and drops to fail. I cannot recall why it was working/needed before. but currently it appears to make the input field unavailable again.

webcompat/static/js/lib/bugform.js
```diff
@@ -442,30 +442,37 @@ function BugForm() {
...
         uploadWrapper.removeClass("is-hidden");
         removeBanner.off("click");
 
         // remove the last embedded image URL
         // Note: this could fail in weird ways depending on how
         // the user has edited the descField.
-        this.stepsToReproduceField.val(function(idx, value) {
-          return value.replace(/\[!\[[^\]]+\]\([^\)]+\)\]\([^\)]+\)$/, "");
-        });
+        // this.stepsToReproduceField.val(function(idx, value) {
```
miketaylr: Is commenting this out intentional?

webcompat/static/js/lib/bugform.js
```diff
@@ -485,7 +492,7 @@ function BugForm() {
       // grab the data URI from background-image property, since it's already
       // in the DOM: 'url("data:image/....")'
       this.previewEl.get(0).style.backgroundImage.slice(5, -2),
-      function(response) {
+      _.bind(function(response) {
```
miketaylr: Why are we adding _.bind here? I think I just removed it during a recent review.

webcompat/static/css/development/components/upload-form.css
miketaylr: Thanks. If we decide to keep this, I'd like the comment in the CSS to explain some of this, or at least provide a link to a longer explanation.

miketaylr:
> I wonder if some code got lost, or not pushed.

`fix: #1571; update to use this.hasImage from #1603` from the commit message makes me think this is likely... I don't see hasImage getting used in this changeset at all.

webcompat/static/js/lib/bugform.js
cch5ng: I think there was miscommunication, I remember asking in the review about why a _.bind() was used but I was not asking to remove it, just asking for clarification. when I was taught, "this" and "bind()" it was done in a really confusing way so I still have to think about it in usage.

in the current master version, I'm getting an error like this when trying to save an issue:
```
TypeError: this.addImageURL is not a function bugform.js:491:9
```

miketaylr: Yeah, my bad. It's back on master.

webcompat/static/js/lib/bugform.js
miketaylr: Hm, I'll have to think this over. It feels a bit too accidental to hide a different DOM element to get this to work. I still think we should be listening for `drop` or `dropend` events and handling things from there, if we want to do this in a more robust way.

But maybe I'm missing something. Will think more about this.

cch5ng:
re: comment below, that whole function was removed in the latest commit
(squashed). I think it was only really useful for an alternative solution
to prevent 2nd drag/drop event from changing the browser url. thx. I never
pushed that alternative solution to github though.

cch5ng:
@miketaylr I think I've made all requested changes on this branch

miketaylr:
OK, this doesn't address my comments at https://github.com/webcompat/webcompat.com/pull/1634#discussion_r127544600, but it does work. So let's merge it for now and I'll file a follow up issue.

Also, I can handle this right now, but in the future for a single commit, can you clean up the commit message?

The following has a ton of irrelevant info for the changeset that is landing:

```
 fix: #1571; update js and styles to handle multiple drag and drop events
cch5ng committed 16 days ago

also update after #1306

fix: #1571; add handler for multiple drag and drop events; update some styling

fix: #1571; update to use this.hasImage from #1603

fix: #1571; clean up addImageURL(); add _.bind() wrapper in maybeUploadImage() to fix error for this.addImageURL() call

fix: #1571; clean up code added for #1574

fix: #1571; remove obsolete handler, handleImageDrop(); for some reason after #1603 this breaks the intended functionality

cleanup: #1571; remove old logic to update text area

doc: #1571; update comment for dnd fix
```

Just something like `Issue #1571. Handle multiple images being dropped into upload area` (or similar). You can also break things up into different commits if they need more explanation, or the single commit is doing too much. Thanks!

cch5ng:
@miketaylr 
> Also, I can handle this right now, but in the future for a single commit, can you clean up the commit message?

thanks for the fix/merge. yes, I was not sure how much to edit all the old commit messages but in the future I will just leave one.

> OK, this doesn't address my comments at #1634 (comment), but it does work. So let's merge it for now and I'll file a follow up issue.

one item I forgot to mention in email (re branch issues/1571/4 alt solution): for me the ideal solution would have been to handle/listen for the 2nd image drag/drop event and get the new file info from event,  then update the file input field with the new file data. however, I followed the current docs (html drag and drop api) and was unable to get the file info from the event (maybe too experimental?). I can write up more details in the wiki if useful (mdn link, etc).

miketaylr:
> however, I followed the current docs (html drag and drop api) and was unable to get the file info from the event (maybe too experimental?). 

That would be surprising, DnD is pretty old stuff at this point.

> I can write up more details in the wiki if useful (mdn link, etc).

Sure, that sounds good.

Issue #1551, opened 2017-05-10
[feature] type-media bugs - handling duplicates automagically


karlcow:
For certain domains, such as vk.com or yandex.ru in the category `type-media` we have a series of duplicates. A couple of them every day. It would make it easier for us to deal with them if we could triage them automagically.

This is just an idea on how we could maybe handle this.

1. An issue is reported with `type-media` label (we know when it's coming from the media reporting system)
2. It contains a specific Error Code and a specific domain name.
3. Check if the tuple (`domain_name`, `media_error_code`) is in the `type-media` duplicate cache list: `media_bugs_cache_list.txt`.
4. If yes
    - Remove `status-needstriage`
    - Add `status-duplicate`
    - Add comment with "closing as duplicate of `first_bug_number`"
    - option: Add comment on  `first_bug_number` with "`current_bug_number` is a duplicate"
    - Close the issue 
5. If No
    - Add a new line to `media_bugs_cache_list.txt` with the following information `domain_name`, `media_error_code`, `first_bug_number`
    - Remove `status-needstriage`
    - Add `status-needsdiagnosis`

This should remove some of our daily burden.

To be discussed: @adamopenweb @MDTsai some thoughts?

Unknown:
* Issues where type-media is added after the fact.
* Old issues do we create a batch processing script doing the above for old bugs we already received.

MDTsai:
One concern is media file in different CDN server, but seems not happened till now. So I think this works.

karlcow:
@MDTsai Thanks! Indeed that might influence the result. For example, for social networks, where external links are authorized.  That could be an additional check. 

I should download all the current data we have and see if we can have a better idea of the variability.

karlcow:
@miketaylr Currently for the webhook, we are using the `/labeler` route which is adding two new labels: browser and needstriage.

(thinking out loud)

There are maybe three options:

* reusing `/labeler` route.  
 Probably modifying the name to be more generic and improving a bit more the logic in the code to be more versatile. The `if` cascade bothers me a bit. 
* <del>OR create a new route for the purpose of triaging type-media issues.  </del>
  How would that work? Can we assign more than one URL per webhook. Heading to the docs. Ah… so a Webhook is only targeting one specific URL. And we already receive all issues events onto labeler. Creating a new route is then not possible without creating a new webhook. Let's strike this option.
* <del>OR create another webhook on the project. </del>  
[...]

PR #1635, opened 2017-07-04
#1565 update issues link to filter out type-media github issues [not ready to merge]

cch5ng:
hmmm, (update 070617), I think the hack is actually broken. I will rethink and retest and come back with better info.

---

this is just a quick hackey initial fix

it might not get used as I believe #1551 is planned to be closed after talking with the media group. but perhaps this logic might be useful later on.

caveats:
* this only affects the issues list from the All Issues nav link
* it results in some prefilled text in the Search input field; but it still allows the user to add more search filter text
* it only works if the user is logged into github
* cases where the filter is not applied (not sure where to find the code right before the api get is made)
  * index.html (triage list + when you search from index.html)
    * probably when you click the View all link from the triage list area
  * my activity page

---
update: as I believe this is a low priority issue, I will put it on the backburner and continue to think and investigate the code to make a cleaner fix. but will probably be working on other issues first starting from this week.

### DIFF
--- a/webcompat/templates/shared/shared-nav-links.html
+++ b/webcompat/templates/shared/shared-nav-links.html
@@ -3,7 +3,7 @@
   <span class="wc-Navbar-link-label">Contribute</span>
 </a>
 <a class="wc-Navbar-link js-issues-link"
-    href="{{ url_for('show_issues') }}?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc"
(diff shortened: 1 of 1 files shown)


miketaylr:
If we decide to go with this, or come back to it -- we probably want to make a similar change to the issue list model: https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/models/issue.js#L151-L158

Maybe we want to touch `updateModelParams` as well... https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/issue-list.js#L480

cch5ng:
closing as outdated

PR #1642, opened 2017-07-06
#1565 - add negative label filter for manual search [not ready to merge]

cch5ng:
int test: #1565; intercept search query for negative label filter

history #1635 (also see MikeT's comments there)

this issue is on hold (related to #1551) but this is just a quick code sample for future record


miketaylr:
I recommend we close this for now -- if/when we re-enable type-media let's re-visit.

PR #1660, opened 2017-07-17
Fixes #1651 - adding Code of Conduct

### LINKED ISSUE #1651: Adding CoC to webcompat.com
As Mozillas Guidelines of Participation are still redone, how about adding a short Code of Conduct to this and our other repositories? For example http://contributor-covenant.org/ aligns with the Mozilla values and could be added easily.

zoepage:
r? @miketaylr

### DIFF
--- a/codeofconduct.md
+++ b/codeofconduct.md
@@ -1,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of experience,
+nationality, personal appearance, race, religion, or sexual identity and
+orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+  address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
+threatening, offensive, or harmful.
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at ogasidlo@mozilla.com. All
+complaints will be reviewed and investigated and will result in a response that
+is deemed necessary and appropriate to the circumstances. The project team is
+obligated to maintain confidentiality with regard to the reporter of an incident.
+Further details of specific enforcement policies may be posted separately.
+
+Project maintainers who do not follow or enforce the Code of Conduct in good
+faith may face temporary or permanent repercussions as determined by other
+members of the project's leadership.
+
+## Attribution
+
+This Code of Conduct is adapted from the [Contributor Covenant][homepage], version 1.4,
+available at [http://contributor-covenant.org/version/1/4][version]
+
+[homepage]: http://contributor-covenant.org
+[version]: http://contributor-covenant.org/version/1/4/
--- a/codeofconduct.md
+++ b/codeofconduct.md
@@ -3,7 +3,7 @@
 ## Our Pledge
 
 In the interest of fostering an open and welcoming environment, we as
-contributors and maintainers pledge to making participation in our project and
+contributors and maintainers pledge to make participation in our project and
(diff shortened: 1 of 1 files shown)


miketaylr:
FYI @webcompat/core-developers.

codeofconduct.md
```diff
@@ -0,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
```
tagawa: `pledge to making` -> `pledge to make`

codeofconduct.md
```diff
@@ -0,0 +1,74 @@
...
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of experience,
+nationality, personal appearance, race, religion, or sexual identity and
+orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
```
tagawa: `that contributes to` -> `that contribute to`

codeofconduct.md
```diff
@@ -0,0 +1,74 @@
...
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
```
tagawa: `behaviors` -> `behavior`

codeofconduct.md
```diff
@@ -0,0 +1,74 @@
...
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
```
tagawa: `abusive, harassing` -> `abuse, harassment`

tagawa:
Added a few grammatical comments. 

Re. the reporting procedure, wouldn't it be better to have a general inbox (e.g. webcompat-support@mozilla.com or similar) rather than a single personal address?

denschub:
@tagawa you should probably file language-related issues upstream at https://github.com/ContributorCovenant/contributor_covenant, although I can recall at least some of your points being discussed before, and it turns out those were perfectly legal and "more right" when watched from a linguistic point of view.

I'm torn between having a generic inbox vs. having a person who is responsible for handling complains. I've heard questions in the past if mails end up on a public mailing list or people being concerned about who can rad complain mails, which isn't really an issue if there is a dedicated person, since people know this is the only person who will have direct access to the mail. No objection against trusting @zoepage with that job from my side.

Also, thanks for doing this. ❤️

zoepage:
@tagawa Thanks! I agree with @denschub on filing issues. :)

*re:general address*
I've added a single person (me) on purpose as a generic contact information is faceless and people tend to not report issues, especially people who get harassed, if they don't see, if they can trust the other side they report to. Not that people think, I'm the most trustworthy person ever, but a face makes things more familiar and puts in more trust. In addition having a female handle those issues is also a plus from a psychological way.  

@denschub ofc <3

tagawa:
Thanks for the comments about upstream changes. I'll take a look at the main project and see what the past discussions were.

I see your point about a personal address seeming more welcoming than a generic address, and I also agree that it mustn't be a public mailing list. This could perhaps be made clear in note in the Code of Conduct. I also have no doubts or prejudices regarding any of the WebCompat team to act as the point of contact. My concerns are:
* What happens if that person goes on holiday or is taken ill? You could redirect mails to someone else's mailbox but then the sender wouldn't know. Alternatively you could change the Code of Conduct temporarily but it's a hassle and liable to be forgotten.
* There's a (hopefully low) chance that a serious complaint is raised with legal implications. To be well prepared, it's sensible to have oversight by a second person from the very start to offer maximum protection. It also avoids the need for human judgement about whether to forward an issue or not.

Incidentally, I don't think we should be making generalisations about which gender is preferable for the point of contact, as this depends on the individual filing a complaint. Probably best to not make assumptions about people's preferences. A compromise might be to set up a mailbox with a neutral human name, e.g. webcompat-alex@mozilla.com but that could be seen as deceptive.

Anyway, there are pros and cons to both approaches so interested to hear other opinions.

zoepage:
@tagawa 
> I also have no doubts or prejudices regarding any of the WebCompat team to act as the point of contact. 

I agree. What about we add @denschub as a second person as he is also very experienced in CoC matters? Would also solve the holiday / legal issue.

> Incidentally, I don't think we should be making generalisations about which gender is preferable for the point of contact, as this depends on the individual filing a complaint.

This is not a generalization or assumptions, but drawn from my experience being a CoC person for years in the FLOSS community. As the most reports (from my experience and from studies I've read > 95%) come from non "cis males", it should be safe to assume a female person as point of contact is crucial those incidents get reported.  

As such matters can be very delicate and in the "wrong hands" create a really hurtful backlash to the reporters professional as well as personal life, it is important to create a safe space for everyone and build a base of trust. When those who report know, who they are reporting to and can even do a bit of research for that person on the web, to make sure the person is trustworthy, it will increase the chance of getting a report immensely. 

This is why, I'd also not just add the email adress, but also the GitHub handle to the CoC.

codeofconduct.md
miketaylr: +1

codeofconduct.md
miketaylr: +1

codeofconduct.md
miketaylr: (while the plural sounds fine to my American English ears...) Behavior is typically considered an uncountable noun, so singular is good here.

codeofconduct.md
miketaylr: This is like a classic example of ambiguous antecedent agreement. Both can be correct, depending on your reading of its structure (e.g., are we describing nouns here, or using adjectives for types of behavior). 

I have no strong preference.

miketaylr:
> @tagawa you should probably file language-related issues upstream at https://github.com/ContributorCovenant/contributor_covenant, although I can recall at least some of your points being discussed before, and it turns out those were perfectly legal and "more right" when watched from a linguistic point of view.

I think we can pull in @tagawa's suggested edits in the meantime.

miketaylr:
Also, if @denschub agrees, I think it's preferable to have 2 contacts.

denschub:
No objections, have been doing the same in other projects for quite a while now.

zoepage:
Added suggested changes.

miketaylr:
Thanks @zoepage!

PR #1667, opened 2017-07-19
Fixes #1564 - Disallow inline scripts via CSP.

### LINKED ISSUE #1564: Do the work to remove CSP script-src 'unsafe-inline' directive
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_inline_script

This means we need to fix all the places we're sending inline `<script>blah()</script>` with script in an actual blah.js file -- which shouldn't be very hard, IMO.

Follow up to #763.

miketaylr:
(I want to run this on localhost for a few days before I pull the trigger and ask for review.)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,7 +16,6 @@
     "Backbone": true,
     "console": true,
     "define": true,
-    "issueNumber": true,
     "markdownitEmoji": true,
     "markdownitSanitizer": true,
     "md": true,
@@ -26,7 +25,6 @@
     "PaginationMixin": true,
     "Prism": true,
     "Promise": true,
-    "repoPath": true,
     "require": true,
     "wcEvents": true,
     "WindowHelpers": true
--- a/webcompat/static/js/lib/flashed-messages.js
+++ b/webcompat/static/js/lib/flashed-messages.js
@@ -1,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// This file is only included in the page when the Flask view is rendered with
+// a flash message. It's responsible for triggering that message as defined
+// in flash-message.js. See layout.html for how category and message data
+// attributes are set.
+
+var currentScript = document.currentScript;
+if (currentScript) {
+  var category = currentScript.dataset.category || "";
+  var message = currentScript.dataset.message || "";
+
+  wcEvents.trigger("flash:" + category, { message: message, timeout: 4000 });
+}
--- a/grunt-tasks/uglify.js
+++ b/grunt-tasks/uglify.js
@@ -28,6 +28,10 @@ module.exports = function(grunt) {
       src: "<%= concat.diagnose.dest %>",
       dest: "<%= jsPath %>/diagnose.min.js"
     },
+    contributors: {
+      src: "<%= jsPath %>/lib/contributors.js",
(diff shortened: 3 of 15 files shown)


.eslintrc
```diff
@@ -26,7 +26,6 @@
     "PaginationMixin": true,
     "Prism": true,
     "Promise": true,
-    "repoPath": true,
```
miketaylr: I think we can remove `issueNumber` as a global as well.

miketaylr:
Some failures to investigate:

```
FAIL: firefox on any platform - Issues - NSFW images are blurred (10553ms)

NoSuchElement: An element could not be located on the page using the given search parameters.

  at nextTickCallbackWith0Args  <node.js:489:9>

  at process._tickCallback  <node.js:418:13>

  at Test.registerSuite.NSFW images are blurred [as test]  <tests/functional/issues-non-auth.js:86:12>

  at nextTickCallbackWith0Args  <node.js:489:9>
```

webcompat/static/js/lib/flashed-messages.js
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// This file is only included in the page when the Flask view is rendered with
+// a flash message. It's responsible for triggering that message as defined
+// in flash-message.js. See layout.html for how category and message data
+// attributes are set.
+
+var currentScript = document.currentScript;
```
miketaylr: @wisniewskit suggested that if we stick the category and message somewhere else like body or html (or main), we don't need to use `currentScript` and it will work in IE. 👍

miketaylr: Hm, this is tricky because we load this script at the bottom of the page to be sure `flash-message.js` is already loaded -- which comes just before `</body>`. So there's an ordering dependency. Also, we can't just set these attributes on `<body>` or `<main>`, then load the script down below because calling `get_flashed_messages` clears them from the session: https://github.com/pallets/flask/blob/a63a4cb034955f5d9c4b14f87a5337d94ba472a1/flask/helpers.py#L425.

There's a polyfill, but it only works for IE6 - 10, and *not* IE11 (which is what ships in Windows 10). Since Edge works with this approach, and Microsoft wants their users on Edge... I'm inclined to just forget about old IE support anyways.... :/

karlcow:
hmmm…

```
→ git checkout -b issues/1564/1 webcompat/issues/1564/1
M	git_modules/css-fixme
Branch issues/1564/1 set up to track remote branch issues/1564/1 from webcompat.
Switched to a new branch 'issues/1564/1'
```

anyway. Let's see.

webcompat/static/js/lib/flashed-messages.js
karlcow: +1 on using `data-*` without polyfills.

karlcow:
@miketaylr r+ 
Thanks for the code.

karlcow:
And I merged with the status: do not merge.

karlcow:
I reverted it. :)

miketaylr:
All good, I forgot to remove that. :)
I'll re-land at some point today -- thanks for review!

miketaylr:
instead, i just pushed the reverting commit to https://github.com/webcompat/webcompat.com/commit/8fadbd6463464c5120586583bef8651980e2350e.

PR #1668, opened 2017-07-20
Fixes #1666 - Get TravisCI functional tests to work with Browserstack.

### LINKED ISSUE #1666: Get TravisCI functional tests to work with Browserstack.
So, Browserstack would let us run functional tests in a bunch of different browsers and OSs. 
For now we could just start with testing in the latest version of Firefox, Chrome, Safari and Edge for Desktop (if that doesn't take too long).

[EDIT]

After discussing this issue in our team meeting, we decided that it's would be best to create a TravisCI cronjob to run periodically with Browserstack, instead of having it run with every PR. The reasons that led us to this decision are that Browserstack would not work for forks, because it requires secure variables, it takes to long and is not very reliable. So what need to be done is:

- [x] Get Browserstack to work locally, so that we can test it from our machines anytime.
- [ ] Get Browserstack to work with TravisCI and setup cronjob.
- [ ] Update CONTRIBUTING.md.

brizental:


tests/browserstack-remote.js
```diff
@@ -0,0 +1,87 @@
...
+      remoteFiles: true
+    },
+
+    // Required for BrowserStack, maximum number of simultaneous integration tests allowed
+    maxConcurrency: 2,
+
+    environments: [
+      {
+        browser: "firefox",
+        browser_version: "54",
```
miketaylr: Do they have any notion of "latest", or does this need to be updated periodically?

brizental: So, according to documentation they do (https://www.browserstack.com/automate/capabilities). You just leave the version blank an it should default to lastest stable. But I tested it and it didn't really work as expected, just returned a random version everytime :/

miketaylr: lol ok

brizental:
Closing this because it doesn't work.

PR #1670, opened 2017-07-22
[not ready to merge] internal testing for another drag drop handling solution

cch5ng:
stubs out code (drag and drop event handlers) for further exploration to #1571


cch5ng:
closing as this stub is only posted so the branch is not deleted; but available for future research/testing

PR #1671, opened 2017-07-22
Fixes #1669 - Gets tests to with Chrome as well as Firefox. Travis and Local.

### LINKED ISSUE #1669: Use TravisCI only for cross-browser func testing (rather than BrowserStack, etc)
Travis has the ability to run Chrome now, in addition to Firefox. And it's fast.

BrowserStack and Sauce are super slow, and have their own set of constraints.

Let's discuss.

+@brizental

brizental:
r? @miketaylr 

Had to also make some changes to some tests because they were failing in Chrome.

### DIFF
--- a/tests/functional/index-non-auth.js
+++ b/tests/functional/index-non-auth.js
@@ -45,7 +45,12 @@ define(
             .findByCssSelector("#js-ReportBug")
             .click()
             .end()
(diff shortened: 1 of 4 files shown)


miketaylr:
👍

PR #1672, opened 2017-07-24
Fixes #1646 - Updates bug form text area placeholder

### LINKED ISSUE #1646: How did you get there example text could be improved
Instead of populating the textarea, maybe we can use `@placeholder` text? Possibly with some validation for empty text? (Or not, there's a good argument that if an issue has an image, it's still useful)...

```
For example,

    I've tried to log in.
    I've filled out the form details.
    I clicked on the submit button.
    Nothing happened.
```

cch5ng:
review requests: 
1 @miketaylr r?
2 @karlcow r? (I guess optional, if you have the time)

switched back to generating text area from the python. it looks like everything gets saved to github ok.


webcompat/templates/home-page/form.html
```diff
@@ -62,7 +62,8 @@ <h3 class="wc-Title--s">Report Site Issue</h3>
             </span>
           </div>
           <div class="wc-Form-group js-Form-group">
-            {{ form.steps_reproduce.label(class_='wc-Form-label') }} {{ form.steps_reproduce(class_='wc-Form-item wc-Form-item--textarea js-autoExpand', rows=4) }}
+            <label class="wc-Form-label" for="steps_reproduce">How did you get there?</label>
+            <textarea class="wc-Form-item wc-Form-item--textarea js-autoExpand" id="steps_reproduce" name="steps_reproduce" rows="4" placeholder="Steps to reproduce the issue"></textarea>
```
miketaylr: Can we include the pre-existing `steps_default` as a placeholder? I think it's good to give the user a hint to give us numbered steps (for example).

webcompat/templates/home-page/form.html
```diff
@@ -62,7 +62,8 @@ <h3 class="wc-Title--s">Report Site Issue</h3>
             </span>
           </div>
           <div class="wc-Form-group js-Form-group">
-            {{ form.steps_reproduce.label(class_='wc-Form-label') }} {{ form.steps_reproduce(class_='wc-Form-item wc-Form-item--textarea js-autoExpand', rows=4) }}
+            <label class="wc-Form-label" for="steps_reproduce">How did you get there?</label>
```
miketaylr: We don't need to hardcode this. `form.steps_reproduce(class_='wc-Form-item wc-Form-item--textarea js-autoExpand', rows=4)` <-- you should be able to pass in `placeholder=foo`.

miketaylr:
(there's a bunch of unrelated merge commits or things referencing #980 that we need to clean up before merging as well, thanks!)

webcompat/templates/home-page/form.html
cch5ng: from Dennis' comment, https://github.com/webcompat/webcompat.com/issues/1646#issuecomment-317375032
(about placeholder text not supporting line breaks), I just made it one line

miketaylr: Yeah, that's unfortunate. But we could try another approach where we include the existing STR in the textarea and just clear it out on textarea `focus`.

denschub: (and make sure we don't submit the "placeholder" then.)

cch5ng: I've tried to include line breaks like:
* `\n`
* `&lt;br&gt;`
* `<br>`

none of these has worked. there is an older SO article:
https://stackoverflow.com/questions/7189223/can-you-have-multiline-html5-placeholder-text-in-a-textarea

but it doesn't seem consistent across browsers. do you still want me to investigate this further?

cch5ng:
moved changes to #1675 and so am closing this
(couldn't figure out how to clear the old commits here)

PR #1674, opened 2017-07-24
Fixes #1673 - remove extra spaces in .travis.yml path to prevent early exit

### LINKED ISSUE #1673: bad file path in .travis.yml file causes travis process to exit early
the travis build/test process is exiting early

error: The command "pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example" failed and exited with 1 during .

looks like unnecessary spaces somehow got inserted in the path:
https://github.com/webcompat/webcompat.com/blob/master/.travis.yml#L46

cch5ng:
@miketaylr r?


miketaylr:
@cch5ng those spaces exist to tell the pep8 binary to search in those different directories. If we remove the spaces, it will then only try to link a non-existent file: `webcompat/tests/config/secrets.py.example`

Issue #1665, opened 2017-07-18
Get tests to pass with PRs from forks.


brizental:
So, we use some encrypted variables in Travis and that is why the tests don't pass for forks.

Talking to @miketaylr, we decided it would be best to just go around the tests that need these secure variables for now. Basically make a subset of tests that don't need them for when the PR is from a fork.

Not ideal, but there is an open issue for encrypted vars to work with forks in the Travis repo (https://github.com/travis-ci/travis-ci/issues/1946). We could watch that and when it's done we could let any PR run all tests.

[EDIT]

So, I figured if we get all tests to always mock all responses and stuff, basically, get them all to work offline, we wouldn't need anymore secure variables and then all tests would always run for everyone.

So now what's needed for this to work is:

- [x] Check that all unittests are working offline, if not, get them to. We already have and open issue for that #1618 
        NOTE: funcional tests are already all working offline (see: #1625 )
- [x] Remove all secure variables and the conditional for different test suites from travis.yml.

PR #1675, opened 2017-07-24
Fixes #1646 - update text area placeholder

### LINKED ISSUE #1646: How did you get there example text could be improved

cch5ng:
@miketaylr r? (didn't know how to clean up all the old commits but this includes change requests from today)

### DIFF
--- a/tests/functional/reporting-non-auth.js
+++ b/tests/functional/reporting-non-auth.js
@@ -15,8 +15,6 @@ define(
     var cwd = intern.config.basePath;
     var VALID_IMAGE_PATH = path.join(cwd, "tests/fixtures", "green_square.png");
     var BAD_IMAGE_PATH = path.join(cwd, "tests/fixtures", "evil.py");
-    var DETAILS_STRING =
-      "Encountered error: NS_ERROR_DOM_MEDIA_DEMUXER_ERR (0x806e000c)%0ALocation: virtual%0ARefPtrMP4Demuxer::InitPromise mozilla::MP4Demuxer::Init()%0AError information:%0AIncomplete MP4 metadata%0AMedia URL: file:///Users/potch/Documents/mozilla/media.mp4";
     var DETAILS_STRING2 = "Encountered+error:+NS_ERROR_DOM_MEDIA_DEMUXER_ERR";
 
     var url = function(path) {
@@ -255,21 +253,6 @@ define(
         );
       },
 
-      // tests a scenario where bug reported from firefox nightly Report Site Issue button
-      "details param adds stylo pref to text area": function() {
-        return FunctionalHelpers.openPage(
-          this,
-          url("/issues/new?details=layout.css.servo.enabled%3A+true"),
-          ".wc-ReportForm-actions-button"
-        )
-          .findByCssSelector("#steps_reproduce")
-          .getProperty("value")
-          .then(function(text) {
-            assert.include(text, "layout.css.servo.enabled: true");
-          })
-          .end();
-      },
-
       // tests a scenario where bug reported from firefox nightly media type tool
       "problem_type param selects problem type": function() {
         this.skip("former media type reporting test");
@@ -327,6 +310,40 @@ define(
               text,
               "Encountered error: NS_ERROR_DOM_MEDIA_DEMUXER_ERR",
               "+ replaced with space"
+            );
+          })
+          .end();
+      },
+
+      "clicking text area clears default value": function() {
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues/new"),
+          ".wc-ReportForm-actions-button"
+        )
+          .findByCssSelector("#steps_reproduce")
+          .click()
+          .sleep(500)
+          .getProperty("value")
+          .then(function(text) {
+            assert.lengthOf(text, 0, "clicking default text area clears value");
+          })
+          .end();
+      },
+
+      "details param: not added to text area": function() {
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues/new?details=" + DETAILS_STRING2),
+          "#description"
+        )
+          .findByCssSelector("#steps_reproduce")
+          .getProperty("value")
+          .then(function(text) {
+            assert.notInclude(
+              text,
+              DETAILS_STRING2,
+              "+ not found in decoded string"
             );
           })
           .end();
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -11,6 +11,7 @@ function BugForm() {
   this.UPLOAD_LIMIT = 1024 * 1024 * 4;
   this.clickedButton = null;
   this.hasImage = null;
+  this.textAreaModified = false;
 
   this.inputs = {
     url: {
@@ -75,13 +76,21 @@ function BugForm() {
       _.bind(this.checkDescriptionValidity, this)
     );
     this.problemType.on("change", _.bind(this.checkProblemTypeValidity, this));
+    this.stepsToReproduceField.on(
+      "change",
+      _.bind(this.updateTextAreaModified, this)
+    );
+    this.stepsToReproduceField.on(
+      "focus",
+      _.bind(this.clearStepsToReproducePlaceholder, this)
+    );
     this.uploadField.on("change", _.bind(this.checkImageTypeValidity, this));
     this.osField
       .add(this.browserField)
       .on("blur input", _.bind(this.checkOptionalNonEmpty, this));
     this.submitButtons.on("click", _.bind(this.storeClickedButton, this));
     this.submitButtons.on("click", _.bind(this.loadingIndicator.show, this));
-    this.form.on("submit", _.bind(this.maybeUploadImage, this));
+    this.form.on("submit", _.bind(this.onFormSubmit, this));
 
     // See if the user already has a valid form
     // (after a page refresh, back button, etc.)
@@ -179,20 +188,6 @@ function BugForm() {
       $("[value=" + problemType[1] + "]").click();
     }
 
-    // If we got a details param, add that to the end of the steps to reproduce field
-    var details = location.href.match(/details=([^&]*)/);
-    if (details !== null) {
-      this.stepsToReproduceField.val(function(idx, value) {
-        return (
-          value +
-          "\n" +
-          // The content of the details param may be encoded via
-          // application/x-www-form-urlencoded, so we need to change the
-          // + (SPACE) to %20 before decoding
-          decodeURIComponent(details[1].replace(/\+/g, "%20"))
-        );
-      });
-    }
   };
 
   this.storeClickedButton = function(event) {
@@ -224,6 +219,20 @@ function BugForm() {
       this.makeInvalid("problem_type");
     } else {
       this.makeValid("problem_type");
+    }
+  };
+
+  this.updateTextAreaModified = function(ev) {
+    if (ev.target.value.length) {
+      this.textAreaModified = true;
+    }
+  };
+
+  this.clearStepsToReproducePlaceholder = function() {
+    if (!this.textAreaModified) {
+      this.stepsToReproduceField.val(function() {
+        return "";
+      });
     }
   };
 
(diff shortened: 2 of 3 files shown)


cch5ng:
@miketaylr in this vers, I tried your suggestion
https://github.com/webcompat/webcompat.com/pull/1672#discussion_r129147324

I tested against my local Nightly (with the servo/stylo pref prepopulated). it appears to work ok except for maybe an extra line break at the top. travis tests currently running...

miketaylr:
> @karlcow r? (if you have time)

Karl is on PTO, reviewing now.

webcompat/static/js/lib/bugform.js
```diff
@@ -227,6 +231,18 @@ function BugForm() {
     }
   };
 
+  this.clearStepsToReproducePlaceholder = function() {
+    var stepsToReproduceStrAr = this.stepsToReproduceField
+      .val()
+      .split("4. Nothing happened.\n");
+    // handle the case where stylo pref prepopulated from FF Nightly
+    if (stepsToReproduceStrAr.length > 1) {
```
miketaylr: @cch5ng We're always going to enter into this if clause if the user hasn't modified the textarea, right?

webcompat/static/js/lib/bugform.js
```diff
@@ -227,6 +231,18 @@ function BugForm() {
     }
   };
 
+  this.clearStepsToReproducePlaceholder = function() {
+    var stepsToReproduceStrAr = this.stepsToReproduceField
+      .val()
+      .split("4. Nothing happened.\n");
+    // handle the case where stylo pref prepopulated from FF Nightly
```
miketaylr: This comment should be moved below, just above `return stepsToReproduceStrAr[1].trim();`. 

(Also, FF Nightly isn't the only browser that has the button, DevEdition does as well. So it might be better to say "handle the case where the user hasn't modified the description, but the Report Site Issue button has added stylo pref information", or something like that)

webcompat/static/js/lib/bugform.js
cch5ng: @miketaylr I guess these are 3 cases that I can come up with:
* default case (only the placeholder text)
* servo case (placeholder text + appended servo/stylo pref str)
* updated case (user already clicked and edited the field; ie placeholder text is gone)

I'll rethink how to handle each case

cch5ng:
@miketaylr 
> What do you think we should do in the case where the user never clicks in the description textarea? Should we run clearStepstoReproducePlaceholder in the submit handler?

hm, yes, I did not test that. make sense to me

cch5ng:
@miketaylr this version should include your change requests +1 functional test. let me know if there should be more changes. travis tests running now...

cch5ng:
@miketaylr hmm, my travis tests are failing (here and for 1662 as well); seeing a variety of python as well as functional tests failures

do you suppose this could be important: https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming?utm_source=web&utm_medium=banner&&utm_campaign=trusty-default

(at the top of the travis report page): This job is running on our Trusty environment, which is gradually becoming our default Linux enviroment. Read all about this in our blog: Trusty as a default Linux is coming and take note that you can add dist: precise in your .travis.yml file to continue using Precise.

(I have been known to experience bad test karma in the past...)

webcompat/static/js/lib/bugform.js
```diff
@@ -75,13 +75,18 @@ function BugForm() {
...
+      _.bind(this.clearStepsToReproducePlaceholder, this)
+    );
     this.uploadField.on("change", _.bind(this.checkImageTypeValidity, this));
     this.osField
       .add(this.browserField)
       .on("blur input", _.bind(this.checkOptionalNonEmpty, this));
     this.submitButtons.on("click", _.bind(this.storeClickedButton, this));
     this.submitButtons.on("click", _.bind(this.loadingIndicator.show, this));
     this.form.on("submit", _.bind(this.maybeUploadImage, this));
+    this.form.on("submit", _.bind(this.clearStepsToReproducePlaceholder, this));
```
cch5ng: @miketaylr is this what you meant for me to do? usually I would write one function as the handler for form submit and place related functions within. but it didn't seem right to put clearStepsToReproducePlaceholder() inside maybeUploadImage().

miketaylr:
> (update 072617) previously when all my travis tests were passing, I had a brief period when I was pushing my branches directly to /webcompat/webcompat.com repo and now (since I updated my dev env on the mozilla laptop) I reset my repo to push to my forked repo /cch5ng/forked-webcompat.com repo... maybe that is the reason for the travis test failures... let me know if that seems important to fix

Yes, all your tests will fail because this PR is from a fork. #1665 and others are trying to make that problem better -- but you'll either need to run all functional and python tests yourself, or develop on the main repo in a branch. Up to you.

webcompat/static/js/lib/bugform.js
miketaylr: No, not quite. I would re-name `maybeUploadImage` to something less image specific (maybe just `onFormSubmit`) and call `clearStepsToReproducePlaceholder` inside that method.

cch5ng: @miketaylr the latest change request is in the latest push. thx

npm-debug.log
```diff
@@ -0,0 +1,47 @@
+0 info it worked if it ends with ok
```
miketaylr: @cch5ng can you add `npm-debug.log` to `.gitignore`, and remove this file from the PR?

tests/functional/reporting-non-auth.js
```diff
@@ -287,6 +287,24 @@ define(
...
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues/new"),
+          ".wc-ReportForm-actions-button"
+        )
+          .findByCssSelector("#steps_reproduce")
+          .click()
+          .end()
+          .sleep(500)
+          .findByCssSelector("#steps_reproduce")
```
miketaylr: After `click()`, the context is still the textarea element, so you can just do something like this:

```js
.click()
.sleep(500)
.getProperty("value")
...
```

No need to re-find the same DOM element.

webcompat/static/js/lib/bugform.js
```diff
@@ -75,13 +75,18 @@ function BugForm() {
...
+    );
     this.uploadField.on("change", _.bind(this.checkImageTypeValidity, this));
     this.osField
       .add(this.browserField)
       .on("blur input", _.bind(this.checkOptionalNonEmpty, this));
     this.submitButtons.on("click", _.bind(this.storeClickedButton, this));
     this.submitButtons.on("click", _.bind(this.loadingIndicator.show, this));
-    this.form.on("submit", _.bind(this.maybeUploadImage, this));
+    this.form.on("submit", _.bind(this.onFormSubmit, this));
+    //this.form.on("submit", _.bind(this.clearStepsToReproducePlaceholder, this));
```
miketaylr: This can be removed.

webcompat/static/js/lib/bugform.js
```diff
@@ -227,6 +232,29 @@ function BugForm() {
...
+      .split("4. Nothing happened.");
+
+    if (stepsToReproduceStrAr.length > 1 && stepsToReproduceStrAr[1] === "\n") {
+      this.stepsToReproduceField.val(function() {
+        // default case with only the placeholder text
+        return "";
+      });
+    }
+
+    if (
```
miketaylr: (Is this the output of `npm run fix`?)

webcompat/static/js/lib/bugform.js
```diff
@@ -227,6 +232,29 @@ function BugForm() {
     }
   };
 
+  this.clearStepsToReproducePlaceholder = function() {
+    var stepsToReproduceStrAr = this.stepsToReproduceField
+      .val()
+      .split("4. Nothing happened.");
+
+    if (stepsToReproduceStrAr.length > 1 && stepsToReproduceStrAr[1] === "\n") {
```
miketaylr: The more I look at this approach, the more I get confused... it's not obvious to me what the linebreak means in the 2nd position of the array unless I open up a console and mess around. 

`stepsToReproduceStrAr.length > 1 && stepsToReproduceStrAr[1].length > 1` is even more opaque.

I wonder if it would be way simple if we tried a different approach... always clear out the placeholder on focus, and only add the `details` param stuff to the end `onFormSubmit().

Or can you think of a way to do this that's going to be easier to understand when we re-visit the code in the future?

webcompat/static/js/lib/bugform.js
miketaylr: looks like it.... :/ oh well.

webcompat/static/js/lib/bugform.js
cch5ng: hi @miketaylr I basically agree with you here. I wasn't totally clear about the logic except I did trial/error logging to find out what worked. maybe it had something to do with using that URLSearchParams in the Nightly change (?)... but I tried to clean up the logic and moved the logic that adds the `details` param to the text area in the onFormSubmit(). there is just a little residual logic in the text area focus handler I needed to retain in order to handle the case where the user had previously edited the field. other changes from Friday's comments were also pushed.

miketaylr:
@cch5ng i see you've added some commits, ready for re-review?

cch5ng:
r? @miketaylr also please see
https://github.com/webcompat/webcompat.com/pull/1675#discussion_r130487797

tests/functional/reporting-non-auth.js
```diff
@@ -255,24 +253,8 @@ define(
         );
       },
 
-      // tests a scenario where bug reported from firefox nightly Report Site Issue button
-      "details param adds stylo pref to text area": function() {
```
miketaylr: Why are we removing this test?

tests/functional/reporting-non-auth.js
```diff
@@ -286,30 +268,23 @@ define(
           .end();
       },
 
-      "details param adds info to description": function() {
```
miketaylr: I don't understand why this test is being changed... we still support the `details` param and need to test that it appends info to the textarea.

tests/functional/reporting-non-auth.js
```diff
@@ -286,30 +268,23 @@ define(
...
-        this.skip(
-          "former media type reporting test - NS_ERROR_DOM_MEDIA_DEMUXER_ERR"
-        );
+      "clicking text area clears default value": function() {
```
miketaylr: Can this be a new test, rather than replacing an unrelated test?

webcompat/static/js/lib/bugform.js
```diff
@@ -227,6 +217,20 @@ function BugForm() {
     }
   };
 
+  this.clearStepsToReproducePlaceholder = function() {
```
miketaylr: I don't see anything changed here from my comment @ https://github.com/webcompat/webcompat.com/pull/1675#discussion_r130193804.

Basically we should clear this *always* on focus. `onFormSubmit` takes care of the `details` param: `this.stepsToReproduceField.val('')`

You could set a property on the BugForm object to know if the user has clicked in here here or not, and then clear it out if they haven't. If they have, just early return. Does that make sense?

webcompat/static/js/lib/bugform.js
```diff
@@ -227,6 +217,20 @@ function BugForm() {
...
+    var stepsToReproduceStrAr = this.stepsToReproduceField
+       .val()
+       .split("4. Nothing happened.");
```
miketaylr: (it's entirely possible a user types in their own reply and it happens to have a step 4 that says "Nothing happened", so we'll end up doing the wrong thing.)

tests/functional/reporting-non-auth.js
miketaylr: edit: this test probably *does* need to be modified, since it gets added at a different time. But not deleted.

cch5ng: in the review comments, you asked me to move adding the details query string to the text area at form submit time. in a different pr, you also explained that we shouldn't write tests that create an actual issue in order to avoid getting labeled as spammers by github. it seems like if I test the form submit (clicking submit buttons), this would  be creating a new github issue? so I believed there was no practical way to test the newly requested behavior(??)

miketaylr: > it seems like if I test the form submit (clicking submit buttons), this would be creating a new github issue? so I believed there was no practical way to test the newly requested behavior(??)

Since we control the entire stack, we can test it without creating new issues on GitHub. But yes, we don't want to *explicitly* test new issue creation, by creating actual new issues for the reasons you mentioned.

A few options would be to mock issue creation when the server is in test mode (via our `@mockable_response` decorator, or using the `mock` module, or possibly intercept all submit events, inspect what we need to, and cancel them via https://github.com/webcompat/webcompat.com/blob/master/tests/functional/lib/window-helpers.js.

webcompat/static/js/lib/bugform.js
cch5ng: > Basically we should clear this always on focus. onFormSubmit takes care of the details param: this.stepsToReproduceField.val('')

> You could set a property on the BugForm object to know if the user has clicked in here here or not, and then clear it out if they haven't. If they have, just early return. Does that make sense?

done.

PR #1676, opened 2017-07-25
Fixes #1662 - update whitelist for type-stylo label

### LINKED ISSUE #1662: update whitelist to get additional url parameter from Nightly bug report
related to https://bugzilla.mozilla.org/show_bug.cgi?id=1381545

after the bug above is merged, a webcompat.com bug reported from the FF Nightly menu button is  expected to include additional URL parameters.

if stylo is enabled, this string will be appended to the webcompat.com bug form url:
"&details=layout.css.servo.enabled: true&label=type-stylo" (plus relevant url escape codes)

- [x] update whitelist
- [x] verify label data is getting sent to github api
- [x] fix text area height
- [x] update functional tests
- [x] add github label for prod issues repo ("type-stylo")

?may want to add a couple functional tests
X - one similar to cur functional test for media report tool (url parameters should populate fields)
* negative case for url param (url without servo pref should leave text area as is)
* one to check the label add for new issue? (not very sure how this should be working)
  * create issue with url param and new issue should have label associated?

cch5ng:
@miketaylr r?

do you think this visual usability bug matters?
https://github.com/webcompat/webcompat.com/issues/1662#issuecomment-316862852

I verified that on submit, the issue gets saved with the appropriate github label so maybe it is not a big deal?

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -65,4 +65,5 @@ for cat_label in cat_labels:
 EXTRA_LABELS = [
     'type-media',
     'type-webvr',
+    'type-stylo'
 ]
--- a/tests/functional/reporting-non-auth.js
+++ b/tests/functional/reporting-non-auth.js
@@ -255,8 +255,24 @@ define(
         );
       },
 
+      // tests a scenario where bug reported from firefox nightly Report Site Issue button
+      "details param adds stylo pref to text area": function() {
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues/new?details=layout.css.servo.enabled%3A+true"),
+          ".wc-ReportForm-actions-button"
+        )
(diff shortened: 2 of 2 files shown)


miketaylr:
> @miketaylr r?

@cch5ng if you use the  `r? @miketaylr` my mail filters catch this more quickly, sorry for the delay here. ^_^

miketaylr:
> do you think this visual usability bug matters?

No, not really. That information is for the triagers and will be used at triage time, rather than report time. Let's land it.

miketaylr:
@cch5ng in the future, if you're making an unrelated change (the changes to tests), please include a separate commit for that. there's a number of good resources out there on "atomic commits", here's one that might be helpful: https://seesparkbox.com/foundry/atomic_commits_with_git

PR #1678, opened 2017-07-26
Revert "Fixes #1564 - Disallow inline scripts via CSP."

### LINKED ISSUE #1564: Do the work to remove CSP script-src 'unsafe-inline' directive

karlcow:
Reverts webcompat/webcompat.com#1667


miketaylr:
Unreverted in https://github.com/webcompat/webcompat.com/commit/8fadbd6463464c5120586583bef8651980e2350e :)~

Issue #1679, opened 2017-07-26
Do the work to remove 'unsafe-eval' from CSP for script-src


miketaylr:
The blocker is using lodash templates. I have a plan to do pre-compiled template functions, which should let us maybe even reduce the size of our lodash build we send over the wire (since we won't need template parsing).

PR #1682, opened 2017-07-27
Fixes: #1679. Remove unsafe-eval directive from our CSP policy.

miketaylr:
This is a *pretty big* change to how we do templates. But it's also 1000% safer than what we currently allow.

<s>Please don't merge yet, but feel free to comment.</s> ready for review/merge

It also probably needs some docs updates to mention that if you ever change a template file, you'll need to re-run `grunt:jst` (or `grunt`). edit: done


grunt-tasks/jst.js
```diff
@@ -0,0 +1,30 @@
...
+  grunt.config("jst", {
+    compile: {
+      options: {
+        namespace: "wcTmpl",
+        prettify: true,
+        processContent: function(src) {
+          // strip the opening and closing <script> tags...
+          // otherwise, the template functions will just inject script elements
+          // that won't render.
+          src = src.split(/\<script type=['"]text\/template['"]>/)[1];
```
miketaylr: string.split as a poor man's parser.... seemed nicer than actually using regexp to parse this. Alternative would be to pull in a [js-dom](https://www.npmjs.com/package/jsdom)-like package and parse the script contents and replace it with `ParentNode.firstElementChild()` or something.

miketaylr:
Better than expected!

```
15/77 tests failed (3 skipped)
```

miketaylr:
Ah yeah, forgot about a tiny inline template in the labels code :))

```
Error: call to Function() blocked by CSP

Content Security Policy: The page’s settings blocked the loading of a resource at self (“script-src http://localhost:5000 https://www.google-analytics.com https://api.github.com”). Source: call to eval() or related function blocked by CSP.
```

miketaylr:
Cool, that fixed it.

grunt-tasks/jst.js
miketaylr: Actually... I don't think that will work, with the JS stuff in the templates. The other option that would work is to wrap the .jst files in an HTML comment, parse that, and pull the text out of the comment node. But that somehow feels more yucky (and will be annoying in text editors when editing templates).

webcompat/templates/issue/issue-labels-sub.jst
```diff
@@ -0,0 +1,8 @@
+<script type="text/template">
+<% _.each(labels, function(label) { %>
+  <span class="wc-Label wc-Label--badge js-Label
+        style="background-color:#<%=label.color%>">
+    <%= label.name %>
+  </span>
+<% }); %>
+<script type="text/template">
```
miketaylr: oops

webcompat/templates/issue/issue-labels.jst
```diff
@@ -1,14 +1,14 @@
-<script type="text/template" id="issue-labels-tmpl">
+<script type="text/template">
   <header class="wc-Issue-labelEditor">
     <div class="wc-Issue-labelEditor-title">
       Labels
     </div>
-    {% if session.user_id and session.avatar_url %}
+    <% if ($("body").data("username")) { %>
```
miketaylr: Note: the python backend will return a 403 if a user isn't authenticated when trying to modify labels, so I feel pretty fine with moving this to the front-end.

webcompat/templates/web_modules/dropdown.html
```diff
@@ -1,3 +0,0 @@
-{% macro dropdown(config) %}
```
miketaylr: Unfortunately this macro had to go -- since it was included at html render time by the jinja2 template parser. Now, with pre-compiled templates, that's way too late. The good news is the macro was longer than the end result -- it's just a single line div. Not a huge loss. :)

grunt-tasks/jst.js
miketaylr: @karlcow any thoughts here?

miketaylr:
OK, tests pass and I get no new csp errors locally when messing around.

r? @zoepage 
r? @magsout

grunt-tasks/jst.js
karlcow: switch to backend rendering 😄  Wrong answer?
Honestly I'm not sure what we are trying to do in this pull request. I'm missing a bit of context, why we need to abandon jinja templates for more jsification (or more exactly how I perceive it after being two weeks into the woods 🌳 ahah)

zoepage:
@miketaylr that is really impressing. I went through the files and everything looks great on first glance including build & tests. But I'd like to take more time, to understand everything. Hope, this is ok.

miketaylr:
(oops, forgot to push docs commit)

webcompat/templates/home-page/browse-issues.html
```diff
@@ -2,16 +2,8 @@
...
-        <% if (issues.length) { %>
-          <% _.each(issues, function(issue) { %>
-            {% include "web_modules/issue-list.html" %}
-          <% }); %>
-        <% } else { %>
-        <p>Nothing here currently, cool!</p>
-      <% } %>
-      </script>
-      </div>
+      {% include "web_modules/issue-list.jst" %}
```
miketaylr: Note: we don't actually need to include this template here in the HTML, since it effectively does nothing, but I thought we should consider keeping it so it's less confusing to know what templates are used where. Or, we could change this to an HTML comment which means less work for the jinja parser.

grunt-tasks/jst.js
miketaylr: @karlcow lol. :)

(below might be useful for @zoepage to understand as well)

Basically 99% of this PR is "pre-compiling" the JS templates and sticking them in a `templates.js` file, so they don't have to be parsed and compiled on page load. 

For example, the following template, in this PR is now in its own file called `state-button.jst`. 

```
<script type="text/template">
  <span class="js-comment-<%= stateClass %>"><%= state %></span>
</script>
```

Before, or rather, currently on master, the same template lived in the HTML file inline:

https://github.com/webcompat/webcompat.com/blob/0f56b8a7aa4cfda06fa7b0fe08367e015638d5c5/webcompat/templates/issue/issue-comment-submit.html#L39-L41

The `jst` command does the pre-compiling, and produces output that looks like this, which lives in the templates.js file:

```js
this["wcTmpl"]["issue/state-button.jst"] = function(obj) {obj || (obj = {});var __t, __p = '', __e = _.escape;with (obj) {__p += '<span class="js-comment-' +((__t = ( stateClass )) == null ? '' : __t) +'">' +((__t = ( state )) == null ? '' : __t) +'</span>';}return __p};
```

So instead of parsing the client-template when we load the page and try to render by passing it data, we just pass this function the data. This also eliminates the need to allow `eval` (or `new Function('blahblahblah')`, which is the same thing, which is a potential XSS attack vector.

The potential downside of this approach is we can't rely on jinja parsing interactions inside the JS templates anymore, since the JS templates are converted into JS before the page is even served.

So there's a few benefits:

1. security -- the real motivation for this PR is to remove `unsafe-eval` from CSP
2. performance -- no more client-side parsing of JS templates, and now we get to rely on `templates.js` living in the HTTP cache.

A few downsides:

1. You need to remember to re-build if you mess with templates, that didn't exist before.
2. We can't mix jinja and JS templates, which was kind of powerful (if not maybe confusing)

As for my original question about using string.split vs regexp... I'm just gonna use regexp since the shape of the opening and closing script tags is constant in the new templates (they don't need ids).

miketaylr:
Just so it's not lost, here's an explanation of what this PR is trying to achieve:

https://github.com/webcompat/webcompat.com/pull/1682#discussion_r130159087

webcompat/static/js/lib/labels.js
```diff
@@ -35,18 +35,10 @@ issues.LabelsView = Backbone.View.extend({
...
-    [
-      "<% _.each(labels, function(label) { %>",
-      '<span class="wc-Label wc-Label--badge js-Label" style="background-color:#<%=label.color%>">',
-      "<%= label.name %>",
-      "</span>",
-      "<% }); %>"
-    ].join("")
-  ),
+  // relavant parts in issue/issue-labels.jst
+  subTemplate: wcTmpl["issue/issue-lables-sub.jst"],
```
miketaylr: oops, typo! should be `issue-labels-sub.jst`. will fix.

zoepage:
@miketaylr The performance improvement is really impressive! :)


I found a few small issues: 

#### CSP issue:
```
Content Security Policy: The page’s settings blocked the loading of a resource at self (“script-src http://localhost:5000 https://www.google-analytics.com https://api.github.com”). Source: onfocusin attribute on DIV element.
```

#### Labels:
The placeholder for label is shown. When a label is added, it's not shown until a reload is fired manually. (See screenshots)

screen shot 2017-07-31 at 2 36 47 pm

screen shot 2017-07-31 at 2 37 04 pm

#### Anchor
http://localhost:5000/#contribute anchor directs not to the correct place as the issues are loaded later (from the all issues list on the index).

miketaylr:
@zoepage thanks for digging around and finding those issues! That `onfocusin` one I'm 99% sure is just your adblocker or lastpass or similar. If you try it in a new profile that error won't be there. Will take a closer look at the rest.

miketaylr:
OK, fixed the label editor issue (which also uncovered another small bug) and filed a follow up for the #contribute link (because that also exists in production): https://github.com/webcompat/webcompat.com/issues/1688

Just need to fix this package.json conflict.

zoepage:
Sweet! Thank you :)

miketaylr:
I just deployed this branch to staging if anyone wants to poke at it some more. Otherwise, will probably merge tomorrow.

grunt-tasks/jst.js
karlcow: @miketaylr thanks for the explanation. regexp vs string.split. If the pattern is regular aka no variation I would have use a split. You could microbenchmark it too :) not sure if it's meaningful performance gain. 

oh it already exists. 
https://jsperf.com/regex-vs-split/2
and https://jsperf.com/slice-vs-substr-vs-substring-vs-split-vs-regexp/2

zoepage:
On stage (after posting an issue). Not just google analytics gets blocked but also the github api?

Things work, but just wanted to make you aware of this, IF this is an issue. :)

screen shot 2017-08-01 at 11 36 16 am

Also the function for the flash message gets blocked, if I saw that right?

miketaylr:
> Things work, but just wanted to make you aware of this, IF this is an issue. :)

Investigating...

miketaylr:
> Investigating...

Yeah oops. Forgot about one last inline template. Fixed now!

miketaylr:
💣s away... thanks for review @karlcow @zoepage!

Issue #1658, opened 2017-07-13
Create an editor to edit images before uploading


miketaylr:
(I can't remember if there's already an issue for this... searched around and found nothing)

It would be good to have the ability to do the following to an image, before uploading:

1. highlight in yellow (paintbrush or rectangle tool)
2. black out an area (paintbrush or rectangle tool)
3. blur an area (rectangle)

This is all doable with modern canvas API support, but the blurring will need a polyfill written (possibly running in a web worker) for Safari and others.

(update) tasks breakdown

- [x] update demo (current canvas code stub) so lines are drawn on drag event (before the mouse up which draws entire rectangle)
- [x] decide where image editor code should go (bugform.js or new file); going with bugform.js for testing ease for now (but may change in future); in case it gets unruly adding in comments like // image editor props to be able to move stuff to another file
- [x] add logic to create a popup/modal (image editor)... need Edit btn (click handler and tracking Edit state)
- [x] add Editor control buttons (remove cur placeholders outside of canvas and create html/css buttons); hook buttons up to existing click event handlers
- [x] integrate cur demo code (canvas stub) into webcompat code
- [x] get screenshot img to show in canvas background
- [x] add Exit/close control button (update state to non Edit mode); hook up handler
- [x] update logic to save final canvas image on form post
- [x] fix screenshot and canvas resizing (half working, when I can get the original image dimensions, the calculations are working but currently having a challenge with getting small image dimensions, ie screenshots or non-complex photos)
- [x] follow up screenshot/canvas resizing; couple of cases: 1 orig screenshot width/height somehow unavailable, then a default canvas width/height should be set and img doesn't need resizing; 2 orig screenshot is much smaller than the canvas width; then screenshot doesn't need resizing
- [x] fix controls (figure out better buttons/design); 
* X- add css for hover/active styles (use cur submit button styles as reference)
* X- toolbar buttons for Image Editor (Highlight btn, Redact btn); 
* X- Form File Input > filled state (Edit btn); 
* X- need apply flexbox to canvas and toolbar area (in place of float right)
- [x] shape drawing (black rectangle) on drag
- [ ] fix line drawing (yellow rectangle) on drag; currently working but not perfect
- [x] internal issues (bugs tracked in the prd currently)
* X - handling ending mouseup position on a smaller shape size than originally drawn during drag motions
* X - canvas placement looks different on FF vs Chrome
[...]

PR #1683, opened 2017-07-28
[not ready to merge] rough prototype image editor

cch5ng:
related to #1658 

right now the functionality covers the checkbox goals (first comment in issue)
r? @miketaylr I'm not totally sure it's ready for code review but do you have any initial feedback? it is still pretty rough...

(is it missing any big picture functionality? does the line drawing part behave as you'd expect?)

main issues with dynamic canvas size (in relation to screenshot dimensions and screen width)

the gui is rough and buttons need to be cleaned up, etc

(update) I'm tracking current known issues within the issue (#1658)
(072817) today's issues tracked in doc:
https://docs.google.com/document/d/11KZICiXwfU8znYtkDzK3y7MFYyEznWKKQXzknLwrb1k/edit#heading=h.lazkmpwrhtsq

### DIFF
--- a/webcompat/static/css/development/components/form.css
+++ b/webcompat/static/css/development/components/form.css
@@ -98,7 +98,11 @@
 }
 
 /* state */
-.wc-Form-item.is-hidden {
+.is-hidden {
+/*.wc-Form-item.is-hidden,
+.wc-Image-Editor-edit-btn.is-hidden,
+.modal.is-hidden {
+*/  
   display: none;
 }
 
@@ -172,4 +176,121 @@ input[type = "radio"] { /* stylelint-disable selector-no-attribute */
 .wc-Form-group.is-error .wc-Form-item:focus {
   border-color: var(--base-variantColorDark);
   box-shadow: none;
+}
+
+/* Image Editor section */
+.wc-Icon--edit-img::before {
+  content: "\270e";
+}
+
+.modal {
+    position: fixed;
+    z-index: 20;
+    left: 0;
+    top: 0;
+    width: 100%;
+    height: 100%;
+    overflow: auto;
+    background-color: var(--base-colorDarkest);
+    background-color: var(--base-variantColorExtraDarkOpacity);
+}
+
+/* Modal Content/Box */
+.modal-content {
+    display: flex;
+    flex-flow: row wrap;
+    background-color: var(--base-variantColorMedLight);
+    margin: 15% auto;
+    border: 1px solid var(--base-variantColorMed);
+    border-radius: 8px;
+    width: 80%;
+}
+
+/* The Close Button */
+.close {
+    color: var(--base-variantColorLight);
+    font-size: 28px;
+    font-weight: bold;
+}
+
+.close:hover,
+.close:focus {
+    color: var(--base-colorDarkest);
+    text-decoration: none;
+    cursor: pointer;
+}
+
+.wc-Image-Editor-header {
+  display: flex;
+  justify-content: flex-end;
+  width: 100%;
+  padding: 0 12px;
+}
+
+.wc-Image-Editor-toolbar {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  width: 100%;
+  padding: 0 12px;
+  margin-bottom: 5px;
+}
+
+/* Image Editor buttons */
+.wc-Image-Editor-edit-btn {
+  margin: 3px;
+  padding-left: 5px;
+  width: 28px;
+  height: 28px;
+  border: 1px solid var(--base-variantColorLight);
+  border-radius: 5px;
+}
+
+.wc-Image-Editor-edit-btn:hover {
+  background-color: var(--base-variantColorLight);
+}
+
+.wc-Image-Editor-button {
+  width: 40px;
+  height: 30px;
+  background-color: var(--base-variantColorMed);
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  align-items: center;
+  border-radius: 5px;
+}
+
+.wc-Image-Editor-button:nth-of-type(1) {
+  margin-right: 3px;
+}
+
+.wc-Image-Editor-button:nth-of-type(2) {
+  margin-left: 3px;
+}
+
+.wc-Image-Editor-button:hover {
+  background-color: var(--base-variantColorLight);
+}
+
+.wc-Image-Editor-button:active,
+.wc-Image-Editor-button.is-active {
+  border-color: var(--base-stateSitewait);
+  border: 1px solid var(--base-stateSitewait);
+  box-shadow: 0 1px 1px var(--base-activeBoxShadow) inset, 0 0 3px var(--base-activeBoxShadow2);
+}
+
+.wc-Image-Editor-yellow-rect {
+  width: 30px;
+  height: 20px;
+  border: 2px solid var(--base-highlightBtnCenter);
+  margin: 3px auto;
+  padding-top: 3px;
+}
+
+.wc-Image-Editor-black-rect {
+  width: 30px;
+  height: 20px;
+  background-color: var(--base-colorDarkest);
+  margin: 5px auto;
 }
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+//var doc = global.document;
 function BugForm() {
   this.form = $("#js-ReportForm form");
   this.submitButtons = $("#js-ReportForm .js-Button");
@@ -11,6 +12,18 @@ function BugForm() {
   this.UPLOAD_LIMIT = 1024 * 1024 * 4;
   this.clickedButton = null;
   this.hasImage = null;
+  // image editor props
+  this.editImageButton = $(".js-edit-img-open-btn");
+  this.editImageModal = $(".modal");
+  this.editImageCloseButton = $(".js-edit-img-close-btn");
+  this.editImageHighlightButton = $(".js-highlight-button");
+  this.editImageRedactButton = $(".js-redact-button");
+  this.canvas = $("#mCanvas");
+  this.curEditorStateOn = true;
+  this.curEditorActiveControl = "highlight";
+  this.initMouseDownPos = {};
+  this.mouseClickDragActive = false;
+  this.mouseDown = false;
 
   this.inputs = {
     url: {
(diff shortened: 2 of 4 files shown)


webcompat/static/js/lib/bugform.js
```diff
@@ -573,6 +605,136 @@ function BugForm() {
...
+    canvas.height = 500;
+    img.src = this.dataURI;
+
+    var modalContent = document.querySelector('.modal-content');
+
+    this.editImageModal.removeClass("is-hidden");
+    modalContent.appendChild(canvas);
+
+    // TODO ask Mike if there is a better way to share this globally, like this.ctx
+    globalCtx = ctx;
```
cch5ng: @miketaylr I kind of understand global vars are not good but this ctx needs to be shared. should I be using something like this.ctx instead?

webcompat/static/js/lib/bugform.js
```diff
@@ -573,6 +605,136 @@ function BugForm() {
...
+    globalCtx.drawImage(img, 0, 0);
+
+    // TODO ask Mike if there is a better way
+    // having to define the canvas event listeners here because at time
+    // of init definition, the canvas does not exist
+    var domCanvas = $("#mCanvas");
+    domCanvas.on("mousedown", _.bind(this.canvasMouseDown, this));
```
cch5ng: @miketaylr for this group of event listeners, originally I tried to place them in the init() but the canvas doesn't get written until this function so I had to move the listeners down here. is this an ok way to handle that?

webcompat/static/js/lib/bugform.js
```diff
@@ -573,6 +605,136 @@ function BugForm() {
...
+    img.src = this.dataURI;
+
+    var modalContent = document.querySelector('.modal-content');
+
+    this.editImageModal.removeClass("is-hidden");
+    modalContent.appendChild(canvas);
+
+    // TODO ask Mike if there is a better way to share this globally, like this.ctx
+    globalCtx = ctx;
+    globalCtx.drawImage(img, 0, 0);
```
miketaylr: Depending on how big the dataURI is, this might race. To be safe, you should consider doing this inside a `img.onload` handler.

webcompat/static/js/lib/bugform.js
miketaylr: Yeah, we don't want this to be global. Why not just pass the `ctx` as an argument to your mouse handlers?

webcompat/static/js/lib/bugform.js
```diff
@@ -573,6 +605,136 @@ function BugForm() {
...
+
+    // TODO ask Mike if there is a better way
+    // having to define the canvas event listeners here because at time
+    // of init definition, the canvas does not exist
```
miketaylr: It's fine to add handlers here, it doesn't have to be on init. Just before we expect to do anything useful with them.

webcompat/static/js/lib/bugform.js
```diff
@@ -573,6 +605,136 @@ function BugForm() {
...
+    var domCanvas = $("#mCanvas");
+    domCanvas.on("mousedown", _.bind(this.canvasMouseDown, this));
+    domCanvas.on("mousemove", _.bind(this.canvasMouseDrag, this));
+    domCanvas.on("mouseup", _.bind(this.canvasMouseUp, this));
+
+  }
+
+  this.closeImageEditor = function() {
+    var myCanvas = document.querySelector('canvas');
```
miketaylr: nit: I'm not a fan of `myFoo` variables names. It's OK to just call it `canvas`.

webcompat/static/js/lib/bugform.js
```diff
@@ -11,6 +14,21 @@ function BugForm() {
...
   this.hasImage = null;
+  // image editor props
+  this.editImageButton = $(".js-edit-img-open-btn");
+  this.editImageModal = $("#imageEditor");
+  this.editImageCloseButton = $(".js-edit-img-close-btn");
+  this.editImageHighlightButton = $(".js-highlight-button");
+  this.editImageRedactButton = $(".js-redact-button");
+  this.canvas = $("#mCanvas");
+  this.curEditorStateOn = true;
+  this.ACTIVE_CONTROL_OPTIONS = [
```
miketaylr: Using this array throughout the code feels a bit awkward -- I'm not sure what it really buys us (and later on reading `this.ACTIVE_CONTROL_OPTIONS[0]` is harder to understand than something like `this.HIGHLIGHT_CONTROL` (or something)).

webcompat/static/js/lib/bugform.js
```diff
@@ -573,6 +605,136 @@ function BugForm() {
     });
   };
 
+  // functions for Image Editor
+  this.openImageEditor = function() {
+    //var doc = global.document;
+    var canvas = document.createElement('canvas');
```
miketaylr: At some point we need to destroy this `canvas` element, otherwise we'll end up just making more elements if you open and close the editor multiple times.

webcompat/static/js/lib/bugform.js
```diff
@@ -573,6 +605,136 @@ function BugForm() {
...
+    // having to define the canvas event listeners here because at time
+    // of init definition, the canvas does not exist
+    var domCanvas = $("#mCanvas");
+    domCanvas.on("mousedown", _.bind(this.canvasMouseDown, this));
+    domCanvas.on("mousemove", _.bind(this.canvasMouseDrag, this));
+    domCanvas.on("mouseup", _.bind(this.canvasMouseUp, this));
+
+  }
+
+  this.closeImageEditor = function() {
```
miketaylr: I suppose we want to nuke the canvas element here.

webcompat/static/js/lib/bugform.js
```diff
@@ -11,6 +14,21 @@ function BugForm() {
   this.UPLOAD_LIMIT = 1024 * 1024 * 4;
   this.clickedButton = null;
   this.hasImage = null;
+  // image editor props
+  this.editImageButton = $(".js-edit-img-open-btn");
+  this.editImageModal = $("#imageEditor");
```
miketaylr: Why not just `$('.modal')? Then you can get rid of the id from the element as well.

webcompat/static/js/lib/bugform.js
```diff
@@ -573,6 +605,136 @@ function BugForm() {
...
+  this.setImageEditorHighlight = function() {
+    this.curEditorActiveControl = this.ACTIVE_CONTROL_OPTIONS[0];
+  }
+
+  this.setImageEditorRedact = function() {
+    this.curEditorActiveControl = this.ACTIVE_CONTROL_OPTIONS[1];
+  }
+
+  this.canvasMouseDown = function(ev) {
+    var myCanvas = document.querySelector('canvas');
```
miketaylr: You should be able to get the canvas element from the `ev` object, rather than needing to query the DOM again. Same for other places.

webcompat/static/js/lib/bugform.js
```diff
@@ -573,6 +605,136 @@ function BugForm() {
...
+    //}
+
+  }
+
+  this.canvasMouseUp = function(ev) {
+    var myCanvas = document.querySelector('canvas');
+    var mPos = this.getMousePos(myCanvas, ev);
+
+    // only do this if event was not mouseup on control button
+    if (mPos.y >= 160) {
```
miketaylr: I'm not sure I understand this. What's 160?

webcompat/templates/home-page/form.html
```diff
@@ -159,5 +160,18 @@ <h3 class="wc-Title--s">Report Site Issue</h3>
...
       {{ form.hidden_tag() }}
     </form>
+  <!-- The Modal -->
+  <div id="imageEditor" class="modal is-hidden">
+
+    <!-- Modal content -->
+    <div class="modal-content">
+      <span class="close js-edit-img-close-btn">&times;</span>
+      <div class="wc-Image-Editor-toolbar">
+        <div class="js-highlight-button">Highlight Rect</div>
```
miketaylr: (I assume we'll want small icons rather than text... maybe)

cch5ng: @miketaylr agreed. I did some sketchy wireframes in the am.

https://docs.google.com/document/d/11KZICiXwfU8znYtkDzK3y7MFYyEznWKKQXzknLwrb1k/edit?usp=sharing

webcompat/static/js/lib/bugform.js
cch5ng: @miketaylr thanks, yes I was getting a bug as you described before that change.

webcompat/templates/home-page/form.html
miketaylr: Looks great!

screen shot 2017-07-28 at 4 57 24 pm

webcompat/static/css/development/components/form.css
```diff
@@ -98,7 +98,9 @@
 }
 
 /* state */
-.wc-Form-item.is-hidden {
+.wc-Form-item.is-hidden,
+.js-edit-img-open-btn.is-hidden,
```
cch5ng: for the .is-hidden class, I agree with you the selectors should be simpler but the current pattern across the css is that it is combined with other selectors. do you mind if I update here (form.css) and open a separate issue to fix the other selectors in remaining *.css?

webcompat/static/js/lib/bugform.js
cch5ng: my bad, this was something I should have deleted from the original demo code... done.

webcompat/static/js/lib/bugform.js
```diff
@@ -417,6 +444,8 @@ function BugForm() {
     var reader = new FileReader();
     reader.onload = _.bind(function(event) {
       var dataURI = event.target.result;
+      // storing URI to use for Image Editor background
+      that.dataURI = dataURI;
```
cch5ng: ended up using `img.src = this.previewEl.get(0).style.backgroundImage.slice(5, -2);` in the code to display the Image Editor modal. let me know if I still need to do clean up on the long value tho...

webcompat/static/js/lib/bugform.js
```diff
@@ -573,6 +599,179 @@ function BugForm() {
...
+  this.canvasMouseDown = function(ev) {
+    var canvas = ev.target;
+    var mPos = this.getMousePos(canvas, ev);
+    var canvasWidth = canvas.offsetWidth + 1;
+    var canvasHeight = canvas.offsetHeight + 1;
+
+    this.initMouseDownPos = mPos;
+    this.mouseDown = true;
+    // used to store image version before shapes drawn during drag events
+    this.canvasImageData = ev.data.ctx.getImageData(0, 0, canvasWidth, canvasHeight);
```
cch5ng: @miketaylr this seems like a good way to get a snapshot of the image before all the on drag shape drawing occurs. should I be concerned with the size of the data and memory for it?

miketaylr: nah, this looks good!

PR #1691, opened 2017-08-01
Fixes #1641 - check if response is defined

### LINKED ISSUE #1641: TypeError: response.responseJSON is undefined
In issues.js:

```
if (response.responseJSON.message === "API call. Not Found") {
          location.href = "/404";
          return;
        }
```

We need to check that the `responseJSON` property exists before trying to grab its `message` property.

`if (foo && foo.bar && foo.bar.baz === "lol") {}`

zoepage:
r? @miketaylr


zoepage:
Fixing that lint issue (which was fixed locally...?)

miketaylr:
Thank you!

Issue #1696, opened 2017-08-02
The Labels' Save & Close button is not working 


softvision-sergiulogigan:
**Environment:**
Browser / Version: Chrome 59.0.3071.115 (Desktop), Firefox 54.0.1 Release (Desktop)
Operating System: Windows 10

**Steps to Reproduce:**
1. Navigate to: (eg.)https://webcompat.com/issues/8621
2. Click on the Gear icon, on the right.
3. Select some labels.
4. CLick the Save & Close button.

**Expected Behavior:**
The Labels list is closed.

**Actual Behavior:**
Nothing happens.

**Notes:**
1. The selected labels are applied.

sv;

softvision-sergiulogigan:
Might be related to #1680

miketaylr:
@sergiorojasa i probably broke this! as a workaround, either hit `esc` or click outside the label editor and it will save label edits. Will fix today. 🙈

gvanwaelvelde:
This issue still exists for me in Firefox 54, Firefox 57, Google Chrome 59 and Google Chrome Canary 62 on Android 6.0.1

PR #1698, opened 2017-08-02
Issue #1696. Unbreak Save & close label editor button.

miketaylr:
Totally my bad... I removed this fix from Bea's original code. But now we have tests... :)

r? @brizental


miketaylr:
Thanks!

PR #1700, opened 2017-08-03
Fixes #1647 - split font short form for webkit

### LINKED ISSUE #1647: Some icons have disappeared
Some icons have disappeared: for example, Home, log in (header)

zoepage:
r? @miketaylr @magsout

### DIFF
--- a/webcompat/static/css/development/components/icon-font.css
+++ b/webcompat/static/css/development/components/icon-font.css
@@ -13,8 +13,16 @@
 .wc-Icon {
   display: inline-block;
   vertical-align: middle;
-  font: normal normal 400 normal 1em/1 webcompat;
+
+  /* to prevent a webcompat regression issue in Safari 10.1.1., we do not use the short form
+  https://bugs.webkit.org/show_bug.cgi?id=165594 */
+  font-style: normal;
+  font-variant: normal;
+  font-weight: 400;
+  font-stretch: normal;
+  font-size: 1em;
   line-height: 1;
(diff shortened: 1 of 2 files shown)


webcompat/static/css/development/components/icon-font.css
```diff
@@ -13,8 +13,16 @@
 .wc-Icon {
   display: inline-block;
   vertical-align: middle;
-  font: normal normal 400 normal 1em/1 webcompat;
+
+  /* to prevent a webcompat issue, we do not use the short for font form here right now
+  webkit has a css2 implementation in 10.1.1. regarding to that spec */
```
miketaylr: What is the actual issue? `webkit has a css2 implementation` is a confusing comment -- is it a bug in a particular part of css3 shorthand value parsing? Or something else?

miketaylr: (is this the bug? https://bugs.webkit.org/show_bug.cgi?id=165594)

zoepage:
Yes, it is that one

webcompat/static/css/development/components/icon-font.css
miketaylr: @zoepage wrote:
> Yes, it is that one

OK, can we update the comment then. It's not accurate to say Safari implements CSS2 for `font`  -- it's just a regression that we need to work around. Thanks!

zoepage:
Will do. Thanks

miketaylr:
👍

PR #1703, opened 2017-08-03
Fixes #1701 - Change default-src policy to self and set object-src to none

### LINKED ISSUE #1701: webcomat issue with CSP LVL3 in webkit
Follow up for https://github.com/webcompat/webcompat.com/issues/1647#issuecomment-319665648

@miketaylr 
In https://github.com/webcompat/webcompat.com/blob/master/webcompat/helpers.py#L496
`manifest-src` is not supported in webkit as it has a partial implementation of CSP LVL3: https://webkit.org/status/#?search=csp

The `manifest-src` is not supported and jumps to `default-src`, which is set to `none`. 
https://w3c.github.io/webappsec-csp/#directive-manifest-src

When we change the `default-src` to `self` anything else without an explicit directive is set on self too. So this is why we set `object-src`to none to prevent that issue. This also seems like a "best practice" thing. 

In addition, because it's a handy tool && for documentation purpose: https://csp-evaluator.withgoogle.com/

zoepage:


zoepage:
@miketaylr Changed the message (hope I didn't mess up anything)

zoepage:
r? @miketaylr 

(or did I miss something?)

miketaylr:
> (or did I miss something?)

Nah, I just missed the ping. Looks like you could rebase, but I'll just squash as I merge in. Thanks!

zoepage:
Thanks!

PR #1710, opened 2017-08-11
Fixes #1665 - Gets tests to pass with PRs from forks.

### LINKED ISSUE #1665: Get tests to pass with PRs from forks.

brizental:
r? @miketaylr 

I removed the screenshot capability because it was making it harder to read the Travis console. We can put it back after this, but I didn't see much use to it in the first place. What do you think, @miketaylr ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,14 +18,9 @@ cache:
 git:
   depth: 5
 
-addons:
-  firefox: "50.1.0"
-
 env:
   global:
     - ISSUES_REPO_URI=webcompat/webcompat-tests/issues
-    - FAKE_ID=e80a666fbf1fa9ea18db
-    - FAKE_SECRET=688c4546f09624f8c44773b22268064dfca19a59
     # grab latest 4.x release, which is LTS for now
     - TRAVIS_NODE_VERSION="4"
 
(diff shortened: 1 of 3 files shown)


miketaylr:
> I removed the screenshot capability because it was making it harder to read the Travis console.

yeah, +1. not as cool as i thought it would be.

.travis.yml
```diff
@@ -18,14 +18,9 @@ cache:
 git:
   depth: 5
 
-addons:
```
miketaylr: (note to self, we don't need this because we tell intern what to use in its config)

run.py
```diff
@@ -108,6 +106,13 @@ def config_validator():
                         help='backup existing issues.db.')
     args = parser.parse_args()
 
+    if not args.testmode:
+        # testing the config file.
+        # this file is only important in non-test mode.
+        # in test mode everything must be mocked, 
+        # so there is no external api communication.
+        config_validator()
```
miketaylr: 💯

tests/functional/lib/helpers.js
```diff
@@ -20,15 +20,6 @@ define(
       return params ? base + params : base;
     };
 
-    function takeScreenshot() {
```
miketaylr: (In the future, since this change is unrelated to getting forks to pass tests, it's OK to include a separate commit for it.)

brizental: noted!

PR #1713, opened 2017-08-13
Fixes 1650. the Report Anonymously button should have gray border

rheaparekh:


miketaylr:
Thanks for the patch @rheaparekh!

r? @magsout

miketaylr:
Guillaume is swamped with work after his holidays -- thanks for being patient! r+.

PR #1706, opened 2017-08-08
Fixes #1704 - Fixes Safari testing failures.

brizental:
r? @miketaylr 

This is not ready for merge, it's just an example of the way I found out to fix Safari testing issues.

So, I realized that most of the Safari issues that I have been getting are Promises that don't finish in time (maybe all of them are, but I'm not yet sure). So I just applied the method I thought about when trying to fix #1470 . I think maybe this has something to do with the way Safari handles promises, and since Intern.js is all promises, that could be an issue. I'll look into it and edit here later.

It works, but the code gets kinda confusing. It's what I was talking about when I mentioned callback hell in #1470 lol.

PR #1716, opened 2017-08-15
Fixes #1704 - Exemplifies fix for Safari errors using async / await.

### LINKED ISSUE #1704: Get funcional tests to run in Safari and Edge locally.
So, since we are going to test these two browsers in Browserstack, I thought it would be good to be able to also test them locally.

Now, it wouldn't be really nice to have 3+ browsers run every time we test something. So, to address that issue, I thought about adding an argument where we list all browsers we want to test with when testing. Something like `browsers=firefox, chrome`.

For that we just need to get tests to run in Safari and Edge, the status right now for Safari is:

**Saf 12603 : [✓✓✓×~~✓✓✓✓××××××××✓×✓✓✓×××✓×✓××✓××××××××] 79/79, 37 fail, 3 skip**

Will add Edge here as soon as I test it, but from what I remember in testing with Browserstack, there are some errors, not as much as Safari though.

I'm ignoring IE for now because I suppose not many people from our user base are still using it. But I will confirm that info in Google Analytics.

brizental:
So, this was kind of a fail.

I thought using `async / await` would fix intermittent errors, but the test for `Posting a comment` still gives an error every now and then. I thought of two possibilities:

1. `await` is not working. But from what I've tested it is, maybe I'm testing wrong, idk.
2. Leadfoot functions just fail for no reason sometimes.

Not very optimistic, but yeah, lol.

Also the `Add a screenshot to a comment` test is not working for Safari. It gives back an `ElementNotVisible` error. 

Now, that I think might be a Leadfoot bug, because Leadfoot claims to be a library that fixes cross-browser inconsistencies for the SeleniumWebDriver. And this test works for Firefox and Chrome, so it should work for Safari I guess.

What makes me unsure about that, though, is the fact that in my previous PR for this issue (#1706) this test worked. 

Anyways, thoughts @miketaylr ?


tests/functional/lib/helpers.js
```diff
@@ -4,135 +4,49 @@
...
-            return null;
-          }
-          elem = elem[0];
-          return elem.offsetWidth > 0 && elem.offsetHeight > 0 ? true : null;
-        },
-        [selector],
-        10000
-      );
-    }
+  async function login(context) {
```
miketaylr: It seems like this login method doesn't actually do same stuff that the original one does, right?

(fill out forms, look for occasional verification page...)

brizental: Ah, yes! 

I removed that just to make the transition faster for me, since this function was not the one that I actually wanted to test. 

But now I'm thinking, since we have the mock login in place is that really needed?

miketaylr: Hmm..... probably not needed/used right now, but I wonder if it might be useful if we're trying to test w/o mock data? Is that even useful I wonder...

brizental: I can't think of a situation where that would be needed.  But it might happen. 

Do you think I should bring it back?

brizental:
Closing this because it doesn't work.

Issue #537, opened 2015-01-15
[testing] login to github will redirect to a different URI


karlcow:
Step to reproduce.

With a clean profile (not logged to github, not logged to webcompat.com)

1. Starts your local instance `python run.py`
2. Start a clean profile browser
3. Enter `http://127.0.0.1:5000/`
4. click on login
5. The user is redirected to github Web site.
6. login/password. Enter
7. The user is redirected to `http://localhost:5000/`
8. fix code.
9. restart the server.
10. Reenter `http://127.0.0.1:5000/`
11. Try to login again.

It will not work a callback is happening. There's a 200 OK in the log but nothing else happened. The UI is not changed to logout and the avatar icon.

What I suspect is that `http://127.0.0.1:5000/` != `http://localhost:5000/`

Suggestions:
* Either having a way to redirect to the right localhost or 127.0.0.1 when started on one of these.
* Either having the starting message of `python run.py` be localhost.

miketaylr:
We probably just need to document `localhost` instead of 127.0.0.1. GitHub requires us to put in the OAuth callback location "Authorization callback URL", which is set to http://localhost:5000/callback for our test application. But people could create their own test applications and use 127.0.0.1 if they want. It just needs to be consistent @ https://github.com/webcompat/webcompat.com/blob/master/config.py.example#L66 and in the application settings.

But maybe just telling people to use localhost is simplest. :)

karlcow:
Yes probably.

PR #1720, opened 2017-08-17
Fixes #1719 - Removes unused issues.db from the code

### LINKED ISSUE #1719: Remove (or use?) issues.db
It doesn't do anything useful, right? Or would it be useful for RSS feeds @karlcow?

karlcow:
- It also cleans up a bit the syntax. Preparing for linting and future pep stuff.
- It changes a bit the logic on run.py. It became obvious that we were not checking dependencies when in test mode (ooops)

@miketaylr r?

### DIFF
--- a/run.py
+++ b/run.py
@@ -6,12 +6,13 @@
 
 import argparse
 import pkg_resources
-from pkg_resources import DistributionNotFound, VersionConflict
 from sqlalchemy import create_engine
 from sqlalchemy.orm import scoped_session
 from sqlalchemy.orm import sessionmaker
 import time
 import os
+from pkg_resources import DistributionNotFound
+from pkg_resources import VersionConflict
 import sys
 
 IMPORT_ERROR = '''
@@ -35,7 +36,7 @@ https://github.com/webcompat/webcompat.com/blob/master/CONTRIBUTING.md#configuri
 
 try:
     from webcompat import app
-except ImportError, e:
+except ImportError as e:
     if 'import_name' in e and e.import_name == 'config':
         # config not found, did you forget to copy config.py.example?
         raise ImportError('{0}\n\n{1}'.format(e, NO_CONF_ERROR))
--- a/run.py
+++ b/run.py
@@ -6,11 +6,6 @@
 
 import argparse
 import pkg_resources
-from sqlalchemy import create_engine
-from sqlalchemy.orm import scoped_session
-from sqlalchemy.orm import sessionmaker
-import time
-import os
 from pkg_resources import DistributionNotFound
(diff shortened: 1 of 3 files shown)


webcompat/db/__init__.py
```diff
@@ -4,15 +4,18 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+"""Databases initialization."""
+
+from hashlib import sha512
```
miketaylr: Any reason this import is above `import os`?

run.py
```diff
@@ -102,54 +98,20 @@ def config_validator():
...
-            for old_file in backup_files[:-3]:
-                os.remove(app.config['BACKUP_DEFAULT_DEST'] + old_file)
+    if check_pip_deps():
+        if not args.testmode:
+            # testing the config file.
+            # this file is only important in non-test mode.
+            # in test mode everything must be mocked,
+            # so there is no external api communication.
+            config_validator()
+            app.run(host='localhost')
```
miketaylr: (this isn't related to this review... but I was trying to remember why we made this change. I can't recall... it was something related to tests failing maybe? Like, if you used 127.0.0.1 instead of localhost? All I found in history was the commit message @ https://github.com/webcompat/webcompat.com/commit/feddc52c66c60a9f6b1baf164e495405613caa34....)

webcompat/db/__init__.py
karlcow: because I'm a 🤡 
I need to fix this. Thanks.

run.py
karlcow: @mike This #537

karlcow: @miketaylr this #537

miketaylr: Thank you! Unsure why my searching-fu was so bad (promise I looked...).

webcompat/db/__init__.py
karlcow: @miketaylr  oh wait…
there's no mistake here 👺 :)

```
from hashlib import sha512
import os
from uuid import uuid4
```
h -> o -> u. :)

karlcow:
@ miketaylr It's mergeable. :)

karlcow:
@miketaylr It's mergeable. :)

karlcow:
(merged myself because @miketaylr is traveling and approved.)

webcompat/db/__init__.py
miketaylr: Ah, OK. I thought we grouped the "bare imports" (where you imported the entire module), then method imports.

PR #1722, opened 2017-08-18
Fixes #1581- [optimization] change urlparse to urlsplit

### LINKED ISSUE #1581: [optimization] change urlparse to urlsplit
In [form.py ](https://github.com/webcompat/webcompat.com/blob/master/webcompat/form.py), there is the function `domain_name` which extracts the domain name. It uses `urlparse.urlparse()` which is less performant than `urlparse.urlsplit()`.

```python
def domain_name(url):
    '''Extract the domain name of a sanitized version of the submitted URL.'''
    # Removing leading spaces
    url = url.lstrip()
    # testing if it's an http URL
    if url.startswith(SCHEMES):
        domain = urlparse.urlparse(url).netloc
    else:
        domain = None
    return domain
```

rheaparekh:
r @karlcow?

### DIFF
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -161,7 +161,7 @@ def domain_name(url):
     url = url.lstrip()
     # testing if it's an http URL
     if url.startswith(SCHEMES):
-        domain = urlparse.urlparse(url).netloc
+        domain = urlparse.urlsplit(url).netloc
     else:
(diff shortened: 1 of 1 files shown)


karlcow:
Super! Thanks @rheaparekh perfect contribution. 💯

PR #1725, opened 2017-08-24
Fixes #1586 - Adds a data/ path logic to the code.

### LINKED ISSUE #1586: [proposal] Adds a data/ directory under .gitignore
Should we have a `data/` directory where we stash things like

* `issues.db`
* `sessions.db`
* Any other data useful for running webcompat.com but not part of the commits. (I'm thinking about the cache type-media list).

(Updated on August 17, 2017)

* [x] Create a `data/` directory
* [x] Add a `DATA_PATH` in the config file
* [x] Check in the code where we rely on `session.db` and `topsites*.db` adjust the paths  (@MDTsai)
* [x] Add `data/` to `.gitignore`
* [ ] Move `topsites*.db` and `session.db` (erase probably)

karlcow:
The data/ is meant to be ignored per #1586.

I modified the code to make sure that the db could live in the new directory.

@miketaylr r?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,9 +13,8 @@ uploads/
 config/secrets.py
 package-lock.json
 
-# backup folder contains the issues.db backup file with issues dump
-# for the previous versions
-backups/*.db
+# The data folder contains information that shouldn't live in version control.
+data/*
 
 # these are our 'dist' files
 # checking them into the repo is silly
--- a/config/environment.py
+++ b/config/environment.py
@@ -9,6 +9,10 @@ import os
 
 # Define the application base directory
 BASE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))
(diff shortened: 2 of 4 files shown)


karlcow:
```
→ nosetests
...............................................................
Ran 63 tests in 0.945s

OK
```

.gitignore
```diff
@@ -13,9 +13,8 @@ uploads/
 config/secrets.py
 package-lock.json
 
-# backup folder contains the issues.db backup file with issues dump
-# for the previous versions
-backups/*.db
+# data folder contains everything we are not interested
```
miketaylr: Can we update this comment to something like `the data folder contains information that shouldn't live in version control` (or just fix the typo at the end and say `interested in`)

karlcow: oops

karlcow: @miketaylr Done! :) Thanks for the review.

miketaylr:
Thanks!

PR #1729, opened 2017-08-27
fixes #1567 - Remove brackets in browser-label if any

### LINKED ISSUE #1567: When the browser name contains (), remove them
This is a followup on #1228 
Currently if the browser name is "Firefox (tablet)" we end up as a tag with `browser-firefox-(tablet)` instead of `browser-firefox-tablet` 

I'm happy to mentor this.
The code to modify is here
https://github.com/karlcow/webcompat.com/blob/0fea5ebe3cfd46efaf6568c35a358e86525f6a41/webcompat/webhooks/helpers.py#L28-L30

And probably an additional test to add to 
https://github.com/karlcow/webcompat.com/blob/0fea5ebe3cfd46efaf6568c35a358e86525f6a41/tests/test_webhook.py#L117

aryanshar:
Closes #1567

### DIFF
--- a/webcompat/webhooks/helpers.py
+++ b/webcompat/webhooks/helpers.py
@@ -27,6 +27,7 @@ def extract_browser_label(body):
         # right now, match_list.groups(0) should look like:
         # ('browser', 'firefox')
         browser = match_list.groups(0)[1].lower()
+        browser = browser.translate(None, '()')
         dash_browser = '-'.join(browser.split())
         return 'browser-{name}'.format(name=dash_browser)
     else:
--- a/tests/test_webhook.py
+++ b/tests/test_webhook.py
@@ -52,6 +52,9 @@ class TestWebhook(unittest.TestCase):
         self.issue_body2 = """
         <!-- @browser: Foobar -->
         """
+        self.issue_body3 = """
+        <!-- @browser: Firefox Mobile (Tablet) 40.0 -->
+        """
 
     def tearDown(self):
         pass
@@ -122,6 +125,8 @@ class TestWebhook(unittest.TestCase):
         self.assertEqual(browser_label, 'browser-firefox')
         browser_label_none = helpers.extract_browser_label(self.issue_body2)
         self.assertEqual(browser_label_none, None)
+        browser_label_paren = helpers.extract_browser_label(self.issue_body3)
+        self.assertEqual(browser_label_paren, 'browser-firefox-mobile-tablet')
 
     def test_is_github_hook(self):
         """Validation tests for GitHub Webhooks."""


webcompat/webhooks/helpers.py
```diff
@@ -27,6 +27,7 @@ def extract_browser_label(body):
         # right now, match_list.groups(0) should look like:
         # ('browser', 'firefox')
         browser = match_list.groups(0)[1].lower()
+	browser = browser.translate(None, "()")
```
karlcow: A couple of changes here mostly syntactic.

1. The indentation is not correct. Python is strict about indentation for defining blocks of code. As it is it will break. Make sure to indent with spaces.
2. Use simple quotes around the paranthesis. `translate(None, '()')`
3. Big +1 on using `translate` instead of `replace` (this is faster).

aryanshar: Hey @karlcow Thanks a lot :) And yeah I actually don't know what exactly this test, you're talking about is for ? Its i guess due to the reason that I don't know what these functions are doing ! A help is needed for sure.

karlcow: so usually it is better to start with tests to know what the function should do with the data. Let's say you are at the root of the project:

```bash
# Let's now start python IDLE
python
# Python 2.7.10 (default, Feb  7 2017, 00:08:15) 
# [GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
# Type "help", "copyright", "credits" or "license" for more information.
# >>> 
```

Now we are in the python environment, to quit you can just type ctrl+D anytime, but for now let's continue. In `webhooks/helpers.py`, we want to use the extract browser label function. This is the current form before your modifications.

https://github.com/webcompat/webcompat.com/blob/18c327aedf76aae5207b0bf69fcb33bd6ac90d1f/webcompat/webhooks/helpers.py#L16-L33

```python
>>> from webcompat.webhooks.helpers import extract_browser_label
>>> # as you can see the function is taking a body argument. Let's make one.
>>> body = "<!-- @browser: Firefox Mobile 40.0 -->"
>>> # Now we can pass this body to the function we imported:
>>> extract_browser_label(body)
'browser-firefox-mobile'
>>> # We can see how it extracted the @browser information.
... # And made this as a string with dashes.
>>> # What if the body contains something with parenthesis.
>>> body = "<!-- @browser: Firefox Mobile (Tablet) 40.0 -->"
>>> # Let's try again!
... extract_browser_label(body)
'browser-firefox-mobile-(tablet)'
>>> # OK so this is the issue we are trying to solve.
>>> # In this case we want the extract_browser_label(body) function
>>> # to return: 'browser-firefox-mobile-tablet'
>>> # This is our test.
```

Let's go back to https://github.com/webcompat/webcompat.com/blob/ffb0dc3932871c7576cbfb9bc82f76c024464983/tests/test_webhook.py#L119-L124

```python
    def test_extract_browser_label(self):
        """Extract browser label name."""
        browser_label = helpers.extract_browser_label(self.issue_body)
        self.assertEqual(browser_label, 'browser-firefox')
        browser_label_none = helpers.extract_browser_label(self.issue_body2)
        self.assertEqual(browser_label_none, None)
        browser_label_paren = helpers.extract_browser_label(
            '<!-- @browser: Firefox Mobile (Tablet) 40.0 -->')
        self.assertEqual(browser_label_paren, 'browser-firefox-mobile-tablet')
```

by having this third test we make sure that the project will not fail in the future if we modify something in the function. If I run this test with the current structure before your modification, we get:

```bash
nosetests -v --nocapture tests/test_webhook.py
# Extract browser label name. ... FAIL
# POST without bogus signature on labeler webhook is forbidden. ... ok
# POST with event not being 'issues' or 'ping' fails. ... ok
# POST without signature on labeler webhook is forbidden. ... ok
# POST with an unknown action fails. ... ok
# GET is forbidden on labeler webhook. ... ok
# Extract the right information from an issue. ... ok
# Validation tests for GitHub Webhooks. ... ok
# POST with PING events just return a 200 and contains pong. ... ok
# 
# ======================================================================
# FAIL: Extract browser label name.
# ----------------------------------------------------------------------
# Traceback (most recent call last):
#   File "/Users/karl/code/webcompat.com/tests/test_webhook.py", line 127, in test_extract_browser_label
#     self.assertEqual(browser_label_paren, 'browser-firefox-mobile-tablet')
# AssertionError: 'browser-firefox-mobile-(tablet)' != 'browser-firefox-mobile-tablet'
# 
# ----------------------------------------------------------------------
# Ran 9 tests in 0.056s
# 
# FAILED (failures=1)
```

The test is failing because indeed we don't take care of the parenthesis. Let's use the function you have done.


```python
def extract_browser_label(body):
    """Parse the labels from the body in comment:

    <!-- @browser: value -->.
    Currently this only handles a single label,
    because that's all that we set in webcompat.com.
    """
    match_list = re.search(r'<!--\s@(\w+):\s([^\d]+?)\s[\d\.]+\s-->', body)
    if match_list:
        # perhaps we do something more interesting depending on
        # what groups(n)[0] is in the future.
        # right now, match_list.groups(0) should look like:
        # ('browser', 'firefox')
        browser = match_list.groups(0)[1].lower()
	browser = browser.translate(None, "()")
        dash_browser = '-'.join(browser.split())
        return 'browser-{name}'.format(name=dash_browser)
    else:
        return None
```

Let's run the test again

```bash
nosetests -v --nocapture tests/test_webhook.py
# Extract browser label name. ... ok
# POST without bogus signature on labeler webhook is forbidden. ... ok
# POST with event not being 'issues' or 'ping' fails. ... ok
# POST without signature on labeler webhook is forbidden. ... ok
# POST with an unknown action fails. ... ok
# GET is forbidden on labeler webhook. ... ok
# Extract the right information from an issue. ... ok
# Validation tests for GitHub Webhooks. ... ok
# POST with PING events just return a 200 and contains pong. ... ok
# 
# ----------------------------------------------------------------------
# Ran 9 tests in 0.054s
# 
# OK
```

Working. No test failure.

I hope it helps.
Could you modify your code so to add the requested changes. Thanks.

aryanshar:
Hey @karlcow Thanks a heap for investing so much. That really helped a lot. I have finally updated the diff , but it is still having indentation issues. Thats weird because i corrected them like in 2-3 editors and sent 2-3 PR update, still indentation is improper.

karlcow:
@aryanshar Your commit is good. Your spaces are correct.

```
→ git fetch webcompat
remote: Counting objects: 5, done.
remote: Total 5 (delta 4), reused 4 (delta 4), pack-reused 1
Unpacking objects: 100% (5/5), done.
From https://github.com/webcompat/webcompat.com
 * [new branch]      issues/1695/1 -> webcompat/issues/1695/1

→ git merge webcompat/master
Already up-to-date.

→ git checkout -b aryanshar-Issue/1567 master
Switched to a new branch 'aryanshar-Issue/1567'

→ git pull https://github.com/aryanshar/webcompat.com.git Issue/1567
remote: Counting objects: 7, done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 7 (delta 5), reused 7 (delta 5), pack-reused 0
Unpacking objects: 100% (7/7), done.
From https://github.com/aryanshar/webcompat.com
 * branch            Issue/1567 -> FETCH_HEAD
Updating ffb0dc3..bea502d
Fast-forward
 tests/test_webhook.py         | 5 +++++
 webcompat/webhooks/helpers.py | 1 +
 2 files changed, 6 insertions(+)

→ nosetests -v --nocapture
API access to comments greater than 30 returns pagination in Link ... ok
API issue search with bad keywords returns JSON 404. ... ok
The data body sent to GitHup API. ... ok
Checks that appropriate radio button label is returned. ... ok
Check ETAG for issues. ... ok
test_build_query_string (tests.test_topsites.TestTopsites) ... ok
test_build_uri (tests.test_topsites.TestTopsites) ... ok
test_node_text (tests.test_topsites.TestTopsites) ... ok
Test issues and integer for: ... ok
/issues/new POST exit with 400 if missing parameters. ... ok

Ran 63 tests in 1.452s

OK

```

Everything is on track.
Let's move on.

Thanks a lot @aryanshar

Issue #1734, opened 2017-08-30
templates.js tracked or not tracked.


karlcow:
In #1730 I had a bit of troubles with this

This is tracked here:
https://github.com/webcompat/webcompat.com/blame/master/webcompat/static/js/templates.js

But https://github.com/webcompat/webcompat.com/blob/dd8102a3d0f89d18bea2a86228afac2fcfeead44/.gitignore#L27
says to not track it…

Madness ensues with pre-commit.

karlcow:
This was added by this commit by @miketaylr 
https://github.com/webcompat/webcompat.com/commit/260c05b3294689bc9b62d3d34557738a4307ec25#diff-f5f2115c1d9da8f288334dd32f2cf1e4

karlcow:
Oh the .gitignore was added after…
https://github.com/webcompat/webcompat.com/blame/master/.gitignore#L27

karlcow:
```
→ git status
On branch master
Your branch is up-to-date with 'webcompat/master'.
nothing to commit, working directory clean

→ grunt
Running "checkDependencies:default" (checkDependencies) task

Running "jst:compile" (jst) task
File webcompat/static/js/templates.js created.

Running "concat:dist" (concat) task

Running "concat:diagnose" (concat) task

Running "concat:issues" (concat) task

Running "concat:issueList" (concat) task

Running "concat:userActivity" (concat) task

Running "concat:cssFixmeLibs" (concat) task

Running "uglify:dist" (uglify) task
>> 1 file created 545.13 kB → 386.68 kB

Running "uglify:issues" (uglify) task
>> 1 file created 41.36 kB → 23.64 kB

Running "uglify:issueList" (uglify) task
>> 1 file created 45.83 kB → 25.18 kB

Running "uglify:userActivity" (uglify) task
>> 1 file created 22.74 kB → 11.87 kB

Running "uglify:diagnose" (uglify) task
>> 1 file created 21.65 kB → 11.63 kB

Running "uglify:contributors" (uglify) task
>> 1 file created 1.05 kB → 737 B

Running "uglify:cssFixmeLibs" (uglify) task
>> 1 file created 131.46 kB → 63.7 kB

Running "postcss:dist" (postcss) task
>> 1 processed stylesheet created.

Running "cssmin:combine" (cssmin) task
>> 1 file created. 271.68 kB → 59.8 kB

Done.
```

miketaylr:
[...]

PR #1739, opened 2017-08-30
Issue #1734. Remove templates.js

miketaylr:
It shouldn't be tracked (.gitignore was added after it was committed
to the repo (mistakenly)).

r? @karlcow


karlcow:
@miketaylr r+

Issue #1590, opened 2017-06-10
https://github.com/mozilla/geckodriver/issues/308 fixed


zoepage:
@miketaylr https://github.com/mozilla/geckodriver/issues/308 has been fixed. 

Related to:
https://github.com/webcompat/webcompat.com/blob/master/tests/functional/lib/helpers.js#L62

Possible this is related to the failing build? 
https://travis-ci.org/webcompat/webcompat.com#L579

brizental:
Closing this because the code for it was removed in #1743

Issue #1385, opened 2017-03-03
Create tests to verify logging in/out works from issue page + home page


miketaylr:
So we don't regress again like: https://github.com/webcompat/webcompat.com/issues/1384

We assume logging in and out will work, but don't have any explicit tests for it.

miketaylr:
- [ ] home page
- [ ] /issues route
- [ ] issue page
... more? all routes? dunno.

jeanhl:
I'd like to claim this issue. To clarify:
- testing the login and logout functions in views.py
- Python unittests.

I have a framework in mind but I'd like to discuss it with someone before I go hogwild with it.

miketaylr:
Interesting, I had functional Intern tests in mind since we're doing a GitHub OAuth exchange (hence it not being simple to just mock out). 

What did you have in mind @jeanhl?

jeanhl:
Ah, my bad. I thought I could set up a mock OAuth. I was trying to access github looking at this line in view.py:
"from webcompat import github"
but I can't find a file called "github."

miketaylr:
Ah! 

https://github.com/webcompat/webcompat.com/blob/master/webcompat/__init__.py#L23 is where `github` is defined, it's a reference to the GitHub Flask addon https://github-flask.readthedocs.io/en/latest/.

Mocking it out would be awesome. It would make offline (functional) tests possible.

jeanhl:
I'll look into how to do a mockup of this. The github docs are very useful. Thank you

jeanhl:
I'm looking at the code and I'm thinking of doing the following:
- use flask's test_quest_context to activate a request context temporarily. This will create a temp request, g and session. This should let us test g.referer and request
- use flask's follow_redirects to make sure the redirect works fine

not too sure how to test github.authorize. I get the following when I print github.authorize('public_repo') :
<Response 393 bytes [302 FOUND]>
type: <class 'werkzeug.wrappers.Response'>

Would you like me to check set_referer in helpers.py as well?

If I'm going way off base or if there is a better way, let me know.

karlcow:
+1 to mocking in general. It is something I have been looking at for a longtime, but didn't take the energy to do. 

But before we go off rails 🛤 . Let's make sure we test what we want to test. 

☎️  @miketaylr 
Note that this test is a good thing but would not have detected the issue in #1384 The **login/logout worked**, the issue was with the caching of the resource. Aka basically what should be tested for this #1384 is the following scenario:

[...]

PR #1743, opened 2017-09-02
Fixes #1631 - Warns user and aborts tests if server isn't right or if it's not in test mode.

### LINKED ISSUE #1631: [tests] Starting "python run.py -t" from the functional tests
So we have a pattern. 

Very often in the team, we forget (or at least I do) to start `python run.py -t` for running the functional tests. And indeed we have quite a lot of things to do before starting the tests. Having the right version of java, starting the server, the long command for launching the tests.

I wonder if this should not be handled with a script.

brizental:
So, to check that the server is in test mode, I make a request to our API and check that it's returning fixture data. This took a little while longer, because I had to find a workaround to add external modules to the `intern.js` config file, since you can't add them as you would on the test suites, because it's not loaded as an AMD module. I don't feel that my solution looks very good, but it works and I asked in the intern Gitter chat and the guys said it was a fine way to do it.

Anyways, I also removed code to login with real data to GitHub. I know that @miketaylr was kind of against it, but I found out that we actually removed the code that did that in #1604 , so it didn't work anyways... I think we could just open a new issue for that if you guys think it's necessary to have the option to test without fixture data. 

I also removed the `visibleByClass` method, because it wasn't used anywhere and Leadfoot has a similar method ([findDisplayedByClassName](https://theintern.io/leadfoot/module-leadfoot_Command.html#findDisplayedByClassName)).

Lastly, I noticed that when you start the server in non-test mode it runs grunt before starting the server, but not in test mode. So I fixed that.

Hope it's alright to add all this to the same PR. I didn't think it was necessary to open new issues for these things.

r? @miketaylr @zoepage

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -5,13 +5,8 @@
 /*eslint no-console: ["error", { allow: ["log", "error"] }] */
 
 define(
-  [
-    "intern",
-    "intern!object",
-    "require",
-    "intern/dojo/node!leadfoot/helpers/pollUntil"
-  ],
-  function(intern, registerSuite, require, pollUntil) {
+  ["intern", "intern!object", "require", "intern/dojo/node!http"],
+  function(intern, registerSuite, require, http) {
     "use strict";
     var config = intern.config;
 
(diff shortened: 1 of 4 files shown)


zoepage:
@brizental great work :) there is an issue somewhere mentioning the geckobug. Could you reference this in the commit? (Sorry, on my phone right now which makes it hard to search for it).

Thanks!

brizental:
@zoepage #1590

tests/functional/lib/helpers.js
```diff
@@ -49,90 +44,68 @@ define(
...
+          if (response.headers.hasOwnProperty("etag")) {
+            reject(
+              new Error(
+                `
+              =======================================================
+              It seems like you didn't start the server in test mode.
+              Open another terminal and window type: 
+             \x1b[32m npm run start:test\x1b[0m
+              or
+             \x1b[32m source env/bin/activate && python run.py -t\x1b[0m
```
miketaylr: Can we have this second line just mention `python run.py -t`? The first does effectively the same, but assumes you used `virtualenv` in a specific way. This second way is more general.

tests/functional/lib/helpers.js
```diff
@@ -49,90 +44,68 @@ define(
...
+    /* 
+    This method makes a call to our API to check that the server is returning fixture data,
+    it will also check if there's anything wrong with the server.
+  */
+    function checkServer() {
+      return new Promise(function(resolve, reject) {
+        var request = http.get(url("/api/issues/100"), function(response) {
+          // If the headers are from non-fixture data,
+          // they will have the "etag" property.
+          if (response.headers.hasOwnProperty("etag")) {
```
miketaylr: Another way would be to look for the `_fixture` property on the first object in the array:

https://github.com/webcompat/webcompat.com/blob/03b0da41a8615f160a90529fef51dea64352d0c6/tests/fixtures/api/issues/100/comments.json#L3

(pretty sure I added that to all fixtures manually... )

tests/functional/lib/helpers.js
```diff
@@ -49,90 +44,68 @@ define(
       );
     }
 
-    // TODO: when https://github.com/mozilla/geckodriver/issues/308 is fixed,
-    // remove this ugliness.
-
-    function login(context) {
```
miketaylr: Can you file a follow up issue to create a way to test logging in, please?

tests/functional/lib/helpers.js
```diff
@@ -49,90 +44,68 @@ define(
...
+        request.on("error", function() {
+          reject(
+            new Error(
+              `
+            ======================================================
+            Oops, something went wrong. Try restarting the server.
+            Open another terminal and window type: 
+           \x1b[32m npm run start:test\x1b[0m
+            or
+           \x1b[32m source env/bin/activate && python run.py -t\x1b[0m
```
miketaylr: Same here.

tests/intern.js
```diff
@@ -25,17 +19,28 @@ define(["intern"], function(intern, topic) {
...
+    },
+
     // The port on which the instrumenting proxy will listen
     proxyPort: 9090,
 
     // A fully qualified URL to the Intern proxy
     proxyUrl: "http://127.0.0.1:9090/",
     siteRoot: siteRoot,
     tunnel: "SeleniumTunnel",
-    tunnelOptions: {
```
miketaylr: Can you explain this part to me? Why don't we want to specify the drivers?

tests/functional/lib/helpers.js
brizental: That's cool! I didn't know that. I'll change it :)

tests/functional/lib/helpers.js
brizental: I think we already have and issues like that, wait...
Yeah! #1385

tests/intern.js
brizental: That was a mistake. I removed it while testing something, but should have put it back. Thanks for noticing.

tests/functional/lib/helpers.js
brizental: Yeah, sure. Makes sense.

miketaylr: Thanks!

tests/functional/lib/helpers.js
miketaylr: Ah, perfect. Thanks!

brizental:
@miketaylr I made the changes you asked for, Mike. Had to squash all commits together because I did something stupid while rebasing the first time, lol sorry.

miketaylr:
> Had to squash all commits together because I did something stupid while rebasing the first time, lol sorry.

Heh no worries. Been there!

miketaylr:
You rock, Bea -- thanks!

PR #1765, opened 2017-09-06
Fixes #1761 - Creates a mapping in between statuses and milestones

### LINKED ISSUE #1761: Create mapping for milestones to API integers
So it's less annoying to deal with them.

#886.

karlcow:
r? @miketaylr

### DIFF
--- a/run.py
+++ b/run.py
@@ -3,6 +3,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""Starter file for the project."""
 
 import argparse
 import pkg_resources
@@ -70,11 +71,11 @@ pip install name==1.2.1
 
 
 def check_pip_deps():
-    '''Check installed pip dependencies.
+    """Check installed pip dependencies.
 
     Make sure that the installed pip packages match what is in
     requirements.txt, prompting the user to upgrade if not.
-    '''
+    """
     for req in open("./config/requirements.txt"):
         try:
             pkg_resources.require(req)
@@ -87,7 +88,7 @@ def check_pip_deps():
 
 
 def config_validator():
-    '''Make sure the config file is ready.'''
+    """Make sure the config file is ready."""
     # Checking if oauth token is configured
     if app.config['OAUTH_TOKEN'] == '':
         sys.exit(TOKEN_HELP)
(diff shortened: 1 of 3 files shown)


run.py
```diff
@@ -94,6 +105,49 @@ def config_validator():
...
 
+def initialize_status(statuses):
+    """Map the status name with the milestone id on GitHub.
+
+    The project needs the mapping in between milestones and their id
+    to be able to query status related things. It uses a backup version
+    when the request to GitHub fails.
+    """
+    print('Statuses Initialization…')
+    REPO_ROOT = app.config['ISSUES_REPO_URI'].rpartition('/issues')[0]
```
miketaylr: Reminder: We should probably create a config or global variable for this, in all the various places we do these things. For future cleanup.

run.py
```diff
@@ -94,6 +105,49 @@ def config_validator():
...
+        r = requests.get(milestones_url)
+        milestones_content = r.content
+        if r.status_code == 200:
+            with open(milestones_file, 'w') as f:
+                f.write(r.content)
+            print('Milestones saved in data/')
+        r.raise_for_status()
+    except requests.exceptions.HTTPError as error:
+        # Not working, let's use the cached copy
+        print(MILESTONE_ERROR.format(msg=error))
```
miketaylr: It might be interesting to log this at some future point.

run.py
```diff
@@ -87,12 +99,55 @@ def check_pip_deps():
...
+        with open(milestones_file, 'r') as f:
+            milestones_content = f.read()
+    finally:
+        # save in data/ the current version
+        print('Milestones in memory')
+        app.config.update(STATUSES=convert_milestones(milestones_content))
+        app.config.update(JSON_STATUSES=json.dumps(app.config['STATUSES']))
```
miketaylr: If we don't want to create an additional config object that's JSON encoded, the other option is to dump it it as a template variable here:

https://github.com/webcompat/webcompat.com/blob/master/webcompat/views.py#L235

But, it might be useful to have this in JSON elsewhere, so

miketaylr: If we don't want to create an additional config object that's JSON encoded, the other option is to dump it it as a template variable here:

https://github.com/webcompat/webcompat.com/blob/master/webcompat/views.py#L235

But, it might be useful to have this in JSON elsewhere, so let's go ahead and merge. If you want to change it or file an issue later up to you! 🥇

run.py
```diff
@@ -87,12 +99,55 @@ def check_pip_deps():
...
+
+
+def convert_milestones(milestones_file):
```
miketaylr: nit: might be good to file a follow-up "good first bug" for someone to write a simple test for this.

PR #1770, opened 2017-09-06
Fixes #1751 - Remove status-needstriage label in webhook

### LINKED ISSUE #1751: Update new issue webhooks
Currently it adds a `status-needstriage` label. We want to add the correct milestone on issue creation.

#886.

MDTsai:
Use milestone to replace status-* so don't use webhook to add
status-needstriage when new issue created.

### DIFF
--- a/webcompat/webhooks/helpers.py
+++ b/webcompat/webhooks/helpers.py
@@ -106,9 +106,8 @@ def new_opened_issue(payload):
     '''When a new issue is opened, we set a couple of things.
 
     - Browser label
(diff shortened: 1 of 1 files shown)


webcompat/webhooks/helpers.py
```diff
@@ -108,7 +108,7 @@ def new_opened_issue(payload):
     - Browser label
     - status-needstriage
```
miketaylr: nit: can we update this comment so it's accurate now plz?

PR #1775, opened 2017-09-06
Fixes #1749 - Switches from labels to milestones

### LINKED ISSUE #1749: Create API backend for milestones
Either create new methods in endpoints.py, or update existing methods to tweak the `milestones` param, etc.

#886.

karlcow:
- Actually switch the code from labels to milestones
- Removes unused code from different parts because we were doing things differently
  (They are in their own separate commits)
- Fixes syntax and code.


karlcow:
@miketaylr r?

karlcow:
```
→ nosetests
...............................................................
Ran 63 tests in 0.848s

OK
```

karlcow:
@miketaylr I also realized that this PR is not enough. I probably need to set up what will interact with the milestone editor. I'll check that later.

Issue #1768, opened 2017-09-06
Create milestones model for milestoneseditorview widget thingy


miketaylr:
Right now it's fixture data, and @karlcow added real data in #1765.

#886.

PR #1781, opened 2017-09-07
Fixes: #1768. Create a model for milestones.

miketaylr:
We populate the model with the statuses info that is injected into the template.

r? @MDTsai


miketaylr:
So many test failures that seem unrelated...

webcompat/static/js/lib/milestones.js
```diff
@@ -4,6 +4,24 @@
 
 var issues = issues || {}; // eslint-disable-line no-use-before-define
 
+issues.MilestonesModel = Backbone.Model.extend({
+  initialize: function(options) {
+    // transform the format from the server into something that our templates
+    // are expecting.
+    var milestones = [];
+    _.forOwn(options.statuses, function(value, key) {
```
miketaylr: https://lodash.com/docs/3.10.1#forOwn

Issue #1782, opened 2017-09-07
Update /edit API endpoint to handle milestone in addition to state


karlcow:
The widget status editor will talk to the endpoint to change the status of the issue. 
This is a followup on #1749. 

aka milestone/state combination.

Issue #1753, opened 2017-09-05
Wire up milestone editor with milestone backend


miketaylr:
So when you change a milestone, it communicates that change with GitHub and syncs the UI.

#886. 

Depends on #1745, related to #1748.

miketaylr:
(essentially #1766)

miketaylr:
Need to send a dict with milestone integer and state to the /issues/N/edit endpoint:

`{'milestone': 2, 'state': 'open'}`

miketaylr:
Depends on #1782.

karlcow:
Fixed by #1787

Issue #1780, opened 2017-09-07
Make it so you can only check a single milestone in the milestone editor


miketaylr:
So selecting a new one will unselect the current one.

We have similar code here:

https://github.com/webcompat/webcompat.com/blob/0ea702ebba61fd2e887a24768c4f79e4b0613875/webcompat/static/js/lib/labels.js#L146-L162

MDTsai:
basically it's here https://github.com/webcompat/webcompat.com/blob/milestones/webcompat/static/js/lib/milestones.js#L73
also get checked one then uncheck all the others.

karlcow:
Why don't we use `type="radio"`

miketaylr:
> Why don't we use type="radio"

Let's re-visit once we have a working version as-is.

PR #1787, opened 2017-09-07
Issue #1753. Send updated milestone and status to API backend.

miketaylr:
Not ready for review. I want #1782 to land first so I can test if my assumptions were true.

### DIFF
--- a/webcompat/static/js/lib/milestones.js
+++ b/webcompat/static/js/lib/milestones.js
@@ -17,6 +17,24 @@ issues.MilestonesModel = Backbone.Model.extend({
 
     this.set("milestones", milestones);
   },
+  updateMilestones: function(data) {
+    $.ajax({
+      contentType: "application/json",
+      data: JSON.stringify(data),
+      type: "PATCH",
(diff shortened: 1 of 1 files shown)


miketaylr:
r? @karlcow 

We tested that this works, and it does. \o/

webcompat/static/js/lib/milestones.js
```diff
@@ -83,14 +101,19 @@ issues.MilestoneEditorView = issues.CategoryEditorView.extend({
   },
   closeEditor: function(e) {
     if (!e || (e && (e.keyCode === 27 || !e.keyCode))) {
-      // var checked = $("input[type=checkbox]:checked");
-      // var milestonesArray = _.pluck(checked, "name");
-      this.issueView.editorButton.removeClass("is-active");
-      // TODO: make this work
-      // this.issueView.model.updateMilestones(milestonesArray);
+      var checked = $("input[type=checkbox]:checked").prop("name");
```
miketaylr: (this assumes only one item will be checked, which is true when #1780 lands)

PR #1791, opened 2017-09-07
Fixes #1780 - Single check in milestone editor widget

### LINKED ISSUE #1780: Make it so you can only check a single milestone in the milestone editor

MDTsai:
Clone from label editor widget and use milestone.name in milestone-editor.jst
to prevent mis-uncheck label.
r?=@miketaylr

### DIFF
--- a/webcompat/static/js/lib/milestones.js
+++ b/webcompat/static/js/lib/milestones.js
@@ -70,7 +70,30 @@ issues.MilestoneEditorView = issues.CategoryEditorView.extend({
     this.issueView = options.issueView;
   },
   template: wcTmpl["web_modules/milestone-editor.jst"],
-  updateView: function() {}, // no-op
+  updateView: function(evt) {
+    // We try to make sure only one milestone is set
+    // enumerate all checked milestones and uncheck the others.
+    var checked = $(
+      'input[type=checkbox][data-remotename^="milestone"]:checked'
+    );
+    _.each(checked, function(item) {
+      if (item !== evt.target) {
+        item.checked = false;
+      }
+    });
+    // we do the "real" save when you close the editor.
+    // this just updates the UI responsively
+    checked = $("input[type=checkbox]:checked");
+    // build up an array of objects that have
+    // .name and .color props that the templates expect
+    var modelUpdate = [];
+    _.each(checked, function(item) {
+      //item already has a .name property
+      item.color = $(item).data("color");
+      modelUpdate.push(item);
+    });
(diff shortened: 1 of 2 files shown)


webcompat/templates/web_modules/milestone-editor.jst
```diff
@@ -14,7 +14,7 @@
           <input class="wc-CategoryEditor-list-item-checkbox"
                  type="checkbox" name="<%= milestone.name %>"
                  data-color="orangered"
-                 data-remotename="<%= milestone.remoteName %>"
+                 data-remotename=milestone-"<%= milestone.name %>"
```
miketaylr: This results in:

```
<input class="wc-CategoryEditor-list-item-checkbox" 
            name="worksforme" 
            data-color="orangered" 
            data-remotename="milestone-"worksforme"" 
            tabindex="-1" 
type="checkbox">
```

Note the extra quotes. It works, but could we change it to something like:

`"data-remotename=milestone-<%= milestone.name %>"` (moving the opening quote before `milestone`)

webcompat/static/js/lib/milestones.js
```diff
@@ -70,7 +70,30 @@ issues.MilestoneEditorView = issues.CategoryEditorView.extend({
...
+    checked = $("input[type=checkbox]:checked");
+    // build up an array of objects that have
+    // .name and .color props that the templates expect
+    var modelUpdate = [];
+    _.each(checked, function(item) {
+      //item already has a .name property
+      item.color = $(item).data("color");
+      modelUpdate.push(item);
+    });
+    this.reRender({ labels: _.uniq(modelUpdate) });
```
miketaylr: Ah, I realized I made a mistake here.

reRender shouldn't want `labels`, but probably `milestones`. BUT, we're missing a `issue-milestones-sub.jst` template, and `subTemplate` property points to the label template. >_<

https://github.com/webcompat/webcompat.com/blob/dddfd9b68d0582c1fdcf2efae516416c3a2d61c9/webcompat/static/js/lib/milestones.js#L51

miketaylr: BUT, don't worry about that. I'll file a follow up bug.

miketaylr:
Actually, let's just merge and I'll fix the quote issue in #1792.

Issue #1798, opened 2017-09-08
Create a `jst` npm script task


miketaylr:
It sucks to have to re-run `npm run build` to get the jst task to run, let's make a task that only runs `grunt jst`.

miketaylr:
(taking this, because it's driving me nuts)

zoepage:
Closing as automated closing after merge failed.

PR #1799, opened 2017-09-08
Issue #1798. Add a jst npm script.

miketaylr:
r? @zoepage


zoepage:
Although test fail, they shouldn't be related to the PR. 

👍

PR #1815, opened 2017-09-08
Fixes #1810 - If no existing topsites.db, no archive

### LINKED ISSUE #1810: If there is no existing topsites.db, we can skip archive old one
When first time run this would happen.

MDTsai:
Do check when archive if no existing topsites.db.
Modify the path used in DB library.
r?=@karlcow


miketaylr:
Ping @karlcow

Issue #1848, opened 2017-10-24
Missing automatic Priority Labeling 


softvision-oana-arbuzov:
URL: www.webcompat.com
Browser / Version: Firefox Mobile Nightly 58.0a1 (2017-10-22), Firefox Release 56.0.1
Operating System: Windows 10 Pro

**Steps to Reproduce**
1. Navigate www.webcompat.com.
2. Type in all required info and report the issue.
3. Observe added labels.

**Expected Behavior:**
Corresponding Priority label is automatically added.

**Actual Behavior:**
Priority label is not automatically added.

**Note**
1. Also reproducible on staging (www.staging.webcompat.com).
2. Workaround: Manually setting the Priority labels is performed.

**Watchers:**
@softvision-sergiulogigan
@softvision-oana-arbuzov

sv; 

![missingprioritylabels](https://user-images.githubusercontent.com/12184325/31953114-0457989c-b8eb-11e7-884f-f9a772a65153.png)

miketaylr:
Thanks for reporting @softvision-oana-arbuzov.

Yesterday I deployed https://github.com/webcompat/webcompat.com/pull/1815. @MDTsai?

MDTsai:
Yes, I found that and manually restore the DB. But I don't think #1815 break that, it's not running. So I guess the deploy process did something I don't know, maybe?

MDTsai:
DB is still there after I manually restore, but the priority label is not working, check this today.

miketaylr:
This is fixed now.

softvision-sergiulogigan:
Happened again for https://github.com/webcompat/web-bugs/issues/13436
I did like 4-5 refresh actions and waited about a minute, but still no label were displayed.
Added the browser-firefox-mobile priority-normal labels manually.

![image](https://user-images.githubusercontent.com/16573771/32727997-5a7d44be-c887-11e7-90d7-8065b032960c.png)

miketaylr:
cc @MDTsai 

Uh I hope I didn't break this again? In theory https://github.com/webcompat/webcompat.com/commit/c427c32064666005df6056f96442e3c71c3193a8 would have prevented that.

MDTsai:
I don't think it's broken, the domain is not in topsites.db. Priority tag function works as far as on issue 13453.
I try to search the URL https://www.alexa.com/siteinfo/indoxxi.com, the global ranking is 10753 for me, so it's not in priority-normal (1000 to 10000).

[...]

PR #1849, opened 2017-10-25
No issue. Ignore data/ when doing deploys.

miketaylr:
See Issue #1848.

r? @karlcow


karlcow:
❤️

miketaylr:
(we don't need to deploy this change.. so leaving out of changelog probably)

PR #1853, opened 2017-10-31
Fixes #1852: Render an issue even if milestone failed to get set.

### LINKED ISSUE #1852: Don't fail hard if milestone wasn't set
We should solve the problem of milestone not getting set in #1851, but we should be more forgiving to the user on the frontend if it does fail.

If there's no milestone, we get the following error and the page fails to render:

`TypeError: milestoneColors[milestone] is undefined`

miketaylr:
I was gonna wait on review for this, but we just got another mention on IRC that their issue failed for them.


miketaylr:
Tests passed, so r=me. Please review @zoepage and file any follow up issues if you find anything.

PR #1854, opened 2017-10-31
Fixes #1851. Add milestone for new issues via our new issue webhook.

### LINKED ISSUE #1851: Filed issue details not displayed (because logged-in non-collaborator users cannot set milestones)
Went to webcompact and reported a bug the reported page is blank. I should have found the sharing option and other details.
screenshare

miketaylr:
Otherwise, this will fail for non-anonymous users, unless they are repo collaborators.

Note: I didn't change the name of the `/labeler` webhook. If someone feels strongly about that, please file a followup issue. Something like `/newissue` would make more sense.

r? @karlcow 
r? @zoepage 

(Note: I tested this on staging (after I removed basic auth so the webhook would work))

### DIFF
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -110,12 +110,12 @@ class TestForm(unittest.TestCase):
         self.assertEqual(actual, expected)
 
     def test_build_formdata(self):
-        """The data body sent to GitHup API."""
+        """The data body sent to GitHub API."""
         form_object = {'foo': 'bar'}
         actual = form.build_formdata(form_object)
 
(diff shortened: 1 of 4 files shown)


karlcow:
I don't think the labeler should change name. it's ok if it does a bit more things and labels/milestones are anyway just boxes that we use.

karlcow:
I see that @zoepage is in PTO for the rest of the week. Let me merge this.

miketaylr:
Thanks @karlcow

PR #1864, opened 2017-11-07
Fixes #1858 - Adds essential routes, empty templates and tests for Triage Dashboard

### LINKED ISSUE #1858: Triage dashboard HTML page and route
Context: #1857 

* [x] Create tests for the new route `/dashboard/triage`
* [x] Create tests for the content returned by this route
* [x] Create html template for `/dashboard/triage` no CSS required at this point

A followup issue will handle the helpers to query the right information to inject in the HTML page

karlcow:
This fixes the #1858

### DIFF
--- a/tests/test_urls.py
+++ b/tests/test_urls.py
@@ -4,7 +4,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-'''Tests for our URL endpoints.'''
+"""Tests for our URL endpoints."""
 
 import os.path
 import sys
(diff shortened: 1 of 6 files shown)


tests/test_urls.py
```diff
@@ -128,6 +132,29 @@ def test_new_issue_should_not_crash(self):
...
+        """Request to /dashboard should be 404.
+
+        For now, the dashboard route has no purpose.
+        """
+        rv = self.app.get('/dashboard/')
+        content_test = 'Lost in Punk Cat Space (404)' in rv.data
+        self.assertEqual(rv.status_code, 404)
+        self.assertTrue('text/html' in rv.content_type)
+        self.assertTrue(content_test)
+        rv = self.app.get('/dashboard')
```
miketaylr: Is there any point in running these lines again? 
Accidental duplication?

webcompat/templates/dashboard/footer.html
```diff
@@ -0,0 +1,3 @@
+<footer>
+@@The shot in the footer@@
```
miketaylr: 😄

tests/test_urls.py
karlcow: @miketaylr The first series is testing `/dashboard/` and then the second one testing `/dashboard` which are two different routes. My initial thought was to be sure to make 404 is the intent here. But many it's unnecessary as any unknown route is 404.

miketaylr: ohhh, my bad. I didn't notice the trailing `/`. All good.

PR #1868, opened 2017-11-12
Fixes #1867 - Fixes time dependent test failure. 

### LINKED ISSUE #1867: [Dashboard] Time dependent test failure
One test was working because it was uncannily in the right time gap. 
It has to be fixed.

karlcow:
- Adds a more generic function to compute the threshold for time.
- Makes the fixture in the past so we don't run in an uncanny situation.
- Modifies the patch so it mocks the return value for the control date.

### DIFF
--- a/tests/fixtures/dashboard/triage.json
+++ b/tests/fixtures/dashboard/triage.json
@@ -16,21 +16,21 @@
         "name": "status-needinfo"
       }
     ],
-    "number": 1298,
+    "number": 1,
     "assignee": null,
     "repository_url": "https://api.github.com/repos/webcompat/webcompat-tests",
     "closed_at": null,
-    "id": 270043595,
-    "title": "blah blah blah - site is not usable",
+    "id": 1,
+    "title": "example.org - dashboard test",
     "comments": 0,
     "state": "open",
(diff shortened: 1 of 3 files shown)


karlcow:
```
→ nosetests -v --nocapture 
Statuses Initialization…
Milestones saved in data/
Milestones in memory
Pagination in comments Link response header. ... ok
Patching the issue is working only with certain circumstances. ... ok
Check we receive the right list of browsers. ... ok
Assess the filtering is correct. ... ok
Check if a list of labels contains the 'status-needinfo'. ... ok
Check if an issue is older than a certain time gap. ... ok
Checks that appropriate radio button label is returned. ... ok
Check ETAG for issues. ... ok
Testing user activity page title. ... ok
test_build_query_string (tests.test_topsites.TestTopsites) ... ok
test_build_uri (tests.test_topsites.TestTopsites) ... ok
test_node_text (tests.test_topsites.TestTopsites) ... ok
Request to /dashboard should be 404. ... ok
Request to /dashboard/triage should be 200. ... ok
Test if issues are really integer. ... ok
/issues/new POST exit with 400 if missing parameters. ... ok
Extract priority label. ... ok

Ran 71 tests in 2.231s

OK
```

miketaylr:
Good stuff. Thanks!

PR #1869, opened 2017-11-12
Fixes #1862 - removes css-fixme

### LINKED ISSUE #1862: Remove css-fixme
As CSS FixMe seems not fully maintained, the idea came up to remote it from the repository and I second that idea. 
There is also always the option to deploy it separately. 

What do you think @adamopenweb @wisniewskit @MDTsai @denschub @miketaylr @karlcow?

zoepage:
r? @miketaylr


webcompat/views.py
```diff
@@ -330,8 +330,11 @@ def contributors():
 
 @app.route('/tools/cssfixme')
 def cssfixme():
-    """Route for CSS Fix me tool."""
-    return render_template('cssfixme.html')
+    msg = """
+    This resource doesn't exist anymore.
+    See https://github.com/webcompat/css-fixme/
+    for more details."""
+    return (msg, 410, {"content-type": "text/plain; charset=utf-8"})
```
karlcow: `"content-type": "text/plain; charset=utf-8"` single quotes here instead of double quotes.

zoepage: then we'd also need to change it here, I guess?
https://github.com/webcompat/webcompat.com/blob/1c34df50ad8d4f5fc6d40f125101aae3205a321a/webcompat/views.py#L283

miketaylr: > then we'd also need to change it here, I guess?

@zoepage yeah, if you don't mind. We tend to use single quotes for python strings, but the codebase isn't 100% consistent. (except for docstrings and other other places where you need to nest quotes for whatever reason).

r+ with that tweak. Thanks!

miketaylr:
Thanks! I'm going to squash these into a single commit.

Issue #1882, opened 2017-11-16
Adjust CSP to allow temporary hosting of assets on webcompat.github.io


denschub:
As discussed with @zoepage and @miketaylr on IRC, let's add webcompat.github.io to the list of allowed domains for assets so we can use generated resources from https://github.com/webcompat/design/.

denschub:
Done in #1886.

PR #1884, opened 2017-11-16
Issue #1882 - Whitelist webcompat.github.io in the CSP

denschub:
As discussed, this is temporary and will be reverted before releasing the refactored changes.

We want to embed SVGs and CSS files from the design repo. Since SVGs can be both an object and an image, depending on their use, I added both image and object src rules, as well as style white-list entries.

r? @miketaylr


denschub:
Disregard, wrong target branch. Sorry for the spam!

Issue #1889, opened 2017-11-17
Replace current CSS with the Work-in-Progress styles from the refactor


denschub:
CSS is hosted at `https://webcompat.github.io/design/css/main.css`. We'll hotlink this directly to have an easier development process. CSP rules for that were added in #1886.

PR #1890, opened 2017-11-17
Issue #1889. Embed the CSS from webcompat/design for easier refactoring.

denschub:
See #1889 for details.

r? @zoepage

### DIFF
--- a/webcompat/templates/layout.html
+++ b/webcompat/templates/layout.html
@@ -27,11 +27,10 @@
   <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-196x196.png') }}" sizes="196x196">
   <meta name="msapplication-TileColor" content="#e4bc05">
   <meta name="msapplication-TileImage" content="{{ url_for('static', filename='favicon/mstile-144x144.png') }}">
-{%- if config.PRODUCTION or config.STAGING -%}
-  <link href="{{ url_for('static', filename='css/webcompat.min.css')|bust_cache }}" type="text/css" rel="stylesheet">
-{%- else %}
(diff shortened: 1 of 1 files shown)


denschub:
Travis does not like this PR at all - but that's no surprise. Some of the tests depend on CSS to be working - which this PR completely breaks. It's probably best to ignore these failures for now, and go back to a green build once we're done with the refactoring.

webcompat/templates/layout.html
```diff
@@ -27,11 +27,7 @@
   <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-196x196.png') }}" sizes="196x196">
   <meta name="msapplication-TileColor" content="#e4bc05">
   <meta name="msapplication-TileImage" content="{{ url_for('static', filename='favicon/mstile-144x144.png') }}">
-{%- if config.PRODUCTION or config.STAGING -%}
-  <link href="{{ url_for('static', filename='css/webcompat.min.css')|bust_cache }}" type="text/css" rel="stylesheet">
-{%- else %}
-  <link href="{{ url_for('static', filename='css/webcompat.dev.css') }}" type="text/css" rel="stylesheet">
-{%- endif %}
+  <link href="https://webcompat.github.io/design/css/main.css" type="text/css" rel="stylesheet">
```
zoepage: Could you also add following? So we have it all done and the header works? :)

`<link href="https://fonts.googleapis.com/css?family=Open+Sans:400,600|Source+Sans+Pro:300,400|PT+Mono" rel="stylesheet">`

denschub: Sure, done! Added a little `TODO` reminder as well, just to make sure we don't forget! :)

Issue #1917, opened 2017-11-18
Refactor: /contributors


zoepage:
This is a list of issues per unit or page.

Please grab one issue and push this into the issues/1917 branch with the commit message mentioning Issue #IssueNumber to get the reference in the correct issue.

- [x] #1911 - contribute.html
- [x] #1913 - diagnose-bug.html
- [x] #1914 - reproduce-bug.html
- [x] #1915 - site-outreach.html
- [x] #1916 - contributors.html

More context: https://github.com/webcompat/design/issues/48

zoepage:
sweeeeeet!

zoepage:
@denschub we need to add more routes here. could you help?

zoepage:
done by #1917, follow up for content is #1933

PR #1920, opened 2017-11-18
Issue #1917 - Add routes for new contributor sites

denschub:
Add dedicated routes for the new contributor pages as part of #1917.

r? @zoepage


denschub:
Part of #1922 now, so let's close this.

PR #1922, opened 2017-11-18
Fixes #1917 - Add routes for new contributor sites

### LINKED ISSUE #1917: Refactor: /contributors

stefanjudis:
This adds the new structure for the contributors pages which looks like 👇🏻

contributors

The templates could have some logic to avoid code duplication, but I'm not sure if that's needed at this stage.

Also this branch is rebased with the added routes added by @denschub.

I only copied the content of each specific route that was there before. :)

### DIFF
--- a/webcompat/templates/contributors.html
+++ b/webcompat/templates/contributors.html
@@ -2,80 +2,29 @@
 {% block title %}Contributors{% endblock %}
 {% block body %}
 <div>
-{% include "shared/nav.html" %}
-<main role="main">
-  <div class="wc-Hero wc-Hero--contributors">
-    <div class="wc-UIContent">
-      <div class="wc-Hero-wrapper">
-        <div class="wc-Hero-textContributors ">
-          <h1 class="wc-Title wc-Title--xl">
-            <span class="wc-Hero-title js-Hero-title">Welcome aboard!<br/>Get started below...</span>
-          </h1>
-        </div>
-        <div class="wc-Hero-SVG is-active js-Hero-svg">
-          {# TODO - we removed outdated images - include "shared/LightbulbwithBug.svg" #}
-        </div>
-      </div>
-    </div>
-  </div>
-  <div class="contributors">
-    <div class="contributors__item">
-      <div class="contributors__item__header">
-        <div class="wc-UIContent contributors__item__title">
-          <h2 id="reproduce">
-            <button class="contributors__item__btn is-active r-ResetButton" type="button">
-              How to Reproduce a Bug
-              <span class="wc-Icon wc-Icon--contributors wc-Icon--contributors--disabled wc-Icon--minus" aria-hidden="true"></span>
-              <span class="wc-Icon wc-Icon--contributors wc-Icon--contributors--active wc-Icon--plus" aria-hidden="true"></span>
-            </button>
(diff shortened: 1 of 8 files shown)


webcompat/templates/contributors.html
```diff
@@ -2,80 +2,30 @@
...
+    <section class="grid grid-show-gap">
+      <div class="grid-row">
+        <aside class="grid-cell x1-fix">
+          <ol class="sub-nav" class="sub-nav">
+            <li class="sub-nav-header">Table of Contents</li>
+            <li class="sub-nav-active"><a href="/contributors">Join Us</a></li>
+            <li>
+              Get Started
+              <ol class="nested-sub-nav">
+                <li><a href="/contributors/reproduce-bug">Reproduce a Bug</a></li>
```
denschub: Please add `title`s to the `<a>` to the navigation. You can ignore the actual site contents, since those will get rewritten anyway, but the nav links should have titles. :)

stefanjudis: Thx for the feedback. 😊 Can you explain why the nav should have titles? Afaik for assistive technology it doesn't bring any benefit for links. For sighted users it's the same as they see the link text.

Sites that deal with a11y also show that there are no titles included:
https://www.paciellogroup.com/
https://inclusive-components.design/

And quick google research returns this: https://silktide.com/i-thought-title-text-improved-accessibility-i-was-wrong/. :)

zoepage:
@denschub wanna resolve the merge conflict? Also happy to do it. :)

webcompat/templates/contributors.html
zoepage: +1 on the title (in general, not as accessibility issue). Also happy to push a fix here. :)

karlcow: if the title on the anchor is the same as the content of the anchor. It is not very useful and probably even annoying for non-sighted users.

webcompat/templates/contributors.html
```diff
@@ -2,80 +2,30 @@
...
-              <span class="wc-Icon wc-Icon--contributors wc-Icon--contributors--active wc-Icon--plus" aria-hidden="true"></span>
-            </button>
-          </h2>
+  {% include "shared/nav.html" %}
+  <main role="main">
+    <section class="grid grid-show-gap">
+      <div class="grid-row">
+        <aside class="grid-cell x1-fix">
+          <ol class="sub-nav" class="sub-nav">
+            <li class="sub-nav-header">Table of Contents</li>
```
karlcow: Do we really need "Table of Contents" here? It sounds a bit like the infamous "Click Here".

zoepage: This is content related and has his own bug: https://github.com/webcompat/webcompat.com/issues/1933

PR #1935, opened 2017-11-23
Fixes #1861 - Triage Dashboard Design

### LINKED ISSUE #1861: Triage Dashboard Design
Context: #1857 
Depends on: #1858 (HTML template)

This is an area which is completely free. I was [trying to explore ideas](http://www.la-grange.net/tmp/triage.html).
@magsout feel free to propose anything.

capture d ecran 2017-11-02 a 10 59 01

magsout:
@karlcow 

Let's try with this PR ;)

and maybe @miketaylr for the small JS part?


magsout:
@miketaylr @karlcow 

this https://github.com/webcompat/webcompat.com/pull/1935/files#diff-23e0a36bbee957147173901eb77913cbR510 (see https://github.com/webcompat/webcompat.com/issues/1861#issuecomment-345355983) is temporary I guess? Seems not safe ..

webcompat/helpers.py
```diff
@@ -507,7 +507,7 @@ def add_csp(response):
         "font-src 'self'; " +
         "img-src 'self' https://www.google-analytics.com https://*.githubusercontent.com data:; " +  # nopep8
         "manifest-src 'self'; " +
-        "script-src 'self' https://www.google-analytics.com https://api.github.com; " +  # nopep8
+        "script-src 'self' 'unsafe-inline' https://www.google-analytics.com https://api.github.com; " +  # nopep8 @todo removed unsafe-inline
```
karlcow: I wonder if we should have a different policy just for this route.

webcompat/templates/dashboard/triage/top-bar.html
```diff
@@ -0,0 +1,11 @@
+<style>
+  .wc-TopBar {
+    font-size: 1rem;
+    background-color: #ffc900;
+    padding: 1em;
+  }
+</style>
+
+<header class="wc-TopBar" role="banner">
+  <h1 style="margin: 0; font-size: 1.3em">Triage Dashboard</h1>
```
karlcow: We break one test with this change. This is fixable. 
Either by modifying the test by adding the style attribute 
OR by pushing the style into the `style` element with a `.wc-TopBar h1 {margin: 0; font-size: 1.3em} rule.`

```
→ nosetests
................................................F......................
FAIL: Request to /dashboard/triage should be 200.
  File "/Users/karl/code/webcompat.com/tests/test_urls.py", line 139, in test_dashboard_triage
    self.assertTrue('<h1>Triage Dashboard</h1>' in rv.data)
AssertionError: False is not true

Ran 71 tests in 2.823s

FAILED (failures=1)
```

webcompat/templates/dashboard/triage.html
```diff
@@ -1,48 +1,109 @@
...
+        alert("tri non supporté")
+    }
+  }
+
+  /* filteringList */
+  const filteringList = (renderActivityIndicator = true) => {
+    if(renderActivityIndicator) {
+      shouldRenderActivityIndicator()
+    }
+    /* liste triage */
```
karlcow: nit: `Triage list`

webcompat/templates/dashboard/triage.html
```diff
@@ -1,48 +1,109 @@
...
+    const Triages = document.getElementById("js-Triages")
+    if(render) {
+      Triages.classList.remove("is-hidden")
+    }
+    if(!render) {
+      Triages.classList.add("is-hidden")
+    }
+  }
+
+  const filteringSort = () => {
```
karlcow: The filtering doesn't seem to work.

webcompat/templates/dashboard/triage.html
magsout: Ahah french 😏

webcompat/templates/dashboard/triage.html
magsout: hum it seems flex direction behavior does not work in a grid

webcompat/templates/dashboard/triage.html
```diff
@@ -1,48 +1,109 @@
...
+    const Triages = document.getElementById("js-Triages")
+    switch (select.value) {
+      case "":
+        Triages.classList.remove("is-oldest")
+        break;
+      case "oldest":
+        Triages.classList.add("is-oldest")
+        break;
+      default:
+        alert("tri non supporté")
```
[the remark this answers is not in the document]
magsout: 🇫🇷

webcompat/templates/dashboard/triage/filters.html
```diff
@@ -0,0 +1,67 @@
...
+  >
+    <div class="wc-Filter">
+      <select
+        class="wc-Filter-select"
+        name="browsers"
+        id="js-Filter-browsers"
+      >
+        <option value="">All browsers</option>
+        <option value="firefox">Firefox</option>
+        <option value="chrome">Chrome</option>
```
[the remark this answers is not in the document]
magsout: we have a list somewhere on the python side to hydrate this list maybe ?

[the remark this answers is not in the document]
karlcow: We don't but I can try to make sure that the list is limited to a certain number of things. I need to check something.

so currently in the tests I do:

https://github.com/webcompat/webcompat.com/blob/dashboard/tests/test_dashboard.py#L80-L83

But I don't test for `browser-firefox-mobile`, it will return `firefox-mobile` for now, but I can make it return `firefox` only.

What do you think?

@miketaylr @adamopenweb @MDTsai for the main triagers.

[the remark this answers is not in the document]
magsout: yes, sounds good.  👍 

For now I add edge and safari.

webcompat/helpers.py
magsout: Yes, maybe, I guess, I don't know how I can do this .. do you could ?

magsout:
test and typo are fixed. Just need to figure out how to fix the behaviour of the Sort Filter

magsout:
@karlcow  I fixed sort filtering, but it let one issue about layout with this filter. I Don't know how I can avoid this behavior using flexbox. 

Maybe flexbox for this filter is not a good idea, and I have to handle it with JavaScript itself. If you have any idea ;)

webcompat/helpers.py
miketaylr: > I wonder if we should have a different policy just for this route.

I would prefer this, and it would be simpler than implementing nonce support.

webcompat/templates/dashboard/triage.html
```diff
@@ -1,48 +1,126 @@
...
+      {% include "dashboard/triage/stats.html" %}
+      {% include "dashboard/triage/filters.html" %}
+      {% include "dashboard/triage/activity-indicator.html" %}
+      {% include "dashboard/triage/needstriage-list.html" %}
+      {% include "dashboard/triage/no-result.html" %}
+    </div>
   </main>
+<script type="text/javascript">
+  /* Added an event onSubmit form */
+  document.getElementById("js-Filters").addEventListener("submit", (e) => {
```
miketaylr: Just a note: using fat arrow functions here means this will only work in new browsers. I think that's fine for triage, since that's a special type of task, rather than reporting or looking at bugs.

webcompat/helpers.py
miketaylr: Looking at the JS again -- why can't we just stick it in a file in `webcompat/static/js/lib` and link to it in the template? That would mean we don't need to special case it.

@magsout is there some non-obvious reason?
(Also, it will mean you'll need to run prettier on it...).

miketaylr:
JS looks fine, just have the question why we can't stick it in an external file, rather than special case the CSP policy.

webcompat/helpers.py
karlcow: no reason, and yes we can. We can create dashboard.js and put it there.
I guess one of the reasons was to try to minimize the interactions with refactor branch. But I would be in favor of deploying the dashboard (once we merged everything). It seems refactor will take longer than expected.

magsout:
@miketaylr thanks for the review, I will stick all the js in an external file.

For fat arrow, the compatilibity seems ok right now, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Browser_compatibility but if you want I can replace by simple functions

magsout:
@karlcow @miketaylr done

magsout:
@karlcow @miketaylr I guess I let css in html files to avoid conflict with the refactor, what do you think ?

miketaylr:
Awesome -- thanks! Yeah, CSS should be fine. r+ from me. I'll let @karlcow merge in case he has any final thoughts.

webcompat/templates/dashboard/triage/needstriage-list.html
```diff
@@ -0,0 +1,156 @@
...
+            <span>{{ browser }} </span>
+          {% endfor %}
+        </div>
+        <div>
+          <a
+            class="wc-Triage-link"
+            href="/issues/{{ issue['number'] }}"
+            title="{{ issue['title'] }}"
+          >
+            More informations
```
karlcow: ah. @magsout I had missed that. Sorry. 

More information. Without s.
There's no plural for information.

miketaylr:
I might take over the history and tweak for @magsout -- I know he's taking care of sick family right now. And I'd like to deploy this today/tomorrow. Thanks!

miketaylr:
I'm just gonna squash it all, the commits touch quite a bit of unrelated things.

Issue #1826, opened 2017-09-19
Create UI for a "detect bugs with similar domain names" feature


zoepage:
This is a new approach to #307. 
Before I can started on this, I'd like to clarify the requirements and some questions. 

### Intro:
The idea is to create an UI to find issues with similar domain names and maybe even more. This feature should be seen as "power user feature", so it would be enabled for users, who are logged in.  

For the requests the easiest way for a first version would be to use the Github search API.

### Description:
After adding an URL (min. 3 chars) to the input in our bug report form, a dropdown would show up and suggest related issues. Click an issue to "link" it to the one you are reporting right now. Focus on input (always a min. of 3 chars) gives you again the dropdown, if there are still issues that can be suggested. A selected issue can be deleted (see mock).

Suggested issues:
![img_20170919_105920](https://user-images.githubusercontent.com/562354/30584734-f0af3272-9d2a-11e7-801a-8fcb735ec4aa.jpg)

After selecting issues:
![img_20170919_105914](https://user-images.githubusercontent.com/562354/30584696-d2aa589c-9d2a-11e7-8efc-566acaeebb21.jpg)

List of issues (imagine several of those in an actual list ;) ):
![img_20170919_110058](https://user-images.githubusercontent.com/562354/30584772-11c06292-9d2b-11e7-810b-8d86d2d3fdde.jpg)

Single issue:
![img_20170919_105937](https://user-images.githubusercontent.com/562354/30584800-2a7ed7f0-9d2b-11e7-980f-b8b76a29a53d.jpg)

#### What is the criteria? 
- Do we filter by `URL`?
- Do we treat `www.firefox.org` and `www.firefox.org/subpage` the same way? (Search API does, also includes mentions of a string in the comments into the search results)
- Do we filter just by `title` or `content including comments`? (Please remember we changed the working of the types + the titles are not very explicit as they are autogenerated.)
- Do we filter by `URL and keyword`? If both, how would we include this in the search?

### Concerns / Improvements (after 1st version is done):
- Performance probably will become an issue at some point due to the search API requests. This is why saving the issue list in a local database or in-browser storage would be a good way to go. 
We could also add a crobjob at some point to reference the issues and minimize the needed requests.

- What happens when we hit the max of API requests? (Reset after 24h)

### References:
Older info can be found here #307, #864 and #865.

adamopenweb:
[...]

PR #1937, opened 2017-11-27
Fixes #1846

### LINKED ISSUE #1846: Create UI for a "detect bugs with similar descriptions" feature
It's from https://github.com/webcompat/webcompat.com/issues/1826#issuecomment-333793412

**Intro:**
When there is Firefox regression bugs, we'll see users report same issues but they have same root cause. It comes into my mind if we can let user know other users already reported similar issue, they can see what's happening and vote/+1/vouch same issue happen to him/her.

**Description:**
While user is typing in the description and more than 2 words, use the keyword to search existing issue description and find similar description.

Click on the suggested issue opens a new tab, see what's going on and do the following. Agree with +1 or disagree then close the tab.

**TBD:**
- What's the better way to guide users have discussion similar description in an issue?
- It's doable via github search API or we need another DB/end point to do this?

MDTsai:
First prototype to make sure this is what we want for similar bug suggestion. r?=@zoepage

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -71,6 +71,7 @@ function BugForm() {
     this.disableSubmits();
     this.urlField.on("blur input", _.bind(this.checkURLValidity, this));
     this.urlField.on("input", _.bind(this.showSimilarBugs, this));
+    this.urlField.on("blur", _.bind(this.hideSimilarBugs, this));
     this.descField.on(
       "blur input",
       _.bind(this.checkDescriptionValidity, this)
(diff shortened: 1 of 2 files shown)


webcompat/static/js/lib/bugform.js
```diff
@@ -216,6 +219,60 @@ function BugForm() {
...
+    }
+
+    if (this.suggestDelay) {
+      clearTimeout(this.suggestDelay);
+    }
+    var bugform = this;
+    this.suggestDelay = setTimeout(function() {
+      try {
+        var url = new URL(bugform.urlField.val());
+        if (url.hostname.length > 7) {
```
zoepage: https://github.com/webcompat/webcompat.com/issues/1826#issuecomment-332374279 > "About the search API" 

> "... fire the event starting at the 5th character ..."

Would be great, if we could change the 7 to a 4 then. thanks!

zoepage:
@MDTsai one small change. otherwise the code looks good. :)

webcompat/static/js/lib/bugform.js
```diff
@@ -216,6 +219,60 @@ function BugForm() {
...
+          var query = `https://api.github.com/search/issues?page=1&per_page=10&stage=all&sort=created&q=${url.hostname}+in:title+repo%3Awebcompat%2Fweb-bugs&order=desc`;
+
+          fetch(query)
```
miketaylr: This is fine for a prototype, but we need to be maximally compatible with old browsers for bug reporting:

https://caniuse.com/#feat=fetch

MDTsai: +1, so this is only for prototyping, let's discuss this in all hands to know if this really improve bug quality? Any change could make it better?

zoepage: +1

webcompat/static/js/lib/bugform.js
zoepage: @MDTsai Would you like to make the small fix? Or shall I add a follow up? I don't want to block the merge with this small request. :)

zoepage: let's discuss the char length at some other point. Merging, so team can review.

miketaylr:
@zoepage @MDTsai FYI

https://travis-ci.org/webcompat/webcompat.com/builds/315353544#L612

zoepage:
That happens due the ES6 features as `grunt-contrib-uglify` v2 is all about ES5. On it!

PR #1944, opened 2017-11-28
Fixes #1938: Add "extra labels" in new issue webhook

### LINKED ISSUE #1938: type-stylo label added, then removed
I plan on removing the `type-stylo` label stuff this week, but I'd like to understand what's going on and fix it (because we want to be able to add labels via GET params in the future).

For example, in https://github.com/webcompat/web-bugs/issues/13877:

1. `type-stylo` label is added
2. bug is added to needstriage milestone
3. priority label and browser label are added (which removes `type-stylo`).

miketaylr:
Tested with this branch deployed to staging:

with &label=type-stylo in params:

https://staging.webcompat.com/issues/1310

without label param (just to prove things still work):

https://staging.webcompat.com/issues/1311

### DIFF
--- a/tests/test_webhook.py
+++ b/tests/test_webhook.py
@@ -50,11 +50,13 @@ class TestWebhook(unittest.TestCase):
         <!-- @browser: Firefox 55.0 -->
         <!-- @ua_header: Mozilla/5.0 (what) Gecko/20100101 Firefox/55.0 -->
         <!-- @reported_with: web -->
+        <!-- @extra_label: type-media -->
 
         **URL**: https://www.example.com/
         **Browser / Version**: Firefox 55.0
         <!-- @browser: Chrome 48.0 -->
         """
+
         self.issue_body2 = """
         <!-- @browser: Foobar -->
         """
@@ -131,17 +133,37 @@ class TestWebhook(unittest.TestCase):
         self.assertEqual(rv.mimetype, 'text/plain')
         self.assertEqual(rv.data, 'Not an interesting hook')
 
+    def test_extract_metadata(self):
+        expected = {'reported_with': 'web', 'extra_label': 'type-media',
+            'ua_header': 'Mozilla/5.0 (what) Gecko/20100101 Firefox/55.0',
+            'browser': 'Firefox 55.0'}
+        actual = helpers.extract_metadata(self.issue_body)
+        self.assertEqual(expected, actual)
+
     def test_extract_browser_label(self):
         """Extract browser label name."""
-        browser_label = helpers.extract_browser_label(self.issue_body)
-        self.assertEqual(browser_label, 'browser-firefox')
-        browser_label_none = helpers.extract_browser_label(self.issue_body2)
-        self.assertEqual(browser_label_none, None)
-        browser_label_paren = helpers.extract_browser_label(self.issue_body3)
-        self.assertEqual(browser_label_paren, 'browser-firefox-mobile-tablet')
-        browser_label_unicode = helpers.extract_browser_label(self.issue_body4)
-        self.assertEqual(
-            browser_label_unicode, 'browser-firefox-mobile-tablet')
+        metadata_tests = [
+            ({'browser': 'Firefox'}, None),
+            ({'browser': 'Firefox Mobile'}, None),
+            ({'browser': 'Firefox99.0'}, None),
+            ({'browser': 'Firefox (tablet)'}, None),
+            ({'browser': 'Firefox 30.0'}, 'browser-firefox'),
+            ({'browser': 'Firefox Mobile 30.0'}, 'browser-firefox-mobile'),
+            ({'browser': 'Firefox Mobile (Tablet) 88.0'}, 'browser-firefox-mobile-tablet')
+        ]
+        for metadata_dict, expected in metadata_tests:
+            actual = helpers.extract_browser_label(metadata_dict)
+            self.assertEqual(expected, actual)
+
+    def test_extract_extra_label(self):
+        """Extract 'extra' label."""
+        metadata_tests = [
+            ({'extra_label': 'type-media'}, 'type-media'),
+            ({'burgers': 'french fries'}, None)
+        ]
+        for metadata_dict, expected in metadata_tests:
+            actual = helpers.extract_extra_label(metadata_dict)
+            self.assertEqual(expected, actual)
 
     def test_extract_priority_label(self):
         """Extract priority label."""
--- a/tests/test_webhook.py
+++ b/tests/test_webhook.py
@@ -143,6 +143,16 @@ class TestWebhook(unittest.TestCase):
         self.assertEqual(
             browser_label_unicode, 'browser-firefox-mobile-tablet')
 
+    def test_extract_priority_label(self):
+        """Extract priority label."""
+        with patch('webcompat.db.site_db.query') as db_mock:
+            db_mock.return_value.filter_by.return_value = [
+                Site('google.com', 1, '', 1)]
+            priority_label = helpers.extract_priority_label(self.issue_body3)
+            self.assertEqual(priority_label, 'priority-critical')
+        priority_label_none = helpers.extract_priority_label(self.issue_body)
+        self.assertEqual(priority_label_none, None)
+
     def test_is_github_hook(self):
         """Validation tests for GitHub Webhooks."""
         json_event, signature = event_data('new_event_invalid.json')
@@ -195,16 +205,6 @@ class TestWebhook(unittest.TestCase):
                     'domain': 'www.chia-anime.tv'}
         actual = helpers.get_issue_info(payload)
         self.assertDictEqual(expected, actual)
-
-    def test_extract_priority_label(self):
-        """Extract priority label."""
-        with patch('webcompat.db.site_db.query') as db_mock:
-            db_mock.return_value.filter_by.return_value = [
-                Site('google.com', 1, '', 1)]
-            priority_label = helpers.extract_priority_label(self.issue_body3)
-            self.assertEqual(priority_label, 'priority-critical')
-        priority_label_none = helpers.extract_priority_label(self.issue_body)
-        self.assertEqual(priority_label_none, None)
 
 
 if __name__ == '__main__':
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -210,9 +210,13 @@ def build_formdata(form_object):
     else:
         summary = '{0} - {1}'.format(normalized_url, problem_summary)
 
+    metadata_keys = ['browser', 'ua_header', 'reported_with']
+    extra_label = form_object.get('extra_label', None)
+    if extra_label in app.config['EXTRA_LABELS']:
+        metadata_keys.append('extra_label')
+
     formdata = {
-        'metadata': get_metadata(('browser', 'ua_header', 'reported_with'),
-                                 form_object),
+        'metadata': get_metadata(metadata_keys, form_object),
         'url': form_object.get('url'),
         'browser': form_object.get('browser'),
         'os': form_object.get('os'),
@@ -245,7 +249,4 @@ def build_formdata(form_object):
     # Append "from webcompat.com" message to bottom (for GitHub issue viewers)
     body += u'\n\n{0}'.format(GITHUB_HELP)
     rv = {'title': summary, 'body': body}
-    extra_label = form_object.get('label', None)
-    if extra_label and extra_label in app.config['EXTRA_LABELS']:
-        rv.update({'labels': [extra_label]})
     return rv
--- a/webcompat/webhooks/helpers.py
+++ b/webcompat/webhooks/helpers.py
@@ -55,6 +55,26 @@ def extract_extra_label(metadata_dict):
     return extra_label
 
 
+def extract_priority_label(body):
+    """Parse url from body and query the priority labels."""
+    hostname = domain_name(extract_url(body))
+    if hostname:
+        priorities = ['critical', 'important', 'normal']
+        # Find host_name in DB
+        for site in site_db.query(Site).filter_by(url=hostname):
+            return 'priority-{}'.format(priorities[site.priority - 1])
+        # No host_name in DB, find less-level domain (>2)
+        # If host_name is lv4.lv3.example.com, find lv3.example.com/example.com
+        subparts = hostname.split('.')
+        domains = ['.'.join(subparts[i:])
+                   for i, subpart in enumerate(subparts)
+                   if 0 < i < hostname.count('.')]
+        for domain in domains:
+            for site in site_db.query(Site).filter_by(url=domain):
+                return 'priority-{}'.format(priorities[site.priority - 1])
+    return None
+
+
 def update_issue(payload, issue_number):
     """Does a GitHub PATCH request to set labels and milestone for the issue.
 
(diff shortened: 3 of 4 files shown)


webcompat/form.py
```diff
@@ -210,9 +210,13 @@ def build_formdata(form_object):
     else:
         summary = '{0} - {1}'.format(normalized_url, problem_summary)
 
+    metadata_keys = ['browser', 'ua_header', 'reported_with']
+    extra_label = form_object.get('extra_label', None)
+    if extra_label and extra_label in app.config['EXTRA_LABELS']:
```
karlcow: @miketaylr You do not need to test both here. See

```python
>>> # app.config['EXTRA_LABELS'] == EXTRA_LABELS for the demo.
>>> EXTRA_LABELS = [
...     'type-media',
...     'type-webvr',
...     'type-stylo'
... ]
>>> metadata_keys = ['browser', 'ua_header', 'reported_with']
>>> # Case for an expected value
>>> extra_label = 'type-media'
>>> if extra_label in EXTRA_LABELS:
...     metadata_keys.append('extra_label')
... 
>>> metadata_keys
['browser', 'ua_header', 'reported_with', 'extra_label']
>>> # Reinitialize
>>> metadata_keys = ['browser', 'ua_header', 'reported_with']
>>> # Case for a not accepted value
>>> extra_label = 'lol'
>>> if extra_label in EXTRA_LABELS:
...     metadata_keys.append('extra_label')
... 
>>> metadata_keys
['browser', 'ua_header', 'reported_with']
>>> # Case for None
>>> extra_label = None
>>> if extra_label in EXTRA_LABELS:
...     metadata_keys.append('extra_label')
... 
>>> metadata_keys
['browser', 'ua_header', 'reported_with']
```

webcompat/form.py
```diff
@@ -210,9 +210,13 @@ def build_formdata(form_object):
...
+        metadata_keys.append('extra_label')
+
     formdata = {
-        'metadata': get_metadata(('browser', 'ua_header', 'reported_with'),
-                                 form_object),
+        'metadata': get_metadata(metadata_keys, form_object),
```
karlcow: +1  for the readibility!

webcompat/webhooks/helpers.py
```diff
@@ -140,11 +157,11 @@ def new_opened_issue(payload):
...
     browser_label = extract_browser_label(issue_body)
+    extra_label = extract_extra_label(issue_body)
     priority_label = extract_priority_label(issue_body)
-    if browser_label:
-        labels.append(browser_label)
-    if priority_label:
-        labels.append(priority_label)
+    for label in (browser_label, extra_label, priority_label):
+        if label:
+            labels.append(label)
```
karlcow: so here it smells list comprehension :)
Let's remove the top `labels = []`

```python
>>> # Let's define the labels.
>>> browser_label = 'browser-firefox'
>>> extra_label = None
>>> priority_label = 'priority-important'
>>> # Without an if, we would get None in the list.
>>> labels = [label for label in (browser_label, extra_label, priority_label)]
>>> labels
['browser-firefox', None, 'priority-important']
>>> # Not what we want, but we can test for this with a conditional list comprehension
>>> labels = [label for label in (browser_label, extra_label, priority_label) if label is not None]
>>> labels
['browser-firefox', 'priority-important']
>>> # Yeah. Oh and if you are worried about the case where everything is None. We got you covered here.
>>> [label for label in (None, None, None) if label is not None]
[]
```

miketaylr:
> There are missing tests.

Ha yeah, just came here to comment that I still have tests to write. :)

Thanks for review so far!

karlcow:
huh… I had a full comment about changing the strategy on helpers.py and it's not there. huh.

webcompat/webhooks/helpers.py
```diff
@@ -39,6 +39,23 @@ def extract_browser_label(body):
...
 
 
+def extract_extra_label(body):
+    """Parse the "extra label" from the body in comment:
+
+    <!-- @extra_label: value -->.
+    Currently this only handles a single label,
+    because that's all that we set in webcompat.com.
+    """
+    match_list = re.search(r'<!--\s@extra_label:\s([\w-]+)\s-->', body)
```
karlcow: So I guess I have forgotten this part here. There was a long comment which is gone. Sniff. 

Let me try to reassess what I was saying. We have two extract using regex for parsing the body. Twice basically. We could change the strategy and code a bit differently.

This is our issue_body. Instead of parsing twice, We could parse once with a more generic function.

```python
>>> issue_body = """
...         <!-- @browser: Firefox 55.0 -->
...         <!-- @ua_header: Mozilla/5.0 (what) Gecko/20100101 Firefox/55.0 -->
...         <!-- @reported_with: web -->
...         <!-- @extra_label: type-media -->
...          
...         **URL**: https://www.example.com/
...         **Browser / Version**: Firefox 55.0
...         <!-- @browser: Chrome 48.0 -->
...         """
>>> match_list = re.findall(r'<!--\s@(\w+):\s(.+)\s-->', issue_body)
>>> match_list
[('browser', 'Firefox 55.0'), ('ua_header', 'Mozilla/5.0 (what) Gecko/20100101 Firefox/55.0'), ('reported_with', 'web'), ('extra_label', 'type-media'), ('browser', 'Chrome 48.0')]
```

This would return a list of tuples that we could use to extract the data for browser name etc. 

but there is better, this is a list of tuples with unique names.

```python
>>> dict(match_list)
{'reported_with': 'web', 'extra_label': 'type-media', 'ua_header': 'Mozilla/5.0 (what) Gecko/20100101 Firefox/55.0', 'browser': 'Chrome 48.0'}
>>> metadata = dict(match_list)
>>> metadata['extra_label']
'type-media'
```

You have a dictionary fully accessible. 
This would require to modify the tests. but could be a nice way of doing.

miketaylr: I like it.

webcompat/webhooks/helpers.py
miketaylr: Heh, I think nearly every Python review has some instance of me not remembering to use a list comprehension. :)

miketaylr:
OK, there's a few things I don't *love* about this, will comment inline. But it works, yay?

tests/test_webhook.py
```diff
@@ -133,15 +139,41 @@ def test_fails_on_not_known_action(self):
 
     def test_extract_browser_label(self):
         """Extract browser label name."""
-        browser_label = helpers.extract_browser_label(self.issue_body)
+        browser_label = helpers.extract_browser_label(
+            helpers.extract_metadata(self.issue_body))
         self.assertEqual(browser_label, 'browser-firefox')
-        browser_label_none = helpers.extract_browser_label(self.issue_body2)
+        browser_label_none = helpers.extract_browser_label(
+            helpers.extract_metadata(self.issue_body2))
```
miketaylr: I guess we had relied on the notion that `<!-- @browser: Foobar -->` would be invalid, because it doesn't have a version number. I don't think anything would actually break, if we allowed it. At least not for labels.

But I had to enforce this in `extract_browser_label`, and added the extra `issue_body5` to test this (as well as test the ordering stuff).

webcompat/webhooks/helpers.py
```diff
@@ -17,20 +17,30 @@
...
-    if match_list:
-        # perhaps we do something more interesting depending on
-        # what groups(n)[0] is in the future.
-        # right now, match_list.groups(0) should look like:
-        # ('browser', 'firefox')
-        browser = match_list.groups(0)[1].lower()
+    match_list = re.findall(r'<!--\s@(\w+):\s(.+)\s-->', body)
+    # Build a dict but only for the first seen instance of a key
+    # (in case there are duplicates added by the user after)
+    metadata_dict = {}
```
miketaylr: /me images Karl suggesting some cool dict comprehension to use here....

So, we can't just `dict(match_list)`, because if there were duplicate keys, the `dict()` constructor swallows them all except the last in the list. Which is the opposite of what we want, I think... 

https://github.com/webcompat/webcompat.com/blob/54e0d5d5a9f2bcfe143fa652894b4703591678c1/tests/test_webhook.py#L56

I don't recall if this was on purpose, but I think the logic here is in case the user somehow adds their own (incorrect) metadata comment, we are interested in the first that the app added. Which makes sense to me. Hence, this constructing the `metadata_dict` by hand.

miketaylr: BTW, I double checked and `re.findall` will insert into the list from left to right, in order. So that's good.

webcompat/webhooks/helpers.py
```diff
@@ -17,20 +17,30 @@
...
+    for key, value in match_list:
+        if key not in metadata_dict:
+            metadata_dict[key] = value
+    return metadata_dict
+
+
+def extract_browser_label(metadata_dict):
```
miketaylr: This method didn't get simplified that much... hooray for tests though. :)

webcompat/webhooks/helpers.py
```diff
@@ -17,20 +17,30 @@
...
+    """Return the browser label from metadata_dict."""
+    browser = metadata_dict.get('browser', None)
+    # Only proceed if browser looks like "FooBrowser 99.0"
+    if browser and re.search(r'([^\d]+?)\s[\d\.]+', browser):
```
miketaylr: My first attempt didn't have this regexp... but that would return the wrong result for something like `Firefox Mobile` (`browser-firefox`). So.

webcompat/webhooks/helpers.py
karlcow: @miketaylr :) Ready for the fun?

```python
>>> foo = [('love', 'wonderful'), ('peace', 'awesome'), ('love', 'overrated')]
>>> foo
[('love', 'wonderful'), ('peace', 'awesome'), ('love', 'overrated')]
>>> dict(foo)
{'peace': 'awesome', 'love': 'overrated'}
>>> # Not what we want :) ok let's do a trick
>>> dict(reversed(foo))
{'peace': 'awesome', 'love': 'wonderful'}
```

So basically we reverse the order of the list so the last (once first) tuples will be the ones we keep. ;)

```python
metadata_dict = dict(reversed(match_list))
```

webcompat/webhooks/helpers.py
```diff
@@ -39,6 +49,36 @@ def extract_browser_label(body):
...
 
+def extract_extra_label(metadata_dict):
+    """Return the 'extra' label from metadata_dict"""
+    extra_label = metadata_dict.get('extra_label', None)
+    if extra_label:
+        extra_label = extra_label.lower()
+        return extra_label.encode('utf-8')
+    else:
+        return None
+
```
karlcow: Why the `else`? the line before the `if`, `None is already assigned`

```python
def extract_extra_label(metadata_dict):
    """Return the 'extra' label from metadata_dict"""
    extra_label = metadata_dict.get('extra_label', None)
    if extra_label:
        extra_label = extra_label.lower()
        extra_label = extra_label.encode('utf-8')
    return extra_label
```

Let's test this

```python
>>> def extract_extra_label(metadata_dict):
...     """Return the 'extra' label from metadata_dict"""
...     extra_label = metadata_dict.get('extra_label', None)
...     if extra_label:
...         extra_label = extra_label.lower()
...         extra_label = extra_label.encode('utf-8')
...     return extra_label
... 
>>> metadata_dict = {'extra_label': 'yohoo', 'foo': 'bar'}
>>> extract_extra_label(metadata_dict)
'yohoo'
>>> metadata_dict = {'foo': 'bar'}
>>> extract_extra_label(metadata_dict)
>>> 
```

webcompat/webhooks/helpers.py
karlcow: What about we do this instead:

```python
def extract_browser_label(metadata_dict):
    """Return the browser label from metadata_dict."""
    browser = metadata_dict.get('browser', None)
    if browser:
        browser = browser.encode('utf-8')
        browser = browser.lower()
        browser = browser.translate(None, '()0123456789.')
        browser = '-'.join(browser.split())
        browser = 'browser-{name}'.format(name=browser)
    return browser
```

And tests.

```python
>>> metadata_dict = {'browser' : 'Firefox'}
>>> extract_browser_label(metadata_dict)
'browser-firefox'
>>> metadata_dict = {'browser' : 'Firefox 30.0'}
>>> extract_browser_label(metadata_dict)
'browser-firefox'
>>> metadata_dict = {'browser' : 'Firefox Mobile 30.0'}
>>> extract_browser_label(metadata_dict)
'browser-firefox-mobile'
>>> metadata_dict = {'browser' : 'Firefox Mobile'}
>>> extract_browser_label(metadata_dict)
'browser-firefox-mobile'
>>> metadata_dict = {'browser' : 'Firefox (tablet)'}
>>> extract_browser_label(metadata_dict)
'browser-firefox-tablet'
```

tests/test_webhook.py
```diff
@@ -133,15 +139,41 @@ def test_fails_on_not_known_action(self):
 
     def test_extract_browser_label(self):
         """Extract browser label name."""
-        browser_label = helpers.extract_browser_label(self.issue_body)
+        browser_label = helpers.extract_browser_label(
+            helpers.extract_metadata(self.issue_body))
```
karlcow: chaining functions in unittest is usually not a good idea. Just test with the input which is  expected to isolate the tests. So here just dictionaries.

```python
def test_extract_browser_label(self):
    """Extract browser label name."""
    metadata_tests = [
        ({'browser': 'Firefox'}, 'browser-firefox'),
        ({'browser': 'Firefox Mobile'}, 'browser-firefox-mobile'),
        ({'browser': 'Firefox 30.0'}, 'browser-firefox'),
        ({'browser': 'Firefox Mobile 30.0'}, 'browser-firefox-mobile'),
        ({'browser': 'Firefox (tablet)'}, 'browser-firefox-tablet')
        ]
    for metadata_dict, expected in metadata_tests:
        actual = helpers.extract_browser_label(metadata_dict)
```

tests/test_webhook.py
karlcow: see my comments in creating the metadata_dict

tests/test_webhook.py
```diff
@@ -133,15 +139,41 @@ def test_fails_on_not_known_action(self):
...
+        self.assertEqual(browser_label_none2, None)
+
+    def test_extract_extra_label(self):
+        """Extract 'extra' label."""
+        extra_label = helpers.extract_extra_label(
+            helpers.extract_metadata(self.issue_body))
+        self.assertEqual(extra_label, 'type-media')
+        extra_label_none = helpers.extract_extra_label(
+            helpers.extract_metadata(self.issue_body2))
+        self.assertEqual(extra_label_none, None)
```
karlcow: same comment here about chaining. Not good for unittests.

webcompat/webhooks/helpers.py
miketaylr: This doesn't work, `{'browser' : 'Firefox'}` should return `None`, because it's missing a version number. Same for the last 2. I think we're splitting hairs on this part, I recommend we land it as-is and if you'd like to file a follow up issue to refactor away from regexp, that's fine.

webcompat/webhooks/helpers.py
miketaylr: yep, thanks.

tests/test_webhook.py
miketaylr: > see my comments in creating the metadata_dict

Yeah, saw that. I'm going to leave it as is for now. Thanks.

miketaylr:
OK, round 3. r? @karlcow

karlcow:
There is a check fail because of indentation and line length in test_webhook, I'll fix it in a follow up issue.

karlcow:
Thanks a lot @miketaylr

miketaylr:
> There is a check fail because of indentation and line length in test_webhook, I'll fix it in a follow up issue.

(I pushed a commit that fixed this on master... sorry about that)

webcompat/webhooks/helpers.py
karlcow: Followup issue for this part https://github.com/webcompat/webcompat.com/issues/1949
Also we forgot to remove the  `else: return None` ;)

PR #1956, opened 2017-12-04
Fixes #1850. Remove redundant test for details param.

### LINKED ISSUE #1850: "layout.css.servo.enabled: true" Should probably not be in the form.
Currently when opening a new issue the form has 
capture d ecran 2017-10-26 a 16 03 35

"layout.css.servo.enabled: true"

This should probably not be editable by the user. Just be a label (maybe it's already is).
Or added to the report after the user puts the description and submit the issue.

miketaylr:
This is already tested @ https://github.com/webcompat/webcompat.com/blob/5b6cdd9c0bf8cde7f9a35095f081c8a41d6aa37b/tests/functional/reporting-non-auth.js#L288-L303

### DIFF
--- a/tests/functional/reporting-non-auth.js
+++ b/tests/functional/reporting-non-auth.js
@@ -255,21 +255,6 @@ define(
         );
       },
 
-      // tests a scenario where bug reported from firefox nightly Report Site Issue button
-      "details param adds stylo pref to text area": function() {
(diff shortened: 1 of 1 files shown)


zoepage:
LTGM.

miketaylr:
Thanks.

PR #1962, opened 2017-12-05
Fixes #1960. Add type-webrender-enabled to EXTRA_LABELS list.

### LINKED ISSUE #1960: Add webrender labels to EXTRA_LABELS whitelist
We'd like to track if people have webrender prefs enabled.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1422985

miketaylr:
(also alphabetize the list)

r? @karlcow

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -138,8 +138,9 @@ for cat_label in cat_labels:
 # creating an issue.
 EXTRA_LABELS = [
     'type-media',
+    'type-stylo',
+    'type-webrender-enabled',
     'type-webvr',
-    'type-stylo'
 ]
 
 from webcompat import app


config/__init__.py
```diff
@@ -138,8 +138,9 @@ def convert_milestones(milestones_content):
 # creating an issue.
 EXTRA_LABELS = [
     'type-media',
+    'type-stylo',
+    'type-webrender-enabled',
     'type-webvr',
```
karlcow: I don't know if it's on purpose. The trailing comma doesn't matter. It's not improper both in style and validity, just noticing because it was not like this just before.

karlcow:
Thanks @miketaylr

PR #1964, opened 2017-12-05
Fixes #1893 - removed unused template

### LINKED ISSUE #1893: refactor template: issue-item.html
To see this, you need to log in... https://webcompat.com/activity/

Refactor following:
https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/shared/issue-item.html

with:

(if the line is wrong, look for id="issues-list" > class="issue")

Please be aware to add the placeholders for links as they are in the old templates.

zoepage:
r? @miketaylr


miketaylr:
Thanks.

Issue #1529, opened 2017-04-25
Audit our image loader usage


miketaylr:
Where do we show loader images and in what circumstances?

Related to #746 and #1000.

miketaylr:
Ping @zoepage

karlcow:
So going through
https://www.webpagetest.org/result/170727_M0_FAA/

I noticed that we were sending two upload images.
One is http://localhost:5000/img/upload-loader.svg
and the other one is http://localhost:5000/img/loader.gif

The SVG is used now on the site.

But there is still a request done for loader.gif from the CSS.

```css
.wc-Loader {
 display:none;
 position:fixed;
 background-image:url(../../../img/loader.gif);
 z-index:10;
 background-color:rgba(255,255,255,0);
 top:0;
 left:0;
 right:0;
 bottom:0;
 background-position:center center;
 background-repeat:no-repeat
}
```

miketaylr:
Thanks @karlcow! We def don't need 2. :)

zoepage:
The issues/1529 shall also tackle #1847

PR #1965, opened 2017-12-05
Fixes #1897 - refactor template new-issues

### LINKED ISSUE #1897: refactor template: new-issue.html
Refactor following:
https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/new-issue.html

with following semantics:

```
{%- extends "layout.html" -%}
{% block title %}New Issue{% endblock %}
{%- block body -%}
{% include "topbar.html" %}
{% include "home-page/form.html" %}
{% endblock %}
```

zoepage:
r?  @miketaylr

### DIFF
--- a/webcompat/templates/new-issue.html
+++ b/webcompat/templates/new-issue.html
@@ -2,8 +2,5 @@
 {% block title %}New Issue{% endblock %}
 {%- block body -%}
 {% include "topbar.html" %}
-<main role="main">
-  <div class="wc-Loader js-Loader"></div>
-  {% include "home-page/form.html" %}
-</main>
+{% include "home-page/form.html" %}
 {% endblock %}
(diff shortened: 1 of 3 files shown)


webcompat/templates/new-issue.html
```diff
@@ -2,8 +2,5 @@
 {% block title %}New Issue{% endblock %}
 {%- block body -%}
 {% include "topbar.html" %}
-<main role="main">
-  <div class="wc-Loader js-Loader"></div>
```
miketaylr: https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L83

Is there an issue on file to make sure this doesn't regress when the refactor lands?

I know we have https://github.com/webcompat/webcompat.com/issues/1529, but that's very general... and I notice the refactor is either commenting out loader images, or removing them.

miketaylr: We chatted about this in a 1:1, @zoepage is going to move the loader one level deeper.

zoepage: Just to add some context here, the loaders have been commented out to we don't forget them and I can look for them, when working on #1529.

miketaylr:
r_

Issue #1971, opened 2017-12-06
Investigate "browser-firefox-mobile-nightly-59.0a1" label on issue


miketaylr:
I don't know why this label would be applied, rather than just `browser-fireffox-mobile`. We have a few `firefox-mobile-nightly` labeled issues as well: https://github.com/webcompat/web-bugs/labels/browser-firefox-mobile-nightly

https://github.com/webcompat/web-bugs/issues/14071 is the issue reported by @softvision-oana-arbuzov.

Here's the comment metadata:

```
<!-- @browser: Firefox Mobile Nightly 59.0a1 (2017-12-04) -->
<!-- @ua_header: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:59.0) Gecko/20100101 Firefox/59.0 -->
```

Maybe the user added it manually to the browser field? Something for us to normalize in webhook/labeler code.

karlcow:
<img width="738" alt="capture d ecran 2017-12-07 a 00 06 06" src="https://user-images.githubusercontent.com/505230/33668221-78048016-dae2-11e7-9e83-e57179ac2ad0.png">

miketaylr:
I wonder if fixing https://github.com/webcompat/webcompat.com/issues/1949 would solve this... if `@browser` only ever contained the name, presumably version info and dates wouldn't sneak in.

karlcow:
```python
>>> browser = 'Firefox Mobile Nightly 59.0a1 (2017-12-04)'
>>> import re
>>> re.search(r'([^\d]+?)\s[\d\.]+', browser)
<_sre.SRE_Match object at 0x101a4fd50>
>>> browser = browser.lower()
>>> browser = browser.rsplit(' ', 1)[0]
>>> browser
'firefox mobile nightly 59.0a1'
>>> browser = browser.encode('utf-8')
>>> browser
'firefox mobile nightly 59.0a1'
>>> browser = browser.translate(None, '()')
>>> browser
'firefox mobile nightly 59.0a1'
>>> dash_browser = '-'.join(browser.split())
>>> dash_browser
'firefox-mobile-nightly-59.0a1'
>>> 'browser-{name}'.format(name=dash_browser)
'browser-firefox-mobile-nightly-59.0a1'
```

The `browser.rsplit(' ', 1)[0]` doesn't answer this use case.

karlcow:
yes I want to rationalize things and probably #1949 can solve it. I could propose something tomorrow.

[...]

PR #1974, opened 2017-12-07
Fixes #1973 - Adds normalization of metadata values.

### LINKED ISSUE #1973: New issue form needs to be stricter
Currently we accept a couple of things that maybe we should be stricter about in terms of validation.

* [ ] browser string. https://github.com/webcompat/webcompat.com/blob/e93f0f9921556f725676b9033724cc4a7a373889/webcompat/form.py#L68
* [ ] os string https://github.com/webcompat/webcompat.com/blob/e93f0f9921556f725676b9033724cc4a7a373889/webcompat/form.py#L69

karlcow:
Following discussions in #1971, the goal of this pull request is to sanitize a bit. This is not perfect but should remove some of the issue. It goes on the side of safe play. For example if it detects there is `>` or `<` it just sends back an empty string instead of trying to clean up.

### DIFF
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-'''Tests for form validation.'''
+"""Tests for form validation."""
 
 import unittest
 
@@ -12,16 +12,19 @@ FIREFOX_UA = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:48.0) Gecko/20100
 
 
 class TestForm(unittest.TestCase):
+    """Module for testing the form."""
 
     def setUp(self):
+        """Set up."""
         webcompat.app.config['TESTING'] = True
         self.app = webcompat.app.test_client()
 
     def tearDown(self):
+        """Tear down."""
         pass
 
     def test_normalize_url(self):
-        """Checks that URL is normalized."""
+        """Check that URL is normalized."""
         r = form.normalize_url('http://example.com')
         self.assertEqual(r, 'http://example.com')
 
@@ -50,7 +53,7 @@ class TestForm(unittest.TestCase):
         self.assertIsNone(r)
 
     def test_domain_name(self):
-        """Checks that domain name is extracted."""
+        """Check that domain name is extracted."""
         r = form.domain_name('http://example.com')
         self.assertEqual(r, 'example.com')
 
@@ -61,7 +64,7 @@ class TestForm(unittest.TestCase):
         self.assertIsNone(r)
 
     def test_metadata_wrapping(self):
-        """Checks that metadata is processed and wrapped."""
+        """Check that metadata is processed and wrapped."""
         TEST_DICT = {'cool': 'dude', 'wow': 'ok'}
         EXPECTED_SINGLE = '<!-- @cool: dude -->\n'
         EXPECTED_MULTIPLE = '<!-- @cool: dude -->\n<!-- @wow: ok -->\n'
@@ -73,7 +76,7 @@ class TestForm(unittest.TestCase):
         self.assertEqual(r, EXPECTED_MULTIPLE)
 
     def test_radio_button_label(self):
-        '''Checks that appropriate radio button label is returned.'''
+        """Check that appropriate radio button label is returned."""
         TEST_LABELS_LIST = [
             (u'detection_bug', u'Desktop site instead of mobile site'),
             (u'unknown_bug', u'Something else')
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -112,6 +112,20 @@ class TestForm(unittest.TestCase):
         expected = u'<!-- @sky: blue -->\n<!-- @earth: None -->\n'
         self.assertEqual(actual, expected)
 
+    def test_normalize_metadata(self):
+        """Avoid some type of strings."""
+        cases = [('blue sky -->', 'blue sky'),
+                 ('blue sky ---->>', 'blue sky'),
+                 ('', ''),
+                 ('blue sky ', 'blue sky'),
+                 ('bad_bird <script>', ''),
+                 ('bad_bird <script-->>', ''),
+                 ('a' * 300, ''),
+                 (None, None)
+                 ]
+        for meta_value, expected in cases:
+            self.assertEqual(form.normalize_metadata(meta_value), expected)
+
     def test_build_formdata(self):
         """The data body sent to GitHub API."""
         form_object = {'foo': 'bar'}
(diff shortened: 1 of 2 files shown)


webcompat/form.py
```diff
@@ -154,6 +159,22 @@ def normalize_url(url):
     return url
 
 
+def normalize_metadata(metadata_value):
+    """Normalize the metadata received from the form."""
+    # Removing closing comments.
```
zoepage: Why do we remove just the closing comments? Just wondering about the context here.

karlcow: @zoepage good question.
The reason is that an opening comment will not do much as just being hidden by the other opening comment, but a closing comment will terminate the opening comment.

This is what our issues look like.

```
<!-- @browser: Firefox 59.0 -->
<!-- @ua_header: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:59.0) Gecko/20100101 Firefox/59.0 -->
<!-- @reported_with: desktop-reporter -->
<!-- @extra_label: type-stylo -->

**URL**: https://web.telegram.org/

**Browser / Version**: Firefox 59.0
**Operating System**: Windows 10
**Tested Another Browser**: Yes

**Problem type**: Site is not usable
**Description**: Can't type text, can't delete.
**Steps to Reproduce**:
layout.css.servo.enabled: true



_From [webcompat.com](https://webcompat.com/) with ❤️_
```

And also note that if someone still try to put slide in an opening `<!--` comment. It will be caught by the `<` which will remove everything.

Maybe we need a better strategy, like a dropdown controlled list, though it will limit people. Freedom and expressiveness or Control and security. Life 🗺

zoepage: > The reason is that an opening comment will not do much as just being hidden by the other opening comment, but a closing comment will terminate the opening comment.

But wouldn't the missing closing comment mean, we never close, so everything that follows is also a comment? So nothing would actually gets displayed?

> caught by the < which will remove everything.

Everything = the value of the input  as a param? So `< Firefox 59.0` -> ``?

karlcow: In the body up there

```
<!-- @browser: Firefox 59.0 -->
```

the part we sanitized is `Firefox 59.0` or if you prefer the `string_here`:

```
<!-- @browser: string_here -->
```

The `<!-- @browser: ` and ` -->` are coming from 

https://github.com/webcompat/webcompat.com/blob/a18d81afc2c75f620ba64d4c8b50228fe622e2b8/webcompat/form.py#L117-L122

So let's say we got 
```python
string_here = '--> blah blah'
```
then the resulting string will be 

```
<!-- @browser: --> blah blah -->
```

which will be rendered `blah blah -->` with a hidden useless `<!-- @browser: -->`

Now if 

```python
# opening comment
string_here = 'blah <!-- foo'
```

it will become

```
<!-- @browser: blah <!-- foo -->
```
which does nothing apart of staying hidden.

Try this in your URL bar and do view source.

```
data:text/html,<!-- @browser: blah <!-- foo -->
```

zoepage: Ah! That makes sense. Thanks for the explanation. :)

I'd +1 the PR && say we chat in Austin about ways to improve?

karlcow:
```
→ nosetests --nocapture -v tests/test_form.py
Statuses Initialization…
Milestones saved in data/
Milestones in memory
Check that domain name is extracted. ... ok
Check that metadata is processed and wrapped. ... ok
Avoid some type of strings. ... ok
Check that URL is normalized. ... ok
Check that appropriate radio button label is returned. ... ok

Ran 8 tests in 0.007s

OK
```

and the full test suite is still running.

```
nosetests
........................................................................
Ran 72 tests in 2.468s

OK
```

I didn't run the functional tests though. Maybe to double check before.

tests/test_form.py
```diff
@@ -109,6 +112,20 @@ def test_get_metadata(self):
...
 
+    def test_normalize_metadata(self):
+        """Avoid some type of strings."""
+        cases = [('blue sky -->', 'blue sky'),
+                 ('blue sky ---->>', 'blue sky'),
+                 ('', ''),
+                 ('blue sky ', 'blue sky'),
+                 ('bad_bird <script>', ''),
+                 ('bad_bird <script-->>', ''),
+                 ('a' * 300, ''),
```
miketaylr: well if i can't rely on this site to do weird string math anymore, what's even the point

webcompat/form.py
```diff
@@ -154,6 +159,22 @@ def normalize_url(url):
...
+    if metadata_value is None:
+        return None
+    if '-->' in metadata_value:
+        metadata_value = metadata_value.replace('-->', '')
+        metadata_value = normalize_metadata(metadata_value)
```
miketaylr: so if we found a `-->`, remove it and recurse checking the modified value. I guess `blue sky ---->>` from the test catches where you might have two closing quote strings (in a more clever way). Cool.

webcompat/form.py
```diff
@@ -154,6 +159,22 @@ def normalize_url(url):
...
+    # Let's avoid html tags in
+    if ('<' or '>') in metadata_value and '-->' not in metadata_value:
```
miketaylr: At this point in the algorithm, when will we still have `'-->'` in `metadata_value`? Wouldn't it have been removed just above?

webcompat/form.py
```diff
@@ -154,6 +159,22 @@ def normalize_url(url):
...
+            metadata_value = ''
+    if len(metadata_value) > 200:
```
miketaylr: Interesting. Hadn't thought about length attacks/spam. There's also an option to truncate but I can't think of a great reason why you would have something that long in a UA string or browser name and it be worth keeping any of it. Other than accidental cmd+v or something.

tests/test_form.py
karlcow: heh. This generates a string of 300 `a` characters  ;) 
It is to test the length :)

webcompat/form.py
karlcow: exactly for for this test: `blue sky ---->>`

1. `blue sky ---->>` (first pass `'-->' in metadata_value`). True. replace
2. `blue sky -->` (second pass `'-->' in metadata_value`). True. replace
3. `blue sky ` (third pass `'-->' in metadata_value`) False. next test.

webcompat/form.py
karlcow: yup. Caught in the recursive test. There is no more `-->`

webcompat/form.py
karlcow: Yes I hesitated in between the two, aka truncating or empty, but as you said I don't see a lot of reason to keep something insanely long which might be misleading anyway when it's time to generate a label.

webcompat/form.py
karlcow: if you want I can add another test. but note that in the case of two closing comments non-nested. They will be removed right away in one pass.

```python
>>> metadata_value = 'blue sky -->-->'
>>> metadata_value.replace('-->', '')
'blue sky '
```

tests/test_form.py
miketaylr: :))

miketaylr:
Looks good, thanks @karlcow.

PR #1978, opened 2017-12-13
 Fixes #1977 - Adds optional browser argument for functional tests

### LINKED ISSUE #1977: Add optional "browsers" argument for functional tests
I had already done that while trying to get tests to run in Safari, but didn't merge it in since those tests failed. I think this is helpful, specially when working with the tests, troubleshooting, etc.

My idea is to add the `browsers` argument and if the user doesn't specify anything, just run all the available ones.

It would look like this: `npm run test:js browsers=safari,chrome,firefox`

brizental:
I also removed  the temporary fix for Firefox 56, because I noticed that the issue (https://github.com/mozilla/geckodriver/issues/858) had already been closed, didn't think it was worth it creating a whole new issue for that.

r? @miketaylr 

Chrome tests are failing, but I don't think it has something to do with this PR. Since I tested them before changing anything and they were already failing... So I think the build will fail in Travis, unfortunately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,10 @@
 # ref: https://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/
 sudo: required
 dist: trusty
-group: travis_latest
+group: deprecated-2017Q4
 
-env:
-  - MOZ_HEADLESS=1
-  
 addons:
-  firefox: latest
+  firefox: "55.0.2"
 
 # Only run travis on the master branch (and PRs), not all branches
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,13 @@
 # ref: https://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/
 sudo: required
 dist: trusty
+group: travis_latest
+
+env:
+  - MOZ_HEADLESS=1
+  
+addons:
+  firefox: latest
 
 # Only run travis on the master branch (and PRs), not all branches
 branches:
@@ -26,21 +33,19 @@ git:
 env:
   global:
     - ISSUES_REPO_URI=webcompat/webcompat-tests/issues
-    # grab latest 4.x release, which is LTS for now
-    - TRAVIS_NODE_VERSION="4"
 
 before_install:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 
 install:
-  - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install $TRAVIS_NODE_VERSION
+  - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install node
+  - node --version
   - travis_retry npm install
   - travis_retry npm run pip
   - travis_retry npm run config
   - which firefox
   - java -version
-  - firefox --version
   # lint python
   - pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example
   - python run.py -t &
@@ -48,8 +53,10 @@ install:
 before_script:
   - "sleep 2"
   # lint JS in default eslint task
-  - npm run lint
-  - npm run build
+  - npm run lint --force
+  - npm run build --force
+  # Check that server is actually running
+  - curl localhost:5000
 
 # now run the tests!
 script:
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -101,6 +101,18 @@ To run a single test suite, where foo.js is the file found in the `tests/functio
 node_modules/.bin/intern-runner config=tests/intern functionalSuites=tests/functional/foo.js 
 ```
 
+Right now the tests are running in Chrome and Firefox as a default. You can specify which browsers you want to test with using the `browsers` argument. Like this:
+
+```bash
+npm run test:js browsers=safari,firefox
+
+or 
+
+node_modules/.bin/intern-runner config=tests/intern browsers=safari,firefox
+```
+
+For a list of the recognized browser names, just refer to [Browser enum](http://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_Browser.html)
+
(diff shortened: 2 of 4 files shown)


brizental:
Don't know why build is failing since I removed chrome from the default testing browsers... Tested locally and everything seems to be alright. If anyone has any idea I would appreciate the insight :)

Meanwhile I'll keep on digging 👷

brizental:
Just some notes on what we have learned with all these debugging commits so far:

- The problem may lie in the `openPage` function in the `helpers.js` file. When `findByCssSelector` is called it return an "Unknown error", every time.
- We don't know why this happens, since from the screenshots we know that the DOM has been loaded, even before the `findByCssSelector` is called.
- I am testing in MacOS and TravisCI is using Ubuntu, so that is the next thing I will try and test.

brizental:
So... this last build passed... but...

screen shot 2017-12-14 at 22 08 19

I can't see the logs...

Anyways, probably the problem was with the new Trusty environment, I don't know exactly what the problem was with it, but changing it made it work. Link: ["If you encounter a spurious build failure because of the update..."](https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch)

brizental:
Actually this build is not passing, Travis is crazy.

brizental:
Ok, now it's passing. Probably the errors had something to do with the new Travis Trusty image, but it's not clear to me why. I think it would be cool to find out what it was and send them the bug report, since they are actively reviewing them (https://github.com/travis-ci/travis-ci/issues/8870).

The crucial changes that I made to the Travis config were:
- Went back to the previous Trusty image.
- Set Firefox to 55.0.2, because that was the one that was working. 

I also made some other changes:

- Set node version to latest always. We were actually using like node version 4, right now node is in version 8.
- I check that the server is running before running the tests.
- I force the `build` and `lint` actions because that is necessary if we are using the deprecated Trusty image.

.travis.yml
```diff
@@ -2,6 +2,10 @@
 # ref: https://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/
 sudo: required
 dist: trusty
+group: deprecated-2017Q4
```
miketaylr: Can we add a comment above this explaining what it does?

.travis.yml
```diff
@@ -26,30 +30,30 @@ git:
...
-    # grab latest 4.x release, which is LTS for now
-    - TRAVIS_NODE_VERSION="4"
 
 before_install:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 
 install:
-  - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install $TRAVIS_NODE_VERSION
+  - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install node
```
miketaylr: I think Travis has a nicer way to do this now, just using a `.nvmrc` file in the root.

https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions-using-.nvmrc

But maybe we just need the following, if we pick latest LTS and don't have to install nvm ourselves?

```
node_js:
    lts/*
```
maybe? Or maybe Travis will hate that, and we have to use .nvmrc. I don't know.

.travis.yml
```diff
@@ -26,30 +30,30 @@ git:
...
   # lint python
   - pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example
   - python run.py -t &
 
 before_script:
   - "sleep 2"
   # lint JS in default eslint task
-  - npm run lint
-  - npm run build
+  - npm run lint --force
```
miketaylr: Any reason you added `--force` here? I think ideally we want lint to fail a build, that way errors don't get into master.

.travis.yml
```diff
@@ -26,30 +30,30 @@ git:
 env:
   global:
     - ISSUES_REPO_URI=webcompat/webcompat-tests/issues
```
miketaylr: Uh weird, I had a comment and it disappeared. It was more or less this:

`- TRAVIS_NODE_VERSION="4"`

What if we picked the current LTS to be a bit more conservative? I imagine the tools we need will continue to work, but might result in fewer annoying regressions. I don't know.

.travis.yml
brizental: Hm, I added force, because it would't run these two scripts and it said the reason was because it didn't have `--force` in them. Since we are using a deprecated image, it seems to be required.

Indeed, it defeats the purpose if we force it... I'll look into it.

.travis.yml
brizental: Yes!

.travis.yml
brizental: Ah yes, I made so many changes to this I probably forgot to mention this one. We were using node version 4, yes. I changed it to get the latest version always: https://github.com/webcompat/webcompat.com/pull/1978/files/8bb75f79b02a621af543bd683da7bc1be1b45b5a#diff-354f30a63fb0907d4ad57269548329e3R39

.travis.yml
brizental: Nice, I'll look into it.

.travis.yml
miketaylr: Hmmmmm, interesting.

miketaylr: Can you paste the error message that it gives without `--force`?

miketaylr:
(adding the do not merge label just as a reminder to myself that there's some pending work here)

brizental:
@miketaylr I'm finishing the Intern 4 migration before I merge this in, in case they work with Travis just fine without using the deprecated image, hehe. Should have mentioned it here, sorry.

miketaylr:
No worries, thanks!

miketaylr:
> @miketaylr I'm finishing the Intern 4 migration before I merge this in, in case they work with Travis just fine without using the deprecated image, hehe. Should have mentioned it here, sorry.

I might mess around with something related to the deprecated image now... it sucks that all our tests are failing on master. Hopefully I don't create too much work for you after the Intern 4 migration is done.

miketaylr:
@brizental I'm gonna steal this PR, Hope that's OK. :)

miketaylr:
Actually... it makes sense to open a new PR with code from this, and code from #2009. Seems less messy.

miketaylr:
(dont delete this branch just yet plz)

PR #1983, opened 2017-12-14
Fixes #1976 - added test cases for feature

### LINKED ISSUE #1976: Add tests to similar bug feature
Context #1846 

Just functional tests for now.

zoepage:
- [x] added 3 test cases for feature
- [x] fixed build bug (uglify) https://github.com/webcompat/webcompat.com/pull/1937#issuecomment-351546497

### DIFF
--- a/tests/functional/reporting-auth.js
+++ b/tests/functional/reporting-auth.js
@@ -39,6 +39,68 @@ define(
             assert.include(text, "Report as"); //Report as FooUser (logged in)
           })
           .end();
+      },
+
+      "Duplicate issue - buglist shows on focus, hides on blur": function() {
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues/new"),
+          ".wc-ReportForm-actions-button"
+        )
+          .findById("url")
+          .click()
+          .type("https://test.")
+          .sleep(500)
+          .end()
+          .findById("description")
+          .click()
+          .end()
+          .sleep(500)
+          .findById("bugList")
+          .getVisibleText()
+          .then(function(text) {
+            assert.equal(text, "");
+          });
+      },
+
+      "Duplicate issue - 3 issues with url found": function() {
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues/new"),
+          ".wc-ReportForm-actions-button"
+        )
+          .findById("url")
+          .click()
+          .type("https://test.")
+          .sleep(500)
+          .end()
+          .findById("bugList")
+          .findAllByCssSelector("li a")
+          .getVisibleText()
+          .then(function(texts) {
+            assert.isAtLeast(texts.length, 3, "found 3 issues");
+          })
+          .end();
+      },
+
+      "Duplicate issue - no issue with url found": function() {
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues/new"),
+          ".wc-ReportForm-actions-button"
+        )
+          .findById("url")
+          .click()
+          .type("https://webctest1234567890")
+          .sleep(500)
+          .end()
+          .findById("bugList")
+          .findAllByCssSelector("li a")
+          .getVisibleText()
+          .then(function(texts) {
+            assert.equal(texts.length, 0, "No results returned.");
+          })
+          .end();
       }
     });
   }


zoepage:
@miketaylr I can't figure out the delay for `reporting-auth.js` > `Show double feature - buglist shows on focus, hides on blur` test.

The `#buglist` should be empty as it gets cleared after loosing focus of the `input#url`. Manually it works. Just can't figure out the syntax here for the delay.

zoepage:
@MDTsai The test should find 3 issues I've created, but it doesn't. 
https://github.com/webcompat/webcompat.com/blob/e7d7393ba16136ecea245f8938adb422a42027f9/tests/functional/reporting-auth.js#L81

You know, what might cause it?

tests/functional/reporting-auth.js
```diff
@@ -39,6 +39,68 @@ define(
...
+        )
+          .findById("url")
+          .click()
+          .type("https://test.")
+          .sleep(500)
+          .end()
+          .findById("description")
+          .click()
+          .end()
+          .sleep(9500)
```
miketaylr: This is a super long delay, are we just waiting until `bugList` is shown? By default all `find*` commands will wait for the DOM element to exist before failing (as long as it happens before the find timeout). Does it work without this `sleep` call?

tests/functional/reporting-auth.js
```diff
@@ -39,6 +39,68 @@ define(
             assert.include(text, "Report as"); //Report as FooUser (logged in)
           })
           .end();
+      },
+
+      "Show double feature - buglist shows on focus, hides on blur": function() {
```
miketaylr: Can we change `Show double feature` to something more descriptive? Is the double feature here referring to related bugs suggestion UI? Duplicate bugs UI?

tests/functional/reporting-auth.js
```diff
@@ -39,6 +39,68 @@ define(
...
+          .findById("url")
+          .click()
+          .type("https://webctest.")
+          .sleep(500)
+          .end()
+          .findById("bugList")
+          .findAllByTagName("li a")
+          .getVisibleText()
+          .then(function(texts) {
+            assert.isAtLeast(texts.length, 3, "found 3 issues");
```
miketaylr: Do you expect there to be more than 3 issues? If not, I would recommend just using `assert.equal` if more than 3 is a meaningful failure.

(I don't know how many suggestions are shown yet)

tests/functional/reporting-auth.js
```diff
@@ -39,6 +39,68 @@ define(
...
+          url("/issues/new"),
+          ".wc-ReportForm-actions-button"
+        )
+          .findById("url")
+          .click()
+          .type("https://webctest.")
+          .sleep(500)
+          .end()
+          .findById("bugList")
+          .findAllByTagName("li a")
```
miketaylr: I don't think `findAllByTagName` takes a CSS selector, but just a single `tagName`:

https://theintern.io/leadfoot/module-leadfoot_Element.html#findAllByTagName

(I could be wrong!)

If you want to filter with a selector, maybe `findAllByCssSelector` is what you want:

https://theintern.io/leadfoot/module-leadfoot_Element.html#findAllByCssSelector

tests/functional/reporting-auth.js
```diff
@@ -39,6 +39,68 @@ define(
...
+          .findById("url")
+          .click()
+          .type("https://webctest1234567890")
+          .sleep(500)
+          .end()
+          .findById("bugList")
+          .findAllByTagName("li a")
+          .getVisibleText()
+          .then(function(texts) {
+            assert.isBelow(texts.length, 1, "works");
```
miketaylr: If you expect 0 results, can we do `assert.equal(text.length, 0, "No results returned")`?

("works also seems like leftover debug text, can we have it be more descriptive so future failures are less confusing?)

webcompat/static/js/lib/bugform.js
```diff
@@ -221,49 +221,57 @@ function BugForm() {
...
-                for (var item in issues.items) {
-                  var issue = issues.items[item];
-                  $("#bugList").append(
-                    `<li><a href='https://webcompat.com/issues/${issue.number}' target='_blank'>#${issue.number} - ${issue.title}</a></li>`
-                  );
+      var bugform = this;
+
+      this.suggestDelay = setTimeout(function() {
+        try {
+          // url got removed the http protocol, so the url works as a keyword. Shortest url.hostname is something like www.
```
miketaylr: I don't understand this comment. Can it be clarified?

webcompat/static/js/lib/bugform.js
```diff
@@ -256,7 +256,13 @@ function BugForm() {
                   for (var item in issues.items) {
                     var issue = issues.items[item];
                     $("#bugList").append(
-                      `<li><a href='https://webcompat.com/issues/${issue.number}' target='_blank'>#${issue.number} - ${issue.title}</a></li>`
+                      "<li><a href='https://webcompat.com/issues/" +
```
miketaylr: We have lodash loaded, it might be a bit nicer to read if we use `_.template`?

https://lodash.com/docs/3.10.1#template

something like:

```
(outside of for loop)
var makeLi = _.template("<li><a href='https://webcompat.com/issues/<%- issue.number %>' target='_blank'>#<%- issues.number %> - <%- issue.title %></a></li>")

$("#bugList").append(makeLi({'issue': issue})
```

miketaylr:
> @MDTsai The test should find 3 issues I've created, but it doesn't.

@zoepage I believe it's due to the wrong argument used for `.findAllByTagName("li a")` (see my other comment). You could check it by inspecting the resolved value of the Promise that it returns.

miketaylr:
Ah, one other nit:

`Issues #1976 - added 3 test cases` - can you make it more descriptive, so it's obvious what the test cases are for? This is super helpful when skimming history. THanks.

tests/functional/reporting-auth.js
zoepage: Is "duplicate issue" ok?

tests/functional/reporting-auth.js
zoepage: The delay was just for testing as there is a bug anyway... 

The DOM element is always there (so finding the DOM element "works" with and without `sleep`). The issue here is, the DOM element with the ID `#bugList` is filled with list items (the items are related issues). When the input loses focus, the list is removed again. 

duolicate_issue

The issue is that in the test, the list items still exist.

I thought that `.end()` would dismiss the scope it has before and the next element you'd look for (e.g. `findById('description')`) would be grabbed "fresh", but the test seems to get an "old" state? (I hope, this was kind of understandable.)

tests/functional/reporting-auth.js
zoepage: yes, perfect. thanks!

tests/functional/reporting-auth.js
zoepage: +1

webcompat/static/js/lib/bugform.js
zoepage: > url got removed the http protocol, so the url works as a keyword. 

The type URL normalizes the value of the `#url` input. Here for it splits the value in sections.`url.hostname` is the hostname and can be used as a keyword.
`https://test` -> `test`

`test` would be the keyword here.

> Shortest url.hostname is something like www.

We added a restriction to the min. char length, so the functionality to look for a duplicate issue does not start unnecessarily and we don't waste requests or get blocked by Github API.
So, we were thinking about the shortest possible URL people would probably use... 

The usecase `www.` is something that sounds reasonable as well as `x.io`, so we chose the min char of 4.

webcompat/static/js/lib/bugform.js
zoepage: Thats a good idea, but CSP is blocking lodash here (which is weird).

tests/functional/reporting-auth.js
```diff
@@ -39,6 +39,68 @@ define(
...
+          .findById("url")
+          .click()
+          .type("https://test.")
+          .sleep(500)
+          .end()
+          .findById("bugList")
+          .findAllByCssSelector("li a")
+          .getVisibleText()
+          .then(function(texts) {
+            assert.isAtLeast(texts.length, 3, "found 3 issues");
```
miketaylr: Do we ever expect more than 3 ever?

zoepage: Sorry, I'm my mind I replied. 

We have more than 3 issues that come back in this case... As issues will be added, the number will vary, but will be more than 3.

zoepage:
Follow up issue https://github.com/webcompat/webcompat.com/issues/1990

tests/functional/reporting-auth.js
miketaylr: OK, thanks. We probably want to limit the upper bounds of what we display (and when we do, add a test for that), so we're not displaying 20+ dupe issues on a mobile device, etc.

miketaylr:
@zoepage can you rebase this branch against the latest similarbugs and re-push, so we can get travis to run these? Thanks.

webcompat/static/js/lib/bugform.js
miketaylr: Ah right, we don't allow eval since we move to jst stuff. So that's another option.

zoepage:
@miketaylr rebased. tests ran, where a few are failing ("Duplicate issue" tests are green).

tests/functional/reporting-auth.js
zoepage: I've added this to #1990

miketaylr:
Thanks!

PR #1987, opened 2017-12-15
Fixes #1986 - Start migration to Intern 4

### LINKED ISSUE #1986: Update from Intern 3 to Inter 4
Why? 

Because it seems to be better and more reliable.

I already tested with one of our tests it and it looks good.

brizental:
Just a test so you guys can see how everything will look like. I will continue the migration from this branch.


zoepage:
Great work Bea! :)

s/guys/folks

miketaylr:
We can add Chrome back as a default once this lands, since it will put us on a modern digdug that supports pointing to the latest Chromedriver (the 2nd part of the issue that broke tests (the first being the travis trusty image update)).

brizental:
@miketaylr Yes, I think so. From the tests that I have already migrated, none of them have had any issues with Chorme, and surprise surprise, they don't even have many issues with Safari, we might be able to add both browsers by default, maybe.

brizental:
one happy sight.
screen shot 2018-01-12 at 00 00 05

What still needs to be done:

- Merge in the latest changes to the tests.
- Try to sort out one last bug in Safari (yeah, only one bug).

miketaylr:
> one happy sight.

aww wyeah. :)

tests/intern.js
```diff
@@ -1,79 +1,9 @@
...
-    functionalSuites: ["tests/functional-all"],
-    excludeInstrumentation: true
-  };
+intern.configure({
+  functionalSuites: ["./tests/javascript/functional/test.js"],
+  environments: { browserName: "firefox" },
+  reporters: "runner"
 });
+
+intern.run();
```
vladikoff: Make sure to catch errors here and exit(1). Ref: https://github.com/mozilla/fxa-content-server/blob/master/tests/intern.js#L133

brizental: Will do!

brizental:
> Done. Your build exited with 0.

Oh yeah!

r? @miketaylr @vladikoff

brizental:
Forgot to mention: I will look at Safari issues after this is done. Even though we have much less errors, we are still getting a bunch of intermittents. Thought it would be best to send this in and finish this PR with Chrome and Firefox only for now :)

miketaylr:
> Forgot to mention: I will look at Safari issues after this is done. Even though we have much less errors, we are still getting a bunch of intermittents. Thought it would be best to send this in and finish this PR with Chrome and Firefox only for now :)

Awesome agreed. Let me file an issue so we have a paper trail.

miketaylr:
> Awesome agreed. Let me file an issue so we have a paper trail.

(bea pointed out that an issue already exists: https://github.com/webcompat/webcompat.com/issues/1704)

tests/functional/_intern.js
```diff
@@ -0,0 +1,68 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
miketaylr: (silly question, any reason this file is named `_intern.js`, rather than `intern.js`? Just leftover from renaming everything during the migration?)

brizental: Yeah it's leftover from renaming, but I thought it was nice that with the underscore it is always the first file listed and has some visual separation from tests files, so I left it like this. But it's nothing important, haha, we can change it back if needed.

miketaylr: Ah cool. I like your reasoning -- thanks!

miketaylr:
Now I guess I should read the Intern 4 changelog to see what cool new stuff we can do. :)

karlcow:
@brizental was there a reason to change the folder structure for unit tests?

brizental:
@karlcow I thought that, since I was putting everything about the functional tests inside a `functional` folder, it would be nice to put everything about the unit tests inside a `unit`folder.

PR #1989, opened 2017-12-20
Fixes #1988 - fixes build for #1983

### LINKED ISSUE #1988: Fix build
Related to #1976 and #1846.
Needs to be merged before #1983, so the build can pass.

zoepage:
This PR fixes the ES6 template issue with the build dependency.
Please merge before #1983 

r? @miketaylr

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -232,11 +232,9 @@ function BugForm() {
 
       this.suggestDelay = setTimeout(function() {
         try {
-          // url got removed the http protocol, so the url works as a keyword. Shortest url.hostname is something like www.
-
           var url = new URL(bugform.urlField.val());
 
-          if (url.hostname.length > 7) {
+          if (url.hostname.length > 3) {
             var query =
               "https://api.github.com/search/issues?page=1&per_page=10&stage=all&sort=created&q=" +
               url.hostname +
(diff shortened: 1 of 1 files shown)


webcompat/static/js/lib/bugform.js
```diff
@@ -221,49 +221,61 @@ function BugForm() {
...
-        } else {
+              })
+              .then(function(issues) {
+                if (issues.items) {
+                  $("#bugListWrapper").show();
+                  $("#bugList").empty();
+                  for (var item in issues.items) {
+                    var issue = issues.items[item];
+                    $("#bugList").append(
+                      "<li><a href='https://webcompat.com/issues/" +
```
miketaylr: I already left some feedback about this in the last PR, I think using `_.template` from lodash would be better here. No hard objections in a proof of concept branch, but it should be fixed before it gets merged into master.

miketaylr:
I think @MDTsai should review, since he authored the original stuff.

r? @MDTsai

MDTsai:
Looks good, as merged into master, I think 2 things to do:
- [ ] change search end point from github search API to ours?
- [ ] alternative of fetch for more compatibility.

webcompat/static/js/lib/bugform.js
zoepage: I agree. Will add this to the follow up issue.

zoepage:
@MDTsai I agree. Will add this to the follow up issue when we merge to master.

Issue #1997, opened 2018-01-04
[cleaning] form.py extracts multiple times some values in the same function.


karlcow:
Realized when working on #1995, form.py contains multiple `url` and we should take care also of other values.

https://github.com/webcompat/webcompat.com/issues/1995#issuecomment-355180517

karlcow:
So this is a prototype on what I want to do with the form code. This is not yet the pull request just an exploration at trying to 

1. make it DRY (Do not Repeat Yourself)
2. Group by areas
3. Make it slightly more modular
4. have a consistency in variables.

Still some issues to fix here and there and some tests to fix and add.
https://github.com/karlcow/webcompat.com/commit/d12695da19e6692c1ef153f5491139a9122ffd17?diff=split

karlcow:
This was fixed by #2474

Issue #1995, opened 2018-01-03
[bug] 502 Bad Gateway on entering a utf-8 string in URL field.


karlcow:
@softvision-oana-arbuzov found an interesting issue.
1. Go to https://staging.webcompat.com/ 
2.  Enter "價格" in the URL field… and boom.

<img width="778" alt="capture d ecran 2018-01-03 a 22 20 47" src="https://user-images.githubusercontent.com/505230/34522081-5fe375c0-f0d4-11e7-84f8-d7dda2063f27.png">

karlcow:
Additional tests. 

* `é` fails too. So anything which is not ascii. 
* `http://愛.com/` fails too. (which is a problem, it should be converted to http://xn--vgu.com/) aka IDN
* `http://example.com/愛` works.

karlcow:
Ah! good. I was afraid it would be an issue with url parsing and libraries not handling idn, but it seems to be a lot simpler.  We are sometimes a bit loose on strings. We should straighten that or switch to python 3.

```
http://™.com/
127.0.0.1 - - [04/Jan/2018 09:49:19] "POST /issues/new HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
[...]

PR #2001, opened 2018-01-04
Fix #1740 - Handles double schemes URL.

### LINKED ISSUE #1740: Handle http://https patterns in bug reporting form
I don't why it is happening, but we should probably handle this case.
Such as https://github.com/webcompat/web-bugs/issues/9473
The URL was
```
 http://https://app.igenapps.com/2360806#0/
```
and the title became

```
https: - see bug description
```

SeeAlso https://github.com/webcompat/web-bugs/issues?utf8=%E2%9C%93&q=is%3Aissue%20http%20in%3Atitle

karlcow:
Added tests and working.

```
→ nosetests -v --nocapture tests/test_form.py
Statuses Initialization…
Milestones saved in data/
Milestones in memory
Check that domain name is extracted. ... ok
Check that metadata is processed and wrapped. ... ok
Avoid some type of strings. ... ok
Check that URL is normalized. ... ok
Check that appropriate radio button label is returned. ... ok

Ran 8 tests in 0.005s

OK
```

The rest of the unit tests are passing.

```
→ nosetests 
........................................................................
Ran 72 tests in 2.347s

OK
```

This helps a bit with #1997 (by chance).

Because it touches the same area than issue #1995 
Once #1998 is merged, I will be happy to rebase this pull request upon the latest version of the master.

r? @miketaylr 
r? @zoepage

### DIFF
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -140,7 +140,10 @@ def normalize_url(url):
         return None
     url = url.strip()
     parsed = urlparse.urlparse(url)
-
+    # Handle the case when URL has the form http://https://example.com
+    if parsed.netloc in ['http:', 'https:'] and parsed.path.startswith('//'):
+        url = url.split('//', 1)[1]
+    # Clean the URL.
     if url.startswith(BAD_SCHEMES) and not url.startswith(SCHEMES):
         # if url starts with a bad scheme, parsed.netloc will be empty,
         # so we use parsed.path instead
@@ -238,7 +241,7 @@ def build_formdata(form_object):
 
     formdata = {
         'metadata': get_metadata(metadata_keys, form_object),
-        'url': form_object.get('url'),
+        'url': normalized_url,
         'browser': normalize_metadata(form_object.get('browser')),
         'os': normalize_metadata(form_object.get('os')),
         'problem_type': get_radio_button_label(
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -17,6 +17,7 @@ class TestForm(unittest.TestCase):
     def setUp(self):
         """Set up."""
         webcompat.app.config['TESTING'] = True
+        self.maxDiff = None
         self.app = webcompat.app.test_client()
 
     def tearDown(self):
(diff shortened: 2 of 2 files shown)


miketaylr:
> 81/81 tests failed

Something is up with Travis... but totally unrelated to this PR and the code @karlcow touched. Do you have any ideas @brizental?

webcompat/form.py
```diff
@@ -140,7 +140,10 @@ def normalize_url(url):
         return None
     url = url.strip()
     parsed = urlparse.urlparse(url)
-
+    # Handle the case when URL has the form http://https://example.com
+    if parsed.netloc in ['http:', 'https:'] and parsed.path.startswith('//'):
+        url = url.split('//', 1)[1]
```
miketaylr: So we take the "inner" value, right?

brizental:
> Something is up with Travis... but totally unrelated to this PR 

Looks like it's the same problem I had in my PR. Travis just updated the Trusty image, probably it's something related to that. To fix it, I just went back to the previous Trusty image and set firefox to 55, because that one is working fine.

miketaylr:
Ah, OK. So we should probs default to 55 for now... thanks @brizental.

webcompat/form.py
karlcow: @miketaylr yes. I assumed that if people are botching this thing is because they had to an already existing string and not insert in the middle of it… ahah. We might get surprises. People are creative.

karlcow:
```python
>>> url = 'http://https://bad.example.com'
>>> # Splitting on all doubles '//'
>>> url.split('//')
['http:', 'https:', 'bad.example.com']
>>> # Splitting on the first instance from the left (for the right rsplit)
>>> url.split('//', 1)
['http:', 'https://bad.example.com']
>>> # Takes the second item on the list only
>>> url.split('//', 1)[1]
'https://bad.example.com'
```

Another assumption: This will happen only once. I don't expect people to do multiple `http://http://http://` if we are wrong, we can still go recursive. :)

karlcow:
@miketaylr you didn't merge in the wrong order. :) It was expected. It's why I said. :) 

>  Once #1998 is merged, I will be happy to rebase this pull request upon the latest version of the master.

Let me fix that pull request.

karlcow:
@miketaylr conflicts resolved.

karlcow:
waiting for travis to finish its checks.

karlcow:
hehe. travis is failing all tests but that seems totally unrelated to the modifications here. :)

/me sees @miketaylr discussed this with @brizental

miketaylr:
> @miketaylr you didn't merge in the wrong order. :) It was expected. It's why I said. :)

> ....

heh, yeah. my brain isn't firing at 100% just yet. maybe by march. :)

thanks.

PR #2004, opened 2018-01-05
Issue#1763

Obtin:
Add watch task for templates


Obtin:
r ? @miketaylr @zoepage, done with the desired changes

PR #2007, opened 2018-01-09
Fixes #2006: Add support for multiple `label` and `details` params.

### LINKED ISSUE #2006: Support multiple `label` and `details` params
In https://bugzilla.mozilla.org/show_bug.cgi?id=1422985, I'm adding support for `type-webrender` labels, for the built-in "report site issue" page action item. While we're still adding `type-servo` labels, it's possible for there to be 2 `label` and 2 `details` params.

There's a few ways to do that, with no clear standard, so I'm choosing the following:

`a=b&a=c` (as opposed to something like `a=b,c`).

Since we are already using `URLSearchParams`, using `URLSearchParams.append` makes sense...

This bug tracks the client (for `details`) and server (for `label`) side work to parse those params correctly.

miketaylr:
Our test situation seems FUBAR in Firefox 57+ right now, but these new functional tests pass in Chrome, and nosetests are good.

screen shot 2018-01-08 at 10 37 03 pm

screen shot 2018-01-08 at 10 40 54 pm

r? @karlcow 
r? @zoepage

### DIFF
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -81,10 +81,14 @@ class TestForm(unittest.TestCase):
         """Check that metadata is processed and wrapped."""
         TEST_DICT = {'cool': 'dude', 'wow': 'ok'}
         EXPECTED_SINGLE = '<!-- @cool: dude -->\n'
+        EXPECTED_SINGLE_COMMA = '<!-- @cool: dude, wow -->\n'
         EXPECTED_MULTIPLE = '<!-- @cool: dude -->\n<!-- @wow: ok -->\n'
 
         r = form.wrap_metadata(('cool', 'dude'))
         self.assertEqual(r, EXPECTED_SINGLE)
+
+        r = form.wrap_metadata(('cool', 'dude, wow'))
+        self.assertEqual(r, EXPECTED_SINGLE_COMMA)
 
         r = form.get_metadata(('cool', 'wow'), TEST_DICT)
         self.assertEqual(r, EXPECTED_MULTIPLE)
@@ -135,7 +139,7 @@ class TestForm(unittest.TestCase):
                  ('bad_bird <script>', ''),
                  ('bad_bird <script-->>', ''),
                  ('a' * 300, ''),
-                 (None, None)
+                 (None, None),
                  ]
         for meta_value, expected in cases:
             self.assertEqual(form.normalize_metadata(meta_value), expected)
--- a/tests/test_webhook.py
+++ b/tests/test_webhook.py
@@ -135,6 +135,7 @@ class TestWebhook(unittest.TestCase):
         self.assertEqual(rv.data, 'Not an interesting hook')
 
     def test_extract_metadata(self):
+        """Extract dictionary of metadata for an issue body."""
         expected = {'reported_with': 'web',
                     'extra_labels': 'type-media, type-stylo',
                     'ua_header': ('Mozilla/5.0 (what) Gecko/20100101 '
--- a/tests/test_webhook.py
+++ b/tests/test_webhook.py
@@ -50,7 +50,7 @@ class TestWebhook(unittest.TestCase):
         <!-- @browser: Firefox 55.0 -->
         <!-- @ua_header: Mozilla/5.0 (what) Gecko/20100101 Firefox/55.0 -->
         <!-- @reported_with: web -->
-        <!-- @extra_label: type-media -->
+        <!-- @extra_labels: type-media, type-stylo -->
 
         **URL**: https://www.example.com/
         **Browser / Version**: Firefox 55.0
@@ -59,6 +59,7 @@ class TestWebhook(unittest.TestCase):
 
         self.issue_body2 = """
         <!-- @browser: Foobar -->
+        <!-- @extra_labels: type-foobar -->
         """
         self.issue_body3 = """
         **URL**: https://www.google.com/
@@ -134,7 +135,8 @@ class TestWebhook(unittest.TestCase):
         self.assertEqual(rv.data, 'Not an interesting hook')
 
     def test_extract_metadata(self):
-        expected = {'reported_with': 'web', 'extra_label': 'type-media',
+        expected = {'reported_with': 'web',
+                    'extra_labels': 'type-media, type-stylo',
                     'ua_header': ('Mozilla/5.0 (what) Gecko/20100101 '
                                   'Firefox/55.0'),
                     'browser': 'Firefox 55.0'}
@@ -157,14 +159,16 @@ class TestWebhook(unittest.TestCase):
             actual = helpers.extract_browser_label(metadata_dict)
             self.assertEqual(expected, actual)
 
-    def test_extract_extra_label(self):
+    def test_extract_extra_labels(self):
         """Extract 'extra' label."""
         metadata_tests = [
-            ({'extra_label': 'type-media'}, 'type-media'),
-            ({'burgers': 'french fries'}, None)
+            ({'extra_labels': 'type-media'}, ['type-media']),
+            ({'extra_labels': 'cool, dude'}, ['cool', 'dude']),
+            ({'extra_labels': u'weather-☁'}, ['weather-\xe2\x98\x81']),
+            ({'burgers': 'french fries'}, None),
         ]
         for metadata_dict, expected in metadata_tests:
-            actual = helpers.extract_extra_label(metadata_dict)
+            actual = helpers.extract_extra_labels(metadata_dict)
             self.assertEqual(expected, actual)
 
     def test_extract_priority_label(self):
@@ -176,6 +180,17 @@ class TestWebhook(unittest.TestCase):
             self.assertEqual(priority_label, 'priority-critical')
         priority_label_none = helpers.extract_priority_label(self.issue_body)
         self.assertEqual(priority_label_none, None)
+
+    def test_get_issue_labels(self):
+        """Extract list of labels from an issue body."""
+        labels_tests = [
+            (self.issue_body, ['browser-firefox', 'type-media', 'type-stylo']),
+            (self.issue_body2, ['type-foobar']),
+            (self.issue_body3, ['browser-firefox-mobile-tablet'])
+        ]
+        for issue_body, expected in labels_tests:
+            actual = helpers.get_issue_labels(issue_body)
+            self.assertEqual(sorted(expected), sorted(actual))
 
     def test_is_github_hook(self):
         """Validation tests for GitHub Webhooks."""
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -128,8 +128,15 @@ def wrap_metadata(metadata):
 
 def get_metadata(metadata_keys, form_object):
     """Return relevant metadata hanging off the form as a single string."""
+    extra_labels = []
+    if 'extra_labels' in metadata_keys:
+        extra_labels = [normalize_metadata(label)
+                        for label in form_object.get('extra_labels')
+                        if label in app.config['EXTRA_LABELS']]
+        metadata_keys.remove('extra_labels')
     metadata = [(key, form_object.get(key)) for key in metadata_keys]
     metadata = [(md[0], normalize_metadata(md[1])) for md in metadata]
+    metadata.extend(extra_labels)
     # Now, "wrap the metadata" and return them all as a single string
     return ''.join([wrap_metadata(md) for md in metadata])
 
@@ -236,9 +243,9 @@ def build_formdata(form_object):
         summary = u'{0} - {1}'.format(normalized_url, problem_summary)
 
     metadata_keys = ['browser', 'ua_header', 'reported_with']
-    extra_label = form_object.get('extra_label', None)
-    if extra_label in app.config['EXTRA_LABELS']:
-        metadata_keys.append('extra_label')
+    extra_labels = form_object.get('extra_labels', None)
+    if extra_labels:
+        metadata_keys.append('extra_labels')
 
     formdata = {
         'metadata': get_metadata(metadata_keys, form_object),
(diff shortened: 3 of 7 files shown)


webcompat/webhooks/helpers.py
```diff
@@ -47,13 +47,17 @@ def extract_browser_label(metadata_dict):
...
-def extract_extra_label(metadata_dict):
-    """Return the 'extra' label from metadata_dict."""
-    extra_label = metadata_dict.get('extra_label', None)
-    if extra_label:
-        extra_label = extra_label.lower()
-        extra_label = extra_label.encode('utf-8')
-    return extra_label
+def extract_extra_labels(metadata_dict):
+    """Return the 'extra_labels' labels from metadata_dict, as a list."""
+    # we have to handle the ', ' foo biz here.
```
miketaylr: Oops, leftover comment.

webcompat/webhooks/helpers.py
```diff
@@ -47,13 +47,17 @@ def extract_browser_label(metadata_dict):
...
+    extra_labels = metadata_dict.get('extra_labels', None)
+    if extra_labels:
+        extra_labels = extra_labels.split(', ')
+        print(extra_labels)
```
miketaylr: Leftover debug to be removed.

webcompat/webhooks/helpers.py
```diff
@@ -149,6 +153,23 @@ def get_issue_info(payload):
...
 
 
+def get_issue_labels(issue_body):
+    """Extract the list of labels from an issue body to be sent to GitHub."""
+    metadata_dict = extract_metadata(issue_body)
+    browser_label = extract_browser_label(metadata_dict)
+    extra_labels = extract_extra_labels(metadata_dict)
+    priority_label = extract_priority_label(issue_body)
+    labelslist = []
+    print(browser_label, extra_labels, priority_label)
```
miketaylr: Leftover debug to be removed.

webcompat/webhooks/helpers.py
```diff
@@ -149,6 +153,23 @@ def get_issue_info(payload):
...
+    for label in (browser_label, extra_labels, priority_label):
```
miketaylr: I can imagine @karlcow coming up with a sweet list comprehension here... but with the conditionals, a plain ol' for loop seems more readable (to me, anyways).

karlcow:
ok let's review this.

```
git checkout master
git fetch upstream
git merge upstream/master
git checkout -b issues/2006/1 upstream/issues/2006/1
```

Let's play with the code now.

tests/test_webhook.py
```diff
@@ -177,6 +180,17 @@ def test_extract_priority_label(self):
...
+    def test_get_issue_labels(self):
+        """Extract list of labels from an issue body."""
+        labels_tests = [
+            (self.issue_body, ['browser-firefox', 'type-media', 'type-stylo']),
+            (self.issue_body2, ['type-foobar']),
+            (self.issue_body3, ['browser-firefox-mobile-tablet'])
+        ]
+        for issue_body, expected in labels_tests:
+            actual = helpers.get_issue_labels(issue_body)
+            self.assertEqual(expected, actual)
```
karlcow: Here you are always testing a list. So we could use.

`self.assertListEqual(expected, actual)` but this would fail if the list items have a different order. 

`self.assertEqual(sorted(expected), sorted(actual))` will fix that. Then we have a more robust test.

webcompat/webhooks/helpers.py
karlcow: After modifying the test for testing lists.

```
→ nosetests -v --nocapture tests/test_webhook.py
Statuses Initialization…
Milestones saved in data/
Milestones in memory
Extract 'extra' label. ... ok
test_extract_metadata (tests.test_webhook.TestWebhook) ... ok
Extract priority label. ... ok
Extract list of labels from an issue body. ... ok

Ran 13 tests in 0.073s

OK
```

webcompat/form.py
```diff
@@ -164,6 +164,17 @@ def normalize_url(url):
 
 def normalize_metadata(metadata_value):
     """Normalize the metadata received from the form."""
+    # If it's a list, we are dealing with @extra_labels,
+    # so convert into a comma separated string.
```
karlcow: Not sure, I like this part. It seems a lot of juggling because of the type.

webcompat/form.py
```diff
@@ -236,9 +247,10 @@ def build_formdata(form_object):
         summary = u'{0} - {1}'.format(normalized_url, problem_summary)
 
     metadata_keys = ['browser', 'ua_header', 'reported_with']
-    extra_label = form_object.get('extra_label', None)
-    if extra_label in app.config['EXTRA_LABELS']:
-        metadata_keys.append('extra_label')
+    extra_labels = form_object.get('extra_labels', None)
+    if extra_labels:
+        # Note: we check if the label is allowed in normalize_metadata
```
karlcow:
```python
    if extra_labels:
        extra_labels = [normalize_metadata(label)
                        for label in extra_labels
                        if label in app.config['EXTRA_LABELS']]
        metadata_keys.append('extra_labels')
```

This way we add only two instructions instead of all the juggling in normalize_metadata

webcompat/form.py
karlcow: What about leaving it as it was before and see the modification in `if extra_labels` https://github.com/webcompat/webcompat.com/pull/2007/files#diff-f1975977b06efb529b8c60f23a7bc4eeR252

tests/test_form.py
```diff
@@ -128,14 +132,22 @@ def test_get_metadata(self):
...
                  ('a' * 300, ''),
-                 (None, None)
+                 (None, None),
+                 (['dance', 'powder'], 'dance, powder'),
+                 (['dance ', 'powder'], 'dance, powder'),
+                 ([' dance -->', 'powder -->'], 'dance, powder'),
+                 (['dance<script>', 'powder -->'], 'powder'),
+                 (['dance', None], 'dance'),
+                 (['dance<script>', ''], ''),
+                 (['not', 'allowed'], '')
```
karlcow: with the modifications proposed you can remove these additional tests, or add the ones which you think are necessary.

webcompat/views.py
```diff
@@ -176,9 +176,10 @@ def create_issue():
         # HTML <form>, so we stick them in the session cookie so they survive
         # the scenario where the user decides to do authentication, and they
         # can then be passed on to form.py
-        for param in ['src', 'label']:
-            if request.args.get(param):
-                session[param] = request.args.get(param)
+        if request.args.get('src'):
+            session['src'] = request.args.get('src')
+        if request.args.get('label'):
+            session['src'] = request.args.getlist('label')
```
karlcow: I'm confused here. Could you explain?

webcompat/webhooks/helpers.py
karlcow: these three lines are a bit dangerous. They are changing the type of the variable.
```python
labels = metadata_dict.get('extra_labels', None)
if labels:
    extra_labels = labels.split(', ')
```

(which also makes me thing if we should return an empty list instead of `None` I wonder if that would simplify things down the pipe. But not urgent)


```

File: webcompat/webhooks/helpers.py [hunk 10]
```diff
@@ -47,13 +47,17 @@ def extract_browser_label(metadata_dict):
...
+def extract_extra_labels(metadata_dict):
+    """Return the 'extra_labels' labels from metadata_dict, as a list."""
+    # we have to handle the ', ' foo biz here.
+    extra_labels = metadata_dict.get('extra_labels', None)
+    if extra_labels:
+        extra_labels = extra_labels.split(', ')
+        print(extra_labels)
+        for idx, label in enumerate(extra_labels):
+            extra_labels[idx] = label.lower()
+            extra_labels[idx] = label.encode('utf-8')
```
karlcow:
```python
>>> extra_labels = [u'Foo', u'愛']
>>> extra_labels = [label.lower().encode('utf-8') for label in extra_labels]
>>> extra_labels 
['foo', '\xe6\x84\x9b']
```

webcompat/webhooks/helpers.py
```diff
@@ -47,13 +47,17 @@ def extract_browser_label(metadata_dict):
...
+    return extra_labels
```
karlcow: so the full function becomes:

```python
def extract_extra_labels(metadata_dict):
    """Return the 'extra_labels' labels from metadata_dict, as a list."""
    labels = metadata_dict.get('extra_labels', None)
    if labels:
        extra_labels = labels.split(', ')
        extra_labels = [label.lower().encode('utf-8')
                        for label in extra_labels]
    else:
        extra_labels = labels
    return extra_labels
```

tests/test_webhook.py
```diff
@@ -157,14 +159,15 @@ def test_extract_browser_label(self):
             actual = helpers.extract_browser_label(metadata_dict)
             self.assertEqual(expected, actual)
 
-    def test_extract_extra_label(self):
+    def test_extract_extra_labels(self):
         """Extract 'extra' label."""
         metadata_tests = [
-            ({'extra_label': 'type-media'}, 'type-media'),
+            ({'extra_labels': 'type-media'}, ['type-media']),
+            ({'extra_labels': 'cool, dude'}, ['cool', 'dude']),
```
karlcow: in the function you seem to expect unicode data (aka you do `encode('utf-8')`) so it's probably a good idea to make the test even with what is expected. and even create a test for a funky unicode string.

karlcow:
I have not reviewed the JavaScript. @zoepage do you want to do it?

tests/test_webhook.py
karlcow: Ah I also realized here that 

* `'cool,dude'` returns `['cool,dude']`
* `'cool , dude'` returns `['cool ', 'dude']`
* `' cool , dude'` returns `[' cool ', 'dude']`
* `', cool , dude'` returns `['', ' cool ', 'dude']`

just fyi

zoepage:
The JS looks 👍 to me.

webcompat/webhooks/helpers.py
miketaylr: Oh... much simpler. :)

webcompat/views.py
miketaylr: > I'm confused here. Could you explain?

@karlcow oh, just a typo. should be `session['label']` (i didn't catch that because this code is only exercised when doing auth reporting).

webcompat/webhooks/helpers.py
miketaylr: I prefer `None`, but I don't think it makes much of a difference...since we need to filter out `None` anyways for `priority_label`. 

If someone feels strongly, please file a follow up bug. :)

webcompat/webhooks/helpers.py
miketaylr: > There is an issue though when extra_labels is None. Here a proposal.

(https://github.com/webcompat/webcompat.com/blob/master/webcompat/webhooks/helpers.py#L76 returns `None` as well, so we have to handle `None` anyways).

webcompat/form.py
miketaylr: This is nicer. We don't do anything with that list you created... but I think we can create a separate `extra_labels` and merge with the `metadata` list before joining as a string inside `get_metadata`.

tests/test_form.py
miketaylr: Let's just remove them, they're not important now.

webcompat/form.py
miketaylr: OK, I can remove the `# Note` comment. I took your suggestion, but moved it here:

https://github.com/webcompat/webcompat.com/pull/2007/files#diff-f1975977b06efb529b8c60f23a7bc4eeR131

tests/test_webhook.py
```diff
@@ -177,6 +182,17 @@ def test_extract_priority_label(self):
...
+    def test_get_issue_labels(self):
+        """Extract list of labels from an issue body."""
+        labels_tests = [
+            (self.issue_body, ['browser-firefox', 'type-media', 'type-stylo']),
+            (self.issue_body2, ['type-foobar']),
+            (self.issue_body3, ['browser-firefox-mobile-tablet'])
+        ]
+        for issue_body, expected in labels_tests:
+            actual = helpers.get_issue_labels(issue_body)
+            self.assertEqual(sorted(expected), sorted(actual))
```
miketaylr: FWIW, according to Python docs, speaking of `assertListEqual` and `assertTupleEqual`:

`These methods are used by default when comparing lists or tuples with assertEqual().`

miketaylr:
Ready for re-review, Karl.

r? @karlcow

tests/test_webhook.py
```diff
@@ -134,7 +135,9 @@ def test_fails_on_not_known_action(self):
         self.assertEqual(rv.data, 'Not an interesting hook')
 
     def test_extract_metadata(self):
-        expected = {'reported_with': 'web', 'extra_label': 'type-media',
+        """Extract dictionary of metadata for an issue body.""".
```
karlcow: cough cough  🎃 

```
Failure: SyntaxError (invalid syntax (test_webhook.py, line 138)) ... ERROR

ERROR: Failure: SyntaxError (invalid syntax (test_webhook.py, line 138))
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Users/karl/.virtualenvs/webcompatcom/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/karl/code/webcompat.com/tests/test_webhook.py", line 138
    """Extract dictionary of metadata for an issue body.""".
                                                           ^
SyntaxError: invalid syntax
```

miketaylr: Lol, what a lame mistake.

miketaylr:
nose tests passing on Travis: https://travis-ci.org/webcompat/webcompat.com/builds/327334792#L639

Let's merge. My plan is to not deploy to prod until Travis is green on master though.

PR #2020, opened 2018-01-12
Fixes #2019. Append extra labels in correct format.

### LINKED ISSUE #2019: extra labels not applied to new issues
```
<!-- @browser: Firefox 58.0 -->
<!-- @ua_header: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:58.0) Gecko/20100101 Firefox/58.0 -->
<!-- @reported_with: desktop-reporter -->
<!-- @t: y -->
```

Looks like I broke this! Investigating...

miketaylr:
The intermediary metadata list expects a list of tuples, with the
metadata key being the first member of the tuple, and its value the
second.

Also, adds an assertion to explicitly test getting `extra_labels` metadata, since it gets special handling.

### DIFF
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -5,6 +5,8 @@
 
 import unittest
 
+from werkzeug import MultiDict
+
 import webcompat
 from webcompat import form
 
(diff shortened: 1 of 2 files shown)


tests/test_form.py
```diff
@@ -129,6 +131,17 @@ def test_get_metadata(self):
         actual = form.get_metadata(metadata_keys, form_object)
         expected = u'<!-- @sky: blue -->\n<!-- @earth: None -->\n'
         self.assertEqual(actual, expected)
+        form_object = MultiDict([
```
miketaylr: (using a MultiDict here is just me being lazy and copying some of the values the app sends...rather than re-formatting into a plain old dict)

karlcow: just fyi if you need again in the future, (I don't think it creates an issue here apart of the dependency.) `MultiDict()` has `form_object.to_dict()`, which converts to a normal dictionary.

```python
>>> form_object.to_dict()
{'url': u'http://localhost:5000/issues/new', 'reported_with': u'desktop-reporter', 'ua_header': u'Mozilla/5.0...Firefox 59.0', 'extra_labels': [u'type-stylo', u'type-webrender-enabled'], 'browser': u'Firefox 59.0'}
```

karlcow: Interesting to see we had a missing test.

PR #2029, opened 2018-01-20
Fixes #1311: Return 403 for PR issues.

### LINKED ISSUE #1311: Pull requests showing on webcompat.com
Thanks to @zoepage for bringing this up.

We can see pull requests to the web-bugs repo on webcompat.com.
https://webcompat.com/issues/4544

This alone isn't a big deal and seems like a limitation of our GH integration. But it probably shouldn't appear on the "all open issues" page.
https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc

Thoughts?

miketaylr:
r? @karlcow

### DIFF
--- a/tests/unit/test_urls.py
+++ b/tests/unit/test_urls.py
@@ -144,6 +144,26 @@ class TestURLs(unittest.TestCase):
         self.assertTrue('text/html' in rv.content_type)
         self.assertTrue(content_test)
 
+    def test_error_routes(self):
+        """Request to /error/<int> should return proper statuses.
+
+        Undefined <int> statuses return 500."""
+        rv = self.app.get('/error/404')
+        self.assertEqual(rv.status_code, 404)
+        rv = self.app.get('/error/403')
+        self.assertEqual(rv.status_code, 403)
+        rv = self.app.get('/error/4000')
+        self.assertEqual(rv.status_code, 500)
+        rv = self.app.get('/error/🤷')
+        self.assertEqual(rv.status_code, 404)
+
+    def test_catchall_404(self):
+        """Unknown routes should always return 404."""
+        rv = self.app.get('/xkxkdjdjdllkj;a;slkdjf')
+        self.assertEqual(rv.status_code, 404)
+        rv = self.app.get('/🤷')
+        self.assertEqual(rv.status_code, 404)
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -455,6 +455,13 @@ def api_request(method, path, params=None, data=None):
     resource = request_method(method, path, headers=request_headers,
                               params=params, data=data)
     if resource.status_code != 404:
+        try:
+            if 'pull_request' in json.loads(resource.content):
+                return (json.dumps({'message': 'Pull Requests are Forbidden.',
+                                    'status': 403}),
+                        403, get_response_headers(resource))
+        except:
+            pass
         return (resource.content, resource.status_code,
                 get_response_headers(resource))
     else:
--- a/webcompat/static/js/lib/issues.js
+++ b/webcompat/static/js/lib/issues.js
@@ -434,19 +434,17 @@ issues.MainView = Backbone.View.extend({
         }, this)
       )
       .error(function(response) {
-        var msg;
-        if (
-          response &&
-          response.responseJSON &&
-          response.responseJSON.message === "API call. Not Found"
-        ) {
-          location.href = "/404";
-          return;
-        } else {
-          msg =
-            "There was an error retrieving the issue. Please reload to try again.";
-          wcEvents.trigger("flash:error", { message: msg, timeout: 4000 });
+        if (response && response.status === 404) {
+          return (location.href = "/error/404");
         }
+
+        if (response && response.status === 403) {
+          return (location.href = "/error/403");
+        }
+
+        var msg =
+          "There was an error retrieving the issue. Please reload to try again.";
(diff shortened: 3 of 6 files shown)


webcompat/helpers.py
```diff
@@ -455,6 +455,12 @@ def api_request(method, path, params=None, data=None):
     resource = request_method(method, path, headers=request_headers,
                               params=params, data=data)
     if resource.status_code != 404:
+        try:
+            if 'pull_request' in json.loads(resource.content):
+                return ('Pull Requests are Forbidden.', 403,
+                         get_response_headers(resource))
```
miketaylr: Hm, I think we're sending back a JSON mimetype at this point, but the content is not valid JSON. To fix, even though nothing in this PR relies on it.

webcompat/static/js/lib/issues.js
```diff
@@ -434,19 +434,17 @@ issues.MainView = Backbone.View.extend({
         }, this)
       )
       .error(function(response) {
-        var msg;
-        if (
-          response &&
-          response.responseJSON &&
-          response.responseJSON.message === "API call. Not Found"
```
miketaylr: Turns out this was broken. Caring about the `message`, rather than the `status` was a silly idea.

webcompat/helpers.py
miketaylr: Updated.

tests/unit/test_urls.py
```diff
@@ -144,6 +144,26 @@ def test_dashboard_route(self):
...
+
+        Undefined <int> statuses return 500."""
+        rv = self.app.get('/error/404')
+        self.assertEqual(rv.status_code, 404)
+        rv = self.app.get('/error/403')
+        self.assertEqual(rv.status_code, 403)
+        rv = self.app.get('/error/4000')
+        self.assertEqual(rv.status_code, 500)
+        rv = self.app.get('/error/🤷')
+        self.assertEqual(rv.status_code, 404)
```
karlcow: @miketaylr (Pre-review question) Was there a need for creating routes for HTTP errors? HTTP errors code are supposed to inform on the specific URI you are hitting at. They don't exist with a specific URI. If we redirect to a specific route we loose the benefits of the HTTP code.

miketaylr: @karlcow yeah. before these changes, we had/have a "catchall" 404, controlled by nginx. so if you want to send someone to a 404 via JS, because the API told you 404, that's easy: redirect to `https://webcompat.com/404`. It's sort of cheating, but it works (because that `/404` truly doesn't exist).

The difficulty here, AFAICT, is we don't know about the PR status until one of many HTTP requests (the API request to the `/api/issues/` endpoint) returns a 403. We can't know beforehand, unless we hard-code.

Now we need to send a 403 for a certain response. What are the options?

* Send 403, nothing happens on the page. Looks like a bug, because well, it is one. 
* Send 403, redirect to 404 page. That seems wrong.
* Send 403, redirect to a 403 page. That requires the ability to do that. That's the approach I took.

The other way to deal with this would be to display a 403 or 404 message entirely via the client. 

> If we redirect to a specific route we loose the benefits of the HTTP code.

I'm not sure I understand this point of view. If you curl `localhost:5000/api/issues/516` you get what you would expect:

```
http -I http://localhost:5000/api/issues/516
HTTP/1.0 403 FORBIDDEN
Content-Length: 58
Content-Security-Policy: default-src 'self'; object-src 'none'; connect-src 'self' https://api.github.com; font-src 'self'; img-src 'self' https://www.google-analytics.com https://*.githubusercontent.com data:; manifest-src 'self'; script-src 'self' https://www.google-analytics.com https://api.github.com; style-src 'self' 'unsafe-inline'; report-uri /csp-report
Date: Mon, 22 Jan 2018 23:04:45 GMT
Server: Werkzeug/0.12.2 Python/2.7.10
X-Frame-Options: DENY
cache-control: private, max-age=60, s-maxage=60
content-type: application/json
etag: W/"2d026dee32c4e3332b9f859081319ee4"

{
    "message": "Pull Requests are Forbidden.", 
    "status": 403
}
```

(it's also true if you curl `localhost:5000/error/403`, you get a 403. it feels a bit philosophical... maybe there's actually something there, but i won't let you see it).

karlcow: @miketaylr See my comment in the issue itself so we have a history on possible outcomes.
https://github.com/webcompat/webcompat.com/issues/1311#issuecomment-359689001

miketaylr: Sounds like you agree with "The other way to deal with this would be to display a 403 or 404 message entirely via the client." (or rather, tell the user that they can't see that issue).

If someone wants to pick up the pieces of this PR, be my guest. Have bigger fish to fry. 🐟

karlcow: no 403, no 404. Just sending 200. but with an appropriate message. I'm happy to take over this :)

miketaylr: > 403 or 404 message

Meaning, tell the user the issue is verboten. HTTP returns a 200 for HTML + other assets. The API endpoint returns a 403, which the JS reacts to and populates the template, no?

karlcow: 403/404 means the URI is forbidden. We indeed return 200 for everything because of the way we are set up. Let me make the suggestion in a different PR. Do not destroy this one. :)

miketaylr: (I think it makes sense that a URI that points to a PR issue should be considered forbidden by webcompat.com. 🍎 🐛

but also, not a hill i want to die on 😄 )

PR #2034, opened 2018-01-24
Fixes #1891: Get form working on refactor branch.

### LINKED ISSUE #1891: refactor template: form.html
Refactor following: 
https://github.com/webcompat/webcompat.com/blob/master/webcompat/templates/home-page/form.html

with: 

The form in http://webcompat.github.io/design/

Please be aware to add the placeholders for links as they are in the old templates.

Left do do:

- [x] Fix screenshot uploader
- [x] Make sure validation works

miketaylr:
WIP.


miketaylr:
OK, image upload preview functionality is working again.

@zoepage I added some styles to start from, feel free to fix/change/whatever after merging. Obviously the error one is pretty terrible. :)

screen shot 2018-01-25 at 11 53 59 am

screen shot 2018-01-25 at 11 53 32 am

There's a few issues that aren't addressed in this PR, in case @zoepage is already tracking them. But to highlight them:

* [ ] form errors are impossible to see as-is. i presume we'll add `--color-error` styles?
* [ ] once we add a style for that,  the existing`wc-Form-helpMessage` class can be replaced with it everywhere (tests, js, templates)
* [ ] the upload spinner is commented out, because there's future work to re-do the spinner (as i understand, before the refactor lands). so that will need to be re-enabled and positioned.
* [ ] the current positioning of the `<input type=file>` and its associated label is sub-optimal:

screen shot 2018-01-25 at 12 40 06 pm

It's possible to click above or below the label and then the filepicker doesn't appear. So possibly we want to make that 100% of height as well. 



r? @zoepage

webcompat/static/css/webcompat.dev.css
```diff
@@ -58,3 +58,7 @@
 .is-hidden {
   display: none;
 }
+
+.visually-hidden {
```
miketaylr: I added this style because if we `display: none` the label/svg, the container collapses and it looks super broken.

webcompat/templates/home-page/form.html
```diff
@@ -89,41 +89,38 @@ <h1 class="headline-1">Report Site Issue</h1>
       {% endfor %}
     </div>
     <div class="form-upload form-element js-image-upload js-Form-group">
-      <label class="wc-UploadForm wc-UploadForm--homePage js-image-upload-label" for="image">
```
miketaylr: Interesting fact: we had `<label for=image>`, but the associated `<input>` never had an `id=image`. But it didn't matter because the input is associated by the label that wraps it anyways. 🌠

webcompat/templates/home-page/form.html
```diff
@@ -89,41 +89,38 @@ <h1 class="headline-1">Report Site Issue</h1>
...
-        <div class="wc-UploadForm-wrapper">
-          <label class="form-label label-upload wc-UploadForm-label">
-            <svg class="icon icon-big wc-UploadForm-icon" role="presentation">
-              <use class="icon-upload" href="#svg-upload-2">Upload</use>
-            </svg>
-            Upload screenshot
-          </label>
-        </div>
+        <label class="form-label label-upload" for="image">
+          <svg class="icon icon-big wc-UploadForm-icon" role="presentation">
```
miketaylr: I don't think `wc-UploadForm-icon` is used anymore, so it could be removed later.

zoepage: Pretty much all `wc-`  styles can be removed.

miketaylr: Just grep the tests and update them when you do that -- the odds are pretty good tests are relying on them. :)

miketaylr:
@zoepage I rebased against latest refactor and fixed the merge conflicts in the tests -- all yours to take over, thanks!

webcompat/static/js/lib/flash-message.js
```diff
@@ -32,20 +32,22 @@ var FlashMessageView = Backbone.View.extend({
     var timeout = opts.timeout || 4000;
     var message = opts.message;
 
+    this.$el.addClass("notification-information");
     this.render(message);
     setTimeout(_.bind(this.hide, this), timeout);
   },
   showError: function(data) {
-    this.$el.addClass("is-error");
+    this.$el.addClass("is-error notification-alert");
```
miketaylr: This means we have `notification-bar js-flashmessage is-error notification-alert notification-information` classes. Is that expected? That is, are `-alert` and `-information` designed to coexist?

webcompat/static/js/lib/flash-message.js
```diff
@@ -32,20 +32,22 @@ var FlashMessageView = Backbone.View.extend({
...
   },
   showForever: function(opts) {
     var message = opts.message;
+    this.$el.addClass("notification-information");
     this.render(message);
   },
   showThanks: function(opts) {
     var buildTemplate = wcTmpl["issue/thanks.jst"];
-    this.$el.addClass("is-active wc-FlashMessage--thanks");
+    this.$el.addClass("is-active notification-information");
```
miketaylr: This will get a double `notification-information` class, because `show` adds one too.

webcompat/templates/home-page/form.html
```diff
@@ -88,48 +83,47 @@ <h1 class="headline-1">Report Site Issue</h1>
...
-          >
-          Remove image upload
-        </button>
-      </label>
     </div>
+
     <div class="form-button form-element js-Form-group">
       <button
-        class="button button-primary wc-Button wc-Button--default js-Button"
+        class="button button-third js-Button"
```
miketaylr: What does `button-third` mean? I don't see it anywhere else.

zoepage:
@miketaylr the 3 states:

screen shot 2018-02-02 at 2 27 00 pm
screen shot 2018-02-02 at 2 26 21 pm
screen shot 2018-02-02 at 2 26 09 pm

webcompat/static/css/src/notification-bar.css
```diff
@@ -1,6 +1,6 @@
 .notification-bar {
-  position: absolute;
-  z-index: 101;
+  position: fixed;
+  z-index: 3101;
```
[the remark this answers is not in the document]
zoepage: It does, but this is doesn't seem obvious enough then. I'll fix it.

webcompat/static/img/svg/icons/svg-cross2.svg
```diff
@@ -1,5 +1,12 @@
...
-<title>cross2</title>
-<path fill="#bf0101" d="M30.173 24.76v0l-8.76-8.76 8.76-8.76c0.094-0.094 0.162-0.204 0.206-0.321 0.119-0.321 0.050-0.697-0.206-0.955l-4.138-4.138c-0.258-0.258-0.633-0.327-0.955-0.206-0.117 0.043-0.227 0.112-0.321 0.206v0l-8.76 8.76-8.76-8.76c-0.094-0.094-0.204-0.162-0.321-0.206-0.321-0.119-0.697-0.052-0.955 0.206l-4.138 4.138c-0.258 0.258-0.327 0.633-0.206 0.955 0.043 0.117 0.112 0.227 0.206 0.321v0l8.76 8.76-8.76 8.76c-0.094 0.094-0.162 0.204-0.206 0.321-0.119 0.321-0.052 0.697 0.206 0.955l4.138 4.138c0.258 0.258 0.633 0.327 0.955 0.206 0.117-0.043 0.227-0.112 0.321-0.206v0l8.76-8.76 8.76 8.76c0.094 0.094 0.204 0.162 0.321 0.206 0.321 0.119 0.697 0.052 0.955-0.206l4.138-4.138c0.258-0.258 0.327-0.633 0.206-0.955-0.043-0.117-0.112-0.227-0.206-0.321z"></path>
-</svg>
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="30px" height="30px" viewBox="0 0 30 30" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 43.2 (39069) - http://www.bohemiancoding.com/sketch -->
+    <title>svg-x</title>
+    <desc>Created with Sketch.</desc>
```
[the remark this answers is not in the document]
zoepage: We actually have that in place. Probably just forgot to run it.

webcompat/templates/home-page/form.html
zoepage: Thanks for the hint. This needs to go.

webcompat/static/css/src/form.css
```diff
@@ -75,7 +104,7 @@
   background-image: url("../../img/svg/icons/svg-checkmark.svg");
 }
 
-.error::after {
+.is-error .error::after {
```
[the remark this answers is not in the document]
zoepage: Hm, I'm not sure if I understand this completely. 🤔

`is-error` is the state, `error` the error element itself. (Maybe this gives a bit of context already?)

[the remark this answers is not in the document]
miketaylr: OK. I'm not sure what an error element with a non-error state means, but I trust it all makes sense. :)

miketaylr:
@zoepage looks good! One concern is what if the screenshot is solid black, will it be possible to see the X Remove Screenshot? That's why we had the yellow overlay before. This could be a follow up issue, but I think something we want to fix before shipping the refactor.

webcompat/static/js/lib/bugform.js
```diff
@@ -116,6 +117,11 @@ function BugForm() {
     // so, bytes = (encoded_string.length - 814) / 1.37)
     // see https://en.wikipedia.org/wiki/Base64#MIME
     if (String(dataURI).length - 814 / 1.37 > this.UPLOAD_LIMIT) {
+      wcEvents.trigger("flash:error", {
```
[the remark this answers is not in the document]
zoepage: This was removed, but showed after I fixed the detached head.

webcompat/static/css/src/form.css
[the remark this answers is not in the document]
zoepage: I'll try to explain a bit more. :)

`is-error` is set by the JS and accounts for a whole node. So, e.g. it is responsible for an error-message that is shown or / and a svg icon in the input field, change of text color. So this is the `state` of the node.

The `error` here is the error indicator itself. If you feel like grouping the size and `display` attribute in a class, we can do that!

zoepage:
@miketaylr the remove x turns yellow on hover, so we have covered light and dark images.

miketaylr:
Excellent! r+

(normally I would recommend fixing up some of the history, but let's just merge).

PR #2035, opened 2018-01-25
[not for merging] Issue #2029 - Testing a progressive DB cache idea for issues.

karlcow:
"Slow" first, Blazing fast subsequent requests.

All of this is just prototype pseudo-working-code
to explain the concept.

This pull request will be scraped for a proper one.

webcompat-tests has a pull request on  516.

We create an *empty* issues.db.

1. The first request for a specific issue will trigger a call
   to the DB which will return None.
2. The None will create a request to the API endpoint to get
   the JSON of the issue.
   It gathers the title and the status of the issue (PR or not)
   It then adds it to the issues.db
   If the status is_issue = False, we can directly send a 403.
3. Subsequent requests for the same will trigger the issues.db
   which this time will have the content.

Some benefits:

* We do not have to run an out of process to build the issues.db
* The issues.db can be scraped anytime (on purpose or by accident),
  it will rebuild itself. The system doesn't fail.
* It becomes better little by little with what is really used.
* We can progressively enhance the issues.db
* It could help us to be less dependent on JS for the issue little by little
  (if we think it is a good idea)
* We can test performances to make sure we are not making mistakes.
* We could add more stuff to it and slowly have an issue db usable for search,
  that would speed up our full process and make search easy

### DIFF
--- a/webcompat/db/__init__.py
+++ b/webcompat/db/__init__.py
@@ -10,6 +10,7 @@ from hashlib import sha512
 import os
 from uuid import uuid4
 
+from sqlalchemy import Boolean
 from sqlalchemy import Column
 from sqlalchemy import create_engine
 from sqlalchemy.ext.declarative import declarative_base
@@ -20,18 +21,26 @@ from sqlalchemy import String
 
 from webcompat import app
 
+# Sessions database for logged in users
 session_engine = create_engine('sqlite:///' + os.path.join(
     app.config['DATA_PATH'], 'session.db'))
 session_db = scoped_session(sessionmaker(autocommit=False,
                                          autoflush=False,
                                          bind=session_engine))
 
+# Amazon Top Sites priority ranking database
 site_engine = create_engine('sqlite:///' + os.path.join(
     app.config['DATA_PATH'], 'topsites.db'))
 site_db = scoped_session(sessionmaker(autocommit=False,
                                       autoflush=False,
                                       bind=site_engine))
 
+# Issues database
+issues_engine = create_engine('sqlite:///' + os.path.join(
+    app.config['DATA_PATH'], 'issues.db'))
+issues_db = scoped_session(sessionmaker(autocommit=False,
+                                        autoflush=False,
+                                        bind=issues_engine))
 
 UsersBase = declarative_base()
 UsersBase.query = session_db.query_property()
@@ -62,6 +71,7 @@ SiteBase.query = site_db.query_property()
 
 class Site(SiteBase):
     """SQLAchemy base object for an Alexa top site."""
+
     __tablename__ = 'topsites'
 
     url = Column(String, primary_key=True)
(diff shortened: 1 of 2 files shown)


karlcow:
@miketaylr this is really code for discussions. It's another take on the issue. I'm happy to explain the concepts. We can hop on a call if it is easier live.

webcompat/db/__init__.py
```diff
@@ -70,9 +80,32 @@ class Site(SiteBase):
...
+
+    issue_number = Column(Integer, primary_key=True)
+    title = Column(String)
+    is_issue = Column(Boolean, default=True)
+
+    def __init__(self, issue_number, title, is_issue):
+        """Initialize the topsite db parameters."""
+        self.issue_number = issue_number
+        self.title = title
+        self.is_issue = is_issue
```
miketaylr: I guess it would be good to spend time being thoughtful about what info we would need in an issues db to avoid unnecessary migrations in the future. Would we want labels? Milestones? Something else?

webcompat/views.py
```diff
@@ -243,6 +246,46 @@ def show_issue(number):
...
+    # All of this is just prototype pseudo-working-code
+    # to explain the concept. It needs to be distributed
+    # in definitions with their own set of tests.
+    # The pull request on webcompat-tests is 516.
+    # We query the issues.db which is a kind of progressive cache.
+    # aka we do not have to run an out of process to build the issues.db
+    # It becomes better little by little with what is really used.
+    # It also can be empty and it will not fail.
+    # Benefits of the two worlds.
+    # We could add more stuff to it and slowly have an issue db
```
miketaylr: Ah cool. So my previous comment about migrations isn't that important if we're treating it more of a cache than a primary datasource.

webcompat/views.py
```diff
@@ -243,6 +246,46 @@ def show_issue(number):
...
+    # that would speed up our full process and make search easy
```
miketaylr: (Speed makes sense, search seems hard. But search is always hard.)

webcompat/views.py
```diff
@@ -243,6 +246,46 @@ def show_issue(number):
...
+    issue = Issues.query.filter_by(issue_number=number).first()
+    # The issue at first request will never exist.
+    # So we are going to request directly GitHub, but only once.
+    if issue is None:
```
miketaylr: Big question: how do we handle edits to issue body content on GitHub to prevent serving stale content? Or labels or comments changes/additions/deletion (assuming we care to cache those things here). Webhooks? Cron? Etags (somehow)? Dunno, food for thought.

karlcow: So for now, we do not have to. The thing is really to solve the 403 issue with a door opened for more stuff. And for the title I put it here in the prototype just to show how it could evolve. Title was an obvious choice because of the bookmarking and tabs thing with more information for users. 

Yes Webhooks is one possibility for updating the content of the DB. I would not cache the comments yet. And if we were doing it, it would probably be wise to put them in a separate DB.

Another possibility, if we keep the principle of failsafe, aka the DB can be erased all the time and the system continue to work, we can be creative for populating the DB. For example, it's possible to have a script feeding the DB by just querying all the issues once. out of webcompat.com processes.

So I propose two steps.

1. Create a proper pull request for this (with observers for measuring performances) aka DB.
2. Deploy on staging and test
3. Remove the observers
4. Deploy on production if we are satisfied.

miketaylr: @karlcow let's do it.

karlcow:
ok let's close this Pull Request in favor of #2050

Issue #2040, opened 2018-01-26
Handle details as JSON


miketaylr:
Needed for changes related to https://bugzilla.mozilla.org/show_bug.cgi?id=1422985. We'll send:

`details=JSONSTRING`, and consume it in bugform.js.

miketaylr:
Oddly enough, the test from https://github.com/webcompat/webcompat.com/commit/440bad707f8dedde1256d66297d1741ed5e4bd10#diff-db9f827f59656441c082dbc15f76426b is no longer on master... possibly swallowed by the Intern 4 merge. :|

But, I was going to delete it anyways.

karlcow:
@miketaylr we close this?

PR #2041, opened 2018-01-26
Issue #2040. Handle a single details param as JSON encoded data.

miketaylr:
No more multiple details params, but instead it will be sent as a JSON string and parsed by the client.

### DIFF
--- a/tests/functional/reporting-non-auth.js
+++ b/tests/functional/reporting-non-auth.js
@@ -11,9 +11,9 @@ const path = require("path");
 var cwd = intern.config.basePath;
 var VALID_IMAGE_PATH = path.join(cwd, "tests/fixtures", "green_square.png");
 var BAD_IMAGE_PATH = path.join(cwd, "tests/fixtures", "evil.py");
+// DETAILS_STRING is a URL encoded array, stringified to JSON.
 var DETAILS_STRING =
-  "Encountered error: NS_ERROR_DOM_MEDIA_DEMUXER_ERR (0x806e000c)%0ALocation: virtual%0ARefPtrMP4Demuxer::InitPromise mozilla::MP4Demuxer::Init()%0AError information:%0AIncomplete MP4 metadata%0AMedia URL: file:///Users/potch/Documents/mozilla/media.mp4";
-var DETAILS_STRING2 = "Encountered+error:+NS_ERROR_DOM_MEDIA_DEMUXER_ERR";
+  "%5B%22gfx.webrender.all%3A+false%22%2C%22gfx.webrender.blob-images%3A+2%22%2C%22gfx.webrender.enabled%3A+false%22%2C%22image.mem.shared%3A+2%22%2C%22layout.css.servo.enabled%3A+true%22%5D";
 
 var url = function(path) {
   return intern.config.siteRoot + path;
@@ -271,35 +271,17 @@ registerSuite("Reporting (non-auth)", {
         .findByCssSelector("#steps_reproduce")
         .getProperty("value")
         .then(function(text) {
-          assert.include(
-            text,
-            "Encountered error: NS_ERROR_DOM_MEDIA_DEMUXER_ERR (0x806e000c)\nLocation: virtual\nRefPtrMP4Demuxer::InitPromise mozilla::MP4Demuxer::Init()\nError information:\nIncomplete MP4 metadata\nMedia URL: file:///Users/potch/Documents/mozilla/media.mp4"
-          );
-        })
-        .end();
-    },
-
-    "details param: + decoded as space"() {
-      return FunctionalHelpers.openPage(
-        this,
-        url("/issues/new?details=" + DETAILS_STRING2),
-        "#description"
-      )
-        .findByCssSelector("#steps_reproduce")
-        .getProperty("value")
-        .then(function(text) {
           assert.notInclude(
             text,
-            "Encountered+error:+NS_ERROR_DOM_MEDIA_DEMUXER_ERR",
+            "gfx.webrender.all:+false",
             "+ not found in decoded string"
           );
           assert.include(
             text,
-            "Encountered error: NS_ERROR_DOM_MEDIA_DEMUXER_ERR",
+            "gfx.webrender.all: false\ngfx.webrender.blob-images: 2",
             "+ replaced with space"
           );
-        })
-        .end();
+        });
     }
   }
 });
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -176,25 +176,22 @@ function BugForm() {
       $("[value=" + problemType[1] + "]").click();
     }
 
-    // If we have one or more details params,
-    // add that to the end of the steps to reproduce textarea
-    var details = location.href.match(/details=([^&]*)/g);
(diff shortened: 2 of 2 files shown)


miketaylr:
Need to update tests...

webcompat/static/js/lib/bugform.js
```diff
@@ -176,25 +176,23 @@ function BugForm() {
...
-          )
+          JSON.parse(
+            decodeURIComponent(
+              details[1]
+                // The content of the details param may be encoded via
+                // application/x-www-form-urlencoded, so we need to change the
+                // + (SPACE) to %20 before decoding
+                .replace(/\+/g, "%20")
+            )
+          ).join("\n")
```
karlcow: Here I see something related to return `\n`. But I don't see tests trying to handle this or maybe I don't understand how it is working. Should there be a test testing that the steps to reproduce field contains a list of things with hard returns, except of only one value as of now.

webcompat/static/js/lib/bugform.js
```diff
@@ -176,25 +176,23 @@ function BugForm() {
       $("[value=" + problemType[1] + "]").click();
     }
 
-    // If we have one or more details params,
+    // If we have a details params, JSON decode it and
     // add that to the end of the steps to reproduce textarea
-    var details = location.href.match(/details=([^&]*)/g);
-    if (details !== null) {
+    var details = location.href.match(/details=([^&]*)/);
```
karlcow: I guess details is what is sent from the browser side to the form. It's a rather vague name compared to something like `browser_config` but let's fly with it given that it is not really part of this PR. I just realized now how it was called. 

```js
var url = 'https://webcompat.com/issues/new?url=https%3A%2F%2Fadobe.github.io%2FSpry%2Farticles%2Fjson_primer%2Fjson_primer.html&src=desktop-reporter&details=%5B%22gfx.webrender.all%3A+false%22%2C%22gfx.webrender.blob-images%3A+2%22%2C%22gfx.webrender.enabled%3A+false%22%2C%22image.mem.shared%3A+2%22%2C%22layout.css.servo.enabled%3A+true%22%5D&label=type-stylo'
var details = url.match(/details=([^&]*)/);
// Array [ "details=%5B%22gfx.webrender.all%3A+false%22%2C%22gfx.webrender.blob-images%3A+2%22%2C%22gfx.webrender.enabled%3A+false%22%2C%22image.mem.shared%3A+2%22%2C%22layout.css.servo.enabled%3A+true%22%5D", "%5B%22gfx.webrender.all%3A+false%22%2C%22gfx.webrender.blob-images%3A+2%22%2C%22gfx.webrender.enabled%3A+false%22%2C%22image.mem.shared%3A+2%22%2C%22layout.css.servo.enabled%3A+true%22%5D" ]
details[1]
// "%5B%22gfx.webrender.all%3A+false%22%2C%22gfx.webrender.blob-images%3A+2%22%2C%22gfx.webrender.enabled%3A+false%22%2C%22image.mem.shared%3A+2%22%2C%22layout.css.servo.enabled%3A+true%22%5D"
details[1].replace(/\+/g, "%20")
// "%5B%22gfx.webrender.all%3A%20false%22%2C%22gfx.webrender.blob-images%3A%202%22%2C%22gfx.webrender.enabled%3A%20false%22%2C%22image.mem.shared%3A%202%22%2C%22layout.css.servo.enabled%3A%20true%22%5D"
decodeURIComponent(details[1].replace(/\+/g, "%20"))
// "[\"gfx.webrender.all: false\",\"gfx.webrender.blob-images: 2\",\"gfx.webrender.enabled: false\",\"image.mem.shared: 2\",\"layout.css.servo.enabled: true\"]"
JSON.parse(decodeURIComponent(details[1].replace(/\+/g, "%20")))
// Array [ "gfx.webrender.all: false", "gfx.webrender.blob-images: 2", "gfx.webrender.enabled: false", "image.mem.shared: 2", "layout.css.servo.enabled: true" ]
JSON.parse(decodeURIComponent(details[1].replace(/\+/g, "%20"))).join("\n")
/* "gfx.webrender.all: false
gfx.webrender.blob-images: 2
gfx.webrender.enabled: false
image.mem.shared: 2
layout.css.servo.enabled: true" */
```

And all of this is added to the Steps to reproduce field. 

This works as expected.
I have seen a couple of times, people confused about why there was these things in the textarea. It probably looks like something broken.

What about feeding these to a hidden input parameter in the form and to add them to the issue body after POSTING. So people from triage see the parameters, but not reporters at the moment of filing the bug.

We keep the information, but it doesn't get into the way of reporting.

tests/functional/reporting-non-auth.js
```diff
@@ -267,39 +266,21 @@ registerSuite("Reporting (non-auth)", {
...
         .findByCssSelector("#steps_reproduce")
         .getProperty("value")
         .then(function(text) {
           assert.notInclude(
             text,
-            "Encountered+error:+NS_ERROR_DOM_MEDIA_DEMUXER_ERR",
+            "gfx.webrender.all:+false",
             "+ not found in decoded string"
           );
           assert.include(
```
karlcow: maybe an assert.include with two lines instead of only one. so we can test that there hard return worked.

webcompat/static/js/lib/bugform.js
miketaylr: We use `\n` just to make the information easier to digest in the textarea/final report. 
(Before, that was handled by Firefox). IMO, testing the formatting of the output isn't super important.

webcompat/static/js/lib/bugform.js
miketaylr: > What about feeding these to a hidden input parameter in the form and to add them to the issue body after POSTING. So people from triage see the parameters, but not reporters at the moment of filing the bug.

Yep, I like this idea a lot. I've been wanting to do something similar for `about:support` data, but rather than appearing in the issue body, put that in a `.txt` file somewhere on the server. I'll file a follow up bug.

miketaylr: > It's a rather vague name compared to something like browser_config but let's fly with it given that it is not really part of this PR.

Good point. Details originally was added to support `type-media` error messages and stack traces. So it really was more about contextual details.

webcompat/static/js/lib/bugform.js
```diff
@@ -176,25 +176,23 @@ function BugForm() {
       $("[value=" + problemType[1] + "]").click();
     }
 
-    // If we have one or more details params,
+    // If we have a details params, JSON decode it and
```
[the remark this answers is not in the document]
miketaylr: Yeah, it's correct English because it's a singular param now (with a confusingly plural name). Less confusing would be something like `// If we have details, JSON decode it...`. I'll update.

miketaylr:
Thanks @karlcow. Will update before merging.

webcompat/static/js/lib/bugform.js
miketaylr: I think I merged this too quickly! I'm gonna just send a JSON blob from firefox, then build up the entire content on the frontend -- and write tests for that. That'll mean less changes once we move to a hidden input as well (since the input wont know or care about `\n`'s or any particular format). Will send in another quick PR before deploying any of this.

Issue #2042, opened 2018-01-29
Put `details` contents in a hidden input


miketaylr:
Rather than populate the textarea (which can be ugly/confusing), @karlcow suggested we put it in a hidden input. We will display the content the same once the issue is created.

See https://github.com/webcompat/webcompat.com/pull/2041#discussion_r164319434

miketaylr:
Another thing that might be good is once the issue is created, to put this pref inside a `<details>` element so it can be hidden by default. It's not always interesting to stare at.

e.g. https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d

zoepage:
@miketaylr happy to tackle the `detail` issue with #2011

miketaylr:
> @miketaylr happy to tackle the detail issue with #2011

Cool. Let's wait until we've fixed this bug first, before deciding how to do that. Thanks!

miketaylr:
(and that random idea should be its own issue at some point -- this one is about hidden inputs).

PR #2044, opened 2018-01-29
Fixes #2043: Handle incoming details as JSON object.

### LINKED ISSUE #2043: Handle incoming details JSON object
I was a bit hasty with the #2041 PR.

https://github.com/webcompat/webcompat.com/pull/2041#discussion_r164319318

Rather than get a half-processed array from Firefox (created from an object...) then processing some more on webcompat, it seems nicer to just send that format-agnostic object and handle all processing by webcompat.

miketaylr:
This way we consume the data however we want on our side, without having to change Firefox if we want some different format.

r? @karlcow

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -179,22 +179,25 @@ function BugForm() {
     // If we have details, JSON decode it and add it to the end of STR textarea
     var details = location.href.match(/details=([^&]*)/);
     if (details) {
-      this.stepsToReproduceField.val(function(idx, value) {
-        return (
-          value +
-          "\n" +
-          JSON.parse(
-            decodeURIComponent(
-              details[1]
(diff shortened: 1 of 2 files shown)


miketaylr:
(last change, until we change it again in... #2042 :))

webcompat/static/js/lib/bugform.js
```diff
@@ -179,24 +179,27 @@ function BugForm() {
...
-      });
+      this.stepsToReproduceField.val(
+        _.bind(function(idx, value) {
+          return value + "\n" + this.buildDetails(details[1]);
+        }, this)
+      );
     }
   };
 
+  this.buildDetails = function(detailsParam) {
```
karlcow: So much more pythonic 😆

miketaylr: Haha, I think you're rubbing off on me.

Issue #2107, opened 2018-02-16
Fix contributors-non-auth.js


miketaylr:
Blocks #2106.

miketaylr:
We can probably replace this test suite entirely with some python unit tests, just to ensure the endpoints result in a 200.

miketaylr:
Oh wait, @karlcow already wrote those unit tests. I'll just delete the file.

Issue #2106, opened 2018-02-16
Get tests green on refactor branch


miketaylr:
Once this is done, no more PRs into refactor will be allowed without fixing broken tests.

miketaylr:
```
× firefox on darwin 17.4.0 - Comments (auth) - Posting a comment
× firefox on darwin 17.4.0 - Comments (auth) - Add a screenshot to a comment
× firefox on darwin 17.4.0 - Comments (auth) - Pressing 'l' inside of comment textarea *doesn't* open the label editor box
× firefox on darwin 17.4.0 - Contributors - page loads
× firefox on darwin 17.4.0 - Contributors - clicking first section closes it
× firefox on darwin 17.4.0 - Contributors - clicking section toggles it
× firefox on darwin 17.4.0 - Contributors - toggling section toggles lightbulb
× firefox on darwin 17.4.0 - History navigation - Back button works from issues page
× firefox on darwin 17.4.0 - Image Uploads (non-auth) - remove image upload button
× firefox on darwin 17.4.0 - Image Uploads (non-auth) - double image select works
× firefox on darwin 17.4.0 - Index - front page loads
× firefox on darwin 17.4.0 - Index - reporter addon link is shown
× firefox on darwin 17.4.0 - Index - form toggles open then closed
× firefox on darwin 17.4.0 - Index - browse issues (needstriage)
× firefox on darwin 17.4.0 - Issue-list - FilterView renders
× firefox on darwin 17.4.0 - Issue-list - IssueListView renders
× firefox on darwin 17.4.0 - Issue-list - Dropdowns reflect state from URL
× firefox on darwin 17.4.0 - Issue-list - Going back in history updates issue list and URL state
× firefox on darwin 17.4.0 - Issue-list - Clicking on a stage filter adds the correct param to the URL
× firefox on darwin 17.4.0 - Issue-list - Toggling a stage filter doesn't leave the param in the URL
× firefox on darwin 17.4.0 - Issue-list - Toggling between stage filters results in last param in URL
× firefox on darwin 17.4.0 - Issues (auth) - Pressing 'l' opens the label editor box
× firefox on darwin 17.4.0 - Issues - Issue page loads
× firefox on darwin 17.4.0 - Labels (auth) - Label editor opens then closes (key events)
× firefox on darwin 17.4.0 - Milestones (auth) - Milestone editor opens then closes (key events)
× firefox on darwin 17.4.0 - Milestones (non-auth) - Page loads without milestone set
[...]

PR #2145, opened 2018-02-18
Issue #2107. Remove contributors-non-auth.js.

miketaylr:
We no longer have the interactive bits these tests were exercising.

r=me


miketaylr:
#2106

PR #2176, opened 2018-02-21
Fixed #2075 - fixes validation checkmarks

### LINKED ISSUE #2075: Missing green checkmarks when adding URL and description
In the existing webcompat form, when you put in a URL and description, you get a green checkmark.

In the form on staging, nothing happens there. Seems like a bug.

zoepage:
r? @miketaylr

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -77,9 +77,14 @@ function BugForm() {
     );
     this.problemType.on("change", _.bind(this.checkProblemTypeValidity, this));
     this.uploadField.on("change", _.bind(this.checkImageTypeValidity, this));
-    this.osField
-      .add(this.browserField)
-      .on("blur input", _.bind(this.checkOptionalNonEmpty, this));
+    this.osField.on(
+      "blur",
+      _.bind(this.checkOptionalNonEmpty, this, this.osField)
+    );
+    this.browserField.on(
+      "blur",
+      _.bind(this.checkOptionalNonEmpty, this, this.browserField)
+    );
     this.submitButtons.on("click", _.bind(this.storeClickedButton, this));
     this.submitButtons.on("click", _.bind(this.loadingIndicator.show, this));
     this.form.on("submit", _.bind(this.maybeUploadImage, this));
@@ -292,18 +297,14 @@ function BugForm() {
 
   /* Check if Browser and OS are empty or not, only
      so we can set them to valid (there is no invalid state) */
-  this.checkOptionalNonEmpty = function() {
-    _.forEach(
-      [this.browserField, this.osField],
-      _.bind(function(input) {
-        var inputId = input.prop("id");
-        if (input.val()) {
-          this.makeValid(inputId);
-        } else {
-          this.makeInvalid(inputId);
-        }
-      }, this)
-    );
+  this.checkOptionalNonEmpty = function(input) {
+    var inputId = input.prop("id");
+
+    if (input.val()) {
+      this.makeValid(inputId);
+    } else {
+      this.makeInvalid(inputId);
+    }
   };
 
   this.checkForm = function() {
@@ -327,8 +328,11 @@ function BugForm() {
       }
     }
     // Make sure we only do this if the inputs exist on the page
-    if (this.browserField.length || this.osField.length) {
-      this.checkOptionalNonEmpty();
+    if (this.browserField.length) {
+      this.checkOptionalNonEmpty(this.browserField);
+    }
+    if (this.osField.length) {
+      this.checkOptionalNonEmpty(this.osField);
     }
   };
 
--- a/webcompat/templates/home-page/form.html
+++ b/webcompat/templates/home-page/form.html
@@ -65,7 +65,9 @@
     <div class="form-text form-element js-Form-group">
       {{ form.browser.label(class_='form-label') }}
       {{ form.browser(class_='form-field text-field') }}
-      <span class="form-input-validation check js-bug-form-os"></span>
+      <span class="form-input-validation check"></span>
+    </div>
+    <div class="form-text form-element js-Form-group">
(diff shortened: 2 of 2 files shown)


webcompat/static/js/lib/bugform.js
```diff
@@ -292,11 +297,13 @@ function BugForm() {
 
   /* Check if Browser and OS are empty or not, only
      so we can set them to valid (there is no invalid state) */
-  this.checkOptionalNonEmpty = function() {
+  this.checkOptionalNonEmpty = function(types) {
     _.forEach(
-      [this.browserField, this.osField],
-      _.bind(function(input) {
+      [...types],
```
miketaylr: We can't use es6 rest because it will break the reporting form for IE users, and other older browsers. :(

webcompat/static/js/lib/bugform.js
```diff
@@ -292,11 +297,13 @@ function BugForm() {
...
+      _.bind(function(key) {
+        const input = this[key];
```
miketaylr: Can we change this to `var` please? We're doing ES5 until we decide we're ready to ship ES6 (or transpile -- but that's probably a future discussion).

webcompat/static/js/lib/bugform.js
```diff
@@ -77,9 +77,14 @@ function BugForm() {
...
-    this.osField
-      .add(this.browserField)
-      .on("blur input", _.bind(this.checkOptionalNonEmpty, this));
+    this.osField.on(
+      "blur",
+      _.bind(this.checkOptionalNonEmpty, this, ["osField"])
+    );
+    this.browserField.on(
+      "blur",
+      _.bind(this.checkOptionalNonEmpty, this, ["browserField"])
```
miketaylr: Any reason to not just pass in the reference to the DOM elements directory here?

webcompat/static/js/lib/bugform.js
zoepage: Sorry, I shouldn't switch between projects. 😅
Will send a fix.

webcompat/static/js/lib/bugform.js
zoepage: I'm not sure I understand the wording. 🤔

miketaylr: Let me rephrase:

You're changing `checkOptionalNonEmpty` to accept an argument `types`, which is an array of keys. For each of those keys, you grab a reference to the corresponding DOM input element via `input = this[key];`.

Why not just pass in `this.browserField` directly and skip passing the array of strings that ultimately just returns `this.browserField`?

`_.bind(this.checkOptionalNonEmpty, this, [this.browserField])`, etc.

zoepage: This is actually exactly what I did today. Small fix and will update the PR :)

webcompat/static/js/lib/bugform.js
```diff
@@ -77,9 +77,14 @@ function BugForm() {
...
-    this.osField
-      .add(this.browserField)
-      .on("blur input", _.bind(this.checkOptionalNonEmpty, this));
+    this.osField.on(
+      "blur",
+      _.bind(this.checkOptionalNonEmpty, this, this.osField)
+    );
+    this.browserField.on(
+      "blur",
+      _.bind(this.checkOptionalNonEmpty, this, this.browserField)
```
miketaylr: If we're going to do it this way, we don't even need to pass in `this.browserField` -- we can just get that from `event.target` down in `checkOptionalNonEmpty`, right?

webcompat/static/js/lib/bugform.js
```diff
@@ -77,9 +77,14 @@ function BugForm() {
     );
     this.problemType.on("change", _.bind(this.checkProblemTypeValidity, this));
     this.uploadField.on("change", _.bind(this.checkImageTypeValidity, this));
-    this.osField
-      .add(this.browserField)
-      .on("blur input", _.bind(this.checkOptionalNonEmpty, this));
+    this.osField.on(
+      "blur",
```
miketaylr: Is removing the `input` handler intentional?

miketaylr:
> And if we run through the inputs with the if, we get rid of the for loop. Felt bit more efficient.

(I'm not sure looping over a list of 2 items is a significant perf bottleneck here. :))

webcompat/static/js/lib/bugform.js
zoepage: adding both would mean to add a bind to both. I don't feel this is needed. Am I missing something?

webcompat/static/js/lib/bugform.js
zoepage: With `this.browserField` we have already set a reference and I feel like this is more readable for someone not knowing the code, but maybe I am missing something?

miketaylr: Sure, I guess it's just a question of style preference.

webcompat/static/js/lib/bugform.js
miketaylr: Not sure I follow, let me try again:

Before we had `$(foo).on("blur input", bar)`. Now we have $(foo).on("blur", bar)`, meaning we're not handling the `input` event.

What do you mean by adding `bind` to both? I don't follow. Thanks!

zoepage: Sorry for the short and confusing answer. Okay, let me rephrase it. :)

So, we add `blur` and `input` as events to trigger the function. I think, `blur` is enough as this will add / remove the checkmark as needed and we don't fire the function on every change of the input value. So the answer to your question `Is removing the input handler intentional?` is yes. 

If you think, this feels weird or does not cover all use cases, I am ok to put it back. I guess, this is mostly about UX preference.

miketaylr: SGTM. Let's merge now and if it feels wrong later on, file a bug to revisit/change it back.

PR #2186, opened 2018-02-22
(ignore, last time i think?)

miketaylr:


miketaylr:
tests are green: https://travis-ci.org/webcompat/webcompat.com/builds/344841465?utm_source=github_status&utm_medium=notification

PR #2190, opened 2018-02-27
Fix #2032 - Adds icons for filtering on issue age and needinfo

### LINKED ISSUE #2032: [triage dashboard]:  How to handle empty priority?
For an issue without a priority label (uh, unsure if we broke this, but I think it's possible not every issue will end up with one), we show an empty *Priority:*

Maybe we could have the strong "none" or not show it if there's no priority?

screen shot 2018-01-23 at 10 08 46 am

cc @magsout @karlcow

karlcow:
- Adds icons in the card
- Adds icons in the filter select
- Adds titles for the meaning of the icon

Rendering

capture d ecran 2018-02-27 a 16 13 05

r? @miketaylr for design
r? @magsout for code

### DIFF
--- a/webcompat/templates/dashboard/triage/filters.html
+++ b/webcompat/templates/dashboard/triage/filters.html
@@ -45,8 +45,8 @@
         name="state"
         id="js-Filter-state"
       >
-        <option value="priority" selected="selected">Priority</option>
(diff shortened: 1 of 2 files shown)


miketaylr:
r+ for design. :)))

PR #2191, opened 2018-02-27
Fixes #2171 - add test for 404 label search

### LINKED ISSUE #2171: Write functional tests for browser label search suggestion
The refactor broke this for a while, which exposed the fact we don't have tests for it:

https://github.com/webcompat/webcompat.com/pull/2168

STR and Expected behavior were more elegantly written by @softvision-oana-arbuzov in https://github.com/webcompat/webcompat.com/issues/2153#issue-298259452:

> **Steps to Reproduce:**
> 1. Navigate to https://staging.webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc
> 2. Tap the "Search" field and type in a string that returns no results (e.g. Testtesttest).
> 3. After error message is displayed, observe suggestions - popular labels (e.g. android/chrome/firefox/ie).
> 4. Tap one of the popular labels and observe behavior (e.g. chrome).

> **Expected Behavior:**
> Redirect to specific page is performed
> (e.g. https://webcompat.com/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=label%3Abrowser-chrome).

zoepage:
@miketaylr I'm having issues with this test. Maybe you can help?

I've added a comment here: https://github.com/webcompat/webcompat.com/blob/issues/2171/tests/functional/search-auth.js#L230

The issue is that the 404 page is not loaded as the URL directs to `http://localhost:5000/api/issues?...` and not `http://localhost:5000/issues?...` as expected. So we don't have a redirect.

Please don't mind the URL params that are passed into the `openPage` function. Also the test itself contains a few things to try out.

### DIFF
--- a/tests/fixtures/api/issues.2c36ce8c0fec0c185101ba8f6b203f48.json
+++ b/tests/fixtures/api/issues.2c36ce8c0fec0c185101ba8f6b203f48.json
@@ -1,0 +1,6 @@
+{
+	"_fixture": true,
+	"total_count": 0,
+	"incomplete_results": false,
+	"items": []
+}
--- a/tests/functional/_intern.js
+++ b/tests/functional/_intern.js
@@ -59,9 +59,7 @@ intern.configure({
   reporters: [args.reporters ? args.reporters : "pretty"],
 
   functionalSuites: [
-    args.functionalSuites
-      ? args.functionalSuites
-      : "./tests/functional/*.js"
+    args.functionalSuites ? args.functionalSuites : "./tests/functional/*.js"
   ]
 });
 
--- a/tests/functional/_intern.js
+++ b/tests/functional/_intern.js
@@ -59,7 +59,9 @@ intern.configure({
   reporters: [args.reporters ? args.reporters : "pretty"],
 
   functionalSuites: [
-    args.functionalSuites ? args.functionalSuites : "./tests/functional/*.js"
+    args.functionalSuites
+      ? args.functionalSuites
+      : "./tests/functional/*.js"
   ]
 });
 
--- a/tests/functional/search-auth.js
+++ b/tests/functional/search-auth.js
@@ -62,96 +62,15 @@ registerSuite("Search (auth)", {
         .end();
     },
 
-    "Results are loaded from the query params"() {
-      var params =
-        "?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=vladvlad";
-      return FunctionalHelpers.openPage(
-        this,
-        url("/issues", params),
-        ".js-IssueList:nth-of-type(1)"
-      )
-        .findDisplayedByCssSelector(".js-IssueList:nth-of-type(1) a")
-        .getVisibleText()
-        .then(function(text) {
-          assert.include(
-            text,
-            "vladvlad",
-            "The search query results show up on the page."
-          );
-        })
-        .end()
-        .getCurrentUrl()
-        .then(function(currUrl) {
-          assert.include(
-            currUrl,
-            "q=vladvlad",
-            "Our params didn't go anywhere."
-          );
-        })
-        .end();
-    },
-
-    "Search works by icon click"() {
-      return FunctionalHelpers.openPage(
-        this,
-        url("/issues"),
-        ".js-IssueList:nth-of-type(10)"
-      )
-        .findByCssSelector(".js-SearchForm input")
-        .type("vladvlad")
-        .end()
-        .findByCssSelector(".js-SearchForm button")
-        .click()
-        .end()
-        .findDisplayedByCssSelector(".js-IssueList:only-of-type a")
-        .getVisibleText()
-        .then(function(text) {
-          assert.include(
-            text,
-            "vladvlad",
-            "The search results show up on the page."
-          );
-        })
-        .end();
-    },
-
-    "Search works by Return key"() {
-      return FunctionalHelpers.openPage(
-        this,
-        url("/issues"),
-        ".js-SearchForm input"
-      )
-        .findDisplayedByCssSelector(".js-SearchForm input")
-        .type("vladvlad")
-        .type("\uE007")
-        .end()
-        .findDisplayedByCssSelector(".js-IssueList:only-of-type a")
-        .getVisibleText()
-        .then(function(text) {
-          assert.include(
-            text,
-            "vladvlad",
-            "The search results show up on the page."
-          );
-        })
-        .end();
-    },
-
-    "Search from the homepage"() {
-      return (
-        FunctionalHelpers.openPage(this, url("/"), ".js-SearchBarOpen")
-          .get(url("/"))
-          .findByCssSelector(".js-SearchBarOpen")
-          .click()
-          .end()
-          // Wait for animation to complete.
-          .sleep(1000)
-          .findByCssSelector(".js-SearchBar input")
-          .click()
-          .type("vladvlad")
-          .type("\uE007")
-          .end()
-          .findDisplayedByCssSelector(".js-IssueList:only-of-type a")
+      "Results are loaded from the query params"() {
+        var params =
+          "?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=vladvlad";
+        return FunctionalHelpers.openPage(
+          this,
+          url("/issues", params),
+          ".js-IssueList:nth-of-type(1)"
+        )
+          .findDisplayedByCssSelector(".js-IssueList:nth-of-type(1) a")
           .getVisibleText()
           .then(function(text) {
             assert.include(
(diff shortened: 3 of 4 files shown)


tests/functional/_intern.js
```diff
@@ -59,7 +59,7 @@ intern.configure({
   reporters: [args.reporters ? args.reporters : "pretty"],
 
   functionalSuites: [
-    args.functionalSuites ? args.functionalSuites : "./tests/functional/*.js"
+    args.functionalSuites ? args.functionalSuites : "./tests/functional/search-auth.js"
```
miketaylr: (tip: when running a single test, you can do the following:

`node ./tests/functional/_intern.js functionalSuites=tests/functional/search-auth.js` 

And then it will only run that test)

miketaylr: Not useful to you now, but I'm going to fix some things in https://github.com/webcompat/webcompat.com/issues/2193 so you can just do `npm run:test:js -- --functionalSuites=blah --grep=interaction`, then you don't have to comment anything out either.

miketaylr:
> The issue is that the 404 page is not loaded as the URL directs to http://localhost:5000/api/issues?... and not http://localhost:5000/issues?... as expected. So we don't have a redirect.

@zoepage can you explain in psuedo test code how you want to test this?

Also, where do you see that linked error message? 

When I try a search with no results, I get a 200 from `https://api.github.com/search/issues?page=1&per_page=50&stage=all&sort=created&q=asdfasdfasdfasdf+state:open+repo:webcompat/webcompat-tests&order=desc`, with a more or less empty response:

```
{
  "total_count": 0,
  "incomplete_results": false,
  "items": [

  ]
}
```

thanks!

miketaylr:
OK, I think I understand now.

```
Expected fixture file: /Users/mitaylor/dev/webcompat.com/tests/fixtures/api/issues.2c36ce8c0fec0c185101ba8f6b203f48.json
127.0.0.1 - - [27/Feb/2018 13:12:55] "GET /api/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=noExpectedResult123 HTTP/1.1" 404 -

```

Since you're running in test mode against a new search URI (q=noExpectedREsults123), you'll need to add a new fixture file called `issues.2c36ce8c0fec0c185101ba8f6b203f48.json`, assuming you stick with that URI (the hash is computed against the params).

See https://github.com/webcompat/webcompat.com/blob/master/docs/tests.md#adding-fixtures

(the easiest way to create the fixture is do that search in non-testmode in Chrome, and copy and paste the response JSON into a new file with the expected name, and add a `"_fixture": true,` key to the first object)

tests/functional/_intern.js
zoepage: 👍

zoepage:
@miketaylr this looks like the missing piece. thanks! :)

tests/functional/search-auth.js
```diff
@@ -198,6 +198,48 @@ registerSuite("Search (auth)", {
...
     },
 
+    "Label have correct URL after search without results"() {
+      // load up a garbage search, so we can more easily detect when
+      // the search values we want are loaded.
+      return FunctionalHelpers.openPage(
+        this,
+        url(
+          "/issues",
+          "?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc"
```
miketaylr: These are the default issues page params, right?

https://github.com/webcompat/webcompat.com/blob/61f4ef1be90d7d90eaec1af7a2fb75b1f055aeb9/webcompat/static/js/lib/models/issue.js#L123-L129

Probably not needed for `openPage` here if so.

tests/functional/search-auth.js
```diff
@@ -198,6 +198,48 @@ registerSuite("Search (auth)", {
...
+        ),
+        ".grid"
+      )
+        .findByCssSelector("#js-SearchForm-input")
+        .clearValue()
+        .click()
```
miketaylr: I believe you don't need `clearValue()` here because by default the value will be empty on page load.

tests/functional/search-auth.js
```diff
@@ -198,6 +198,48 @@ registerSuite("Search (auth)", {
...
+        .type("noExpectedResult123")
+        .submit()
+        .getProperty("value")
+        .then(function(text) {
+          assert.equal(
+            text,
+            "noExpectedResult123",
+            "The q param populated the search input."
```
miketaylr: This assertion description doesn't match what the test is doing -- it's not looking at `q`. But that said, what is the point of this part of the test? You're entering text into the input and making sure that it's still there? Is there a situation where it wouldn't be?

tests/functional/search-auth.js
```diff
@@ -198,6 +198,48 @@ registerSuite("Search (auth)", {
...
+          );
+        })
+        .end()
+        .findByCssSelector("#x-search-bar")
+        .submit()
+        .end()
+        .sleep(2000)
```
miketaylr: Does this test work without the `sleep(2000)`? In theory, there's a default find timeout of 10 seconds.

tests/functional/search-auth.js
```diff
@@ -198,6 +198,48 @@ registerSuite("Search (auth)", {
...
+        .findDisplayedByCssSelector(".label-browser-android")
+        .click()
+        .end()
+        .getCurrentUrl()
+        .then(function(url) {
+          assert.equal(
+            url,
+            "http://localhost:5000/issues?page=1&per_page=50&state=open&stage=all&sort=created&direction=desc&q=label%3Abrowser-android",
```
miketaylr: I *think* it might be possible for this to fail if someone passed in a new `--siteRoot`, other than `http://localhost:5000/`. I'll follow a file up bug, because it's hardcoded in other tests.

Maybe a safer option here would be `assert(url.includes("q=label%3Abrowser-android"))`, since that's the interesting bit of the test.

tests/functional/search-auth.js
zoepage: You're right. This is probably a left over of the hunt for the issue is had with the test before. Will remove.

tests/functional/search-auth.js
zoepage: You're right. Leftover.

tests/functional/search-auth.js
zoepage: I had issues with the input and got in some cases an `undefined`, but this is probably worth a separate test. Will file a follow up issue.

tests/functional/search-auth.js
zoepage: 👍

zoepage: I'll add this to this test, so we have less work later on.

tests/functional/search-auth.js
zoepage: It does not as it looks like the test is loosing the scope through the redirect. The label element is not found.

tests/functional/search-auth.js
zoepage: I can't reproduce right now. If it shows up again, I will add tests for that.

tests/functional/search-auth.js
```diff
@@ -198,6 +198,34 @@ registerSuite("Search (auth)", {
...
+        this,
+        url(
+          "/issues"
+        ),
+        ".grid"
+      )
+        .findByCssSelector("#js-SearchForm-input")
+        .click()
+        .type("noExpectedResult123")
+        .submit()
```
miketaylr: I think we can get rid of this `submit()` too, since you're submitting the form just after.

tests/functional/search-auth.js
miketaylr: Weird, there isn't even a real redirect here, just some XHR requests. But if it works... seems good. :)

tests/functional/search-auth.js
```diff
@@ -198,6 +198,34 @@ registerSuite("Search (auth)", {
...
+        .findByCssSelector("#x-search-bar")
+        .submit()
+        .end()
+        .sleep(2000)
+        .findDisplayedByCssSelector(".label-browser-android")
+        .click()
+        .end()
+        .getCurrentUrl()
+        .then(function(url) {
+          assert(url.includes("q=label%3Abrowser-android"));
```
miketaylr: Can you add a description as a 2nd argument to `assert`?

miketaylr: (oh weird, i already wrote this comment...) 👻

tests/functional/search-auth.js
```diff
@@ -198,6 +198,34 @@ registerSuite("Search (auth)", {
         .end();
     },
 
+    "Label have correct URL after search without results"() {
+      // load up a garbage search, so we can more easily detect when
+      // the search values we want are loaded.
+      return FunctionalHelpers.openPage(
+        this,
```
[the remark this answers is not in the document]
zoepage: This is a weird behaviour I ran into from time to time (UX wise). After fixing the lint issue, I forget to add the fixed file and push the commit without the fix. Maybe something like `npm run lint:js && git add .` would fix this?

tests/fixtures/api/issues.2c36ce8c0fec0c185101ba8f6b203f48.json
```diff
@@ -0,0 +1,6 @@
+{
+	"_fixture": true,
```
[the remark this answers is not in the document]
zoepage: 👍

tests/functional/search-auth.js
zoepage: fixed above.

tests/functional/search-auth.js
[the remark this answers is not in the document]
miketaylr: Ah, interesting.

I finally just configured my text editor to auto-prettier all JS files on save, because I had done the same things as well. I'd be a little nervous about making `npm run lint:js && git add .` a default command, frequently I have a lot of unstaged changes that I want to include in different commits.

PR #2200, opened 2018-02-28
Fixes #2101 - Contrast in between instructions and user input

### LINKED ISSUE #2101: Contrast in between instructions and user input
The new form is generally more elegant than the previous one, but it also has its own set of UX issues.
Once I have filed it, it becomes hard to review and see at first sight what I have entered there. Specifically in the Description and textarea sections.

capture d ecran 2018-02-16 a 12 41 32 2

zoepage:
r? @karlcow 
r? @miketaylr 
screen shot 2018-02-28 at 8 45 56 pm

### DIFF
--- a/webcompat/static/css/src/form.css
+++ b/webcompat/static/css/src/form.css
@@ -23,8 +23,8 @@
 .form-element {
   position: relative;
   max-width: 100%;
-  margin: 0;
   padding: 0;
+  margin: 0 0 20px;
 }
 
 .form-field {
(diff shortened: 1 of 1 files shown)


miketaylr:
The additional "breathing room" looks good to me, though I'm not sure how to quantify is contrast is better or not. Let's merge and maybe do some light user testing before launch to see if this is a stumbling point for people.

(I think the placeholder text makes it obvious where to type things, but it's less clear once there's content in the inputs -- but hard to say if this is a real problem or not).

karlcow:
@zoepage Thanks! Let's see with a bit of usage if it is still an issue. We can improve later with small steps if necessary.

Issue #2083, opened 2018-02-15
Slack as a communication channel


karlcow:
As for now, slack is not our main communication channel.

<img width="332" alt="capture d ecran 2018-02-16 a 07 40 18" src="https://user-images.githubusercontent.com/505230/36284886-d38cdf0e-12ec-11e8-98dd-6ea32fd1b99f.png">

We never came to a decision on this. Personally I don't think we should do it. The more complete discussion we had about this is in #1284

miketaylr:
There's too much slack in my life already, I would prefer if we didn't create a new comms channel beyond irc + github issues.

miketaylr:
(oh, and meetings)

zoepage:
We have the slack channel already and we used it also for the sprints. If we want to get more contributors and especially a more diverse (= skill set, experience level) crowd of people, I do not see a way around this.

karlcow:
> I do not see a way around this.

Yes there is. We can decide that we would not have a slack channel and live with it. That's a choice. I understand the appeal of the "new" tools and the generational impact to it. When the Webcompat IRC channel started, there was barely 5 persons on it on a daily basis. There are more than 65 people right now as I write this. People come and go. They contribute for a while, disappear or stick around. That's life. There is no urgency on changing that. Slow growth for a community is also a good thing. It might seem frustrating, and I kind of understand the appeal of Slack, but that doesn't equate to we should do it.

miketaylr:
> We can decide that we would not have a slack channel and live with it.

The opposite is also true. My opinion on this has evolved to be: if there are people in the slack who can help out that community, great. 

(I don't see myself idling there, fwiw, but that's because I'm already idling in too many slack and irc channels as is).

zoepage:
I'm okay with hanging out in the Slack and answer questions. I'm sure, after the community grows we'll have more people to help out answering questions.

The next steps would be:
- [ ] Add Slack invite link to contact page
- [ ] Add CoC related responses per Slack bot
- [ ] Invite the team & ping people on IRC & Moz Slack

karlcow:
ok. Then be it. 😈 We can revisit the issue later if it's not successful. Or embrace it if it is. 
 
* One of my uncomfortable feeling about it is yet another reliance on a private communication system.
* I'll probably need to find a way to get the archives of the messages out of slack. Maybe this? https://get.slack.help/hc/en-us/articles/201658943-Exporting-your-team-s-Slack-history

[...]

PR #2211, opened 2018-03-07
Issue #2210. Add contact page edits.

miketaylr:
r? @zoepage

### DIFF
--- a/webcompat/templates/contact.html
+++ b/webcompat/templates/contact.html
@@ -9,7 +9,7 @@
 		</div>
 		<div class="grid-cell x2">
 		    <p>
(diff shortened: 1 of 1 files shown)


webcompat/templates/contact.html
```diff
@@ -27,17 +27,17 @@ <h1 class="headline-1 highlight">Say hello!</h1>
...
-		    		Join us on our 
+		    		Join us on our
 		    		<a href="https://lists.mozilla.org/listinfo/compatibility" title="Join us on our mailing list">
 		    			mailing list
 		    		</a>
-		    	</li> 
+		    	</li>
 		    	<li>
 		    		Join our
 		    		<a href="https://webcompat.slack.com/" title="Join our Slack">
```
miketaylr: I know #2083 is not resolved yet, so we should resolve it one way or the other before we ship this refactor.

zoepage: Sounds good. :)

zoepage: Let's merge and I'll put #2083 on the v1 list?

Issue #1924, opened 2017-11-21
make animation of bug in hero section run just once


zoepage:
Follow up for https://github.com/webcompat/design/pull/79

The animation should appear just on the very first load. We can set a status in the localStorage and make sure, if gets checked before the animation starts.

Edit:
We also need a fix for the bug falling out of the scene. Right now it falls all the way down... it should stay in the hero section box and disappear there.

miketaylr:
> The animation should appear just on the very first load. We can set a status in the localStorage and make sure, if gets checked before the animation starts.

Having played around with this locally, I would recommend we only make it animate after some kind of interaction -- I think it's a bit too distracting, even for the first load.

zoepage:
Was thinking about this too recently. How about a hover and just one time, so it has an easter egg feel to it?

miketaylr:
> Was thinking about this too recently. How about a hover and just one time, so it has an easter egg feel to it?

I like this. 👍

zoepage:
Maybe @mpaarating might be interested in working on this after #2061 ? :)

deeptibaghel:
@miketaylr I’d like to work on this for my outreachy contribution!

zoepage:
@deeptibaghel Feel free! Please make sure to branch from the `refactor`  branch. Let me know, if you have questions :)

deeptibaghel:
@zoepage i tried to work branch refactor but got errors as below :
<img width="1450" alt="screen shot 2018-03-06 at 7 34 13 pm" src="https://user-images.githubusercontent.com/448609/37036379-84f8a016-2175-11e8-8369-46e2590a903b.png">

deeptibaghel:
Also i can see the animation in https://staging.webcompat.com/ but not on https://webcompat.com.
Can i get any hint on which js files would be having the animation related code.

deeptibaghel:
@miketaylr , sorry yeah i missed building it. Also i sensed it is coming from svg, but then "lang: js" tag in issue made me think that way.
The animation seems to be running only once to me on page load, without any interaction, so we need to change it to happen on hover ?

zoepage:
Sorry for the confusion with the `lang: JS` tag. The initial idea was to just run it once and store the status in the local Storage. This is why we added the JS tag. It's all CSS now. :)

The animation should just run on hover of the bug. :)

zoepage:
merged #2219

PR #2219, opened 2018-03-07
Fixes Issue #1924 make animation of bug in hero section run just once

deeptibaghel:
The bug now starts moving on mouse over and runs only once.

### DIFF
--- a/webcompat/static/css/src/animations.css
+++ b/webcompat/static/css/src/animations.css
@@ -66,12 +66,12 @@
     transform: translate3d(130px, -20px, 0) rotateZ(5deg) scale3d(1.2, 1.2, 1);
   }
   94% {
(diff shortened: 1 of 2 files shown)


miketaylr:
thanks @deeptibaghel!

r? @zoepage

zoepage:
Thanks @deeptibaghel :)

The initial idea was different, but I actually like it more this way!
(Now you kind of have to hunt the bug towards the bulb.) Could we add a small change and remove the animation of the bug where it falls down? So it would end when it arrives at the bulb and wobble a little bit? Please let me know, if you have any questions.

Thanks for the great idea. That added the last missing piece.

deeptibaghel:
@zoepage , i think now it's better :)

zoepage:
Love it! Thank you!

miketaylr:
This is very cool, thank you @deeptibaghel.

PR #2235, opened 2018-03-14
WIP: Fixes #2205 – Refine margins and paddings around hero element

### LINKED ISSUE #2205: Hero image on non-desktop
screen shot 2018-03-06 at 2 27 35 pm

I was pretty sure this was reported already, but couldn't find it.

Regaddi:
r? @zoepage

### DIFF
--- a/webcompat/static/css/src/grid.css
+++ b/webcompat/static/css/src/grid.css
@@ -1,5 +1,5 @@
 .grid {
-  margin: 9px;
+  margin: 9px 0;
 }
 
 .grid-row {
--- a/webcompat/static/css/src/variables.css
+++ b/webcompat/static/css/src/variables.css
@@ -1,6 +1,6 @@
 :root {
-  --grid-edge-s: 9px;
-  --grid-edge-m: 19px;
+  --grid-outermargin-s: 9px;
+  --grid-outermargin-m: 19px;
   --unit-modul-s: 270px;
   --unit-modul-m: 585px;
   --unit-modul-l: 900px;
--- a/webcompat/static/css/src/variables.css
+++ b/webcompat/static/css/src/variables.css
@@ -1,4 +1,6 @@
 :root {
+  --grid-edge-s: 9px;
+  --grid-edge-m: 19px;
   --unit-modul-s: 270px;
   --unit-modul-m: 585px;
   --unit-modul-l: 900px;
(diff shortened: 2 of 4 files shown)


webcompat/static/css/src/grid.css
```diff
@@ -1,5 +1,5 @@
 .grid {
-  margin: 9px;
+  margin: 9px 0;
```
zoepage: Should we use the var here too?

webcompat/static/css/src/variables.css
```diff
@@ -1,4 +1,6 @@
 :root {
+  --grid-edge-s: 9px;
```
zoepage: As we deal with compatibility issues, edge might be misleading? 😅
How about we change it to `outermargin` or something similar?

webcompat/static/css/webcompat.dev.css
```diff
@@ -40,18 +40,18 @@
 
 #body-webcompat {
   background-color: var(--color-bg-page);
-  margin-top: var(--navi-height);
+  padding: var(--navi-height) var(--grid-edge-s) 25px;
```
zoepage: could we change the `25px` to `--unit-gutter-m` as this should work context wise?

webcompat/static/css/webcompat.dev.css
```diff
@@ -40,18 +40,18 @@
...
 }
 
 @media screen and (min-width: 665px) {
   #body-webcompat {
-    padding: var(--navi-height) 19px 50px;
+    padding: var(--navi-height) var(--grid-edge-m) 50px;
```
zoepage: I know, we used a fix value here, but could we change the `50px` as we are on it anyway to `calc(--unit-gutter-m * 2)`?

webcompat/static/css/src/grid.css
Regaddi: This is probably not related / a side effect. It was the initial overall margin for all `.grid` elements and therefore produced the left and right margin.

zoepage: Just to make sure, we don't misunderstand each other, this was referring to the `9px` and if we should change this to `--grid-outermargin-s`. But this is 100% not blocking and we are free to merge!

zoepage:
Merging this as the last comment is nothing pressuring. 

Great work @Regaddi && thank you so much! :)

PR #2236, opened 2018-03-14
Fixes #2134 - fixes fade in load page

### LINKED ISSUE #2134: Issue page flash of unstyled content
STR: go to an issue page on staging.webcompat.com. that's it.

Before we attempted to fade everything in once content was ready, now it does this:

fouc

magsout:
r @zoepage ?

### DIFF
--- a/webcompat/templates/issue.html
+++ b/webcompat/templates/issue.html
@@ -20,7 +20,7 @@
     </article>
 </section> -->
 
-    <section id="issue" class="grid-row js-Issue">
+    <section id="issue" class="grid-row js-Issue is-hidden">
       <article class="grid-nested x2 issue-labels">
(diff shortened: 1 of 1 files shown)


webcompat/templates/issue.html
```diff
@@ -20,7 +20,7 @@
     </article>
 </section> -->
 
-    <section id="issue" class="grid-row js-Issue">
+    <section id="issue" class="grid-row js-Issue" style="display: none">
```
zoepage: that's great! :)
could we change the inline style to the class `is-hidden`? I've tested it out and it works. 
Otherwise thank you for figuring it out so quickly! <3

magsout:
@zoepage done

PR #2237, opened 2018-03-14
Fixes #2234 - empty search returns undefined

### LINKED ISSUE #2234: Empty search submit shows "undefined" in search bar
**Steps to reproduce:**

1.  Go to https://staging.webcompat.com
2. Click on "Search"
3. Leave input empty and hit enter
4. Searchbar on issue navigator contains "undefined"

bildschirmfoto 2018-03-14 um 14 08 04

zoepage:
r? @miketaylr 

(I know, we could also have prevented the submit here, but we didn't want to touch any JS code. So it just quick fixed it)

### DIFF
--- a/webcompat/static/js/lib/issue-list.js
+++ b/webcompat/static/js/lib/issue-list.js
@@ -266,7 +266,7 @@ issueList.IssueView = Backbone.View.extend(
     },
     // NOTE: these filters don't need "status-" prefixes because appear in URL params
     _filterRegex: /&*stage=(closed|contactready|needscontact|needsdiagnosis|needstriage|sitewait)&*/i,
-    _searchRegex: /&*q=(?:(.+)?)&*/i,
+    _searchRegex: /&*q=(?:([^&=]+))&*/i,
     _githubSearchEndpoint: "https://api.github.com/search/issues",
     _isLoggedIn: $("body").data("username"),
     _loadingIndicator: $(".js-Loader"),
--- a/webcompat/templates/list-issue/issuelist-search.jst
+++ b/webcompat/templates/list-issue/issuelist-search.jst
@@ -1,10 +1,10 @@
 <script type="text/template">
-  <form id="x-search-bar" method="get" action="/issues">
-    <input id="js-SearchForm-input" class="form-field text-field" name="q" placeholder="Search issues by keyword"/>
+  <div id="x-search-bar">
+    <input id="js-SearchForm-input" class="form-field text-field mousetrap" name="q" placeholder="Search issues by keyword"/>
     <button role="button" class="nav-button nav-link js-SearchForm-button" type="button" title="Search for an item.">
       <svg class="icon nav-icon" viewBox="0 0 30 30" role="presentation">
         <use href="#svg-search2">Search</use>
       </svg>
     </button>
-    </form>
+    </div>
 </script>
--- a/webcompat/templates/list-issue/issuelist-search.jst
+++ b/webcompat/templates/list-issue/issuelist-search.jst
@@ -1,10 +1,10 @@
 <script type="text/template">
-  <div id="x-search-bar">
-    <input id="js-SearchForm-input" class="form-field text-field mousetrap" name="q" placeholder="Search issues by keyword"/>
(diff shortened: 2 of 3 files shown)


Regaddi:
The change doesn't solve #2234 for me 😿Still undefined in the input field.

zoepage:
Did you hard reload the page @Regaddi ?

Regaddi:
Yes, I did. I checked out the issues/2234 branch locally, restarted `npm run start`, emptied the browser cache and did the steps as described in #2234.
Still undefined 🤔

Regaddi:
bildschirmfoto 2018-03-15 um 09 02 43

Regaddi:
I think it's some JavaScript inserting the value here. When reloading I can see the placeholder for a very short time.

Regaddi:
[This Regex](https://github.com/webcompat/webcompat.com/blob/refactor/webcompat/static/js/lib/issue-list.js#L269) is not completely correct.
It currently matches `&q=`, but fails to put the empty string in the match result.
We could change the Regex to `/&*q=(?:([^&=]+))&*/i`.

This would also prevent another issue. See this: https://staging.webcompat.com/issues?q=&page=1&per_page=50&state=open&stage=all&sort=created&direction=desc

zoepage:
@miketaylr what do you think?

miketaylr:
> @miketaylr what do you think?

I'll have some time to look at this this afternoon, but at first glance, we should fix the core issue rather than paper over it by changing a `form` to a `div`.

miketaylr:
(it shouldn't even be possible to submit an empty search, IMO)

Regaddi:
As an alternative we can set a default value [here](https://github.com/webcompat/webcompat.com/blob/refactor/webcompat/static/js/lib/issue-list.js#L171).

Just like that:

```javascript
  updateSearchQuery: function(data) {
    this.input.val(data || '');
  }
```

zoepage:
>  we should fix the core issue rather than paper over it by changing a form to a div.

I've adapted the old semantics as we did not want to change any JS. Happy to provide a proper fix.

miketaylr:
> This Regex is not completely correct.

@Regaddi great find, can you file a new bug for that? as i understand it, to fix this bug, we need to make sure the form is not submitting if it's empty (or not letting the default action happen, which for a form when you hit enter is to attempt a POST)

Regaddi:
> we need to make sure the form is not submitting if it's empty (or not letting the default action happen, which for a form when you hit enter is to attempt a POST)

@miketaylr This is partly true. Preventing a form submit, when the search input is empty fixes only one symptome of this issue. The other thing is when the issues page is navigated with an empty `q=` parameter. This can happen independent of the search form behaviour.

So IMO it's irrelevant, if we prevent the form submit, when the handling of the `q=` parameter is correct.

miketaylr:
@Regaddi These still seem like 2 different bugs to me:

1. user can submit an empty search
2. user can navigate to a URL with an empty q param

But if someone wants to fix them both at the same time, that's great. We should have functional tests to cover both cases as well, so we don't regress in the future.

zoepage:
Just to clarify as this is a P1 bug we need to fix until the end of the week, which one of the 2 do we want to fix in this PR?

miketaylr:
@zoepage 1 is the most important, as it's the most likely for a user to run into.

Regaddi:
Issue 1 can be easily solved by adding the `required` attribute to the search input.
Issue 2 can be solved by applying the regex fix, I suggested [here](https://github.com/webcompat/webcompat.com/pull/2237#issuecomment-373311422).

miketaylr:
> Issue 1 can be easily solved by adding the required attribute to the search input.
> Issue 2 can be solved by applying the regex fix, I suggested here.

Those sound like good approaches! Want to take over this PR @Regaddi? I don't think @zoepage will mind. :)

Regaddi:
@miketaylr Have a look on the tests, please 😄

Regaddi:
This case is now not possible anymore because of the required field. We should either negate the test or remove it completely, I guess.

Regaddi:
Should be good now. @miketaylr  ?

PR #2238, opened 2018-03-15
Fixes #2088 - Add tiles section to index

### LINKED ISSUE #2088: Contribute -> Team
When someone clicks on Contribute in the top header.
capture d ecran 2018-02-16 a 08 04 20

it lands to a page where the main content is The team. It feels weird and put us before the tasks. I think we should direct people to ways of contributing, not us.

zoepage:
r? @magsout


zoepage:
<3

Issue #2230, opened 2018-03-14
Normalize naming of npm scripts


zoepage:
In our `package.json` we have different scripts,  e.g. for linting JS or testing. The writing is different (capital letters, small letters...) This is really confusing as I run into that issue all the time. 😅

My suggestion: we should go with small letters all the way.

Here are the lines that need to be changed:
https://github.com/webcompat/webcompat.com/blob/master/package.json#L59-L60
https://github.com/webcompat/webcompat.com/blob/master/package.json#L62
https://github.com/webcompat/webcompat.com/blob/master/package.json#L78

`JS` -> `js`

Thanks! :)

mpaarating:
I got this :D

zoepage:
merged #2243

PR #2243, opened 2018-03-16
Issue #2230 - Modify npm scripts to be uniform

mpaarating:
r? @zoepage 

Makes all the of the npm scripts uniform in name (removes capital letters from script titles).


zoepage:
Oh, you also thought about the `CSS` naming. Perfect!

PR #2246, opened 2018-03-16
Fixes #2245. Avoid inline styles set with JS.

### LINKED ISSUE #2245: label box floating broken on stage
As you can see here: https://staging.webcompat.com/issues/1390 the label box is not on the right side anymore. 

Seems like an issue introduced before that appeared after we merged.

@magsout @Regaddi Any idea?

Regaddi:
r? @zoepage


zoepage:
LGTM! 🎉

PR #2252, opened 2018-03-16
Fixes #2247 -  nested grid in issue details.

### LINKED ISSUE #2247: Issue detail page grid off
The issue detail page got a little bit off after solving #2245.

image

I'd suggest the following layout as a fix (got everything ready locally):

image

@zoepage ?

Regaddi:
Needed some adjustments because of nested grid.

r? @zoepage

### DIFF
--- a/webcompat/static/css/src/issue.css
+++ b/webcompat/static/css/src/issue.css
@@ -15,8 +15,9 @@
 }
 
 .issue-comment {
-  max-width: 100%;
+  max-width: calc(100% - 1 / 2 * var(--unit-space));
   position: relative;
+  width: 100%;
 }
 
 .issue-comment .comment-header {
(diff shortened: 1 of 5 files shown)


webcompat/static/css/src/issue.css
```diff
@@ -60,9 +70,33 @@
...
 
+.comment-body {
+  max-width: 100%;
+  overflow: auto;
+}
+
 /* As we can not add a class to the img right now (markup parsers are complicated), we need to use the element */
 .comment-body img,
 .js-Issue-markdown img {
   max-width: calc(var(--unit-modul-m) - (var(--unit-space) * 4));
```
zoepage: One last detail. The max-width if the image is 550px. The breaking point 650px. The image parent box `comment-body` gets an overflow-x, which maybe is fixable?

If this is an issue, we can also file a follow up. Just want to make sure, we have everything set as there will be work build upon that change.

PR #2259, opened 2018-03-19
Fixes #1847 - add animation for loader

### LINKED ISSUE #1847: No progress indicator for image uploads from a comment
(STR taken from https://github.com/webcompat/webcompat.com/issues/1000#issuecomment-339006338)

1. Go to https://staging.webcompat.com/issues/1294
2. Add a comment with an image around 3Mb (desktop screenshot for example)
3. Start your chrono ;)

Basically because it takes a long time and there is no visual cue to tell the user that an upload is in the process, you might give up or continue typing text or even restart an upload. 

Sometimes it has happen to me that I posted the text comment without remembering that I started an upload. And the markup for the image shows up in the next comment form. :)

zoepage:
R? @miketaylr

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -546,6 +546,8 @@ function BugForm() {
       success: callback,
       error: _.bind(function(response) {
         var msg;
+        this.uploadLoader.removeClass(".is-active");
+
         if (response && response.status === 415) {
           wcEvents.trigger("flash:error", {
             message: this.inputs.image.helpText,
@@ -558,6 +560,7 @@ function BugForm() {
             "The image is too big! Please choose something smaller than 4MB.";
           wcEvents.trigger("flash:error", { message: msg, timeout: 5000 });
         }
+        this.loadingIndicator.removeClass("is-active");
       }, this)
     });
 
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -527,7 +527,7 @@ function BugForm() {
     this.disableSubmits();
     this.uploadLoader.addClass("is-active");
 
-    $('.js-remove-upload').addClass('is-hidden');
+    $(".js-remove-upload").addClass("is-hidden");
 
     var formdata = new FormData();
     formdata.append("image", dataURI);
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -553,8 +553,6 @@ function BugForm() {
             "The image is too big! Please choose something smaller than 4MB.";
           wcEvents.trigger("flash:error", { message: msg, timeout: 5000 });
         }
-
-        this.loadingIndicator.hide;
       }, this)
     });
 
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -5,9 +5,9 @@
 function BugForm() {
   this.form = $("#js-ReportForm form");
   this.submitButtons = $("#js-ReportForm .js-Button");
-  this.loadingIndicator = $(".js-Loader");
+  this.loadingIndicator = $(".js-loader");
   this.reportButton = $("#js-ReportBug");
-  this.uploadLoader = $(".js-Upload-Loader");
+  this.uploadLoader = $(".js-image-loader");
   this.previewEl = $(".js-image-upload");
   this.UPLOAD_LIMIT = 1024 * 1024 * 4;
   this.clickedButton = null;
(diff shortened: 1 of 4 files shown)


zoepage:
FYI The linter changes didn't get staged again. Sorry. Will fix tonight. Feel free to review and I can merge later on.

webcompat/static/js/lib/bugform.js
```diff
@@ -550,7 +553,8 @@ function BugForm() {
             "The image is too big! Please choose something smaller than 4MB.";
           wcEvents.trigger("flash:error", { message: msg, timeout: 5000 });
         }
-        this.loadingIndicator.hide();
+
+        this.loadingIndicator.hide;
```
miketaylr: missing calling parens `.hide()`

miketaylr: Won't this make the screenshot upload thingy disappear now?

`<div class="form-upload js-image-upload js-Form-group loader js-loader js-image-loader">`

webcompat/static/js/lib/bugform.js
```diff
@@ -5,9 +5,9 @@
 function BugForm() {
   this.form = $("#js-ReportForm form");
   this.submitButtons = $("#js-ReportForm .js-Button");
-  this.loadingIndicator = $(".js-Loader");
+  this.loadingIndicator = $(".js-loader");
```
miketaylr: I'm not sure how useful this is for the non-image upload case. If there's no image upload, how will the user know to look there for opacity changes to indicate progress?

zoepage: > If there's no image upload, how will the user know to look there for opacity changes to indicate progress?

I'm not sure I understand this line? 🤔

miketaylr: Right now there are 2 loader elements: loading indicator to indicate the form is being submitted, and loading indicator to indicate an image is being uploaded. You have both of these selectors on the same element, which is the image upload preview area. 

Since the effect of the loader is to apply an animation that "pulses" on opacity, how does the user notice  that if they haven't uploaded an image? Maybe if you upload a screenshot of both loader indicators in action it will make more sense (with and without an image preview).

miketaylr: Another thing that doesn't quite make sense with this approach:

https://github.com/webcompat/webcompat.com/blob/7694b0a5c26b08f01a60680d18d6841c28944cca/webcompat/static/js/lib/bugform.js#L89

On form submit we're going to call `show()` on an already visible image upload preview area. Is that intentional? The previous design worked because there was a display:none animated image, which was shown then hidden.

zoepage: > Maybe if you upload a screenshot of both loader indicators in action it will make more sense (with and without an image preview).

Now I get it. Yes, agreed. Lets give it a try. 

Just some context regarding the spinner / message etc.: 
As we talked about the loader we really wanted to have something that is not blocking the UI what would happen with an overlay. This is why I'd rather try to find a different solution than that.

zoepage:
@miketaylr Wanna check the animation right now? (its quick and dirty just for you to check.)

webcompat/static/js/lib/bugform.js
```diff
@@ -550,7 +558,6 @@ function BugForm() {
             "The image is too big! Please choose something smaller than 4MB.";
           wcEvents.trigger("flash:error", { message: msg, timeout: 5000 });
         }
-        this.loadingIndicator.hide();
       }, this)
```
miketaylr: We still need to handle the error case, otherwise it will continue to pulse forever.

bug

miketaylr:
@zoepage I like the approach! We just need to handle stopping the animation if an error is returned from the server.

Regaddi:
This will also fix #2276

webcompat/static/js/lib/bugform.js
```diff
@@ -550,7 +558,6 @@ function BugForm() {
             "The image is too big! Please choose something smaller than 4MB.";
           wcEvents.trigger("flash:error", { message: msg, timeout: 5000 });
         }
-        this.loadingIndicator.hide();
```
miketaylr: @Regaddi points out that removing this line will fix #2276: https://github.com/webcompat/webcompat.com/issues/2276#issuecomment-375362881

(since loadingIndicator now refers to the form element, we don't want to make it disappear. :))

miketaylr: (but we probably still need to replace it with whatever will make the upload loader stop animating)

webcompat/static/js/lib/bugform.js
```diff
@@ -558,6 +560,7 @@ function BugForm() {
             "The image is too big! Please choose something smaller than 4MB.";
           wcEvents.trigger("flash:error", { message: msg, timeout: 5000 });
         }
+        this.loadingIndicator.removeClass("is-active");
```
miketaylr: I would stick this under `this.uploadLoader.removeClass(".is-active");`, but let's file a follow up "good first bug".

PR #2265, opened 2018-03-20
Fixes #2087 - Show and hide navbar in dependency of scrolling

### LINKED ISSUE #2087: Top sticky header is breaking the anchors
Go to https://staging.webcompat.com/contributors/web-platform-research#surveys
capture d ecran 2018-02-16 a 08 01 27

We should be able to read the header of this section. 

Personally I don't think the header should be sticky. It should just go away when we scroll.

Regaddi:
I had to introduce a scrollHandler for this.

r? @miketaylr @zoepage

### DIFF
--- a/webcompat/static/css/src/nav.css
+++ b/webcompat/static/css/src/nav.css
@@ -10,6 +10,12 @@
   box-sizing: border-box;
   background-color: var(--color-first);
   color: var(--color-second);
+  transition: transform 400ms ease, opacity 400ms ease;
+}
+
+.is-offscreen {
+  opacity: 0;
+  transform: translateX(-50%) translateY(-100%) scale(1.2);
 }
 
 .nav-container {
--- a/webcompat/static/js/lib/homepage.js
+++ b/webcompat/static/js/lib/homepage.js
@@ -72,13 +72,15 @@ function HomePage() {
     var isScrolling = false;
     $(window).on("scroll", function() {
       isScrolling = true;
-      clearTimeout(scrollTimeout);
+      window.clearTimeout(scrollTimeout);
       scrollTimeout = window.setTimeout(function() {
         isScrolling = false;
       }, 500);
     });
     window.setInterval(function() {
-      if (!isScrolling) return;
+      if (!isScrolling) {
+        return;
+      }
       if (window.pageYOffset < navbarHeight) {
         $navbar.removeClass("is-offscreen");
         lastScrollY = window.pageYOffset;
--- a/webcompat/static/js/lib/homepage.js
+++ b/webcompat/static/js/lib/homepage.js
@@ -32,6 +32,7 @@ function HomePage() {
   this.setUpEvents = function() {
     this.searchHandler();
     this.dropDownHandler();
+    this.navbarHandler();
   };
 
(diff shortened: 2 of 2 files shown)


zoepage:
LGTM!

webcompat/static/css/src/nav.css
```diff
@@ -10,6 +10,12 @@
   box-sizing: border-box;
   background-color: var(--color-first);
   color: var(--color-second);
+  transition: transform 400ms ease, opacity 400ms ease;
```
miketaylr: Fancy transition. :)

webcompat/static/js/lib/homepage.js
```diff
@@ -32,6 +32,7 @@ function HomePage() {
   this.setUpEvents = function() {
     this.searchHandler();
     this.dropDownHandler();
+    this.navbarHandler();
```
miketaylr: I guess it feels a little weird to put this navbar controller code in homepage.js, but maybe it's more awkward we include homepage.js on every page. 😄 

We should probably file a followup bug to change that, and have something like navbar.js with fancy code-splitting. I imagine we'll move to a more modern modular design this year.

webcompat/static/js/lib/homepage.js
```diff
@@ -63,6 +64,39 @@ function HomePage() {
...
+    var isScrolling = false;
+    $(window).on("scroll", function() {
+      isScrolling = true;
+      clearTimeout(scrollTimeout);
+      scrollTimeout = window.setTimeout(function() {
+        isScrolling = false;
+      }, 500);
+    });
+    window.setInterval(function() {
+      if (!isScrolling) return;
```
miketaylr: nit: can we add opening and closing braces and put the return on newline?

webcompat/static/js/lib/homepage.js
```diff
@@ -63,6 +64,39 @@ function HomePage() {
...
 
+  this.navbarHandler = function() {
+    var $navbar = $(".navigation");
+    var navbarHeight = $navbar.outerHeight();
+    var lastScrollY = window.pageYOffset;
+    var scrollTimeout = null;
+    var isScrolling = false;
+    $(window).on("scroll", function() {
+      isScrolling = true;
+      clearTimeout(scrollTimeout);
```
miketaylr: I notice you're referring to the `setTimeout` and `setInterval` globals on the `window` object, but not `clearTimeout`. I don't have a hard preference either way (I normally just use global identifier w/o `window`), but it's probably good to be consistent.

Regaddi: I'd probably be explicit and always to refer to `window`.

webcompat/static/js/lib/homepage.js
Regaddi: Sure.

webcompat/static/js/lib/homepage.js
Regaddi: Indeed. I needed some time to find the "correct" place to add this :)

miketaylr:
Filed https://github.com/webcompat/webcompat.com/issues/2268 as a follow up.

PR #2279, opened 2018-03-21
Fixes #2277 - Add missing border colors for status labels

### LINKED ISSUE #2277: milestone labels list has wrong colored borders
The `border-left`  should indicate the color of the status, but here it jumps back to default as it's been overwritten per CSS. We need to adjust the specificity.

Labels (should):
screen shot 2018-03-21 at 11 27 33 am

Milestones (default, but wrong):
screen shot 2018-03-21 at 11 27 24 am

Code can be found here:
https://github.com/webcompat/webcompat.com/blob/refactor/webcompat/static/css/src/labels.css#L69-L97

Regaddi:
There are more status labels than defined colors. So not sure, if this fits the needs.
Haven't spotted `border-color` definitions for status labels in the CSS, only `background-color` definitions, so I added them.

r? @zoepage

### DIFF
--- a/webcompat/static/css/src/labels.css
+++ b/webcompat/static/css/src/labels.css
@@ -140,7 +140,7 @@
 }
 
 .label-box .label-editor-list [class *= "label-"] {
-  border-left-color: var(--color-fourth);
+  border-left-color: var(--label-closed);
 }
 
 .label-box .label-editor-list [class *= "label-type"] {
@@ -188,7 +188,7 @@
   border-left-color: var(--label-needs-contact);
 }
 
-.label-box .label-editor-list .label-ready {
+.label-box .label-editor-list .label-contactready {
   border-left-color: var(--label-outreach);
 }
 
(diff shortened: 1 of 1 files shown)


webcompat/static/css/src/labels.css
```diff
@@ -175,3 +175,27 @@
...
+}
+
+.label-box .label-editor-list .label-needsdiagnosis {
+  border-left-color: var(--label-needs-diagnosis);
+}
+
+.label-box .label-editor-list .label-needscontact {
+  border-left-color: var(--label-needs-contact);
+}
+
```
zoepage: `label-contactready` -> var(--label-outreach)

webcompat/static/css/src/labels.css
```diff
@@ -175,3 +175,27 @@
 .label-box .label-editor-list [class *= "label-status"] {
   border-left-color: var(--label-closed);
```
zoepage: the default should be `var(--label-closed)` but it is `var(--label-fourth)` 🤔

webcompat/static/css/src/labels.css
Regaddi: Adjusted class.

webcompat/static/css/src/labels.css
Regaddi: I set `--label-closed` now as the default color.

PR #2284, opened 2018-03-22
Fixes #2274 - Automatic scrolling of label box when editing

### LINKED ISSUE #2274: Scrolling overflow chops off milestones editor
overflow-weird

STR: go to staging, log in, go to an issue and open the status editor.

On desktop at least everything is chopped off unless you scroll.

Regaddi:
This will automatically set the label-box scroll position, so the freshly opened list of labels or statuses is directly visible to the user.

r? @zoepage @miketaylr


zoepage:
That's quick && clever :) Like it.

miketaylr:
This is fine as a workaround, and so we can ship, but I think we should re-think how overflow works for these editor widget. Scrollbars inside scrollbars is a bit strange.

screen shot 2018-03-22 at 11 49 21 am

PR #2290, opened 2018-03-23
Fixes #2215 - Tweaks needinfo filtering and design

### LINKED ISSUE #2215: The "Older than 48h" bucket in NeedsTriage Dashboard should not take into consideration issues with "Needsinfo"
Mike sent us an e-mail, saying that "I thought if an issue had a “needs info” label, it shouldn’t appear in the “older than 48” bucket… if that’s not the case currently, that’s a bug."

https://webcompat.com/dashboard/triage

karlcow:
This alters slightly the design to take into account needinfo.
It also adjusts a couple of things.

The Pull Request is not finished because I managed to hide the needinfo issues, but I'm unable to make the full list reappears. It seems the initial code is working with a local storage version of the issue instead of the HTML.  I need your help @magsout 

This is basically only the last commit which needs to be fixed. The rest is more or less independent.
https://github.com/webcompat/webcompat.com/commit/18f342f23980c7c30263134baae0fc947473b9ae

r? @magsout

### DIFF
--- a/tests/unit/test_dashboard.py
+++ b/tests/unit/test_dashboard.py
@@ -53,15 +53,16 @@ class TestDashboard(unittest.TestCase):
 
     def test_filter_needstriage(self):
         """Assess the filtering is correct."""
-        expected = [{'needinfo': True, 'title': u'example.org - dashboard test', 'created_at': u'2015-02-01T12:00:00Z', 'updated_at': u'2017-11-10T12:00:00Z', 'number': 1, 'priority': True, 'browsers': ['firefox']}, {'needinfo': False, 'title': u'dashboard.example.com - site is not usable', 'created_at': u'2015-02-20T12:00:00Z', 'updated_at': u'2015-02-20T12:00:00Z', 'number': 2, 'priority': False, 'browsers': ['chrome', 'firefox']}]  # noqa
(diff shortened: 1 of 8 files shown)


karlcow:
capture d ecran 2018-03-23 a 11 26 49

magsout:
@karlcow what I propose to handle needissue base on your very good work https://github.com/webcompat/webcompat.com/commit/4c4e6292c6055b81067cbe914931e288a06d8697#diff-cb1f45ce95dccbecbee0de83a4013e49

karlcow:
@softvision-sergiulogigan This is what it currently looks like. Would you be happy with this?

capture d ecran 2018-03-26 a 07 38 18

magsout:
@karlcow I definitely didn't understand the requirement. Sorry :(

softvision-sergiulogigan:
> This is what it currently looks like. Would you be happy with this?

It looks awesome!
I suppose that, by default, the "Older than 48h" does not include the "NeedInfo" numbers.

karlcow:
>  I suppose that, by default, the "Older than 48h" does not include the "NeedInfo" numbers.

Exactly !

Ok. everyone is happy. Let's ride with it.

Thanks a bunch everyone for comments, push backs, explanations, and code.

PR #2305, opened 2018-03-28
Fixes #2301 - Upload input size equals hoverable area

### LINKED ISSUE #2301: Remove screenshot hover activation possibly misleading
STR:

1. go to staging.webcompat.com/issues/new
2. upload an image
3. hover over the edge of the image and note "Remove screenshot" turning yellow
4. click

Expected: it gets removed
Actual: it only gets removed if you click the `.js-remove-upload` div.

My brain assumed that since the color changed, it was an active clickable area. Should we make the entire preview be the click target?

Regaddi:
r? @zoepage

### DIFF
--- a/webcompat/static/css/src/form.css
+++ b/webcompat/static/css/src/form.css
@@ -199,16 +199,14 @@
 }
 
 .input-upload {
+  display: block;
   position: relative;
   opacity: 0;
   z-index: 30;
   color: var(--color-fourth);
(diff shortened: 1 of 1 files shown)


zoepage:
Not sure, if we misunderstood each other here... :)

The area is too small right now and should cover the input field. (or am I missing something? :) )
screen shot 2018-03-28 at 10 28 12 am

PR #2309, opened 2018-03-28
Fixes #2306 - remove unused CSS

### LINKED ISSUE #2306: remove unsused CSS
- [x] add pkg, grunt task & npm script https://www.npmjs.com/package/grunt-purifycss
- [x] remove unused CSS 
- [x] check for `webcompat-reset.css`
- [x] check `color.css`
- [x] check `styleguide.css`

zoepage:
r? @miketaylr


Gruntfile.js
```diff
@@ -29,7 +29,8 @@ module.exports = function(grunt) {
     "concat",
     "uglify",
     "postcss",
-    "cssmin"
+    "cssmin",
+    "purifycss"
   ]);
```
miketaylr: Can we also add it to the deploy task please?

grunt-tasks/purifycss.js
```diff
@@ -0,0 +1,22 @@
...
+module.exports = function(grunt) {
+  grunt.config("purifycss", {
+    options: {
+      rejected: true,
+      info: true
+    },
+    target: {
+      src: [
+        "webcompat/templates/*.html",
+        "webcompat/templates/**/*.html",
```
miketaylr: `**` matches zero or more directories (when its the only thing in the path), so the first line is redundant. Just `"webcompat/templates/**/*.html"` should be sufficient.

grunt-tasks/purifycss.js
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function(grunt) {
+  grunt.config("purifycss", {
+    options: {
+      rejected: true,
```
miketaylr: So the idea is we create a stylesheet that *only* contains the unused styles? And then presumably we send in PRs from time to time removing them?

Could we add something somewhere in docs to explain this?

miketaylr: (if we're crunched on time, filing a follow up bug to document is this is also cool)

Gruntfile.js
zoepage: Right now we are generating a CSS which is stripped, which we also can use as our `*.min.css` version, but I am not fully trusting that one yet. 

The purify task is showing the not used classes right now.

grunt-tasks/purifycss.js
zoepage: No, we create one without the removed styles. The `rejected` option is showing in the logs the unsused styles that have been removed.

miketaylr: Ah, thanks!

PR #2333, opened 2018-03-28
Fixes #2329 - remove inline style

### LINKED ISSUE #2329: Remove inline styles from screenshots
See the ones added in https://github.com/webcompat/webcompat.com/pull/2328#pullrequestreview-107746708.

It would be good to make a class in a stylesheet somewhere, rather than have inline styles.

zoepage:
r? @miketaylr


miketaylr:
🎉

Issue #2334, opened 2018-03-28
Unable to use "Remove screenshot" on an image preview that comes from in-browser Report Site Issue button


miketaylr:
STR:

1. in Firefox Nightly, go to about:config and change `extensions.webcompat-reporter.newIssueEndpoint` to `https://staging.webcompat.com/issues/new`
2. open a webpage
3. click "Report site issue" from the page actions menu
4. in the image preview in the new tab, click "Remove screenshot"

Expected: it gets removed
Actual: it doesn't. it also doesn't apply the yellow hover styles.

miketaylr:
Seems like this will be a pretty simple fix, it's just a z-index issue because `is-validated` isn't added to the parent div when the image is added via the browser. I'll write a patch and add a test.

PR #2335, opened 2018-03-28
Issue #2334. Ensure that we consider browser-sent images as valid.

miketaylr:
r? @zoepage 

`npm run test:js -- --grep="remove image upload button"`

screen shot 2018-03-28 at 3 45 15 pm


miketaylr:
The bug is that without the `.is-validated` class, it never gets a higher z-index which is controlled by this selector, so you can't actually click it:

```
.is-validated .label-remove {
    visibility: visible;
    z-index: 40;
    cursor: pointer;
    border: 0 none;
    outline: 0;
}
```

zoepage:
LGTM! :)))

PR #2353, opened 2018-03-30
Fixes #2300. Try to respect DNT preference before talking to GA.

### LINKED ISSUE #2300: Respect DNT before sending anything to Google Analytics
See https://hacks.mozilla.org/2016/01/google-analytics-privacy-and-event-tracking/.

miketaylr:
r? @karlcow 

I think... we should probably update the privacy policy to mention that we do not send usage to GA if DNT header is set to 1. Will push a patch <s>on Monday</s> in 4 minutes.


webcompat/static/js/lib/ga.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+  // ie 9 & 10
+  if ("msDoNotTrack" in navigator) {
+    return navigator.msDoNotTrack;
+  }
+
+  return "0";
+}
+
+if (getDNTStatus() !== "1") {
```
karlcow: Why testing different from 1 instead of testing equal to 0?

```js
   if (getDNTStatus() === "0") {…}
```

miketaylr: Dunno, just the way my brain works, negative bias. :P

If they didn't say we can't track then, track them.

miketaylr: (also, it's possible for the API to return "unspecified"... seems easier to just check for one thing)

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack

PR #2357, opened 2018-04-03
fixes #2356 - fixed result when older < needsinfo

### LINKED ISSUE #2356: Older than 48h can be negative
**Environment:**
Browser / Version: Chrome 65.0.3325.181 (Desktop), Firefox Nightly 61.0a1 (Desktop)
Operating System: Windows 10

**Steps to Reproduce:**
1. Go to https://webcompat.com/dashboard/triage
2. Observe "Older than 48h" info

**Expected Behavior:**
The value is >= 0.

**Actual Behavior:**
The value can be < 0.

**Notes:**
1. Screen capture attached.
2. Reproducible both on Chrome and Firefox.

Watchers:
@softvision-sergiulogigan
@softvision-oana-arbuzov

sv;

image

magsout:
A ternary condition would be more beautiful, But I don't know how to do that in python and don't find anything 

r @karlcow ?

### DIFF
--- a/webcompat/templates/dashboard/triage/stats.html
+++ b/webcompat/templates/dashboard/triage/stats.html
@@ -74,7 +74,8 @@
       <div class="wc-Stat-container">
         <h2 class="wc-Stat-title">Older than 48h</h2>
         <div class="wc-Stat-total">
-          {{ stats.older - stats.needsinfo }}
+          {% set total = stats.older - stats.needsinfo %}
+          {{ 0 if total <= 0 else total }}
         </div>
       </div>
     </div>


webcompat/templates/dashboard/triage/stats.html
```diff
@@ -74,7 +74,12 @@ <h2 class="wc-Stat-title">Needstriage issues</h2>
       <div class="wc-Stat-container">
         <h2 class="wc-Stat-title">Older than 48h</h2>
         <div class="wc-Stat-total">
-          {{ stats.older - stats.needsinfo }}
+          {% set olderThan48h = stats.older - stats.needsinfo %}
+          {% if olderThan48h <= 0 %}
+            0
+          {% else %}
+            {{ olderThan48h }}
+          {% endif %}
```
karlcow: @magsout you were asking to do [ternary expression in Python](https://docs.python.org/2.7/reference/expressions.html#conditional-expressions).

> Conditional expressions (sometimes called a “ternary operator”) have the lowest priority of all Python operations.
> The expression x if C else y first evaluates the condition, C (not x); if C is true, x is evaluated and its value is returned; otherwise, y is evaluated and its value is returned.


```python
total = stats.older - stats.needsinfo
0 if total <= 0 else total
```


Maybe for consistency we should keep it in python.

magsout: @karlcow the backend handle the total and sent the var to the frond end ?

karlcow: @magsout It depends on what you call back-end and front-end. 

There is a python function which creates a data structure, namely `stats`. When the view is called by an HTTP request, the data structure is passed to the template. The page is generated and finally sent to the browser (which is the front-end).

So basically. You can do 

```python
        <div class="wc-Stat-total">
          {% set total = stats.older - stats.needsinfo %}
          {{ 0 if total <= 0 else total }}
        </div>
```

Does it answer your question?

magsout: yes it does, thanks, let me update my PR so

magsout:
done

last r ? @karlcow

PR #2365, opened 2018-04-05
Fixes #2225 - Add stylelint-order. Fix ordering in CSS src files

### LINKED ISSUE #2225: Add stylelint-order to .stylelintrc
This is possibly totally opinionated, but in the past I made great experiences using [stylelint-order](https://www.npmjs.com/package/stylelint-order) in addition to the standard stylelint config.
I usually prefer the alphabetical order for CSS properties, so instead of wildly mixed property orders like

```css
.some-selector {
  font-size: 14px;
  color: black;
  width: 100px;
  height: 50px;
}
```

You'd get:

```css
.some-selector {
  color: black;
  font-size: 14px;
  height: 50px;
  width: 100px;
}
```

I personally think, that the alphabetical order is the most natural order we as a humans are used to.

Just wanted to drop this idea here 😄

What do you think?

@zoepage

Regaddi:
The ordering was fixed with `npm run lint:fix:css`
r? @miketaylr @magsout

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,5 +1,8 @@
 {
     "extends": "stylelint-config-standard",
+    "plugins": [
+      "stylelint-order"
+    ],
     "rules": {
         "indentation": 2,
         "string-quotes": "double",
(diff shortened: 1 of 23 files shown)


magsout:
👍 @Regaddi

PR #2387, opened 2018-04-11
Fixes #2337: Refactor contributor routes for simpler nav/sub-nav templates.

### LINKED ISSUE #2337: Simplify footer.html template conditionals
In the following 3 spots in footer.html, it should be possible to make this a little more DRY by doing something like `if request.url_rule.endpoint.includes('foo_')`

https://github.com/webcompat/webcompat.com/blob/93be91b6acddbe42f447e4017e31231d30a61772/webcompat/templates/shared/footer.html#L19

https://github.com/webcompat/webcompat.com/blob/93be91b6acddbe42f447e4017e31231d30a61772/webcompat/templates/shared/footer.html#L43

https://github.com/webcompat/webcompat.com/blob/93be91b6acddbe42f447e4017e31231d30a61772/webcompat/templates/shared/footer.html#L61

We probably need to come up with meaningful groupings and rename the route method names so this trick would work:

https://github.com/webcompat/webcompat.com/blob/93be91b6acddbe42f447e4017e31231d30a61772/webcompat/views.py#L342

miketaylr:
r? @karlcow

### DIFF
--- a/webcompat/templates/contributors/sub-nav.html
+++ b/webcompat/templates/contributors/sub-nav.html
@@ -5,30 +5,30 @@
     </a>
   </li>
 
-  <li class="sub-nav-item {% if request.url_rule.endpoint in ['contributors_report_bug', 'contributors_reproduce_bug', 'contributors_diagnose_bug', 'contributors_site_outreach'] %} active {% endif %}">
+  <li class="sub-nav-item {% if 'contributors_bug_' in request.url_rule.endpoint %} active {% endif %}">
     <a class="link-text" href="/contributors/report-bug">
       Working with bugs <span class="nested-sub"></span>
     </a>
     <ol class="nested-sub-nav">
-      <li class="nested-sub-nav-item {% if request.url_rule.endpoint == 'contributors_report_bug' %} sub-nav-active{% endif %}">
+      <li class="nested-sub-nav-item {% if request.url_rule.endpoint == 'contributors_bug_report' %} sub-nav-active{% endif %}">
         <a class="link-text" href="/contributors/report-bug">
           Report a Bug
         </a>
       </li>
 
-      <li class="nested-sub-nav-item {% if request.url_rule.endpoint == 'contributors_reproduce_bug' %} sub-nav-active{% endif %}">
+      <li class="nested-sub-nav-item {% if request.url_rule.endpoint == 'contributors_bug_reproduce' %} sub-nav-active{% endif %}">
         <a class="link-text" href="/contributors/reproduce-bug">
           Reproduce a Bug
         </a>
       </li>
 
-      <li class="nested-sub-nav-item {% if request.url_rule.endpoint == 'contributors_diagnose_bug' %} sub-nav-active{% endif %}">
+      <li class="nested-sub-nav-item {% if request.url_rule.endpoint == 'contributors_bug_diagnosis' %} sub-nav-active{% endif %}">
         <a class="link-text" href="/contributors/diagnose-bug">
           Diagnose a Bug
         </a>
       </li>
 
-      <li class="nested-sub-nav-item {% if request.url_rule.endpoint == 'contributors_site_outreach' %} sub-nav-active{% endif %}">
+      <li class="nested-sub-nav-item {% if request.url_rule.endpoint == 'contributors_bug_outreach' %} sub-nav-active{% endif %}">
         <a class="link-text" href="/contributors/site-outreach">
           Site outreach
         </a>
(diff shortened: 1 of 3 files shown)


webcompat/views.py
```diff
@@ -350,7 +350,7 @@ def contributors():
 
 @app.route('/contributors/report-bug')
 @cache_policy(private=True, uri_max_age=0, must_revalidate=True)
-def contributors_report_bug():
+def contributors_bug_report():
```
karlcow: @miketaylr Probably these changes in `views.py` are not necessary. See my comment on the templates.

webcompat/templates/contributors/sub-nav.html
```diff
@@ -5,59 +5,59 @@
     </a>
   </li>
 
-  <li class="sub-nav-item {% if request.url_rule.endpoint in ['contributors_report_bug', 'contributors_reproduce_bug', 'contributors_diagnose_bug', 'contributors_site_outreach'] %} active {% endif %}">
+  <li class="sub-nav-item {% if 'contributors_bug_' in request.url_rule.endpoint %} active {% endif %}">
```
karlcow: @miketaylr first of all. Thanks. This is working.

But I was feeling a bit itchy to make this dependent on the function names instead of the routes.

`request.url_rule.rule` returns the route instead of the function name returned by `request.url_rule.endpoint`

```python
<li class="sub-nav-item {% if  request.url_rule.rule.startswith('/contributors/') %} active {% endif %}">
    <a class="link-text" href="/contributors/report-bug">
      Working with bugs <span class="nested-sub"></span>
    </a>
  ```

File: webcompat/templates/contributors/sub-nav.html [hunk 3]
```diff
@@ -5,59 +5,59 @@
...
     <a class="link-text" href="/contributors/report-bug">
       Working with bugs <span class="nested-sub"></span>
     </a>
     <ol class="nested-sub-nav">
-      <li class="nested-sub-nav-item {% if request.url_rule.endpoint == 'contributors_report_bug' %} sub-nav-active{% endif %}">
+      <li class="nested-sub-nav-item {% if request.url_rule.endpoint == 'contributors_bug_report' %} sub-nav-active{% endif %}">
```
karlcow: We could apply the same logic here and so on.

```python
    <ol class="nested-sub-nav">
      <li class="nested-sub-nav-item {% if request.url_rule.rule == '/contributors/report-bug' %} sub-nav-active{% endif %}">
        <a class="link-text" href="/contributors/report-bug">
          Report a Bug
        </a>
      </li>
```

webcompat/templates/contributors/sub-nav.html
```diff
@@ -5,59 +5,59 @@
     </a>
```
karlcow: And this one too (even if it was not part of your commit.)

```
<ol class="sub-nav" class="sub-nav">
  <li class="sub-nav-item {% if request.url_rule.rule == '/contributors' %} sub-nav-active{% endif %}">
    <a class="link-text" href="/contributors">
      Join Us
    </a>
```

webcompat/templates/contributors/sub-nav.html
miketaylr: This makes sense.

webcompat/templates/contributors/sub-nav.html
miketaylr: @karlcow the one area where this becomes less desirable, IMO, is that we can't nicely create groups without changing route names (which we don't want to do, because we have to support redirects forever).

Right now there's 2 "groups", beyond the root `/contributors` route:

"Bug" group

| (proposed) method_name  | route |
| ------------- | ------------- |
| contributors_bug_report  | /contributors/report-bug |
| contributors_bug_reproduce | /contributors/reproduce-bug |
| contributors_bug_diagnosis | /contributors/diagnose-bug | 
| contributors_bug_outreach | /contributors/site-outreach |

"Other" group

| (proposed) method_name  | route |
| ------------- | ------------- |
| contributors_other_tools  | /contributors/build-tools |
| contributors_other_research | /contributors/web-platform-research  |
| contributors_other_events |/contributors/organize-webcompat-events| 

So we can simplify a *little*, but not as much as we could if we rename the methods. Any reason to not change the method names? There are no external dependencies (like changing URLs would create).

webcompat/templates/shared/footer.html
```diff
@@ -16,8 +16,7 @@
             About
           </a>
         </li>
-        <li class="footer-item{% if request.url_rule.endpoint in
-        ['contributors_build_tools', 'contributors_web_platform_research', 'contributors_organize_webcompat_events', 'contributors', 'contributors_report_bug', 'contributors_reproduce_bug', 'contributors_diagnose_bug', 'contributors_site_outreach'] %} active{% endif %}">
+        <li class="footer-item{% if 'contributors' in request.url_rule.endpoint %} active{% endif %}">
```
miketaylr: We can just check `/contributors` here in the rule

webcompat/templates/shared/footer.html
```diff
@@ -40,33 +39,33 @@
       </ul>
 
 <!-- sub nav for contribute > working with bugs -->
-      {% if request.url_rule.endpoint in ['contributors_report_bug', 'contributors_reproduce_bug', 'contributors_diagnose_bug', 'contributors_site_outreach'] %}
+      {% if 'contributors_bug_' in request.url_rule.endpoint %}
```
miketaylr: For this, using `rule`, I would have to look for `-bug` and `site-outreach`, rather than just `contributors_bug_`

webcompat/templates/shared/footer.html
```diff
@@ -40,33 +39,33 @@
...
-          <li class="footer-sub-item {% if request.url_rule.endpoint == 'contributors_site_outreach' %}active{% endif %}">
+          <li class="footer-sub-item {% if request.url_rule.endpoint == 'contributors_bug_outreach' %}active{% endif %}">
             <a class="footer-item-link" href="/contributors/site-outreach">Site Outreach</a>
           </li>
         </ul>
       {% endif %}
 
 <!-- sub nav for contribute > other contributions -->
-      {% if request.url_rule.endpoint in ['contributors_build_tools', 'contributors_web_platform_research', 'contributors_organize_webcompat_events'] %}
+      {% if 'contributors_other_' in request.url_rule.endpoint %}
```
miketaylr: For this I have to look for all 3 "other" routes, because there's nothing common there to match on:

| (proposed) method_name  | route |
| ------------- | ------------- |
| contributors_other_tools  | /contributors/build-tools |
| contributors_other_research | /contributors/web-platform-research  |
| contributors_other_events |/contributors/organize-webcompat-events|

miketaylr:
Thanks @karlcow for your review, I didn't know about `request.url_rule.rule`. 🎈

karlcow:
As a kind of followup on this #2391

PR #2392, opened 2018-04-13
Fixes #2388 - Simplify border definition on active sub nav item

### LINKED ISSUE #2388: Weird border on iOS 11 Safari on contributions page
image

The border around "Join us" should be only on the left.

Regaddi:
r? @zoepage


zoepage:
LGTM!
looks good

Issue #1834, opened 2017-10-03
Move all "built" scripts into a build folder


miketaylr:
Basically everything that's not in `js/lib` or `js/vendor` should go into a `js/build` directory, and be git ignore'd. We should never edit those files, and it would be easier to hide them from editor search results as well.

zoepage:
This was already discussed in #1352, but I second that! Shall we do this now then (with grunt)?

miketaylr:
Yeah, why not. If and when webpack happens, it would mean less work to modify templates, etc.

zoepage:
@Regaddi is this something you'd be interested in? :)

Regaddi:
`js/build` is already in use and contains `ga.js`.
How about `js/dist` so we're consistent with `css/dist`?

miketaylr:
> Isn’t this done and can be closed? 🤔

Looks like it!

PR #2385, opened 2018-04-10
Fixes #2375. Add some diagnostic logging for checkServer

miketaylr:
(note: we should leave #2375 open, this is just an attempt to understand why this might be failing)

r? @brizental

PR #2394, opened 2018-04-16
Issue #1834 - Move all generated JS files to js/dist

Regaddi:
I think I caught all references, but don't know how to test locally because of `{%- if config.PRODUCTION or config.STAGING -%}`.
Maybe someone has delightful input for me how to test PRODUCTION or STAGING locally 😄 

r? @zoepage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,15 +20,12 @@ data/*
 
 # these are our 'dist' files
 # checking them into the repo is silly
-webcompat/static/js/build/
-webcompat/static/js/webcompat.js
-webcompat/static/js/diagnose.js
-webcompat/static/js/issues.js
-webcompat/static/js/issue-list.js
-webcompat/static/js/user-activity.js
-webcompat/static/js/templates.js
 webcompat/**/*.min.js
-webcompat/**/**/*.min.css
+webcompat/static/js/*.js
+webcompat/static/js/build/
+webcompat/static/js/dist/
+webcompat/**/*.min.css
+webcompat/static/css/dist/
 
 #selenium server
 *.jar
--- a/grunt-tasks/concat.js
+++ b/grunt-tasks/concat.js
@@ -25,7 +25,7 @@ module.exports = function(grunt) {
         "<%= jsPath %>/lib/bugform.js",
         "<%= jsPath %>/templates.js"
       ],
-      dest: "<%= jsPath %>/<%= pkg.name %>.js"
+      dest: "<%= jsPath %>/dist/<%= pkg.name %>.js"
     },
     diagnose: {
       src: [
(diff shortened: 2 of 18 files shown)


zoepage:
R? @miketaylr (to be sure we caught everything)

miketaylr:
> Travis tests have failed

(bummer, if https://github.com/webcompat/webcompat.com/pull/2385 was merged, we would have more info on this failure)

miketaylr:
> Maybe someone has delightful input for me how to test PRODUCTION or STAGING locally 😄

@Regaddi you can just set a local environment variable: https://github.com/webcompat/webcompat.com/blob/master/config/environment.py#L19

grunt-tasks/concat.js
```diff
@@ -25,15 +25,15 @@ module.exports = function(grunt) {
         "<%= jsPath %>/lib/bugform.js",
         "<%= jsPath %>/templates.js"
```
miketaylr: Do we need to update this as well?

https://github.com/webcompat/webcompat.com/pull/2394/files#diff-56763bbaaf4762aa0cf49808e18d196eR26

.gitignore
```diff
@@ -20,15 +20,12 @@ data/*
...
-webcompat/static/js/build/
-webcompat/static/js/webcompat.js
-webcompat/static/js/diagnose.js
-webcompat/static/js/issues.js
-webcompat/static/js/issue-list.js
-webcompat/static/js/user-activity.js
-webcompat/static/js/templates.js
 webcompat/**/*.min.js
-webcompat/**/**/*.min.css
+webcompat/static/js/*.js
```
miketaylr: Do we need this line? In theory there shouldn't be anything there, right?

(or is it just to ignore the already existing files so people don't have to manually delete them?)

grunt-tasks/concat.js
```diff
@@ -25,15 +25,15 @@ module.exports = function(grunt) {
...
       ],
-      dest: "<%= jsPath %>/<%= pkg.name %>.js"
+      dest: "<%= jsPath %>/dist/<%= pkg.name %>.js"
```
miketaylr: Maybe we could make a new Gruntfile variable`jsDistPath` in Gruntfile.js?

zoepage:
@Regaddi I've merged #2385. Could you rebase please? Thank you! :)

.gitignore
Regaddi: Yes, it's to make sure, that there are no accidental files or leftovers from older project states.

grunt-tasks/concat.js
Regaddi: I can do that, sure.

grunt-tasks/concat.js
Regaddi: Since `templates.js` is also a generated file, I'd put it into the `dist` folder as well.

Regaddi: Ah I didn't see it right. Needs to be updated, sure. This made the tests fail I guess.

miketaylr:
Tests are green, LGTM!

PR #2397, opened 2018-04-16
Fixes #2396 - Move ga.js to dist folder

### LINKED ISSUE #2396: Move ga.js to /dist/ga.js
That way we don't need `/build`.

#2394

Regaddi:
r? @miketaylr


zoepage:
As @Regaddi asked on IRC, +1 from my side. :)

PR #2398, opened 2018-04-17
Fixes #1794 - Prevent submit by enter in bugform

### LINKED ISSUE #1794: Enter in description textbox submits form
As described in this bug: https://github.com/webcompat/web-bugs/issues/9706

If you press enter in the description textbox, it submits the form before you've (potentially) finished filling it out.

Maybe we should prevent this behaviour and force the user to click the button?

Regaddi:
r? @miketaylr @zoepage

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -94,6 +94,17 @@ function BugForm() {
     );
     this.form.on("submit", _.bind(this.maybeUploadImage, this));
 
+    // prevent submit by hitting enter key for single line input fields
+    this.urlField.on("keypress", _.bind(this.preventSubmitByEnter, this));
+    this.problemType.on("keypress", _.bind(this.preventSubmitByEnter, this));
+    this.descField.on("keypress", _.bind(this.preventSubmitByEnter, this));
+    this.browserField.on("keypress", _.bind(this.preventSubmitByEnter, this));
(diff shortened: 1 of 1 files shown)


zoepage:
r+

webcompat/static/js/lib/bugform.js
```diff
@@ -94,6 +94,17 @@ function BugForm() {
     );
     this.form.on("submit", _.bind(this.maybeUploadImage, this));
 
+    // prevent submit by hitting enter key for single line input fields
+    this.urlField.on("keypress", _.bind(this.preventSubmitByEnter, this));
```
miketaylr: Maybe we could make this a little bit more DRY via event delegation, i.e., adding the listener on the form element itself, and checking that its target is an `input` (and not `textarea`) inside `preventSubmitByEnter`?

Regaddi:
@miketaylr I reduced to only one event listener.

webcompat/static/js/lib/bugform.js
```diff
@@ -94,6 +94,13 @@ function BugForm() {
     );
     this.form.on("submit", _.bind(this.maybeUploadImage, this));
 
+    // prevent submit by hitting enter key for single line input fields
+    this.form.on(
+      "keypress",
+      ":input:not(textarea)",
```
miketaylr: clever, i like it!

PR #2412, opened 2018-04-20
Fixes #2411 - Adds Failsafe check for differences in between config and milestones

### LINKED ISSUE #2411: App fails to start because it doesn't know about a milestone
I just deployed v9.0.5, and noticed webcompat.com production was returning 500.

Checking uwsgi logs:

```
  File "./webcompat/__init__.py", line 20, in <module>
    app.config.from_object('config')
  File "/usr/local/lib/python2.7/dist-packages/flask/config.py", line 168, in from_object
    obj = import_string(obj)
  File "/usr/local/lib/python2.7/dist-packages/werkzeug/utils.py", line 406, in import_string
    return __import__(import_name)
  File "./config/__init__.py", line 151, in <module>
    if not initialize_status():
  File "./config/__init__.py", line 62, in initialize_status
    app.config['STATUSES'] = convert_milestones(milestones_content)
  File "./config/__init__.py", line 85, in convert_milestones
    STATUSES[milestone['title']]['id'] = milestone['number']
KeyError: u'work'
unable to load app 0 (mountpoint='') (callable not found or import error)
```

There was a `work` milestone that someone created 10 hours ago (probably on accident, it's pretty easy to do this in github's interface). And I guess we failed hard. 💣

karlcow:
r? @miketaylr 
This should avoid the error 500. 
It is checking in advance that the milestones are really what we expect before trying to update them.

(That said we really need to review the way we do things)


karlcow:
Hopefully this should avoid most of the issue of missing/additional milestones.

config/__init__.py
```diff
@@ -59,7 +59,10 @@ def initialize_status():
     finally:
         # save in data/ the current version
         if milestones_content:
-            app.config['STATUSES'] = convert_milestones(milestones_content)
+            updated_statuses = update_status_config(milestones_content)
+            if not updated_statuses:
+                return False
+            app.config['STATUSES'] = updated_statuses
```
miketaylr: But if `updated_statuses` returns early, we still end up with `sys.exit('Milestones are not initialized.')`, right?

config/__init__.py
```diff
@@ -78,10 +81,18 @@ def milestones_from_file(milestones_path):
...
+def update_status_config(milestones_content):
     """Convert the JSON milestones from GitHub to a simple dict."""
-    milestone_full = json.loads(milestones_content)
-    for milestone in milestone_full:
+    milestones = json.loads(milestones_content)
+    # Test that GitHub milestones and local definitions are equivalent
+    status_names = sorted(STATUSES.keys())
+    milestone_names = sorted([milestone['title'] for milestone in milestones])
+    if milestone_names != status_names:
+        print('A milestone is missing or has been added.\n{names}'.format(
```
miketaylr: Can we add this to the logfile so it can be observed there? That's usually the first place I look when something explodes.

tests/unit/test_config.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python
```
miketaylr: 💯 for tests.

config/__init__.py
karlcow: yup. We can. Before doing so.

Just note if there is an issue, it will be visible right away. The project will not start, without the right set of milestones. 

Now I'm having a question. Is the config file each time a process is spawned (once the project is running by itself through uwgsi)?

Maybe we could also log through the webhook, the milestones creation and destruction and either log it or send a TAKE ACTION email about it. 

My idea for the future of milestones, is that they should be initialized (file generation) as a part of the deployment process more than the config file. Because once they are created, we don't need to touch them again. It seems stronger as a design for me than having them in the config files. It would also means that we should have a deploy script for local devs and/or move part of it back to `run.py`. 

* deploy or `run.py` should generate a file if it doesn't exist.
* config/environment.py should just read the file to load the milestones in memory. 

To think about.

miketaylr: > Now I'm having a question. Is the config file each time a process is spawned (once the project is running by itself through uwgsi)?

Can you clarify? I'm not sure I understand the question.

karlcow: @miketaylr 

yup sure. and make complete sentences. Sorry about that. "Is the config file read".
My question is triggered by [Emperor](https://uwsgi-docs.readthedocs.io/en/latest/Emperor.html). This starts multiple instances of the app. I was wondering if it was reading the config each time. Though maybe it doesn't matter once the milestone file is created on the fs. 

The question comes to assess the reliability of what we have now (with this new pull request). Currently with this pull request, even if someone is erasing or adding a milestone the project will not fail (when already running). In the UI, it will just show if the issue is just milestone-less we took care of that in Taipei. The 500 issue here was because the config was trying to create the initial dictionary with a non-matching list of keys.

The code now just prevent the project to start if there is a difference of dicts and so you will know right away. Would you like to proceed another way?

miketaylr: Thanks for the explanation. I'm not sure how Emperor handles the config, I would assume it reads the config when the app is restarted.. but I don't know for sure.



> The code now just prevent the project to start if there is a difference of dicts and so you will know right away. Would you like to proceed another way?

This seems fine for now, I think it would be nice to have a possible future where the app just starts.... but maybe that makes no sense since we're tied to what exists on GitHub. Makes me a little nervous about any of us having the ability to delete a milestone (on accident) and creating a huge headache.

PR #2439, opened 2018-05-04
Fixes #2409. Don't pre-fill forms for "self reports" (webcompat.com)

### LINKED ISSUE #2409: Do not fill in URL information for Site Report Issue triggered on webcompat.com
Adam and Mike mentioned submissions with webcompat.com as reported URL.

I assume the users' reasons might either be trying out the feature after reading about it or report an issue for another site after.

A solution would be to not fill in site-related fields (URL, screenshot) to avoid any accidental submission.

miketaylr:
r? @Regaddi

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -18,6 +18,7 @@ function BugForm() {
   this.submitButtons = $("#js-ReportForm .js-Button");
   this.submitTypeInput = $("#submit_type:hidden");
   this.uploadLabel = $(".js-label-upload");
+  this.urlParamRegExp = /url=([^&]*)/;
 
   this.UPLOAD_LIMIT = 1024 * 1024 * 4;
 
@@ -88,7 +89,12 @@ function BugForm() {
     this.onReceiveMessage = this.onReceiveMessage.bind(this);
     this.preventSubmitByEnter = this.preventSubmitByEnter.bind(this);
 
-    this.checkParams();
+    // Make sure we're not getting a report
+    // about our own site before checking params.
+    if (!this.isSelfReport()) {
+      this.checkParams();
+    }
+
     this.disableSubmits();
     this.initAutogrowFields();
     this.urlField.on("blur input", this.checkURLValidity);
@@ -160,7 +166,12 @@ function BugForm() {
   };
 
   this.onReceiveMessage = function(event) {
-    // Make sure the data is coming from ~*inside the house*~!
+    // We're getting a report about our own site, so let's bail.
+    if (this.isSelfReport()) {
+      return false;
+    }
+
+    // Make sure the data is coming from a trusted source.
     // (i.e., our add-on or some other priviledged code sent it)
     if (location.origin === event.origin) {
       // See https://github.com/webcompat/webcompat.com/issues/1252 to track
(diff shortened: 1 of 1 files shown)


webcompat/static/js/lib/bugform.js
```diff
@@ -182,18 +193,28 @@ function BugForm() {
...
     if (!location.search) {
       return;
     }
 
-    var urlParam = location.href.match(/url=([^&]*)/);
-    if (urlParam !== null) {
-      // weird Gecko bug. See https://bugzilla.mozilla.org/show_bug.cgi?id=1098037
-      urlParam = this.trimWyciwyg(urlParam[1]);
-      this.urlField.val(decodeURIComponent(urlParam));
+    var url = location.href.match(/url=([^&]*)/);
```
Regaddi: Can we also use `this.urlParamRegExp` here?

webcompat/static/js/lib/bugform.js
```diff
@@ -18,6 +18,7 @@ function BugForm() {
   this.submitTypeInput = $("#submit_type:hidden");
   this.uploadLabel = $(".js-label-upload");
   this.uploadLoader = $(".js-image-loader");
+  this.urlParamRegExp = /url=([^&]*)/;
```
Regaddi: I'd possibly check for `/url=([^&]+)/` and immediately ignore empty values for `url=`.

webcompat/static/js/lib/bugform.js
```diff
@@ -182,18 +193,28 @@ function BugForm() {
     img.src = dataURI;
   };
 
+  // Is the user trying to report a site against webcompat.com itself?
+  this.isSelfReport = function() {
```
Regaddi: Passing in the URL to check as a parameter could improve the testability of this function.

webcompat/static/js/lib/bugform.js
```diff
@@ -182,18 +193,28 @@ function BugForm() {
...
+    var url = location.href.match(this.urlParamRegExp);
+    if (url !== null) {
+      if (_.includes(decodeURIComponent(url[0]), location.origin)) {
```
Regaddi: What exactly is checked here?

If the following URL is given `https://www.webcompat.com/?url=https://www.webcompat.com`, then `url` would contain `["url=https://www.webcompat.com", "https://www.webcompat.com"]`.
So url[0] is `"url=https://www.webcompat.com"`.
`location.origin` is `"https://www.webcompat.com"`.
`_.includes(decodeURIComponent(url[0]), location.origin)` is true in this case.

What if the user visits using this URL?
`https://www.webcompat.com/?url=https://webcompat.com`

The check would fail, because `"url=https://webcompat.com"` does not contain `"https://www.webcompat.com"`.

@miketaylr Any chance to make this failsafe?

webcompat/static/js/lib/bugform.js
miketaylr: Good suggestion. Possibly worth a follow up bug and patch, but if it's not too much work let's do it.

webcompat/static/js/lib/bugform.js
miketaylr: > What exactly is checked here?

We check that the origin is somewhere inside the decoded `url` param. And yeah, in theory someone could visit with `https://www.webcompat.com/?url=https://webcompat.com`, though the likelihood is probably near zero. But let's fix it. :)

miketaylr:
Thanks for review @Regaddi, hoping I can make changes today or tomorrow.

miketaylr:
> hoping I can make changes today or tomorrow.

lol

miketaylr:
> Need help? 🙂 @miketaylr

@Regaddi yeah... would you mind taking over the PR? Been swamped with other projects. >_<

webcompat/static/js/lib/bugform.js
Regaddi: After having another thought on this, I guess this is probably too complicated to solve with too few outcomes. If that's doing problems in the future we should consider fixing it then.

Regaddi:
LGTM now. @miketaylr

miketaylr:
(just confirmed working locally as expected)

PR #2444, opened 2018-05-09
Fix #2437 - Upgrades python ua-parser version to 0.8

### LINKED ISSUE #2437: Update ua-parser to 0.8
A new version is out: https://github.com/ua-parser/uap-python

karlcow:
r? @miketaylr 
see comments in #2437 for details.

### DIFF
--- a/tests/functional/index-non-auth.js
+++ b/tests/functional/index-non-auth.js
@@ -26,16 +26,6 @@ registerSuite("Index", {
         .end();
     },
 
-    "reporter addon link is shown"() {
-      return FunctionalHelpers.openPage(this, url("/"), ".js-hero-title")
-        .findByCssSelector(".js-addon-link")
-        .getVisibleText()
-        .then(function(text) {
-          assert.include(text, "Download");
-        })
-        .end();
-    },
-
     "form toggles open then closed"() {
       return (
         FunctionalHelpers.openPage(this, url("/"), ".js-hero-title")
--- a/tests/unit/test_rendering.py
+++ b/tests/unit/test_rendering.py
@@ -16,8 +16,11 @@ import webcompat  # nopep8
 
 # Any request that depends on parsing HTTP Headers (basically anything
 # on the index route, will need to include the following: environ_base=headers
-headers = {'HTTP_USER_AGENT': ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.9; '
-                               'rv:31.0) Gecko/20100101 Firefox/31.0')}
+CHROME_UA = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3425.0 Safari/537.36'  # noqa
+FIREFOX_UA = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:61.0) Gecko/20100101 Firefox/61.0'  # noqa
+FIREFOX_AND_UA = 'Mozilla/5.0 (Android; Mobile; rv:61.0) Gecko/61.0 Firefox/61.0'  # noqa
+OPERA_UA = 'Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/28.0.1500.52 Safari/537.36 OPR/15.0.1147.100'  # noqa
+headers = {'HTTP_USER_AGENT': FIREFOX_UA}
 
 
 class TestURIContent(unittest.TestCase):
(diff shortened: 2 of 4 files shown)


karlcow:
ok let's see manually at least. This is working.

capture d ecran 2018-05-10 a 05 59 44

And I'm not sure why it would fail, because this is done on the JS side there's no request to the UA parsing in python. 

```
Unable to locate element: {"method":"css selector","selector":".js-addon-link"}
```

The test took 10+ s. I would bet on unresponsive page.

```
× chrome 66.0.3359.139 on Linux - Index - reporter addon link is shown (10.126s)
```

The `ua_parser` is used to prepopulate the form with values.  The only related issue I could imagine is that the page took too long to generate.

miketaylr:
@karlcow do you see a "retrigger build" button from https://travis-ci.org/webcompat/webcompat.com/builds/376706871?utm_source=github_status&utm_medium=notification when you log in?

Normally it's there for me... but I think since I don't have write access to your fork, only can do it. Might be good to see if it's just being slow like you suggest.

karlcow:
Same consistent failure. So weird.

karlcow:
ah it is used here   

https://github.com/webcompat/webcompat.com/blob/7661d4919d60bd8704fdce11f979340a1700c57c/webcompat/views.py#L150-L151

karlcow:
Locally and manually, it displays: chrome using chrome Canary.
but the tests seem to be using chrome 66.0.3359.139

Let's test a couple of things.

karlcow:
Ah! uap-core added headlesschrome parsing one year ago on Mar 15, 2017
https://github.com/ua-parser/uap-core/commit/8319ee986517fa69a81b74e8709772a9abf5b11f

https://github.com/ua-parser/uap-python/releases
* 0.7.2 (current version) is from Oct 31, 2016
* 0.7.3 from Feb 2, 2017
* 0.8 from Apr 10, 208

So we have the combination of a weak functional test and a better UA detection.

ok now I need to fix the functional test :)

miketaylr:
> So we have the combination of a weak functional test and a better UA detection.

:)

We could just search for "chrome" in the string, rather than expect an exact match...

karlcow:
@miketaylr 
We could search in the template, indeed. That's the quick fix. :) 
But that also means the test will fail if we were adding yet another strange test browser name. 

I think these addons link test could be on the unit testing side, as they are not dependent of the JS interaction at all but about the content. 
Testing on the unit testing side would allow us to mock easily the user agent string for testing all types of addon links AND the non-known-browser case.

karlcow:
ok let's see what Travis think about 
https://github.com/webcompat/webcompat.com/pull/2444/commits/13306b98f7444c06caec476e9584ac2337947e37

karlcow:
@miketaylr Yoohoo all checks have passed. Pfew…

tests/functional/index-non-auth.js
```diff
@@ -26,16 +26,6 @@ registerSuite("Index", {
         .end();
     },
 
-    "reporter addon link is shown"() {
```
miketaylr: 🎉

tests/unit/test_rendering.py
```diff
@@ -16,8 +16,12 @@
 
 # Any request that depends on parsing HTTP Headers (basically anything
 # on the index route, will need to include the following: environ_base=headers
-headers = {'HTTP_USER_AGENT': ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.9; '
-                               'rv:31.0) Gecko/20100101 Firefox/31.0')}
+CHROME_UA = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3425.0 Safari/537.36'  # noqa
+FIREFOX_UA = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:61.0) Gecko/20100101 Firefox/61.0'  # noqa
+FIREFOX_AND_UA = 'Mozilla/5.0 (Android; Mobile; rv:61.0) Gecko/61.0 Firefox/61.0'  # noqa
+SAFARI_UA = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_4) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/11.2 Safari/605.1.15'  # noqa
```
miketaylr: This is unused.

miketaylr:
Feel free to merge if you want to remove that safari string (or just merge and leave it in, not a huge issue inside test code).

tests/unit/test_rendering.py
karlcow: I will fix that.

karlcow:
wooot…
capture d ecran 2018-05-11 a 06 53 54

karlcow:
Let's kick it again again… 😭

karlcow:
pfew. all green. 💚

PR #2458, opened 2018-05-18
Fixes #2459 - Add .prettierrc

### LINKED ISSUE #2459: Having a .prettierrc file in user directory causes linting errors when settings vary from default settings
Without an .prettierrc file in the project prettier uses ~/.prettierrc which will cause thousands of linting errors if one has settings different from the default settings.

image

# How to reproduce

Place a `.prettierrc` file with the following content in your user directory `~/`:

```yaml
# See https://prettier.io/docs/en/options.html for available options
# parser: flow
printWidth: 100
useTabs: true
singleQuote: true
bracketSpacing: false
```

--- 

PR over here: https://github.com/webcompat/webcompat.com/pull/2458

mischah:
Seems to be pretty useless because it‘s an empty file.
But without an `.prettierrc` file in the project prettier uses `~/.prettierrc` which will cause thousands of linting errors if one has settings different from the default settings.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+# See https://prettier.io/docs/en/options.html for available options
+
+# We actually don’t need to configure anything in here, because we adapt the default settings of prettier.
+# But we need this file, because otherwise prettier will search up the file tree until to find a config file.
+# This will lead to problems if one has settings different from the default settings defined in ~/.prettierrc.
+# See https://github.com/webcompat/webcompat.com/issues/2459


mischah:
# Before

image

# After

image

# How to reproduce

Place a `.prettierrc` file with the following content in your user directory `~/`:

```yaml
# See https://prettier.io/docs/en/options.html for available options
# parser: flow
printWidth: 100
useTabs: true
singleQuote: true
bracketSpacing: false
```

Regaddi:
Seems absolutely reasonable to me.

r? @miketaylr @zoepage

mischah:
Hej, thanks for your instant feedback ⚡️ 

One question: What means 
> r?

Review?

mischah:
Thanks. Didn’t read all the docs yet 🙈

mischah:
Sorry, just read https://github.com/webcompat/webcompat.com/blob/master/docs/pr-coding-guidelines.md

Let me know if you want me to create an issue to this PR.

karlcow:
@mischah thanks for the PR. 
Yes always start with an issue.

.prettierrc
```diff
@@ -0,0 +1,2 @@
+# See https://prettier.io/docs/en/options.html for available options
+# parser: flow
```
miketaylr: Is this a leftover comment?

mischah: Jup. Will clean this up in a minute.

.prettierrc
```diff
@@ -0,0 +1,2 @@
+# See https://prettier.io/docs/en/options.html for available options
```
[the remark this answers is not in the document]
mischah: You’re right =^.^=
I will some additional info in a minute.

mischah:
Update: 

- Created this issue for reference: #2459
- Cleaned up and extended .prettierrc 
- Force pushed with updated commit message

miketaylr:
> But without an .prettierrc file in the project prettier uses ~/.prettierrc which will cause thousands of linting errors if one has settings different from the default settings.

Ah, now I understand. Makes sense -- thanks!

PR #2457, opened 2018-05-18
Add host and port to server start message

mischah:
I love to click those URLs in iTerm 😘

PR #2462, opened 2018-05-22
Fixes #2445 - Re-enables logging capabilities and initialization message

### LINKED ISSUE #2445: When starting the project in local, logs and debugging mode are not accessible anymore
Before when we were starting the project locally, the logs were displayed on the terminal, 
and we were given a PIN number for debugging live.

This is not happening anymore.

karlcow:
This also fixes the bug reported by @mischah in #2457

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -8,7 +8,6 @@
 
 from collections import namedtuple
 import json
-import logging
 import os
 import sys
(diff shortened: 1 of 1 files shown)


karlcow:
See #2445 comments for  the whys behind this issue.

PR #2471, opened 2018-05-31
Fixes #1971 - Adds stricter parsing for browser-* and a fallback

### LINKED ISSUE #1971: Investigate "browser-firefox-mobile-nightly-59.0a1" label on issue

karlcow:
This PR fixes issue #1971

## Proposed PR background

This does two things:
1. Adds a generic browser-fixme label for unknown pattern
2. Greatly simplify the list of accepted parameters for browsers names.

This is an intermediary steps toward the proposal of @denschub (which I agree with)

r? @miketaylr

### DIFF
--- a/webcompat/webhooks/helpers.py
+++ b/webcompat/webhooks/helpers.py
@@ -17,6 +17,8 @@ from webcompat.form import domain_name
 from webcompat.helpers import extract_url
 from webcompat.helpers import proxy_request
 
+BROWSERS = ['blackberry', 'brave', 'chrome', 'edge', 'firefox', 'iceweasel', 'ie', 'lynx', 'myie', 'opera', 'puffin', 'qq', 'safari', 'samsung', 'seamonkey', 'uc', 'vivaldi']  # noqa
+
 
 def extract_metadata(body):
     """Parse all the hidden comments for issue metadata.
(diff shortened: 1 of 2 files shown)


webcompat/webhooks/helpers.py
```diff
@@ -17,6 +17,8 @@
 from webcompat.helpers import extract_url
 from webcompat.helpers import proxy_request
 
+BROWSERS = ['blackberry', 'brave', 'chrome', 'edge', 'firefox', 'iceweasel', 'ie', 'lynx', 'myie', 'opera', 'puffin', 'qq', 'safari', 'samsung', 'seamonkey', 'uc', 'vivaldi']  # noqa
```
miketaylr: So the idea is, "these are the browsers we know about, we'll manually label anything else"?

webcompat/webhooks/helpers.py
```diff
@@ -38,13 +40,33 @@ def extract_browser_label(metadata_dict):
...
+            browser_name = label_data[0]
+        # Let's find a type for the browser.
+        remainder = label_data[1:]
+        browser_type = None
+        if 'mobile' in remainder:
+            browser_type = 'mobile'
+        if 'tablet' in remainder:
+            browser_type = 'tablet'
+        if browser_type:
+            dash_browser = '{browser_name}-{browser_type}'.format(
```
miketaylr: In theory it would be possible to have an input like `LOL Mobile 55.0`, which would result in `browser-fixme-mobile` --- do we want this?

karlcow: A missing test! ❤️ 
Thanks!

karlcow:
@miketaylr ok. Another round. Added two tests and modify the function. LOL Mobile 55 returns `browser-fixme`

webcompat/webhooks/helpers.py
miketaylr: I like your solution -- I think it covers the 99% use-cases just fine.

PR #2477, opened 2018-06-01
Fixes #2474 - Properly send along the submit_type to the server.

### LINKED ISSUE #2474: 400 on POST on localhost
@miketaylr 
So from the changelog I see that the code has been deployed since April 2018 version 10.0.0
That might be a good thing.

For me currently on localhost.

```bash
git checkout master
python fetch upstream
git merge upstream/master
grunt
python run.py
```

So far so good.

1. Open a browser
2. Go to http://localhost:5000/issues/new
3. File a bug report
4. Submit.

```
127.0.0.1 - - [31/May/2018 17:30:02] "POST /issues/new HTTP/1.1" 400 -
```

First question… is it only me?

miketaylr:
Thanks for finding this bug @karlcow. 

This underscores a critical weakness: we need functional tests for bug creation (ideally that don't create real bugs so we don't get labelled as spammers).

r? @karlcow 

Note: tags v11.0.0 and v11.0.1 aren't safe to deploy, but we'll cut v11.0.2 and deploy that once this is merged.

### DIFF
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -90,7 +90,7 @@ class IssueForm(FlaskForm):
                       [Optional(),
                        FileAllowed(Upload.ALLOWED_FORMATS, image_message)])
     details = HiddenField()
-    submit_type = HiddenField(default="submitanon")
+    submit_type = HiddenField()
 
 
 def get_form(ua_header):
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -267,7 +267,7 @@ function BugForm() {
   };
 
   this.storeClickedButton = function(event) {
-    this.clickedButton = event.target.value;
+    this.clickedButton = event.target.name;
   };
 
   this.trimWyciwyg = function(url) {
--- a/webcompat/templates/home-page/form.html
+++ b/webcompat/templates/home-page/form.html
@@ -119,6 +119,7 @@
         class="button button-secondary js-Button"
         id="submitanon"
         type="submit"
+        name="github-auth-report"
       >
         Report Anonymously
       </button>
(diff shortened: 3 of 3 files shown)


webcompat/form.py
```diff
@@ -90,7 +90,7 @@ class IssueForm(FlaskForm):
                       [Optional(),
                        FileAllowed(Upload.ALLOWED_FORMATS, image_message)])
     details = HiddenField()
-    submit_type = HiddenField(default="submitanon")
```
miketaylr: This is totally just wrong...

webcompat/static/js/lib/bugform.js
```diff
@@ -267,7 +267,7 @@ function BugForm() {
   };
 
   this.storeClickedButton = function(event) {
-    this.clickedButton = event.target.value;
```
miketaylr: Also this... maybe at one point I had a `value` prop on the button, I don't remember. 🙈

karlcow:
> This underscores a critical weakness: we need functional tests for bug creation (ideally that don't create real bugs so we don't get labelled as spammers).

In fact it's probably a couple of things here.
1. Most of the parts in this view should be in helpers functions easily testable. The mistake was a missing key in the `form` dictionary triggering the third abort.
2. The HTTP POST on the webcompat side can be tested in unit-tests, as we know what we want to receive and what we want to send back to the browser. The GitHub data just needs to be mocked. To read again stuff in #396
3. The Abort 400 should maybe come with a log message.
4. In this case I have the feeling that 400 was not the appropriate answer, but 500, because it's a failure of our code. So there's probably something to fix here too.
5. Probably part also of #1997 work

Issue #2488, opened 2018-06-07
webcompat site: search form components receive focus even when not visible


patrickhlauke:
Currently, even when the user hasn't triggered the "Search" in the top navigation to make the search form visible, the text input and "Clear" control receive focus - so for a sighted keyboard user, tabbing after "My Account" gets their focus on the two invisible form elements before they get to the "Report bug" link.

PR #2489, opened 2018-06-07
webcompat site: tweak/remove aria-label in footer links, make search form in top nav only focusable when visible

### LINKED ISSUE #2487: webcompat site: unnecessarily verbose footer link text in assistive technologies
Links in the footer feature unnecessarily verbose `aria-label` text. all text links are prefixed with "go to ...", which is unnecessary (as AT users know that a link "goes" somewhere).
The twitter and github links use `aria-label` to convey that they open in a new tab, but follow the "open X in a new tab" wording - would be better to front-load the text to say the target/site name first.

patrickhlauke:
Fixes #2487 and #2488

Issue #2487 - remove/tweak the `aria-label`s in footer links
Issue #2488 - change between `visibility:hidden`/`visibility:visible` for the search form, which also makes the form non-focusable and hidden from assistive technologies when it's not active

r? @miketaylr
---

### DIFF
--- a/webcompat/static/css/src/nav.css
+++ b/webcompat/static/css/src/nav.css
@@ -87,6 +87,7 @@
   top: 1px;
   transform: rotateY(-90deg);
   transition: all .2s ease 0s;
+  visibility: hidden;
   width: 100%;
   z-index: 1200;
 }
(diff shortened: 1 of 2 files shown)


miketaylr:
sorry about the merge conflict here, I merged in the wrong order. fixing...

PR #2496, opened 2018-06-15
Fixes #2453 - Changes regex to fix progressive label filtering

### LINKED ISSUE #2453: Progressive filtering doesn't work for Labels
**Environment:**
Browser / Version: Chrome 66 (Desktop), Firefox Nightly 62.0a1 (Desktop)
Operating System: Windows 10

**Prerequisites:**
Login onto webcompat with an account that has labels editing privileges.

**Steps to Reproduce:**
1. Go to (for example) https://webcompat.com/issues/16883.
2. Click on the Category Editor Launcher button.
3. Start typing something (eg. firefox).

**Expected Behavior:**
The Labels list is reduced to te relevant items.

**Actual Behavior:**
The list is not filtered.

**Notes:**
1. Screen capture attached.
2. Reproducible both on Chrome and Firefox.

Watchers:
@softvision-sergiulogigan
@softvision-oana-arbuzov

sv;

image

laghee:
This PR fixes issue `#2453`

## Proposed PR background

Progressive label filtering was broken, seemingly because of something wonky in the regex. I simplified the regex used in `editor.js` to escape special characters in label titles. I only included currently used special characters, so if future label titles involve anything other than `/`, `.`, `(`, `)`, `/s`, `-`, `&`, and `:`, this will need to be updated.

---

- [✔︎] I have read the [Pull Request + Code Style Guidelines](https://github.com/webcompat/webcompat.com/blob/master/docs/pr-coding-guidelines.md) thoroughly.

### DIFF
--- a/tests/functional/labels-auth.js
+++ b/tests/functional/labels-auth.js
@@ -109,6 +109,35 @@ registerSuite("Labels (auth)", {
           assert.notInclude("is-active", className);
         })
         .end();
+    },
+
+    "Label editor filters label list based on entered text": function() {
+      var count = 0;
+      return FunctionalHelpers.openPage(
+        this,
+        url("/issues/100"),
+        ".js-LabelEditorLauncher",
+        true /* longerTimeout */
+      )
+        .findByCssSelector(".js-LabelEditorLauncher")
+        .click()
+        .end()
+        .findByCssSelector(".js-label-search")
+        .type("tacos")
+        .end()
+        .sleep(1000)
+        .findAllByXpath(
+          '//label[contains (@class, "label-editor-list-item") and not(@style="display: none;")]'
+        )
+        .then(function(elements) {
+          count = elements.length;
+          assert.deepEqual(
+            1,
+            count,
+            "Entering label name in search box filters to matching label"
+          );
+        })
+        .end();
     }
   }
 });
--- a/webcompat/static/js/lib/editor.js
+++ b/webcompat/static/js/lib/editor.js
@@ -105,7 +105,7 @@ issues.CategoryEditorView = Backbone.View.extend({
     }, 100);
 
     var escape = function(s) {
-      return s.replace(/[-/\\^$*+?:.()|[\]{}]/g, "\\$&");
+      return s.replace(/[-&:/.\s()]/g, "\\$&");
     };
     var re = new RegExp("^" + escape(e.target.value), "i");
     var toHide = _.filter(this.model.toArray(), function(label) {


miketaylr:
OK, let's see. 

When looking at a normal issue (non-fixture):

screen shot 2018-06-21 at 12 22 22 pm

If i type `s`, I get 2 labels.

When looking at fixture data, I get one:

screen shot 2018-06-21 at 12 22 59 pm

It seems like we're matching on the first letter after `status-`... this seems expected, I think?

So if you search for `status-tacos`, it won't match because it's looking for `status-status-tacos`. Does that make sense? If I were writing a test, I would search for `tacos` instead, then do assertions, etc.

https://github.com/webcompat/webcompat.com/blob/d8d18c65b7fd3be1e3a69ed6bd700310f7fdbb22/webcompat/static/js/lib/models/label-list.js#L25 handles all the matching around "namespaces"... I think. It's been a while. :)

laghee:
Wait, so `status` isn't part of the actual label text? Ahhhhhhh, OK. And I suppose that's true for `closed` and `browser`, too? Now I get it.

Sorry, @miketaylr , my bad -- even setting the tests to run in non-headless, I couldn't see the labels themselves or get the inspector to run, so I just assumed everything after `label` was displayed text (along the lines of `status: add to changelog`). 🤦‍♀️  (Now that I go back and look at the live labels with the inspector, I see it's the same there, which totally slipped past me.)

New commit incoming ...

miketaylr:
OK, travis is super busted, let's do this manually:

`npm run test:js -- --functionalSuites=tests/functional/labels-auth.js`

```bash
✓ firefox on darwin 17.6.0 - Labels (auth) - Label editor opens then closes (clicks)
✓ firefox on darwin 17.6.0 - Labels (auth) - Label editor opens then closes (key events)
✓ firefox on darwin 17.6.0 - Labels (auth) - Clicking outside label editor closes it
✓ firefox on darwin 17.6.0 - Labels (auth) - Clicking close button actually closes it
✓ firefox on darwin 17.6.0 - Labels (auth) - Label editor filters label list based on entered text
✓ chrome 67.0.3396.87 on Mac OS X - Labels (auth) - Label editor opens then closes (clicks)
✓ chrome 67.0.3396.87 on Mac OS X - Labels (auth) - Label editor opens then closes (key events)
✓ chrome 67.0.3396.87 on Mac OS X - Labels (auth) - Clicking outside label editor closes it
✓ chrome 67.0.3396.87 on Mac OS X - Labels (auth) - Clicking close button actually closes it
✓ chrome 67.0.3396.87 on Mac OS X - Labels (auth) - Label editor filters label list based on entered text

node
Tunnel: Ready

Total: [✓✓✓✓✓✓✓✓✓✓] 10/10
Passed: 10  Failed: 0   Skipped: 0

Fx:         [✓✓✓✓✓] 5/5
Chr 67 Mac: [✓✓✓✓✓] 5/5
it started, yay!
```

tests/functional/labels-auth.js
```diff
@@ -109,6 +109,35 @@ registerSuite("Labels (auth)", {
...
+        true /* longerTimeout */
+      )
+        .findByCssSelector(".js-LabelEditorLauncher")
+        .click()
+        .end()
+        .findByCssSelector(".js-label-search")
+        .type("tacos")
+        .end()
+        .sleep(1000)
+        .findAllByXpath(
```
miketaylr: https://theintern.io/docs.html#Leadfoot/2/api/Command/finddisplayedbycssselector is probably another way to only grab visible docs, but this is working. So...

laghee: The problem with `findDisplayedByCssSelector` (iirc) is that it will only grab the first element that qualifies, so it could still be displaying others. I tried using `findAllByCssSelector`, then `isDisplayed`, but it wasn't working for some reason (all were displayed). I did eventually put in a short sleep after I switched tactics, though, so it might have worked with that delay.

PR #2504, opened 2018-06-18
Fixes #2375. Wait a little longer before doing some things on Travis.

### LINKED ISSUE #2375: Intermittent Travis failures
See https://travis-ci.org/webcompat/webcompat.com/builds/364187331#L1039

screen shot 2018-04-09 at 1 26 39 pm

https://github.com/webcompat/webcompat.com/blob/8203b7a482cbb78fabb5a458e8565227d15acebd/tests/functional/lib/setup.js#L21

Possibly there's a timing issue here? I wonder if we can just throw in a timeout here. I'd like to add some diagnostic logging to understand the error first, right now we just swallow it.

miketaylr:
This is a stab in the dark...

Before we used to invoke the selenium servers ourselves, but now intern does it via `intern.run()`. I dunno if sleeping here will make a difference, but it won't hurt!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ install:
   - python run.py -t &
 
 before_script:
-  - "sleep 2"
+  - sleep 5
   # lint JS in default eslint task
   - npm run lint
   - npm run build
@@ -58,4 +58,5 @@ before_script:
 # now run the tests!
 script:
   - nosetests
+  - sleep 5
   - npm run test:js -- --reporters="runner" --firefoxBinary=`which firefox`
--- a/tests/functional/lib/setup.js
+++ b/tests/functional/lib/setup.js
@@ -50,16 +50,12 @@ intern.registerPlugin("checkServer", function() {
     });
 
     // Handle connection errors.
-    request.on("error", function() {
+    request.on("error", function(err) {
       reject(
         new Error(
           `
         ======================================================
-        Oops, something went wrong. Try restarting the server.
-        Open another terminal and window type:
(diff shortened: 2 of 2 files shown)


miketaylr:
> something bad happened inside intern.run()

Ok.

miketaylr:
```
======================================================
    Intern checkServer Connection Error: Error: connect ECONNREFUSED 127.0.0.1:5000
======================================================
```

ok, that's more useful.

miketaylr:
So I wonder. We recently upgraded Flask to 1.0 and Intern to 4.2.0, and no real changes to code related to starting intern or tests. It looks like Intern is having a hard time talking to Flask.

miketaylr:
strange, suddenly tests are working again.

miketaylr:
So, tests work again... but a little delay and some better logging is still a good thing for the future.

r? @karlcow

miketaylr:
> So, tests work again...

...and they failed.

.travis.yml
```diff
@@ -50,12 +50,13 @@ install:
...
-  - "sleep 2"
+  - sleep 5
   # lint JS in default eslint task
   - npm run lint
   - npm run build
 
 # now run the tests!
 script:
   - nosetests
+  - sleep 5
```
karlcow: * I searched for similar issues online and one comment was saying that 10s was not enough and they put 20s. 
* Also most of the starting script for servers have `python run.py > /dev/null &`

karlcow: nosetests phase doesn't need the python server to be started.

PR #2516, opened 2018-06-21
Fixes #2515 - Make tmp path if it doesn't already exist.

### LINKED ISSUE #2515: IOError: [Errno 2] No such file or directory: '.../webcompat.com/tmp/webcompat.log'
Noticed by @laghee in https://github.com/webcompat/webcompat.com/pull/2513.

```
  File "run.py", line 35, in <module>
    from webcompat import app
  File "/home/travis/build/webcompat/webcompat.com/webcompat/__init__.py", line 44, in <module>
    file_handler = logging.FileHandler(app.config['LOG_FILE'])
  File "/opt/python/2.7.14/lib/python2.7/logging/__init__.py", line 913, in __init__
    StreamHandler.__init__(self, self._open())
  File "/opt/python/2.7.14/lib/python2.7/logging/__init__.py", line 943, in _open
    stream = open(self.baseFilename, self.mode)
IOError: [Errno 2] No such file or directory: '/home/travis/build/webcompat/webcompat.com/tmp/webcompat.log'
```

I'm guessing if you don't have a `/tmp/` it blows up, so we should create it.

miketaylr:
This PR fixes issue #2515.

If the /tmp/ path didn't exist (because a user or automation hadnt yet built the project), it would throw. This fixes that.

r? @karlcow

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -7,6 +7,7 @@
 """This module handles application configuration and secrets."""
 
 from collections import namedtuple
+import errno
 import json
 import os
 import sys
@@ -136,6 +137,11 @@ CSP_LOG = True
 # 2015-09-14 20:50:19,185 INFO: Thing_To_Log [in /codepath/views.py:127]
 
 # set the tempdir to somewhere predictable, no matter the platform
+try:
+    os.makedirs(os.path.join(os.getcwd(), 'tmp'))
+except OSError as exception:
+    if exception.errno != errno.EEXIST:
+        raise
 tempfile.tempdir = os.path.join(os.getcwd(), 'tmp')
 print('Writing logs to: {}'.format(tempfile.gettempdir()))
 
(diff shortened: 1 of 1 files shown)


config/__init__.py
```diff
@@ -136,6 +137,11 @@ def update_status_config(milestones_content):
 # 2015-09-14 20:50:19,185 INFO: Thing_To_Log [in /codepath/views.py:127]
 
 # set the tempdir to somewhere predictable, no matter the platform
+try:
+    os.makedirs(os.path.join(os.getcwd(), 'tmp'))
+except OSError as exception:
+    if exception.errno != errno.EEXIST:
+        raise
 tempfile.tempdir = os.path.join(os.getcwd(), 'tmp')
```
karlcow: nit:

The code computes twice the same thing with  `os.path.join(os.getcwd(), 'tmp')`
Maybe the first thing to do.

```python
tempdir_path = os.path.join(os.getcwd(), 'tmp')
```
then use it. In case we decide to change the name one day. Who knows.

miketaylr: cool, will do.

Issue #2505, opened 2018-06-19
Put checked labels at the start of the label editor


karlcow:
When a bug has labels, for example:

<img width="321" alt="capture d ecran 2018-06-19 a 09 03 10" src="https://user-images.githubusercontent.com/505230/41568271-b508f3ba-739f-11e8-9a04-f5b606f11339.png">

And we choose to modify a label, the checked labels are not at the start of the list

<img width="337" alt="capture d ecran 2018-06-19 a 09 03 02" src="https://user-images.githubusercontent.com/505230/41568297-d6ade28c-739f-11e8-9e73-aa67d5967bd4.png">

Regaddi:
I think we will need to swap out the checkboxes for custom styled ones.
This way we could restructure the HTML to take advantage of the `:checked + label` selector and re-order the checked labels to the top of the list (with the `order` attribute).

We can simply use the `BALLOT BOX` ☐ and `BALLOT BOX WITH CHECK` ☑︎ Unicode character as a first draft. Maybe we want to have some webcompy-fancy-custom-styling afterwards.
I'll push a PR in a minute, let's discuss it there 😄

miketaylr:
Can this be closed now @Regaddi?

karlcow:
This part is working but it broke the milestones.

PR #2521, opened 2018-06-22
Issue #2505 - Custom checkbox & flexbox order to move checked labels up

Regaddi:
This PR fixes issue #2505 

## Proposed PR background

This will move the checked labels in the label editor to the top of the list.
I introduced some custom styling to achieve this, since I can't reference a parent element in CSS, e.g. to select a label which wraps around the checkbox input element.

---

### DIFF
--- a/webcompat/static/css/src/label-editor.css
+++ b/webcompat/static/css/src/label-editor.css
@@ -33,6 +33,8 @@
 
 /* list category */
 .label-editor-list {
+  display: flex;
+  flex-direction: column;
   max-height: 18.75em;
   -webkit-overflow-scrolling: touch;
   overflow-x: hidden;
@@ -52,7 +54,20 @@
   cursor: pointer;
   display: block;
   margin: 0;
+  order: 2;
   padding: calc(var(--unit-space) * .75) calc(var(--unit-space) * .5);
+}
+
+.label-editor-list-item .small {
+  vertical-align: middle;
+}
+
+.label-editor-list-item::before {
+  content: "☐";
+  display: inline-block;
+  font-size: 1.5em;
+  margin-top: -4px;
+  vertical-align: middle;
 }
 
 .label-editor-list-item:last-child {
@@ -70,8 +85,15 @@
 
 /* checkbox */
 .label-editor-list-item-checkbox {
-  margin: 0;
-  padding: 0;
+  display: none;
+}
+
+.label-editor-list-item-checkbox:checked + .label-editor-list-item::before {
+  content: "︎☑︎";
+}
+
+.label-editor-list-item-checkbox:checked + .label-editor-list-item {
+  order: 1;
 }
 
 /* Launch Category Editor */
--- a/webcompat/static/css/src/label-editor.css
+++ b/webcompat/static/css/src/label-editor.css
@@ -79,13 +79,19 @@
 .label-editor-list-item.focused {
   border-color: Highlight;
   border-style: solid;
-  border-width: 2px;
+  border-width: 2px 2px 2px 4px;
   outline: none;
 }
 
 /* checkbox */
 .label-editor-list-item-checkbox {
-  display: none;
+  clip: rect(0 0 0 0);
+  height: 0;
+  margin: 0;
+  padding: 0;
+  position: absolute;
+  visibility: hidden;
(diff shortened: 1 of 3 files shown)


magsout:
@Regaddi thanks. R+ for the css. 

Could you take a look for the failing test? It seems due to your commits.

@karlcow could you take a look for the UI? If it's ok with what you want?


My thought, it's we only change the order when we open the box, not when we check or  uncheck the inputs

miketaylr:
```
127.0.0.1 - - [26/Jun/2018 20:42:00] "GET /api/issues/100/comments?page=1 HTTP/1.1" 200 -
× chrome 67.0.3396.99 on Linux - Labels (auth) - Label editor filters label list based on entered text (1.35s)
    AssertionError: Entering label name in search box filters to matching label: expected 1 to deeply equal 21
      at Command.<anonymous>  <tests/functional/labels-auth.js:134:18>
      at Test.Label editor filters label list based on entered text [as test]  <tests/functional/labels-auth.js:132:10>
      at <src/lib/Test.ts:260:47>
```

webcompat/static/css/src/label-editor.css
```diff
@@ -70,8 +85,15 @@
 
 /* checkbox */
 .label-editor-list-item-checkbox {
-  margin: 0;
-  padding: 0;
+  display: none;
+}
+
+.label-editor-list-item-checkbox:checked + .label-editor-list-item::before {
+  content: "︎☑︎";
```
karlcow: Something I'm not clear about is why do we need to add a content.

ah… because we can't reorder the parent… in CSS.
and no luck there. https://caniuse.com/#feat=css-has This is the right use case for this.

webcompat/static/css/src/label-editor.css
```diff
@@ -70,8 +85,15 @@
 
 /* checkbox */
 .label-editor-list-item-checkbox {
-  margin: 0;
-  padding: 0;
+  display: none;
```
karlcow: I wonder if this would be better for accessibility @patrickhlauke 

```css
.label-editor-list-item-checkbox {
	/* display: none; */
	opacity: 0;
	order: 1000;
}
```
And do we need to give an outline on focus to the label?

webcompat/static/css/src/label-editor.css
Regaddi: Yes, luckily label and input are connected through the `id` and `for` attributes. So we don't rely on wrapping the label around the input. And CSS order does only a visual reordering, not a structural one.

webcompat/static/css/src/label-editor.css
Regaddi: I personally don't remove the outline on focus, because it drastically reduces the ability for keyboard navigation.
For the input we can also use the `sr-only` helper.

Regaddi:
> My thought, it's we only change the order when we open the box, not when we check or uncheck the inputs

@magsout order is tied to the checked state of the checkboxes. It's a CSS-only solution :)

magsout:
@Regaddi 
>@magsout order is tied to the checked state of the checkboxes. It's a CSS-only solution :)

Yes and it's a good solution ;)

webcompat/static/css/src/label-editor.css
patrickhlauke: i'd need to see this in action and in context somewhere, but from memory (not tested in a while) `opacity:0` was a bit inconsistent on whether or not browser/AT combinations still announced it or not. If the intent here is to have something visually hidden, but announced by AT, CSS off-screen techniques are still the most reliable.

and yes, in cases where things like checkboxes/radio buttons are "faked" (assuming that's the context here?) it's not enough to not remove outline...as the thing that receives focus (the native checkbox/radio) is not on the screen/visible, there needs to be some explicit other way that shows/hides focus on the "faked" bit. randomly, i did some work ages ago on http://wtfforms.com/ which may show some of this approach - hoping it makes sense in this context

miketaylr:
Tests are green, are we ready to merge @magsout?

magsout:
@miketaylr yes we can 👍

Thanks @Regaddi

miketaylr:
cool, thanks @magsout 
i'll go ahead and do another deploy for this.

Issue #2518, opened 2018-06-22
Testing offline - Milestones initialization


karlcow:
Steps to reproduce

1. Switch off your network connection
2. Activate the project environment
3. Enter nosetests.

Actual:
```
🐝 10:02:31 ~/code/webcompat.com
→ nosetests
Milestones are not initialized. Check logs.
```

This should not happen. The tests should be able to run independently of any milestones initialization.
We probably need at least for testing a milestones.json ready to be used so that we do not have to fetch them.

karlcow:
In https://github.com/webcompat/webcompat.com/pull/2527#issuecomment-400772839

@miketaylr said:

> Hmm... @karlcow I guess a Travis IP will hit the rate limit. I wonder if we need to hard-code milestones.json for tests?

karlcow:
I will work on this today, given the failures on Travis and the hurdles for people testing locally with/without network.

karlcow:
Some context for myself.

The milestones need to be understood in the context of #1836

Just to repeat what was said there. When 

* executing tests with nosetest
* running the app online (prod, staging)

`run.py` is **never** used.

What is used is `webcompat/*`. Each time in a module, we get a `import webcompat`, we kick the `__init__.py` which starts the milestones initialization. We probably do not want this or not in that way at least. So it would be probably good to kick it differently. 

At least #2519 reduces the network access bit.

laghee:
So after kicking Travis three times today to get tests to pass, I've become obsessed ... and confused. For one thing, I'm pretty sure this:

> When ... executing tests with nosetest ... run.py is never used.

...isn't true when running either on localhost **or** Travis. (Not sure if we're running nosetests anywhere else?)

As far as I can tell, to get `notestests` to run, we're **always** calling `run.py` or `run.py -t` either directly or by using `run start` or `run start:test`.

Couldn't we just stick a dictionary constant in `run.py` [like the one we use to compare  against in `test_config`](https://github.com/webcompat/webcompat.com/blob/81b79c4ec324cc033a16cfcf21542952fb6375b2/tests/unit/test_config.py#L41)? Since staging and production are using some mysterious `fabfile`, I assume they wouldn't be affected.

Going forward, it seems like it might create less confusion to have a more parallel process in dev/staging/prod that just pivots on environment variables or something (which I guess it probably how it works between staging and production?)...

...but for now at least we could stop Travis from crashing all the time ❓

karlcow:
[...]

PR #2527, opened 2018-06-27
Fixes #2526 - fixed filteringSort function

### LINKED ISSUE #2526: The sort function seems reversed -- like oldest and newest are doing the opposite of the expected:
Open issue for this comment by @laghee https://github.com/webcompat/webcompat.com/issues/2450#issuecomment-400588542

magsout:
This PR fixes issue #2526

## Proposed PR background

fixed `filteringSort` function. The order of issue was not stable. Every time we click on `filtered` button the order of issue changed. I created a global object to save the order and the previous select.

---

### DIFF
--- a/webcompat/static/js/lib/dashboard.js
+++ b/webcompat/static/js/lib/dashboard.js
@@ -36,14 +36,23 @@ const shouldRenderNeedsTriageList = (render = true) => {
 };
 
 /* Filtering (newest or oldest) list of Triage */
-const filteringSort = issuesNumber => {
+const filteringSort = (issuesNumber, initForm = false) => {
   const select = document.getElementById("js-Filter-sort");
-  if ("oldest" === select.value) {
-    issuesNumber.reverse();
+  const lastSelectvalue = issuesNumber.selectValue;
+  // By default the oldest option is selected
+  // But if someone selects oldest option and refresh it browser
+  // the current option selected is save
+  // so we must test the value for the initform
+  if (initForm && select.value === "newest") {
+    issuesNumber.list.reverse();
+  } else if (select.value !== lastSelectvalue && null !== lastSelectvalue) {
+    issuesNumber.list.reverse();
   }
+  // save select value
+  issuesNumber.selectValue = select.value;
   const triageList = Array.from(document.getElementsByClassName("wc-Triage"));
   triageList.forEach(issue => {
-    issue.style.order = issuesNumber.indexOf(issue.dataset["number"]);
+    issue.style.order = issuesNumber.list.indexOf(issue.dataset["number"]);
   });
 };
 
@@ -88,13 +97,11 @@ const handleSpecialFilter = (
 };
 
 /* Filtering List of triage */
-const filteringList = (renderActivityIndicator = true) => {
+const filteringList = (initForm = false, issuesNumber) => {
   /* init nbItem visible */
   let nbItemVisible = 0;
-  /* init an array of list of issue */
-  const issuesNumber = [];
 
-  if (renderActivityIndicator) {
+  if (!initForm) {
     shouldRenderActivityIndicator();
   }
   /* Triage list */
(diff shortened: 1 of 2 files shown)


laghee:
**Disclaimer:** I'm not super familiar with all the code yet, and I haven't got tests back up locally yet to double-check the Travis issue (not sure if those errors are about the new milestone fetching order?), so I'm tagging in a backup reviewer ;p ... r? @miketaylr 

So [from this it seems like the milestone list of issues (`issuesNumber`) come through in order of oldest-first, ascending](https://github.com/webcompat/webcompat.com/blob/a0dde5e397290e20e9cccf81533e64c25c0ed81c/webcompat/views.py#L437). And since we're initializing on `newest`, we need to reverse that order, so that seems sensible 👍.

Saving the sort order between filters seems reasonable, too.

I'm curious about the `sort` option in the dropdown. It doesn't seem to have utility beyond explaining the function, but the user doesn't see it unless they click (which won't have any effect on filtering). Would it make sense, or make any logic simpler/easier if we were to load that as the default option?

miketaylr:
> It failed with 403 Client Error: Forbidden for url: https://api.github.com/repos/webcompat/webcompat-tests/milestones!

Hmm... @karlcow I guess a Travis IP will hit the rate limit. I wonder if we need to hard-code milestones.json for tests?

karlcow:
@miketaylr this is #2518

miketaylr:
> @miketaylr this is #2518

thanks, it felt familiar.

miketaylr:
> I'm curious about the sort option in the dropdown. It doesn't seem to have utility beyond explaining the function, but the user doesn't see it unless they click (which won't have any effect on filtering). Would it make sense, or make any logic simpler/easier if we were to load that as the default option?

@magsout ?

magsout:
@laghee 
>I'm curious about the sort option in the dropdown. It doesn't seem to have utility beyond explaining the function, but the user doesn't see it unless they click (which won't have any effect on filtering). Would it make sense, or make any logic simpler/easier if we were to load that as the default option?

I'm not sure to understand.

What do you mean by `which won't have any effect on filtering` and `if we were to load that as the default option`?

laghee:
Sorry, @magsout, my parentheses sometimes get the better of me!

> What do you mean by `which won't have any effect on filtering` 

Selecting `sort` in the drop down and clicking on `filter` doesn't change the order of the cards.

> and `if we were to load that as the default option`?

If we loaded the page with `Sort` selected rather than `Newest`.
(We could also consider eliminating the `sort` option altogether and have just `Oldest` and `Newest`?)

magsout:
@laghee thanks for the explanations.

> (We could also consider eliminating the sort option altogether and have just Oldest and Newest?)

Good idea 👍

magsout:
@laghee @karlcow https://github.com/webcompat/webcompat.com/blob/a0dde5e397290e20e9cccf81533e64c25c0ed81c/webcompat/views.py#L437 changes the Maybe, changing filter `direction` by `desc` would a good idea?

karlcow:
I'm trying to understand. Right now we get a list where the sort is 

1. by creation date
2. from oldest to newest

The logic at the time it was designed is that the oldest ones are the more urgent.

It's how it is displayed right now.

The way the UI is working right now is that you select the parameters and you have to click "filtered" (which maybe should be "submit") which reorder the full list. 

What could be done probably is that maybe it's not necessary to have the "filtered" button and that the reordering is done with `onchange`

magsout:
@karlcow 
>The logic at the time it was designed is that the oldest ones are the more urgent.

At the moment, the UI not reflect that.

capture d ecran 2018-06-29 a 08 57 55

I need to fix that too. It's a mistake, in this way I make logic simpler in the code.

>What could be done probably is that maybe it's not necessary to have the "filtered" button and that the reordering is done with

Yes, we could do that.

magsout:
What I propose

- Remove default empty option `Sort`
- made by default `oldest` in the select (reflect the design)

little bonus? ;)
- removed button filter
- added an onChange event on every form element.

@karlcow @laghee @softvision-oana-arbuzov @softvision-sergiulogigan , what do you think?

softvision-sergiulogigan:
As far as I can see, the `Sort` and the `Newest` are pretty much the same thing (at least in today's list).
Maybe we should default to `Oldest`

And as for

> added an onChange event on every form element.

I's second this.

magsout:
@miketaylr @laghee 
updated my commits

magsout:
@laghee 
>Looks good and works for me locally -- except the filtered button is still there (and throws a type error onClick). Maybe you forgot to remove it?

oops, too much cmd+z ;)

fixed ;)

laghee:
🎉

Issue #2372, opened 2018-04-06
[Brainstorming] Re-considering the way we apply browser-labels


denschub:
This was initially posted as a reply to #1971, but given the discussion'y and potentially lenghty nature of this subject, having an own issue is more appropriate.

Let's talk about browser labels. We have a lot of them, and the current situation is [really bad](https://github.com/webcompat/web-bugs/labels?utf8=%E2%9C%93&q=firefox) and if we use any form of filtering of issues for our work queue (like I do), or use those labels to build metrics, that's bad.

I understand we trust user input here, and we also want to be able to create label on-demand. However, as I am currently a bit unsatisfied with not being able to reliably get a list of all Firefox issues, I'll ramble a bit.

Here is a list of all 81 browser labels existing right now:

---

<details>

```
browser-all
browser-blackberry
browser-blackberry-browser
browser-blackberry-webkit
browser-brave-browser
browser-chrome
browser-chrome-65-&-ie-11-&-ms
browser-chrome-mobile
browser-chrome-mobile-ios
browser-chromium
browser-current-nightly
browser-edge
browser-edge-mobile
browser-fennec
browser-ffox
browser-firefox
browser-firefox-57.0.4
browser-firefox-57.04
browser-firefox-61.0-spoofed-as-chrome-62-or
browser-firefox-61.0a1
browser-firefox-android
browser-firefox-aurora-mobile
browser-firefox-beta
browser-firefox-developer-edition-59.0b1
browser-firefox-esr
browser-firefox-focus
browser-firefox-for-android
browser-firefox-for-ios
browser-firefox-ios
browser-firefox-mobile
browser-firefox-mobile-(tablet)
browser-firefox-mobile-60.0
browser-firefox-mobile-60.0a1
browser-firefox-mobile-nightly
browser-firefox-mobile-nightly-59.0a1
browser-firefox-mobile-nightly-60.0a1
browser-firefox-mobile-tablet
browser-firefox-nightly
browser-firefox-nightly-(tablet)
browser-firefox-nightly-59.0a1
browser-firefox-nightly-60.0a1
browser-firefox-nightly-60.0a1-2018-02-25-64
browser-firefox-nightly-60.0a1-2018-03-01-64
browser-firefox-nightly-61.0a1
browser-firefox-nightly-in-lineageos
browser-firefox-nightly-mobile
browser-firefox-nightly-mobile-60.0a1
browser-firefox-nightly-version
browser-hsafari
browser-iceweasel
[...]

PR #2542, opened 2018-07-04
Fixes #2537 - Add browser-focus-geckoview to EXTRA_LABELS

### LINKED ISSUE #2537: Add "browser-focus-geckoview" to EXTRA_LABELS allowlist
Right now we can't distinguish between Focus GeckoView and Fennec, because they send the same UA string. 

As a stopgap solution -- if we encounter a bug w/ firefox-focus and firefox-mobile, we'll know it's a GeckoView bug.

miketaylr:
This will likely change in the near future, after the work for https://github.com/webcompat/webcompat.com/issues/2372 and related issues is complete.

We will need to coordinate a PR against Focus when that happens.
This PR fixes issue #2537

## Proposed PR background

This PR adds a label to the extra labels allowlist so we can distinguish between firefox focus (webview) and firefox focus (geckoview), if and only if the report comes in via their report site issue menu. For other web-reported bugs, we won't know.

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -172,6 +172,7 @@ for cat_label in cat_labels:
 # labels that we allow to be added via a `label` GET param, when
 # creating an issue.
 EXTRA_LABELS = [
+    'browser-focus-geckoview',
     'type-media',
     'type-stylo',
     'type-tracking-protection-basic',
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -159,6 +159,10 @@ class TestWebhook(unittest.TestCase):
             ({'browser': 'Firefox Developer Edition 60.0b14 (64-bit)'}, 'browser-firefox'),  # noqa
             ({'browser': 'Firefox Mobile Nightly 61.0 & Firefox PC Nightly'}, 'browser-firefox-mobile'),  # noqa
             ({'browser': 'LOL Mobile 55.0'}, 'browser-fixme'),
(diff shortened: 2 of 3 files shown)


config/__init__.py
```diff
@@ -172,6 +172,7 @@ def update_status_config(milestones_content):
 # labels that we allow to be added via a `label` GET param, when
 # creating an issue.
 EXTRA_LABELS = [
+    'browser-focus-geckoview',
```
karlcow: this adds an extra label of the browser-* category

Will it work? with the new webhook logic. Maybe this requires a test. because there will be a `browser-focus-geckoview` coming from extra_labels AND a `browser-firefox-mobile` coming from browser_label.

Will this end up with a `browser-fixme`?

miketaylr: ...good question. let's go find out.

miketaylr:
r? @karlcow 

Tests are passing. I don't think `extract_browser_label` and `extract_extra_labels` will conflict, given that they look at different keys of the `metadata_dict`.

tests/unit/test_webhook.py
```diff
@@ -169,6 +173,8 @@ def test_extract_extra_labels(self):
         """Extract 'extra' label."""
         metadata_tests = [
             ({'extra_labels': 'type-media'}, ['type-media']),
+            ({'extra_labels': 'browser-focus-geckoview'},
```
miketaylr: Oh wait, typo.

PR #2544, opened 2018-07-06
Fixes #2507 - Display needsinfo requests in user activity page

### LINKED ISSUE #2507: Surface the nicknamed needinfo into the /activity/me page
status-needinfo-nickname exists. We can surface this information into the activity/me page

miketaylr:
This PR fixes issue #2507

## Proposed PR background

Just adding another section to the activity page, showing needsinfo requests.

Need to add some functional tests before I request review.


miketaylr:
Actually... maybe @karlcow can review the python bits before he goes on PTO.

miketaylr:
(and then I'll ping someone else for review on functional tests and JS)

webcompat/api/endpoints.py
```diff
@@ -105,8 +108,12 @@ def get_user_activity_issues(username, parameter):
     # copy the params so we can add to the dict.
     params = request.args.copy()
     params['state'] = 'all'
-    params['{0}'.format(parameter)] = '{0}'.format(username)
+    if parameter == 'needsinfo':
+        params['labels'] = 'status-needsinfo-{0}'.format(username)
+    else:
+        params['{0}'.format(parameter)] = '{0}'.format(username)
```
karlcow: I first wonder about why `'{0}'.format(parameter)` instead of `parameter`, then I though wait this is the same thing for username. (not your code which is fine, but the previous version.)

```python
>>> params = {}
>>> parameter = u'created'
>>> type(parameter)
<type 'unicode'>
>>> username = u'miketaylr'
>>> type(username)
<type 'unicode'>
>>> params[parameter] = username
>>> params
{u'created': u'miketaylr'}
>>> parameter.encode('utf-8')
'created'
>>> params[parameter.encode('utf-8')] = username.encode('utf-8')
>>> params
{u'created': 'miketaylr'}
>>> parameter = parameter.encode('utf-8')
>>> parameter
'created'
>>> params[parameter] = username.encode('utf-8')
>>> params
{u'created': 'miketaylr'}
>>> params['created']
'miketaylr'
```

webcompat/api/endpoints.py
```diff
@@ -105,8 +108,12 @@ def get_user_activity_issues(username, parameter):
...
     path = 'repos/{path}'.format(path=ISSUES_PATH)
+    print(path, params)
```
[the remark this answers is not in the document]
miketaylr: haha yes, thanks.

webcompat/api/endpoints.py
miketaylr: yeah, good question. it's possible it was a more complex formatting during some "in progress" version, and then was trimmed without much thought. :)

let's fix here.

miketaylr:
OK, python nits addressed.

r? @magsout for non-python review parts.

magsout:
@miketaylr it's on my todo list. Lot of working travel ;)

magsout:
@miketaylr goog job.

Just few question about UX.

There is 3 section in this page, so 3 paginations. But whatever the pagination you're clicking every section is updated. It looks weird, nah, what do you think?

magsout:
following my previous question, I think we should hide pagination if there is not result:

capture d ecran 2018-07-13 a 22 21 03

I think is not depends of it view, it seems general, isn't it ?

miketaylr:
> But whatever the pagination you're clicking every section is updated. It looks weird, nah, what do you think?

Yeah, that feels like a bug.

> following my previous question, I think we should hide pagination if there is not result:

Yeah, good idea!

miketaylr:
> following my previous question, I think we should hide pagination if there is not result:
> 
> Yeah, good idea!

In theory, the current behavior is to add a "is-disabled" class. This seems to work on the issue list page, but no here. So, some bug somewhere. If I can find it, let's fix that then open a follow up bug to swap "is-disabled" with "is-hidden" (or something like that).

miketaylr:
> If I can find it, let's fix that then open a follow up bug to swap "is-disabled" with "is-hidden" (or something like that).

Hey, it turns out that's our default behavior. :p

Been a long time since I've looked at this code. The bug was that all the shared stuff was trying to paginate on the "my issues" stuff. I've made sure it operates on the correct instance instead.

r? @magsout

PR #2551, opened 2018-07-12
Fixes #2550 - Adds samesite attribute to session cookie

### LINKED ISSUE #2550: Add SameSite attribute to cookie config to avoid CSRF attacks
([enabled in flask v1.0 a few months ago](http://flask.pocoo.org/docs/1.0/security/#set-cookie-options))

laghee:
This PR fixes issue `#2550`

## Proposed PR background

Making a security tweak for safer compatting (based on [analysis from observatory.mozilla.org](https://observatory.mozilla.org/analyze/webcompat.com) and [new flask capability](http://flask.pocoo.org/docs/1.0/security/#set-cookie-options)).

**Edit:** Added conditional disabling so functional tests don't crash.

---

- [✅ ] I have read the [Pull Request + Code Style Guidelines](https://github.com/webcompat/webcompat.com/blob/master/docs/pr-coding-guidelines.md) thoroughly.

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -122,9 +122,11 @@ DEBUG = False
 if not PRODUCTION:
     DEBUG = True
 
-# localhost runs on HTTP, use secure flag on session cookie otherwise.
+# localhost runs on HTTP
+# use secure and samesite flags on session cookie otherwise
 if not LOCALHOST:
     SESSION_COOKIE_SECURE = True
+    SESSION_COOKIE_SAMESITE = 'Strict'
 
 # By default, we want to log CSP violations. See /csp-report in views.py.
 CSP_LOG = True
--- a/run.py
+++ b/run.py
@@ -114,6 +114,7 @@ if __name__ == '__main__':
             # disable HttpOnly setting for session cookies so Selenium
             # can interact with them. *ONLY* do this for testing.
             app.config['SESSION_COOKIE_HTTPONLY'] = False
+            app.config['SESSION_COOKIE_SAMESITE'] = None
             app.config['TESTING'] = True
             print("Starting server in ~*TEST MODE*~")
             app.run(host='localhost')


config/__init__.py
```diff
@@ -122,9 +122,11 @@ def update_status_config(milestones_content):
 if not PRODUCTION:
     DEBUG = True
 
-# localhost runs on HTTP, use secure flag on session cookie otherwise.
+# localhost runs on HTTP
+# use secure and samesite flags on session cookie otherwise
 if not LOCALHOST:
     SESSION_COOKIE_SECURE = True
+    SESSION_COOKIE_SAMESITE = 'Lax'
```
miketaylr: Does `Strict` break things?

laghee: The flask docs said `Lax` was recommended because
> `Strict` prevents sending cookies with all external requests, including following regular links.

That sounded potentially "breaky" to me. We set cookies to keep track of login state, don't we?

miketaylr: I guess it's worth testing... the cookies we use to track login state only make sense on webcompat.com, them not being sent to GitHub shouldn't matter.

The only open question is if logging in the initial time will still work, but I suspect it will -- because it sends back auth token junk in the URL. But should be easy to test by removing the local oauth app, and re-authenticating with `Strict` set.

In general, I think it's good to be as secure as possible, so it's worth testing.

laghee: OK, so I don't really understand what's going on, but it seems like even with `Lax`, I'm having problems logging into GH, but it's GH giving me the 404. Is that because we're not managing to pass it something? Or is it just me? 😭  I'm so confused by which file powers what in testing vs production/staging (comment: #2518 ), so I'm not sure I disabled/enabled everything I needed to. Advice appreciated.

screenshot 2018-07-17 19 33 20

screenshot 2018-07-17 19 33 41

miketaylr:
Thanks for adding this, @laghee

PR #2558, opened 2018-07-26
Issue #2557 - Remove form-action until we better understand issue #2557.

### LINKED ISSUE #2557: Unable to report issue on webcompat.com using "Report as foolip"
Starting from https://webcompat.com/issues/new?label=type-google (HTTPS) and being logged in to the site using my GitHub account, I see a "Report as foolip" button alongside the "Report Anonymously" button. Using that button doesn't work, the whole form is fading in/out (intentional, I presume) but nothing happens. In the Chrome devtools console:
```
Refused to send form data to 'http://webcompat.com/' because it violates the following Content Security Policy directive: "form-action 'self'".
```

In other words, a problem of mixing http and https.

miketaylr:
This doesn't fix #2557, but it at least unbreaks master so we can deploy the latest there.

### DIFF
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -511,7 +511,6 @@ def add_csp(response):
         "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; " +
         "base-uri 'self'; " +
         "frame-ancestors 'self'; " +
-        "form-action 'self'; " +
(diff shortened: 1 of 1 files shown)


miketaylr:
Going to r=me on this one to unbreak master.

PR #2562, opened 2018-08-02
Fixes #2560 - Add browser-firefox-reality to EXTRA_LABELS

### LINKED ISSUE #2560: figure out a way to identify firefox reality reporters
right now they have a built-in reporter that uses the following link: 

https://webcompat.com/issues/new?src=web-fxr

We should add a label, or parse their UA string.

miketaylr:
This way we can identify them from in-product reports (before UA strings are finalized... and device detection libs updated, etc.).

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -175,6 +175,7 @@ for cat_label in cat_labels:
 # creating an issue.
 EXTRA_LABELS = [
     'browser-focus-geckoview',
+    'browser-firefox-reality',
     'type-google',
     'type-media',
     'type-stylo',


cvan:
@miketaylr: nice, thanks for adding!

can we get the label added to the [Labels in the `web-bugs` repo](https://github.com/webcompat/web-bugs/labels). or if they magically get added, even better! thanks :+1:

miketaylr:
> can we get the label added to the Labels in the web-bugs repo. or if they magically get added, even better! thanks 👍

yep, they get magically added as soon as a bug report comes with that label. 🎉 🎈

PR #2568, opened 2018-08-09
Fixes #2567 - Starts the server for functional tests after nosetests.

### LINKED ISSUE #2567: [travis] nosetests should not be run with python run.py -t
Following the discussion in https://github.com/webcompat/webcompat.com/issues/2518#issuecomment-408359792

> At any rate, you're right, of course -- it's not that nosetests kicks run.py directly... it's that in Travis we always execute run.py -t before we run nosetests (or, rather, all the test scripts). (At least digging into all this, I learned a bunch of stuff about our configuration! (-; )

This should not happen. It's a "bug".

We should start `python run.py -t` after nosetests has run.

karlcow:
This PR fixes issue  #2567 

## Proposed PR background

This intends to fix the fact python run.py starts a server before running nosetests.
Unit testing is independent of any tests running and that would delay the execution time and even could fail when milestones have difficult time to be initialized.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,6 @@ install:
   - firefox --version
   # lint python
   - pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example
-  - python run.py -t &
 
 before_script:
   - sleep 5
@@ -58,5 +57,6 @@ before_script:
 # now run the tests!
 script:
   - nosetests
+  - python run.py -t &
   - sleep 5
   - npm run test:js -- --reporters="runner" --firefoxBinary=`which firefox`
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -47,7 +47,6 @@ install:
   - firefox --version
   # lint python
   - pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example
```
laghee: This seems like a sensible change to me. 👍  
Does it make any sense to move the `pep8` call into the `before_script` section, while we're at it (for consistency with other linting)?

miketaylr: That sounds reasonable.

karlcow:
waiting for travis to do its thing.

PR #2577, opened 2018-09-13
Fixes #1955 - Adds missing tests to webhooks

### LINKED ISSUE #1955: Complete the webhooks tests coverage
The webhook handlers run in our app, and they just react to JSON coming from GitHub, so we should be able to test them (which would prevent us from breaking them in the future :cough: sorry :cough:).

karlcow:
This PR fixes issue #1955

## Proposed PR background

This does a couple of things.

1. It removes the blueprint for webhooks. It was not that useful in this context (one route, no templates) and it made the testing a lot harder.
2. It removes the update_issue which was kind of duplicating what new_opened_issue was doing.
3. There's a room for improving by creating a function to extract the useful information in new_opened_issue.
4. It adds the missing tests for webhooks module.

@miketaylr r?

### DIFF
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -52,3 +52,6 @@ class TestConfig(unittest.TestCase):
         milestones_json = json_data('milestones_content_plus.json')
         actual = update_status_config(milestones_json)
         self.assertEqual(actual, expected)
+
+if __name__ == '__main__':
+    unittest.main()
--- a/webcompat/webhooks/helpers.py
+++ b/webcompat/webhooks/helpers.py
@@ -97,24 +97,6 @@ def extract_priority_label(body):
     return None
 
 
-def update_issue(payload, issue_number):
-    """Send a GitHub PATCH to set labels and milestone for the issue.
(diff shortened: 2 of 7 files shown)


karlcow:
ah syntax. Thanks @TravisBuddy

tests/unit/test_config.py
```diff
@@ -52,3 +52,6 @@ def test_update_status_config(self):
         milestones_json = json_data('milestones_content_plus.json')
         actual = update_status_config(milestones_json)
         self.assertEqual(actual, expected)
+
+if __name__ == '__main__':
```
miketaylr: I guess this allows us to run the test independent of nose?

webcompat/webhooks/helpers.py
```diff
@@ -97,24 +97,6 @@ def extract_priority_label(body):
     return None
 
 
-def update_issue(payload, issue_number):
```
miketaylr: Presumably we're putting this back into `new_opened_issue` so it's easier to test?

Issue #2380, opened 2018-04-09
Remove waffle.io ?


zoepage:
Are we still using waffle.io? If not, we might want to remove it from our README.md?

miketaylr:
I don't think we've been using it very actively. Github projects is a good alternative, if we feel like we need to organize a specific project.

zoepage:
Okay, let's remove it then from the README.md.

Are there any labels that we can also remove that are used by waffle.io?

magsout:
@zoepage The only label used by waffle is 

![capture d ecran 2018-04-13 a 21 31 56](https://user-images.githubusercontent.com/1997108/38754446-3092cec4-3f62-11e8-84d1-10f5e02278b8.png)

zoepage:
@magsout Wasn't `ready` also one? 🤔

magsout:
@zoepage ah, hum is it @karlcow that puts this label there https://github.com/webcompat/webcompat.com/pull/2379 ? Or his account connected with waffle?

zoepage:
Both might be possible. We should just keep this in mind when we remove waffle.io :)

Obtin:
@zoepage @karlcow is this issue still active ? If yes, I would like to take this up.

karlcow:
@Obtin please do ! :) thanks.

karlcow:
The line in question.
https://github.com/webcompat/webcompat.com/blame/d56cd783366d725bcf2cf560723a590faf41e08f/README.md#L11

karlcow:
Closed by #2588

PR #2588, opened 2018-09-21
Removed waffle.io from README.md

Obtin:
This is in Refernce to Issue #2380.


karlcow:
We usually notify about the issue number in the pull request title.
See the example in point 9 of https://github.com/webcompat/webcompat.com/blob/master/docs/pr-coding-guidelines.md#pull-request-guidelines

I have modified this one, but you can follow the proper syntax for the next one!
Thanks. :)

PS: Also usually in the commit message do like this `Issue #nnnn - What the commit does`.

PR #2593, opened 2018-09-23
Fixes #2587 - Adds subpath parameter to the wellknown function

### LINKED ISSUE #2587: The parameter subpath is not passed to the route
Here in views.py

https://github.com/webcompat/webcompat.com/blob/39da3a084bd3d5001dd962a9600674bfd15766e3/webcompat/views.py#L469-L471

we have forgotten to pass the parameter to the function. It should be

```python
@app.route('/.well-known/<path:subpath>')
@cache_policy(private=False, uri_max_age=31104000, must_revalidate=False)
def wellknown(subpath):
```

kritisingh1:
Fixes issue #2587 
I also made minor correction in the file `test_urls.py`
@karlcow Please review.

### DIFF
--- a/tests/unit/test_rendering.py
+++ b/tests/unit/test_rendering.py
@@ -113,5 +113,12 @@ class TestURIContent(unittest.TestCase):
         self.assertTrue('Mac OS X 10.13' in rv.data)
         self.assertTrue('http://example.com/' in rv.data)
 
+    def test_wellknown_subpath(self):
+        """Test that the /.wellknown/subpath route gets 404."""
+        rv = self.app.get('/.well-known/test-route')
+        expected = 'test-route'
+        self.assertEqual(rv.status_code, 404)
+        self.assertTrue(expected in rv.data)
+
 if __name__ == '__main__':
     unittest.main()
--- a/tests/unit/test_urls.py
+++ b/tests/unit/test_urls.py
@@ -193,7 +193,7 @@ class TestURLs(unittest.TestCase):
         self.assertEqual(rv.status_code, 400)
 
     def test_tools_cssfixme(self):
-        """Test that the /tools/cssfixme route gets 200."""
+        """Test that the /tools/cssfixme route gets 410."""
         rv = self.app.get('/tools/cssfixme')
         self.assertEqual(rv.status_code, 410)
 
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -486,7 +486,7 @@ def log_csp_report():
 
 @app.route('/.well-known/<path:subpath>')
 @cache_policy(private=False, uri_max_age=31104000, must_revalidate=False)
-def wellknown():
+def wellknown(subpath):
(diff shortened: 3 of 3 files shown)


tests/unit/test_rendering.py
```diff
@@ -113,5 +113,11 @@ def test_form_rendering(self):
         self.assertTrue('Mac OS X 10.13' in rv.data)
         self.assertTrue('http://example.com/' in rv.data)
 
+    def test_wellknown_subpath(self):
+        """Test that the /.wellknown/subpath route gets 404."""
+        rv = self.app.get('/.well-known/{subpath}')
```
karlcow: here the code is testing a `{subpath}` variable, which is not we are trying to do. The `{subpath}` in the flask route is a keyword we are using later on in the template. See [variable rules](http://flask.pocoo.org/docs/1.0/quickstart/#variable-rules)

We are trying to test if any string after `/.well-known/` will trigger a 404. For example `/.well-known/wonderland` or `/.well-known/jupiter` really anything should trigger a 404.

So you can name any tests routes, the name doesn't matter.

tests/unit/test_rendering.py
```diff
@@ -113,5 +113,11 @@ def test_form_rendering(self):
...
+        self.assertEqual(rv.status_code, 404)
+
```
karlcow: As these tests are about rendering. It would be to check if `subpath` is inside the rendered content.

tests/unit/test_urls.py
```diff
@@ -193,7 +193,7 @@ def test_csp_report_uri_bad_content_type(self):
         self.assertEqual(rv.status_code, 400)
 
     def test_tools_cssfixme(self):
-        """Test that the /tools/cssfixme route gets 200."""
+        """Test that the /tools/cssfixme route gets 410."""
```
karlcow: oh nice litlle fix. :) Thanks.

webcompat/views.py
```diff
@@ -486,7 +486,7 @@ def log_csp_report():
 
 @app.route('/.well-known/<path:subpath>')
 @cache_policy(private=False, uri_max_age=31104000, must_revalidate=False)
-def wellknown():
+def wellknown(subpath):
```
karlcow: This fix is good.

karlcow:
@kritisingh1 The previous commit was `added subpath parameter to the wellknown function`. 
On this project we follow a [consistent style guide](https://github.com/webcompat/webcompat.com/blob/master/docs/pr-coding-guidelines.md). Make sure to carefully read it.

Specifically the way we do commit messages.
```
# add commit message including issue number
git commit -m 'Issue #NumberOfIssue - Fixes broken layout on mobile browsers'  
```

In your case that would be:

`Issue #2587 - Adds subpath parameter to the wellknown function`

After fixing this PR, it would be to rebase it. If you do not know how to do it, I will help you to guide you through the steps.

kritisingh1:
@karlcow I have made the requested changes. Please review. Thanks

PR #2598, opened 2018-09-25
Fixes #2563-Fixes status change instructions

### LINKED ISSUE #2563: Unclear status-change instructions at /contributors/reproduce-bug
It sounds like we're telling people to change to needstriage status **after** they've triaged a bug they can reproduce. `NEEDSTRIAGE` here should be changed to `NEEDSDIAGNOSIS`:

screenshot 2018-08-03 13 41 40

https://github.com/webcompat/webcompat.com/blob/42781cc3a673b622b9fd0d5b07f6250895160a4b/webcompat/templates/contributors/reproduce-bug.html#L104

Also, in the second-to-last sentence here ("If the issue was reported anonymous..."), `anonymous` should be changed to `anonymously`:
https://github.com/webcompat/webcompat.com/blob/42781cc3a673b622b9fd0d5b07f6250895160a4b/webcompat/templates/contributors/reproduce-bug.html#L54

hritvi:
This PR fixes issue #2563 
r? @laghee

### DIFF
--- a/webcompat/templates/contributors/reproduce-bug.html
+++ b/webcompat/templates/contributors/reproduce-bug.html
@@ -51,7 +51,7 @@
           <h2 id="next" class="headline-1 highlight">Possible Next Steps</h2>
 
           <p>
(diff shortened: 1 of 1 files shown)


laghee:
@miketaylr Had to kick Travis -- looks like it was just the milestone thing. It actually looks like it's letting me merge... So I'm going to try it! :sad_there's_no_power_glove_in_girhub_emojis

PR #2613, opened 2018-10-01
Fixes #1297: Upgrade Markdown-it Emoji to v1.4.0

### LINKED ISSUE #1297: Emoji in markdown for specific patterns should be avoided
This is happening in https://webcompat.com/issues/4313
When a user is typing a character in a form which is just after `:` without leaving a space. The converter will interpret it as an emoji. Some strings should be out of parsing because there are part of the system. `Expected behavior:` Should not be parsed for emojis.
 

capture d ecran 2017-01-23 a 11 22 16

miketaylr:
I believe this commit fixed the issue as reported: https://github.com/markdown-it/markdown-it-emoji/commit/79fe7293850abd63f61e5e96192ce8247937b347

r? @karlcow


miketaylr:
http://localhost:5000/issues/1685 will show the fix, if you're running this branch locally:

screen shot 2018-10-01 at 4 14 55 pm

karlcow:
@miketaylr Please rebase. :) thanks.

PR #2618, opened 2018-10-01
fixes #2332 - removed the a11y-en file

### LINKED ISSUE #2332: Use `a11y-en.css` just on dev
We've added `a11y-en.css` a while ago to help us with accessibility issues. 
We commented it out at some point to deploy to staging, but it never found it's way back. 

What we want it to add it just for localhost, so we need a solution here.

The line can be found here:
https://github.com/webcompat/webcompat.com/blob/refactor/webcompat/static/css/webcompat.dev.css#L5

magsout:
This PR fixes issue #nnnn

## Proposed PR background

Removed the A11 css tools.

r @miketaylr ?


magsout:
oups..

magsout:
@miketaylr done

PR #2621, opened 2018-10-02
Fixes #2331. Remove PurifyCSS.

### LINKED ISSUE #2331: Remove PurifyCSS
In #2306, we added https://www.npmjs.com/package/grunt-purifycss to check for unused CSS.

We added the grunt task as well a npm script that can be called with `npm run clean:css`.
It will generate a `tmp/` folder with the reduced css (Just used CSS). It also shows which classes are not used thanks to the `rejected` option and can minify the CSS.

This is how the log looks like:
screen shot 2018-03-28 at 8 37 14 pm

The question now is... Do we want to replace https://www.npmjs.com/package/grunt-contrib-cssmin and use the generated and reduced CSS file and move it to our `dist` folder as the default?

If so, we should not forget to update the docs.

miketaylr:
Unused, and not maintained enough to be used.

r? @magsout


magsout:
👍

PR #2623, opened 2018-10-02
fixes #2622 - break word to avoid overflow with long string

### LINKED ISSUE #2622: URL area is overflowing the body
@magsout had fixed this, but we seem to get into a strange case for this one.

https://webcompat.com/issues/16643

magsout:
This PR fixes issue #2622 

## Proposed PR background

break word to avoid overflow with long string


magsout:
I edit un bug on local http://localhost:5000/issues/1685

PR #2628, opened 2018-10-02
Issue #2624. Get Circle CI working

miketaylr:
r? @karlcow for changes to environment.py and secrets.py

r? @magsout for config.yml parts

### DIFF
--- a/config/environment.py
+++ b/config/environment.py
@@ -53,3 +53,21 @@ WELL_KNOWN_ALL = """
 WELL_KNOWN_SECURITY = """Contact: mailto:kdubost+securitywebc@mozilla.com
 Contact: mailto:miket@mozilla.com
 """
+
+# Database backup path.
+if LOCALHOST:
+    BACKUP_DEFAULT_DEST = BASE_DIR + '/backups/'
+else:
+    BACKUP_DEFAULT_DEST = ''
+
+if LOCALHOST:
+    UPLOADS_DEFAULT_DEST = BASE_DIR + '/uploads/'
+    UPLOADS_DEFAULT_URL = 'http://localhost:5000/uploads/'
+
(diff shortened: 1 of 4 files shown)


miketaylr:
See also #2631

magsout:
hbgtv4y

config/environment.py
```diff
@@ -53,3 +53,20 @@
...
 """
+
+# Database backup path.
+if LOCALHOST:
+    BACKUP_DEFAULT_DEST = BASE_DIR + '/backups/'
+else:
+    BACKUP_DEFAULT_DEST = ''
+
+UPLOADS_DEFAULT_DEST = BASE_DIR + '/uploads/'
+UPLOADS_DEFAULT_URL = 'http://localhost:5000/uploads/'
```
karlcow: I wonder how this will be working in production. In `secrets.py`, the same variable exists. but in environment.py is read after secrets.py everything is falling apart no?

Maybe we should add a `if LOCALHOST` to be sure that the default will not crush the production one.

Another strategy which is being used for these things is through an obvious overriding mechanism with `class Config(object):`.
see http://flask.pocoo.org/docs/1.0/config/#development-production

```python
class Config(object):
    DEBUG = False
    TESTING = False
    DATABASE_URI = 'sqlite:///:memory:'

class ProductionConfig(Config):
    DATABASE_URI = 'mysql://user@localhost/foo'

class DevelopmentConfig(Config):
    DEBUG = True

class TestingConfig(Config):
    TESTING = True
```

And they use `app.config.from_object('configmodule.ProductionConfig')`

What do you think?

miketaylr: Agreed, let's wrap it in `if LOCALHOST`.

miketaylr:
thankssss

PR #2634, opened 2018-10-02
Fixes #1828 - Adds richer descriptive titles

### LINKED ISSUE #1828: Use Description for the title instead of the radio button text when it has been filled
See for example https://webcompat.com/issues/9602
It would probably be better to feed the title with the Description field than the current title after the domain name.

karlcow:
This PR fixes issue #1828 

## Proposed PR background
This creates a title composed from description and domain name. 
This is not perfect and can be improved. 
It takes into account cases where the domain is not available and where the description is not existent.

It would benefit of having `-webkit-line-clamp` (ahahah irony).

It would also be good to make a better study on the most probably length we would need for this.


karlcow:
@miketaylr Thanks for the review
@magsout has kind of already reviewed (pair coding)

PR #2640, opened 2018-10-03
Fixes #2639: Use env vars for prod and staging configuration

### LINKED ISSUE #2639: Production and Staging secrets should be read from environment variables
Localhost can still be configurable via hard coded values.

miketaylr:
r? @karlcow


karlcow:
@miketaylr This probably needs an actual manual ok here that the uwsgi ENV has been properly configured before merging here.

Issue #2629, opened 2018-10-02
Set a max characters number on Description


karlcow:
In the wake of #1828, we should probably set a limit on description. I'm not entirely sure how long is the right limit. I should probably dig into the description given by @softvision-oana-arbuzov and @softvision-sergiulogigan

softvision-sergiulogigan:
I think this should be a combination of the longest string for the URL and the longest string a registered user used for the title

karlcow:
@softvision-sergiulogigan I'll check that. Thanks!

karlcow:
Thinking about this a bit more. 
This is is partly connected to the discussion in https://github.com/webcompat/webcompat.com/pull/2812

Probably to make sense of this feature, there should be flag with the number of characters left when someone is typing.

PR #2642, opened 2018-10-04
Fixes #2630. Tweak description and textarea labels

### LINKED ISSUE #2630: Rename form fields for descriptions and step by step
Related to #1828 
In the current UI, we get:
capture d ecran 2018-10-02 a 13 32 15

* Reporters put long description in the first textarea to the point where sometimes they ignore the steps.

Probably "Please describe what was wrong" should be something "Give a short problem summary".
 And the describe what was wrong step by step in the bigger textarea

miketaylr:
Maybe not perfect, open to suggestions.

r? @karlcow

### DIFF
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -57,12 +57,12 @@ radio_message = u'Problem type required.'
 username_message = u'A valid username must be {0} characters long'.format(
     random.randrange(0, 99))
 
-desc_label = u'Please describe what was wrong'
+desc_label = u'Give a short problem summary'
 desc_message = u'An issue description is required.'
 
 url_label = u'Site URL'
 browser_test_label = u'Did you test in another browser?'
-textarea_label = u'What steps did you take before this problem occurred?'
+textarea_label = u'Please describe what happened, including any steps you took before you saw the problem'  # noqa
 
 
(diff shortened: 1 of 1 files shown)


miketaylr:
Also, r? @adamopenweb for some native English speaker feedback

adamopenweb:
Seems OK to me.

>Give a short problem summary

I might just be tired and can't tell if it reads better, but I also don't know the thoughts behind these changes. 😄 

>Please write a short summary of the problem

I think "write" might translate better than "give". 

Should the language of `desc_message` change with it?

>An issue description is required.

"Issue description" might need to be "problem summary" or similar.

miketaylr:
> but I also don't know the thoughts behind these changes

Yeah, it's a secret. Sorry. 

Basically the motivation came from @karlcow doing some research on the "Description" input, which was intended to be a quick summary -- and finding some people wrote way too much info in there. So we're trying to communicate: "short thing here", "longer thing here".

karlcow:
@adamopenweb to add on what @miketaylr said. There is also #2629

karlcow:
@miketaylr approved on my side. @adamopenweb seems to recommend some changes.

PR #2658, opened 2018-10-16
Fixes #2653. Encode details as JSON before filling input

### LINKED ISSUE #2653: Fix parsing of browser config data
We have console logs now (thanks @wisniewskit!), but it looks like formatting is a little wonky. Probably not hard to fix:

See https://webcompat.com/issues/19843

miketaylr:
This is a bit voodoo magic here. I don't actually know where the `details` POST param gets added as the hidden input value, before this patch. 

It seems like a default Flask / wtform behavior?

Either way, before this patch the hidden input is getting populated with a unicode string that kinda looks like JSON, but is not:

With this patch, we end up with encoded JSON:

screen shot 2018-10-16 at 4 23 23 pm

https://github.com/webcompat/webcompat-tests/issues/1704 is a test bug I filed from my branch.

r? @karlcow 

There's follow up work to make it prettier, but I'll file a new bug for that.

### DIFF
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -117,31 +117,61 @@ def get_form(form_data):
     return bug_form
 
 
-def get_details(details_string):
+def get_details(details):
     """Return details content.
 
-    * If JSON as a formatted string
+    * If a dict, as a formatted string
     * Otherwise as a string as-is.
     """
-    content = details_string
+    content = details
     rv = ''
     try:
-        details = json.loads(content)
         rv = ''.join(['<li>{k}: {v}</li>'.format(k=k, v=get_str_value(v))
                       for k, v in details.items()])
-    except ValueError:
+    except AttributeError:
         return content
     return rv
 
 
+def get_console_section(console_logs):
+    """Return a section for console logs, or the empty string.
+
+    This populates the named argument `{console_section}`
+    inside the formatted string that `build_details` returns.
+    """
+    if not console_logs:
+        return ''
+    return """<p>Console Messages:</p>
+<pre>
+{console_logs}
+</pre>""".format(console_logs=console_logs)
+
+
 def build_details(details):
-    """Populate and return the Browser Configuration section template."""
+    """Populate and return the Browser Configuration section template.
+
+    If we get JSON, we try to pull out the console logs before building the
+    rest of the details.
+    """
+
+    console_logs = None
+    content = details
+    try:
+        content = json.loads(details)
+        console_logs = content.pop('consoleLog', None)
+    except ValueError:
+        # if we got a ValueError, details was a string, so just pass it
+        # into get_details below
+        pass
+
     return """<details>
 <summary>Browser Configuration</summary>
 <ul>
   {details_list_items}
 </ul>
-</details>""".format(details_list_items=get_details(details))
+{console_section}
+</details>""".format(details_list_items=get_details(content),
+                     console_section=get_console_section(console_logs))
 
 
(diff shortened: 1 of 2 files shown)


webcompat/form.py
```diff
@@ -103,6 +103,7 @@ def get_form(form_data):
     ua_header = form_data['user_agent']
     # Populate the form
     bug_form.browser.data = get_browser(ua_header)
+    bug_form.details.data = json.dumps(form_data.get('details'))
```
karlcow: Yeah you ran into the same type of issues I didn't want to go into in 
https://github.com/webcompat/webcompat.com/pull/2582#discussion_r218993713
This will kind of work but leads to unintended consequences one example is:

capture d ecran 2018-10-18 a 10 25 50

`@font-face` is an active link.

unintended Notifications hell for people.

I guess this can be fixed before doing:
https://github.com/webcompat/webcompat.com/pull/2658/files#diff-f1975977b06efb529b8c60f23a7bc4eeL320

because if we do in only on the client side on webcompat.com, we don't solve the notifications issue. :/

I have the feeling there is a missing type in the webplatform to be able to store JSON directly into form values.

I wonder what would do a 

```python
console_textarea = TextAreaField("console_log", default=console_log_dumps)
``` 

then 

```css
textarea.console {visibility: hidden}
```

karlcow:
@miketaylr there is a bizarre thing in your screenshot
screen shot 2018-10-16 at 4 23 23 pm

The input shows double quotes inside double quotes is it really what happens?

This kind of issues are difficult to test without the right client. :)

miketaylr:
> The input shows double quotes inside double quotes is it really what happens?

It's *possible* this is just DevTools not displaying the escaped quotes.

I ran a little test to be sure:

screen shot 2018-10-17 at 10 00 09 pm

miketaylr:
>  Could we make sure to leave a comment to explain why we need the dumps there?

Yeah, this needs a big long comment. :) Will do.

webcompat/form.py
miketaylr: I like this idea!

Tom showed me this JSON formattiner thing: https://azimi.me/json-formatter-js/

We could toss all of the details into a textarea (to avoid weird linkage), and render it fancy-like on the client? It's perhaps overkill, but worth at least attempting in a branch. 

Maybe in the meantime, we just toss the console logs in backticks in this PR, and figure out the nicer solution in #2659 ?

```
no emails for @fontface
```

miketaylr:
Have some local stuff with this approach:

screen shot 2018-10-18 at 5 14 58 pm

I don't love it, but it's better than the current mess. We can make it better in #2659. Hoping to get patches/tests cleaned up for review after the kids go to bed tonight.

karlcow:
let me try something:

```json
{"json": "on one line", "what": "github does", "with them": "no idea"}
```

karlcow:
ah no specific formatting.

karlcow:
@miketaylr  what about doing this before dumping in the textarea

```python
things_for_textarea = json.dumps(details_dict, indent=2)
```

miketaylr:
OK, approach slightly changed. I'm still using a hidden input here, because if we stick it inside `<pre>`, we don't have the linking problems. Will experiment with a textarea and something that isn't a single line in the follow up bug.

r? @karlcow

miketaylr:
screen shot 2018-10-18 at 10 34 41 pm

miketaylr:
> @miketaylr You could have use a dict comprehension in…

haha <3

PR #2664, opened 2018-10-19
Fixes #2663 - Reduce number of untriaged bugs on homepage

### LINKED ISSUE #2663: Reduce the number of untriaged bugs on homepage
We have 10, and I doubt anybody really looks at these. How about we reduce this to 5?

To fix this see, https://github.com/webcompat/webcompat.com/blob/d56cd783366d725bcf2cf560723a590faf41e08f/webcompat/static/js/lib/diagnose.js#L30-L31

While we're editing these files, diagnose.js doesn't make sense anymore (it had this name before we had the notion of "need striage", we could rename this to untriaged.js, or triage.js or something.

But we need to rename it on other places too:

https://github.com/webcompat/webcompat.com/search?q=diagnose.js&unscoped_q=diagnose.js

Anushi1998:
Fixes #2663 - Reduce number of untriaged bugs displayed on homepage and renamed `diagnose` to `untriaged`.

Let me know if I  missed something :)

r? @miketaylr

### DIFF
--- a/grunt-tasks/concat.js
+++ b/grunt-tasks/concat.js
@@ -28,13 +28,13 @@ module.exports = function(grunt) {
       ],
       dest: "<%= jsDistPath %>/<%= pkg.name %>.js"
     },
-    diagnose: {
+    untriaged: {
       src: [
         "<%= jsPath %>/lib/models/label-list.js",
         "<%= jsPath %>/lib/models/issue.js",
-        "<%= jsPath %>/lib/diagnose.js"
+        "<%= jsPath %>/lib/untriaged.js"
       ],
-      dest: "<%= jsDistPath %>/diagnose.js"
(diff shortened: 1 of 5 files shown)


miketaylr:
Ah yeah, whoops. We need to update the test assertions:

```
127.0.0.1 - - [19/Oct/2018 01:53:49] "GET /api/issues/category/needstriage HTTP/1.1" 200 -
× chrome 70.0.3538.67 on Linux - Index - browse issues (needstriage) (0.144s)
    AssertionError: 10 issues should be displayed: expected 5 to equal 10
      at Command.<anonymous>  <tests/functional/index-non-auth.js:64:18>
      at Test.browse issues (needstriage) [as test]  <tests/functional/index-non-auth.js:63:10>
      at <src/lib/Test.ts:263:51>
```

PR #2678, opened 2018-10-25
Fixes #2679 - Reduce the size of HTML sent over network

### LINKED ISSUE #2679: Reduce the size of HTML sent over network
Minifying HTML can improve the network cost while loading a webpage and hence performance improvements! :-)

Anushi1998:
Fixes #2679 Minifying HTML can improve the network cost while loading a webpage and hence performance improvements! :-)

I have not modified the actual HTML file so they are still readable in the source code. `htmlmin` will minify the file that is sent to client.


karlcow:
I propose we close the pull request here for now. :) We can still revive something around this when we have a better idea about the discussion outcome in #2679

PR #2683, opened 2018-10-29
Fixes #2682 - Removes unused markdown-it in issue-list.js

### LINKED ISSUE #2682: Unused markdown-it initialization in issue-list.js
While building my final project I found a part of the code that isn't used anywhere:

https://github.com/webcompat/webcompat.com/blob/fa98afe61f62ed754c27cbde832b320bcec3c887/webcompat/static/js/lib/issue-list.js#L8-L17

marimeireles:
Hi @miketaylr . You think this is valid?
Thanks! :)


marimeireles:
>Ideally in the future we modularize the code and we can make a lot more improvements like this (like, not shipping the markdown-it code at all here (which is lumped in with webcompat.min.js).

I have a surprise at my project then! >:3

PR #2691, opened 2018-10-31
Fixes #2689 - Update minimum node version to 10.13.0

### LINKED ISSUE #2689: Update minimum node version in package.json
Before we merge https://github.com/webcompat/webcompat.com/pull/2685, let's require Node >= 8.6 (because lint-staged requires it).

https://github.com/webcompat/webcompat.com/blob/master/package.json#L14

I just checked out docs, and we say we require the current LTS (which is like 10.13.0), so let's just go ahead and require `>= 10.13.0`

Anushi1998:
Fixes #2689 - Update minimum node version to 10.13.0 because lint-staged requires `Node >= 8.6`

r? @miketaylr

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "url": "https://github.com/webcompat/webcompat.com.git"
   },
   "engines": {
-    "node": ">= 4.x"
+    "node": ">= 10.13.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


miketaylr:
Thank you @Anushi1998!

PR #2696, opened 2018-11-01
Fixes #2688 - Only anonymous users should go through the blacklist domain check

### LINKED ISSUE #2688: Anonymous reporting for Blacklisted domains
@karlcow  The message in [config/environment.py](https://github.com/webcompat/webcompat.com/blob/master/config/environment.py#L82) should not use the word `Anonymous` as I still can't create the issue when I opt to share my Github Id.

Anushi1998:
Fixes #2688 
@karlcow 
I have included the check of `PROXY_REPORT` before checking `is_blacklisted_domain(..)` because it would be better not to go through blacklisted domain check is we don't require it.

Let me know if I should do in the way you [mentioned](https://github.com/webcompat/webcompat.com/issues/2688#issuecomment-434581183)


webcompat/views.py
```diff
@@ -231,7 +231,8 @@ def create_issue():
             ip=request.remote_addr,
             url=form['url'].encode('utf-8')))
         # Checking blacklisted domains
-        if is_blacklisted_domain(urlparse.urlsplit(form['url']).hostname):
+        if form.get('submit_type') == PROXY_REPORT and \
```
karlcow: It's better to avoid the continuation character `\` in general. 
Your idea is good! but we could do that instead:

```python
        # Logging the ip and url for investigation
        log.info('{ip} {url}'.format(
            ip=request.remote_addr,
            url=form['url'].encode('utf-8')))
        # Check if the form is valid
        if not is_valid_issue_form(form):
            abort(400)
        # Anonymous reporting
        if form.get('submit_type') == PROXY_REPORT:
            # Checking blacklisted domains
            domain = urlparse.urlsplit(form['url']).hostname
            if is_blacklisted_domain(domain):
                msg = app.config['IS_BLACKLISTED_DOMAIN'].format(form['url'])
                flash(msg, 'notimeout')
            json_response = report_issue(form, proxy=True)
            session['show_thanks'] = True
            return redirect(
                url_for('show_issue', number=json_response.get('number')))
        # Authenticated reporting
        if form.get('submit_type') == AUTH_REPORT:
```

What do you think?

Anushi1998: Good idea. I will make the changes

Anushi1998:
@karlcow Sorry for the delay. Please review the changes

webcompat/views.py
```diff
@@ -231,10 +231,12 @@ def create_issue():
             ip=request.remote_addr,
             url=form['url'].encode('utf-8')))
         # Checking blacklisted domains
-        if is_blacklisted_domain(urlparse.urlsplit(form['url']).hostname):
-            msg = app.config['IS_BLACKLISTED_DOMAIN'].format(form['url'])
-            flash(msg, 'notimeout')
-            return redirect(url_for('index'))
+        if form.get('submit_type') == PROXY_REPORT:
```
karlcow: @Anushi1998 almost there. In my code suggestion I was proposing also to move the code after the `is_valid_issue` block. :)

Anushi1998:
@karlcow Sorry for the delay. These days Diwali (one of the two big festivals in India) is going on :p 

 My bad, I didn't notice that with care. I have not removed the comment. Let me know if I should remove it :)

webcompat/views.py
```diff
@@ -230,16 +230,17 @@ def create_issue():
...
-        # Checking blacklisted domains
-        if is_blacklisted_domain(urlparse.urlsplit(form['url']).hostname):
-            msg = app.config['IS_BLACKLISTED_DOMAIN'].format(form['url'])
-            flash(msg, 'notimeout')
-            return redirect(url_for('index'))
         # Check if the form is valid
         if not is_valid_issue_form(form):
             abort(400)
-        # Anonymous reporting
+        # Checking blacklisted domains
```
karlcow: Almost there. The comment should be after the `if form.get('submit_type') == PROXY_REPORT` as this `if` is about proxy reporting not blacklisting alone. :)

Also probably better to have the commit messages like:

```
Issue #2688 - Adds blacklist domain check only to anonymous reporter
```

* shorter.
* Says what it does.

Anushi1998: I have made the changes. Please review now :)

Anushi1998:
@karlcow Is it a good idea to add tests for blacklist domain check for authorized users in `reporting-auth.js`?

karlcow:
That would be probably reporting-non-auth.js (as we target only anonymous user) and the test should check the user receives the right UI. Potential issue because of the way this is set up, right now, it will be hard to mock the domain list. aka if one day the list of domains change the tests will fail. 

@Anushi1998 You can at least open an issue about that. But it will not be about just adding the test :) Thanks!

Anushi1998:
So my cents would be to move list of blacklisted domains to some config file and while building test we can pick anyone at random and check. 

What's your suggestion @karlcow?

karlcow:
@Anushi1998 Open a new issue about it, and we discuss it there ;)

PR #2713, opened 2018-11-15
[DO NOT MERGE] Fixes #1822 - Adds input for GitHub username

### LINKED ISSUE #1822: Provide a way to get in touch with anonymous reporters
As mentioned in yesterdays meeting, the number of anonymous reports which get closed because we have troubles understanding or reproducing the issue is very high. Sometimes, it feels like we could miss a lot of possibly important bugs, simply because the initial report didn't provide enough information for us to fully grasp it.

When people use GitHub sign in to report issues, we have much higher success rates, since we can easily ping the reporters if we have questions. It would be nice to offer anonymous reporters a way to opt-in to feedback, without the need of using a GitHub account to report.

Let's collect ideas and Proof-of-Concepts!

karlcow:
This PR fixes issue #1822

## Proposed PR background

this adds a new input for gathering a github nickname. It's a low tech way to help people who don't want to/can't do the GitHub dance.


karlcow:
@miketaylr This is just made as a test, for commenting and discussing ideas around the code and the UI.

webcompat/form.py
```diff
@@ -65,6 +67,9 @@
 browser_test_label = u'Did you test in another browser?'
 textarea_label = u'Please describe what happened, including any steps you took before you saw the problem'  # noqa
 
+contact_message = u'There is a mistake in the username.'  # noqa
+contact_label = u'Anonymous is fine. Giving your Github nickname could help us with diagnosis. (publicly visible)'  # noqa
```
miketaylr: nit: GitHub

webcompat/templates/home-page/form.html
```diff
@@ -85,7 +85,7 @@ <h1 class="headline-1">Report Site Issue</h1>
         {{ subradio.label(class_='form-label label-radio') }}
       {% endfor %}
     </div>
-    <div class="form-upload js-image-upload js-Form-group">
+    <div class="form-upload form-element js-image-upload js-Form-group">
```
miketaylr: What does `.form-element` do for us here?

webcompat/templates/home-page/form.html
```diff
@@ -112,6 +112,20 @@ <h1 class="headline-1">Report Site Issue</h1>
...
+      <div class="js-Form-information form-label-message">
+        {{ form.contact.label(class_='form-label') }}
+        {% if form.contact.errors %}
+          {% for error in form.contact.errors %}
+            <small class="form-message-error">{{ error }}</small>
+          {% endfor %}
+        {% endif %}
+      </div>
+      {{ form.contact(class_='form-field text-field',
+         placeholder='Github nickname') }}
```
miketaylr: nit: GitHub

webcompat/form.py
```diff
@@ -65,6 +67,9 @@
 browser_test_label = u'Did you test in another browser?'
 textarea_label = u'Please describe what happened, including any steps you took before you saw the problem'  # noqa
 
+contact_message = u'There is a mistake in the username.'  # noqa
```
miketaylr: I think we probably need to make similar / duplicate changes in bugform.js to be able to react to this before form submission.

webcompat/form.py
miketaylr: The wording probably needs some tweaking here... but I'm not sure what would be better. :)

webcompat/form.py
miketaylr: See the following for some hints on that

https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L25-L28
https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L313-L320
https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L355

karlcow: @miketaylr are we duplicating in bugform.js or in templates what should be done in one place. Do we have an opportunity to remove in one of the two places?

webcompat/templates/home-page/form.html
karlcow: @miketaylr probably structural spacing if I remember? but the new design kind of messed up my own understanding of the form structure.

karlcow: Maybe/Probably @magsout knows better

webcompat/form.py
karlcow: better wording: "WHO ARE YOU?"

karlcow:
capture d ecran 2018-11-29 a 10 15 04

karlcow:
```diff
diff --git a/webcompat/form.py b/webcompat/form.py
index 5496013..8816f5a 100644
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -68,7 +68,7 @@ browser_test_label = u'Did you test in another browser?'
 textarea_label = u'Please describe what happened, including any steps you took before you saw the problem'  # noqa
 
 contact_message = u'There is a mistake in the username.'  # noqa
-contact_label = u'Anonymous is fine. Giving your Github nickname could help us with diagnosis. (publicly visible)'  # noqa
+contact_label = u'Sharing your GitHub nickname (without signing up) could help us for diagnosis. (publicly visible)'  # noqa
 
 
 class IssueForm(FlaskForm):
diff --git a/webcompat/templates/home-page/form.html b/webcompat/templates/home-page/form.html
index aed962d..ee84b16 100644
--- a/webcompat/templates/home-page/form.html
+++ b/webcompat/templates/home-page/form.html
@@ -122,7 +122,7 @@
         {% endif %}
       </div>
       {{ form.contact(class_='form-field text-field',
-         placeholder='Github nickname') }}
+         placeholder='GitHub nickname') }}
          <span class="form-input-validation check error"></span>
     </div>
 
```

```
→ nosetests
...........................................................................................................
Ran 107 tests in 1.352s

OK
```

Let's push this. I need to work on bugform.js after.

webcompat/form.py
karlcow: @miketaylr ok I added a couple of things here. I think the helptext is not working but I'm not sure why.

https://github.com/webcompat/webcompat.com/pull/2713/commits/3a261f8d6216b873001ec1e7c5a78f714fc595ab

Once this is right I can add a test on the functional side.

miketaylr: > Ah indeed my validation code is pretty much useless here. Given that we do not do anything with it server side, where we would check if the field was valid or not and have an action about it.

It used to be the case that server-only validation would work (before we added all the client side stuff). I'm not 100% sure if it does, because we don't really exercise that code path. But the theory is WTForm will catch the validation errors and Flask will re-render the page using the errors that WTForm gives it:

```
{% if form.contact.errors %}
  {% for error in form.contact.errors %}
  <small class="form-message-error">{{ error }}</small>
  {% endfor %}
{% endif %}
```

webcompat/templates/home-page/form.html
miketaylr: Not a blocker, I was just curious.

webcompat/static/js/lib/bugform.js
```diff
@@ -341,6 +351,26 @@ function BugForm() {
...
+    - no two consecutive hyphens
+  */
+  this.isValidGitHubUsername = function(contact) {
+    return this.githubRegexp.test(contact);
+  };
+
+  /* Check to see if the GitHub username has the right syntax. */
+  this.checkGitHubUsername = function() {
+    var contact = this.contactField.val();
+    if (this.isValidGitHubUsername(contact)) {
```
miketaylr: Right now, this will fail even if the input is empty, right?

miketaylr: My bad, I should have given a better example. We should treat this like `osField` -- which is by default valid, but we check validation once you edit it (or something like that).

https://github.com/webcompat/webcompat.com/blob/3a261f8d6216b873001ec1e7c5a78f714fc595ab/webcompat/static/js/lib/bugform.js#L344-L352

As-is, if you never put anything in the contactField, the valid state is `null`, or if you put in a name, but then delete it all to be the empty string, it will be made invalid. I think we want the intial state to be `valid: true`.

And then when we're validating, we only make it invalid if there is a value, and it doesn't pass the regexp.

webcompat/templates/home-page/form.html
magsout: @miketaylr @karlcow it is used for making space between differents form elements `margin: 0 0 20px`.

magsout: capture d ecran 2018-12-04 a 09 32 55

webcompat/static/js/lib/bugform.js
karlcow: done in https://github.com/karlcow/webcompat.com/commit/4080f289a17e8a561102a0ce403cc10bfb579ed6

karlcow:
two tests failed on chrome unrelated to my commit :o)
Let's try to kick it again.

karlcow:
@miketaylr let's see if this build is passing. 
and happy with a review when you have time.

karlcow:
ok it seems the test failing on chrome only and not on Firefox is usually the one just after the tests I added for the form. Weird.

karlcow:
once everything is running I'll rebase or clean up with a new branch.

miketaylr:
For that Chrome failure, perhaps https://stackoverflow.com/a/19763087/66348 is helpful. I've seen this in the past... but can't remember how we worked around it. My first trick is usually `.sleep(2000)` 🙈 ...

karlcow:
@miketaylr … That's interesting
I changed the tests order and labels… rebased a bit. And the failure is the same and the labels of the tests… are the same like if it was in cache on circleci.
https://circleci.com/gh/webcompat/webcompat.com/237

karlcow:
🤡 
Example:
https://github.com/webcompat/webcompat.com/blob/f43d8fe0c24d831400ba859fb244cbf6f1e026ee/tests/functional/reporting-non-auth.js#L321

vs 

the test key in the fail log.

> ✓ chrome 71.0.3578.98 on Linux - Reporting (non-auth) - GitHub contact name with a leading @ (0.668s)

which was a previous key.

karlcow:
hmmm

capture d ecran 2018-12-14 a 13 18 33

Ah! https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-cache

karlcow:
In this [example](https://circleci.com/docs/2.0/sample-config/#section=configuration), tests are in their specific group  at the same level than `build:` while in our `.circleci/config.yml` it is under `build:` not sure it is important or not.

karlcow:
I'm closing this pull Request as the action is now in #2744

PR #2731, opened 2018-12-10
Fixes #2727 - Removes unecessary use of markdown-it

### LINKED ISSUE #2727: Removes unnecessary use of markdown-it
The Github API already [offers a markdown version](https://developer.github.com/v3/media/#html) for the requests we're making, so it's not necessary to use markdown-it to process our requests.

However we still need it to format users comments, so we're still using it on  all of the pages where people can write comments. It might be worth it to look for a more specific lib to do this job.

marimeireles:
I've removed the library and now webcompat shows emojis exactly like Github, for example `:smile:` will show 😄 .

r? @miketaylr


marimeireles:
But I'm having some issues with CSP. When Github tries to make a request to download an emoji that is a picture, I'm getting blocked. For example the `octocat` :octocat: emoji will cause the following error:

`Content Security Policy: The page’s settings blocked the loading of a resource at https://assets-cdn.github.com/images/icons/emoji/octocat.png (“img-src”).`

So if I understood correctly we have to update our Content-Security-Policy HTTP header adding the following address: https://assets-cdn.github.com/images/icons/emoji

marimeireles:
Sorry to ping you guys, I'm just running in too many issues with the tests & I'll keep trying tomorrow.
The tests are running on my computer but failing on CI, I'll also have to figure it out tomorrow.

marimeireles:
This PR depends on the approval of [PR#2739](https://github.com/webcompat/webcompat.com/pull/2739). Once that one is approved and a new PR fixing the fixtures be approved, this one should work.

miketaylr:
> This PR depends on the approval of [PR#2739](https://github.com/webcompat/webcompat.com/pull/2739). Once that one is approved and a new PR fixing the fixtures be approved, this one should work.

Can we just commit the fixtures generated by that tool here? Then we're not depending on it.

marimeireles:
@miketaylr 

Good idea. I'll try to make it work!

marimeireles:
>To note that #2739 will change as it will move to tools/ as discussed in the PR.

Yes, I'm aware of that. Even so, it wouldn't matter for this PR.
I talked to Mike yesterday and we got to the conclusion that this new tool will only be used to fetch new issues. As you said, some issues were fixed manually to work with the current test mode, so if we fetch them the way they're on `webcompat/webcompat-tests/issues` they're not gonna work. (I tried)

marimeireles:
@miketaylr could you tell me where this fixture come from?
https://github.com/webcompat/webcompat.com/blob/master/tests/fixtures/api/issues/15000.json

I couldn't find it using search.
Thanks!

karlcow:
@marimeireles 

> @miketaylr could you tell me where this fixture comes from?
> https://github.com/webcompat/webcompat.com/blob/master/tests/fixtures/api/issues/15000.json
> 
> I couldn't find it using search.
> Thanks!

So I made it. Let's [look at the GitHub blame](https://github.com/webcompat/webcompat.com/blame/master/tests/fixtures/api/issues/15000.json) to see who was the first author and which issue it belonged to. And it was for the issue #1828 

Are you asking which original issue? If I go to the number in the fixture, I get [an issue](https://github.com/webcompat/web-bugs/issues/15000), but its json response is slightly different. It doesn't matter.

But as you can see I tweaked the description, so that it would fit our use case for the pull request.
```
    "description": "issues which seemed to have been webcompat issues for the bug reporter but are in fact working in most circumstances. For example, network errors.",
```

It's an example of our [discussion about mocking](https://github.com/webcompat/webcompat.com/pull/2739#discussion_r241608605) not being necessary real issue.

marimeireles:
You found it! :)
Thanks for the effort Karl.

>It's an example of our discussion about mocking not being necessary real issue.

Yes. You're right.

Thank you Karl.
I wanted to find the real issue so I could take the Github flavored markdown and convert it to html using the tool I made. But, as you said, it doesn't matter since some issues are completely different.
What I'm doing now is using [curl](https://developer.github.com/v3/markdown/) to fetch the html.

Sorry for the delay to finish this PR. What I'm doing now is dealing with a ton of test errors that are happening because I changed the way we're fetching the issues. But I think it's totally worth it to spend as much time as I need to figure this out because not having to use this `markdown-it` lib will mean a great improvement for performance, specially for mobile users.

11/84 tests aren't passing on firefox and 1/84 on chrome. So I think there are at least some hours ahead of me.

marimeireles:
Ok @miketaylr, this beauty is ready to be approved.

webcompat/static/js/lib/issues.js
```diff
@@ -547,8 +547,10 @@ issues.MainView = Backbone.View.extend(
       var target =
         e.target.nodeName === "IMG"
           ? e.target
-          : e.target.nodeName === "P" && e.target.firstElementChild;
+          : e.target.nodeName === "P" && e.target.getElementsByTagName("img");
```
miketaylr: Can you help me understand this change?

marimeireles: Yes!
When I fetch new images from github using this html format githubs gives me a response that contains an image nested on a link, like this:

```
<p class="issue-details-nsfw">
    <a target="_blank" rel="noopener noreferrer" href="https://cloud.githubusercontent.com/assets/67283/15969039/bd82fae2-2ef3-11e6-800f-535785681b50.jpg">
        <img src="https://cloud.githubusercontent.com/assets/67283/15969039/bd82fae2-2ef3-11e6-800f-535785681b50.jpg" alt="29upbiv" style="max-width:100%;">. 
    </a>
</p>
```

And the original code was expecting something like this:

```
<p class="issue-details-nsfw">
    <img src="https://cloud.githubusercontent.com/assets/67283/15969039/bd82fae2-2ef3-11e6-800f-535785681b50.jpg" alt="29upbiv" style="max-width:100%;">. 
</p>
```

And every time I clicked on the image instead of only toggling the the blur on/off it was also opening the image in a new tab because of the <a>. So... I though I could just deactivate it, since Github fetches this automatically. You think it's bad implementation choice to do that way? And have a link that doesn't go everywhere? 
I could try to remove it when the issue is being fetched, so it wouldn't even be part of the code... Even though I don't know how I'd do that. Maybe with regex?

tests/functional/search-non-auth.js
```diff
@@ -36,7 +36,7 @@ registerSuite("Search (non-auth)", {
       return FunctionalHelpers.openPage(
         this,
         url("/issues"),
-        "[data-remotename=browser-android"
+        "[data-remotename=browser-android]"
```
miketaylr: 🔍 weird.

marimeireles: It was wrong, right?

webcompat/helpers.py
```diff
@@ -36,7 +36,7 @@
                   '127.0.0.1', 'localhost')
 FIXTURES_PATH = os.getcwd() + '/tests/fixtures'
 STATIC_PATH = os.getcwd() + '/webcompat/static'
-JSON_MIME = 'application/json'
+JSON_MIME = 'application/vnd.github.v3.html+json'
```
miketaylr: I think we probably don't want to request `html+json` for all JSON requests here. Perhaps we can add a new `HTML_JSON_MIME` or something, and use that where we need it?

(unless you've already verified we actually do need this for all JSON requests)

webcompat/static/js/lib/issues.js
miketaylr: Hm. I'm not sure about what the right thing to do is here... I'll leave that up to you. :)

But we should use `querySelector` instead of `getElementsByTagName` for the reasons mentioned above.

And [using regex to parse & manipulate HTML is super painful](https://blog.codinghorror.com/parsing-html-the-cthulhu-way/). If we have to go that way, it would make more sense to manipulate the DOM when we render the body/comment, the way you already are doing.

It's maybe possible to just prevent the default action on that `<a>` and dispatch a click on the `, but I'm not sure it's worth the extra effort for such a corner case feature.

tests/functional/search-non-auth.js
miketaylr: (yeah, i'm mostly concerned that we didn't know until you saw it 😄 )

webcompat/helpers.py
miketaylr: Here's all the uses of `JSON_MIME` in this file:

https://github.com/webcompat/webcompat.com/blob/b2619d99a2ad85852ac47fa4c7d1fb24799efb18/webcompat/helpers.py#L190
https://github.com/webcompat/webcompat.com/blob/b2619d99a2ad85852ac47fa4c7d1fb24799efb18/webcompat/helpers.py#L203
https://github.com/webcompat/webcompat.com/blob/b2619d99a2ad85852ac47fa4c7d1fb24799efb18/webcompat/helpers.py#L373

So you're saying it's expected behavior to be requesting `vnd.github.v3.html+json` in all those places? Doesn't that mean we're sending that MIME on every `api_request`, for example? https://github.com/webcompat/webcompat.com/blob/b2619d99a2ad85852ac47fa4c7d1fb24799efb18/webcompat/helpers.py#L450

It's entirely possible it doesn't make a visible difference -- and the fact that you pass all tests is probably indicative of that. What I'm wondering is if we should only request `vnd.github.v3.html+json` when we actually want an html response.

miketaylr: @marimeireles maybe this *is* what we want, don't take my questions to mean you're doing something wrong :). I guess my question is, is there a situation where we aren't requesting HTML from GitHub, just plain JSON? Or is it all HTML now?

miketaylr: Here's one: 

screen shot 2018-12-21 at 9 52 34 am

Perhaps we need @karlcow to chime in on this. I'm not sure what the consequences are for this change (claiming `application/json` is `application/vnd.github.v3.html+json` when it isn't). I think we're breaking HTTP semantics by doing so.

marimeireles: >So you're saying it's expected behavior to be requesting vnd.github.v3.html+json in all those places?

Yes.

>I guess my question is, is there a situation where we aren't requesting HTML from GitHub, just plain JSON? Or is it all HTML now?

It's still a JSON, the same JSON we had before. I made only small changes on the JSONs to replace markdown for HTML. Here is an example of the kind of changes I did:

```
removed: "body": "\n**URL**: www.mmmmm.com\n**Browser**: Firefox\n**Version**: 30\n\nThe site asks me to upgrade",
added: "body_html": "<p><strong>URL</strong>: <a href=\"http://www.mmmmm.com\" rel=\"nofollow\">www.mmmmm.com</a><br> <strong>Browser</strong>: Firefox<br> <strong>Version</strong>: 30</p> <p>The site asks me to upgrade</p>",
```

The only part where we're still using markdown-it is for rendering the comments that the user writes on the website, because we need to send them as markdown to github and also because immediately after the user sends a comment we show the rendered comment on the webpage (without making any request to Github), so we need the lib to do that for us. 

But when we fetch the files from github we're also fetching the comments as HTML, everything we fetch from Github is HTML and IMO it's not necessary to use `application/json` anymore. If you think it's mandatory to use it somewhere because of some reason I didn't think of then I'd have to change the code because as far as I'm concerned the program will only work with JSONs with HTML from now on, instead of JSONs with markdown.

> maybe this is what we want, don't take my questions to mean you're doing something wrong :)

Okay :)

marimeireles: >Perhaps we need @karlcow to chime in on this. I'm not sure what the consequences are for this change (claiming application/json is application/vnd.github.v3.html+json when it isn't). I think we're breaking HTTP semantics by doing so.

Me neither... We can wait for Karl, then. :)
According to [this](https://stackoverflow.com/a/20620606/10756713) answer on stackoverflow it doesn't change anything.
If we want to make things right... I think I'd just have to find how the `/webcompat/static/js/dist/issues.js` is generated and change the part that generates this:

```
    $.ajax({
      contentType: "application/json",
      data: JSON.stringify({
        milestone: statusObject.id,
        state: statusObject.state
      }),
      type: "PATCH",
      url: "/api/issues/" + $("main").data("issueNumber") + "/edit",
      success: _.bind(function() {
        var currentMilestone = _.find(this.get("milestones"), function(status) {
          return status.id === statusObject.id;
        });

        this.set("milestone", currentMilestone);
        this.issueModel.set("milestone", currentMilestone.name);
      }, this),
```

@karlcow am I in the right direction here?

webcompat/static/js/lib/issues.js
```diff
@@ -547,8 +547,10 @@ issues.MainView = Backbone.View.extend(
       var target =
         e.target.nodeName === "IMG"
           ? e.target
-          : e.target.nodeName === "P" && e.target.firstElementChild;
+          : e.target.nodeName === "P" && e.target.querySelectorAll("img");
```
marimeireles: @miketaylr is that what you expected?

miketaylr: @magsout almost! we should use [`querySelector`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector), not [`querySelectorAll`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll). The former returns an Element, and the latter returns a NodeList.

marimeireles: okie, sry for that. I think it's fixed now.
I did something wrong the first time I tested with querySelector and it wanst working that's why I tried with querySelectorAll.

webcompat/helpers.py
miketaylr: > I think I'd just have to find how the `/webcompat/static/js/dist/issues.js` is generated

See https://github.com/webcompat/webcompat.com/blob/master/grunt-tasks/concat.js#L39

> and change the part that generates this

Not sure what you're thinking, but yes, milestone API requests and label API requests should be sending a non-HTML mimetype. I was thinking we could extend `api_request` to take a mime override argument for the cases where we are requesting non-`application/json`.

marimeireles: > milestone API requests and label API requests should be sending a non-HTML mimetype

Why do the milestones and labels can't request for a HTML?
The JSON file looks exactly the same, except for the markdown parts.
The only issue that I see is the one pointed [here](https://github.com/webcompat/webcompat.com/pull/2731#discussion_r243620137) about the content-type being wrong.
Sorry for insisting, but I really didn't understand.

> I was thinking we could extend api_request to take a mime override argument for the cases where we are requesting non-application/json.

Seems like a good solution, if we'll have to fetch application/json.

miketaylr: Here's an example of requesting the labels as `html+json`:

```
🐓 http --print=hH GET https://api.github.com/repos/webcompat/web-bugs/labels **Accept:'application/vnd.github.v3.html+json'**
GET /repos/webcompat/web-bugs/labels HTTP/1.1
Accept: application/vnd.github.v3.html+json
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: api.github.com
User-Agent: HTTPie/1.0.0

HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type
Cache-Control: public, max-age=60, s-maxage=60
Content-Encoding: gzip
Content-Security-Policy: default-src 'none'
**Content-Type: application/json; charset=utf-8**
Date: Fri, 21 Dec 2018 21:35:28 GMT
ETag: W/"0c4c97b020aa6958117b5d8f3cad9bf6"
Link: <https://api.github.com/repositories/17914657/labels?page=2>; rel="next", <https://api.github.com/repositories/17914657/labels?page=5>; rel="last"
Referrer-Policy: origin-when-cross-origin, strict-origin-when-cross-origin
Server: GitHub.com
Status: 200 OK
Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
Transfer-Encoding: chunked
Vary: Accept
X-Frame-Options: deny
X-GitHub-Media-Type: github.v3; param=html; format=json
X-GitHub-Request-Id: CD59:4747:1C34DB6:476FBBA:5C1D5C9F
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 46
X-RateLimit-Reset: 1545431394
```

We requested `...html+json`, but GitHub just sent us back `application/json`, presumably because today that's all they support. I think it's problematic for the server to claim the content type is `...html+json`, because it isn't.

> Why do the milestones and labels can't request for a HTML?

They can, but a follow-up question is *why* would we request `...html+json` for labels or milestones? If one day GitHub sent back an actual HTML+JSON response, and we expect just plain JSON, that would probably break parts of our app. The fact that today they return `application/json` today is useful, but we can't rely on that behavior, especially if we're explicitly requesting a different content type

I'm also concerned about breaking HTTP semantics by **claiming** we're returning `application/vnd.github.v3.html+json` in all API requests, when in fact we're not -- we've just hard-coded that value. I suppose we've just been lucky that we've only requested `application/json` in the past, so hardcoding wasn't an obvious issue.

(and yes, this feels very theoretical and who cares, etc., but I think it's worth doing it right because it might lead to bugs or confusion in the future, in ways we can't predict 😸 ).

I think a path forward here is to be explicit about what we're requesting (sensible defaults make sense, with a way to override), and for responses proxy the **actual** `Content-Type` header value returned from GitHub.

marimeireles: >but we can't rely on that behavior, especially if we're explicitly requesting a different content type

I see your point and I totally agree.

> (and yes, this feels very theoretical and who cares, etc., but I think it's worth doing it right because it might lead to bugs or confusion in the future, in ways we can't predict 😸 ).

I think it's super important to write a good code so we don't struggle with it in the future. We'll probably save time doing thins the right way. :)

marimeireles: Thank you a lot for the discussion guys.
I'll start implementing these changes.

tests/fixtures/api/issues/252.json
```diff
@@ -76,7 +76,7 @@
   "created_at": "2015-03-09T22:01:56Z",
   "updated_at": "2015-06-19T22:10:58Z",
   "closed_at": null,
-  "body": "<!-- @browser: Firefox 38.0 -->\n<!-- @ua_header: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:38.0) Gecko/20100101 Firefox/38.0 -->\n\n**URL**: www.miketaylr.com\n**Browser / Version**: Firefox 38.0\n**Operating System**: Mac OS X 10.10\n**Problem type**: Layout is messed up\n\n**Steps to Reproduce**\n1) Navigate to: www.miketaylr.com\r\n2) …\r\n\r\nExpected Behavior:\r\nActual Behavior:\r\n\r\nNOT A GREAT SITE OK",
+  "body_html": "\nURL</strong>: <a href=\"http://www.miketaylr.com\" rel=\"nofollow\">www.miketaylr.com</a><br>\n<strong>Browser / Version</strong>: Firefox 38.0<br>\n<strong>Operating System</strong>: Mac OS X 10.10<br>\n<strong>Problem type</strong>: Layout is messed up</p>\n<p><strong>Steps to Reproduce</strong></p>\n<ol>\n<li>Navigate to: <a href=\"http://www.miketaylr.com\" rel=\"nofollow\">www.miketaylr.com</a></li>\n<li>…</li>\n</ol>\n<p>Expected Behavior:<br>\nActual Behavior:</p>\n<p>NOT A GREAT SITE OK</p>",
```
karlcow: Is this test fixture obsolete, used somewhere?

tests/fixtures/api/issues/252/comments.fafc13ba597647ded279ea1b663135a6.json
```diff
@@ -26,7 +26,7 @@
     },
     "created_at": "2015-03-10T14:49:52Z",
     "updated_at": "2015-03-10T14:49:52Z",
-    "body": "* one\r\n* two\r\n* three"
+    "body_html": "<ul>\n<li>one</li>\n<li>two</li>\n<li>three</li>\n</ul>"
```
karlcow: this doesn't seem to be used anywhere

tests/fixtures/api/issues/396.json
```diff
@@ -77,6 +77,6 @@
...
...9eWlzZWFyY2g7Zmk9Zml0O2dlPTAwNjYwMDtncz0wMEEzMDA7aD00MDA7dz03MTE-/http://www.educadictos.com/wp-content/uploads/2013/06/color.png.cf.png</a></li>\n<li>…</li>\n</ol>\n<p>Expected Behavior:</p>\n<p>Actual Behavior:</p>\n<p><a target=\"_blank\" rel=\"noopener noreferrer\" href=\"https://camo.githubusercontent.com/4ccd0c8db5142e00518b1d3be82d2fee5b923d10/687474703a2f2f73746167696e672e776562636f6d7061742e636f6d2f75706c6f6164732f323031362f352f30326637373230622d303030662d343966642d396330632d3265363761383433623430352e6a706567\"><img src=\"https://camo.githubusercontent.com/4ccd0c8db5142e00518b1d3be82d2fee5b923d10/687474703a2f2f73746167696e672e776562636f6d7061742e636f6d2f75706c6f6164732f323031362f352f30326637373230622d303030662d343966642d396330632d3265363761383433623430352e6a706567\" alt=\"Screenshot Description\" data-canonical-src=\"http://staging.webcompat.com/uploads/2016/5/02f7720b-000f-49fd-9c0c-2e67a843b405.jpeg\" style=\"max-width:100%;\"></a></p>\n<p><em>From <a href=\"https://webcompat.com/\" rel=\"nofollow\">webcompat.com</a> with <g-emoji class=\"g-emoji\" alias=\"heart\" fallback-src=\"https://github.githubassets.com/images/icons/emoji/unicode/2764.png\">❤️</g-emoji></em></p>",
```
karlcow: yoohoo this is used.

tests/fixtures/api/issues/70.json
```diff
@@ -82,7 +82,7 @@
   "created_at": "2014-06-09T22:43:30Z",
   "updated_at": "2015-07-30T18:59:10Z",
   "closed_at": null,
-  "body": "**URL**: www.💩.com\n**Browser**: Firefox\n**Version**: 30\n\nThe site asks me to upgrade\n",
+  "body_html": "<p><strong>URL</strong>: <a href=\"http://www.%F0%9F%92%A9.com\" rel=\"nofollow\">www.<g-emoji class=\"g-emoji\" alias=\"hankey\" fallback-src=\"https://github.githubassets.com/images/icons/emoji/unicode/1f4a9.png\">💩</g-emoji>.com</a><br>\n</strong>: Firefox<br>\n<strong>Version</strong>: 30</p>\n<p>The site asks me to upgrade</p>",
```
karlcow: not used for body testing.

tests/fixtures/api/issues/70/edit.json
```diff
@@ -52,7 +52,7 @@
   "created_at": "2014-06-09T22:43:30Z",
   "updated_at": "2017-06-28T23:03:43Z",
   "closed_at": "2017-06-28T23:03:43Z",
-  "body": "**URL**: www.💩.com\n**Browser**: Firefox\n**Version**: 30\n\nThe site asks me to upgrade\n",
+  "body_html": "<p><strong>URL</strong>: <a href=\"http://www.%F0%9F%92%A9.com\" rel=\"nofollow\">www.<g-emoji class=\"g-emoji\" alias=\"hankey\" fallback-src=\"https://github.githubassets.com/images/icons/emoji/unicode/1f4a9.png\">💩</g-emoji>.com</a><br>\n</strong>: Firefox<br>\n<strong>Version</strong>: 30</p>\n<p>The site asks me to upgrade</p>",
```
karlcow: same. not used for body testing.

tests/fixtures/api/issues/9.json
```diff
@@ -56,7 +56,7 @@
   "updated_at": "2017-10-31T15:25:08Z",
   "closed_at": null,
   "author_association": "OWNER",
-  "body": "**URL**: google.com\n**Browser**: chrome\n**Version**: 88\n**Problem type**: Unknown\n**Site owner**: No\n\n_Steps to Reproduce_\n1. Do this\n2. Do that\n",
+  "body_html": "<p><strong>URL</strong>: google.com<br>\n<strong>Browser</strong>: chrome<br>\n<strong>Version</strong>: 88<br>\n<strong>Problem type</strong>: Unknown<br>\n<strong>Site owner</strong>: No</p>\n<p><em>Steps to Reproduce</em></p>\n<ol>\n<li>Do this</li>\n<li>Do that</li>\n</ol>",
```
karlcow: not used for body testing?

tests/unit/test_api_urls.py
```diff
@@ -94,7 +94,8 @@ def test_api_labels_without_auth(self):
                 {'status_code': 200, 'content': '[]'})
             rv = self.app.get('/api/issues/labels', environ_base=headers)
             self.assertEqual(rv.status_code, 200)
-            self.assertEqual(rv.content_type, 'application/json')
+            self.assertEqual(
+                rv.content_type, 'application/vnd.github.v3.html+json')
```
karlcow: as we are testing the label json only

webcompat/static/js/lib/models/comment.js
```diff
@@ -23,14 +12,14 @@ issues.Comment = Backbone.Model.extend({
...
-      body: md.render(response.body),
+      body: response.body_html,
       commenter: response.user.login,
       commentLinkId: "issuecomment-" + response.id,
       createdAt:
         moment(Date.now()).diff(response.created_at, "weeks") > 1
           ? moment(response.created_at).format("YYYY-MM-DD")
           : moment(response.created_at).fromNow(),
-      rawBody: response.body
+      rawBody: response.body_html
```
karlcow: do we still need `rawBody`?
Because now we have two keys carrying the same content.
better to call now body, anywhere else in the code.

webcompat/static/js/lib/models/issue.js
```diff
@@ -83,7 +72,7 @@ issues.Issue = Backbone.Model.extend({
       state: response.state,
       title: this.getTitle(
         this.getDomain(response.title),
-        this.getDescription(response.body),
+        this.getDescription(response.body_html),
```
karlcow: This doesn't seem necessary anymore. It was necessary because we wanted the raw version and not the `md.render` one.

`this.getDescription(body)` should work.

webcompat/static/js/lib/models/issue.js
```diff
@@ -102,7 +91,7 @@ issues.Issue = Backbone.Model.extend({
 
   getDescription: function(body) {
     // Get the Description of the body
-    var regex = /\*\*Description\*\*: ([^*]+)/;
+    var regex = /<strong>Description<\/strong>: (.+?(?=<br>))/;
```
karlcow: This gives me butterfly. Regex and html are not good friends. :) Probably better to parse it as an html fragment and then extract what we want.

webcompat/templates/issue.html
```diff
@@ -45,6 +45,9 @@
 {%- if config.PRODUCTION or config.STAGING -%}
 <script src="{{ url_for('static', filename='js/dist/issues.min.js')|bust_cache }}"></script>
 {% else %}
+<script src="{{ url_for('static', filename='js/vendor/markdown-it.js') }}"></script>
+<script src="{{ url_for('static', filename='js/vendor/markdown-it-emoji-1.4.0.js') }}"></script>
+<script src="{{ url_for('static', filename='js/vendor/markdown-it-sanitizer-0.4.1.js') }}"></script>
```
karlcow: Do we still need them?

webcompat/static/js/lib/models/issue.js
marimeireles: okay :)
I also think it's a better idea. thanks for the tip

webcompat/templates/issue.html
marimeireles: Unfortunately yes.
When a user types a comment we need to show it formatted correctly right after he sent the comment, so we need to format the comment ourselves.

Hmm... Or maybe, now that I'm thinking about it, we could make a request to Github right after the user post their comment, maybe it'd be too slow though. What do you think?

webcompat/static/js/lib/models/issue.js
marimeireles: Sorry, I didn't understand what you mean.
We need the `body_html` because is the name of the attribute. 
I tried to replace for `this.getDescription(body)` but it didn't work.

webcompat/static/js/lib/models/comment.js
marimeireles: You're right, they're the same. I'll remove the rawBody thing.
Thanks for noticing that Karl :)

webcompat/static/js/lib/models/issue.js
```diff
@@ -91,9 +91,14 @@ issues.Issue = Backbone.Model.extend({
...
-    return description != null ? description[1].slice(0, 74) : null;
+    var el = document.createElement("html");
+    el.innerHTML = body;
+    el = el.getElementsByTagName("strong");
+    if (el[5]) {
+      var description = el[5].nextSibling.textContent;
+      description = description.replace(": ", "");
+      return description;
+    }
   },
```
marimeireles: Is this more or less what you were expecting @karlcow ?
I had to add the `if` because some mock outs don't have the description attribute and they were breaking the tests.

I don't know how the tests are built... But I don't think we should have so much disparity between tests, it always cause me this kind of problem, and IMO we shouldn't have to add code to the website to deal with things that aren't going to happen in the "real life" for the users.

tests/unit/test_api_urls.py
marimeireles: > So I hope I'm wrong, but I guess many fixtures are not worth fixing. 

@karlcow If we don't leave as `body_html` we won't be able to visualize the issue. If we go to webcompat.com/issues/100 it will show a blank page... So that's why I changed all of the files.
Why do you think we should keep them in the old way?

webcompat/templates/issue.html
karlcow: ah understood. ok let's keep it for now.

webcompat/static/js/lib/models/issue.js
karlcow: >  I had to add the if because some mock outs don't have the description attribute and they were breaking the tests.

and 

> I don't know how the tests are built... But I don't think we should have so much disparity between tests, it always cause me this kind of problem, and IMO we shouldn't have to add code to the website to deal with things that aren't going to happen in the "real life" for the users.

Agreed but check this :)
https://webcompat.com/issues/1000

we didn't always have a description attribute. 
The description appears with [issue 8073](https://webcompat.com/issues/8073) which was created on 2017-07-07.

Let's [check the changelog](https://github.com/webcompat/webcompat.com/blob/master/CHANGELOG.md#530---2017-07-07). It was changed with PR #1578 fixing issue #1547

tests/unit/test_api_urls.py
karlcow: this is a request to `'/api/issues/labels'` so not related to body_html. Probably this comment is a reply to something else. :)
The request to /api/issues/labels should definitely stays as `'application/json'` see the other discussion about mime type. :)

webcompat/static/js/lib/models/issue.js
karlcow: yup. I'm just saying it's already defined just above in the code.

https://github.com/webcompat/webcompat.com/blob/08336441c3c00a2c3189f629e666f4b8ab8b9a6c/webcompat/static/js/lib/models/issue.js#L60-L78

or to make it more compact.

```js
    this.set({
      body: response.body_html,

// removed lines for readability

      title: this.getTitle(
        this.getDomain(response.title),
// body has been assigned above already.
        this.getDescription(body),
        response.title
      )
});
```

That said I'm not a JS expert, and maybe it's not good to do this :)

webcompat/static/js/lib/models/issue.js
marimeireles: But now that we have the attribute, shouldn't we include it on the mock outs? So they're more alike with the actual website?

karlcow: If we had unit test cases for the JS functions :) we would indeed cover it. I don't remember we have functional test cases for checking the title. If we have one, we should have the two cases covered. :)

webcompat/helpers.py
marimeireles: I was reading more about the use of html+json and I learned that "The suffix "+json" MAY be used with any media type whose representation follows that established for "application/json"." [according](https://tools.ietf.org/html/rfc6839#section-3.1) to IETF, in other words a document with `vnd.github.v3.html+json` MIME type is an `application/json` document. 

Besides that Github says that every time there is a comment when we make this  `application/vnd.github.VERSION.html+json` request it will return the HTML version, but that will *only* happen for comments. Github wouldn't start sending HTML when we request a label because it'd mean that they're breaking their API. The version 3 that we use is stable.

So I think the ideal thing would be to always return `application/json` in our responses since `vnd.github.v3.html+json` is a Github vendor MIME type that isn't meaningful to our front end.

marimeireles: Okay Karl, you convinced me with you wisdom. Thank you and Mike for being so patient, you're super nice. :) 

I'm sorry for insisting so much. I've wrote a patch, gonna ping you and Mike there to check if everything is right.

tests/fixtures/api/issues/100.json
```diff
@@ -76,7 +76,7 @@
   "created_at": "2014-06-16T19:12:15Z",
   "updated_at": "2015-10-12T17:20:43Z",
   "closed_at": null,
-  "body": "\n**URL**: www.mmmmm.com\n**Browser**: Firefox\n**Version**: 30\n\nThe site asks me to upgrade",
+  "body_html": "<p><strong>URL</strong>: <a href=\"http://www.mmmmm.com\" rel=\"nofollow\">www.mmmmm.com</a><br> <strong>Browser</strong>: Firefox<br> <strong>Version</strong>: 30</p> <p>The site asks me to upgrade</p>",
```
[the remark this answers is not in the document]
marimeireles: Just to make sure we agreed on maintaining the `body_html` instead of the `body`, right?
Because we wouldn't be able to see the content rendered otherwise.

tests/fixtures/api/issues/101.json
```diff
@@ -76,7 +76,7 @@
   "created_at": "2014-06-16T19:16:03Z",
   "updated_at": "2015-09-29T09:33:15Z",
   "closed_at": "2015-09-29T09:33:15Z",
-  "body": "\n**URL**: www.mmmmm.com\n**Browser**: Firefox\n**Version**: 30\n\nThe site asks me to upgrade",
+  "body_html": "<p><strong>URL</strong>: <a href=\"http://www.mmmmm.com\" rel=\"nofollow\">www.mmmmm.com</a><br>\n<strong>Browser</strong>: Firefox<br>\n<strong>Version</strong>: 30</p>\n<p>The site asks me to upgrade</p>",
```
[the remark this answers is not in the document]
miketaylr: I don't see it being used anywhere. It could be deleted.

webcompat/static/js/lib/models/issue.js
```diff
@@ -102,9 +91,14 @@ issues.Issue = Backbone.Model.extend({
 
   getDescription: function(body) {
     // Get the Description of the body
-    var regex = /\*\*Description\*\*: ([^*]+)/;
-    var description = regex.exec(body);
-    return description != null ? description[1].slice(0, 74) : null;
+    var el = document.createElement("html");
```
miketaylr: @marimeireles can you explain what is happening here? It seems odd to create an `html` element.

webcompat/static/js/lib/models/issue.js
```diff
@@ -102,9 +91,14 @@ issues.Issue = Backbone.Model.extend({
...
+    el.innerHTML = body;
+    el = el.getElementsByTagName("strong");
+    if (el[5]) {
```
miketaylr: This seems fragile -- what if the DOM structure changes? 

What is the 6th `strong` element? Some comments in the code would probably help me understand (and the next person who has to deal with this). Can we get the description with a CSS selector and `querySelector` rather than this kind of DOM traversal? It would also fail if the DOM changed, but it seems easier to understand. 

Another option would be to use the jQuery contains selector: https://api.jquery.com/contains-selector/

marimeireles: Yeah.. I figured it wasn't the best way. :p
What do you think of this new solution?
I thought it was too simple to write a comment explaining it, but I might be wrong.
Thanks for the tip :) I don't know much about jquery.

miketaylr:
>  you think it's better?

The selector bit looks fine, I don't know if it actually works though? You're passing in `body` -- is that just a string? I assume you've verified the changes work as expected?

unrelated(?):

`FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory` that's a new error for me, node ran out of memory?

marimeireles:
Ahm... Weird. I didn't get this error when I was running on my computer.
It worked as expected.
I don't remember if `body` is just a string. Body is the response of a request made a few lines above.
I'm AFK now, but I'll try to reproduce it when I get back. Did you do anything special or just ran the program normally?

miketaylr:
> Did you do anything special or just ran the program normally?

I'm just looking at the test failures on CircleCI

karlcow:
@marimeireles 

>  But I just found this error. The list of issues wasn't showing the bug titles correctly because it was still a markdown file (it was showing the title attribute instead of the title+description). So I had to update them to use the html version. I hope it's not a problem (if it is, it's okay, I can change it back to the markdown version. I've updated them without asking you first because I thought they were related to a bug I was trying to fix, but in fact they're not, so the code won't crash if it stays the old way, but it will be wrong).

no issue! 😄 You are doing good.

miketaylr:
>  you think it's better?

LGTM. once the tests are green, i'm happy with your work here (for the JS bits!)

webcompat/helpers.py
```diff
@@ -447,7 +447,7 @@ def proxy_request(method, path, params=None, headers=None, data=None):
     return req(resource_uri, data=data, params=params, headers=auth_headers)
 
 
-def api_request(method, path, params=None, data=None):
+def api_request(method, path, params=None, data=None, mime_type=JSON_MIME):
```
karlcow: where is `JSON_MIME` defined now?

marimeireles: JSON_MIME is defined at the top of this file, as it was before. 
When it's necessary I pass `JSON_MIME_HTML`as an argument to `mime_type`, like I do [here](https://github.com/webcompat/webcompat.com/pull/2731/files#diff-b9ca02974fd8e51b2c21afc1f20d62b5). Do you think I should write a comment on the code, explaining why do we use two different kind of mimes?

marimeireles:
> I'm just looking at the test failures on CircleCI

Yeah... It's weird.
All of the tests are running normally on my computer. I get all green and the results are as expected.

I think this
screen shot 2019-01-07 at 23 34 41
means CI broke on the commit #2727 acd17c0, right? Maybe it's because it's a huge commit and CI ran out of memory. 
I think we could just skip it and see if it works. But I don't think I have permission to do that 
@miketaylr or @karlcow, what do you think?

webcompat/helpers.py
karlcow: ah I missed it. Apologizes. thanks. No need for comments at this point.

tests/fixtures/api/issues/100.json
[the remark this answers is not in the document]
karlcow: yes

tests/fixtures/api/issues/101.json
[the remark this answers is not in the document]
karlcow: ok to delete.

tests/fixtures/api/issues/101/comments.fafc13ba597647ded279ea1b663135a6.json
```diff
@@ -26,7 +26,7 @@
     },
     "created_at": "2014-06-16T19:16:03Z",
     "updated_at": "2014-06-16T19:16:03Z",
-    "body": "1"
+    "body_html": "<p>1</p>"
```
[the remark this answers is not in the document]
karlcow: to delete

webcompat/api/endpoints.py
```diff
@@ -214,4 +217,4 @@ def get_repo_labels():
     """XHR endpoint to get all possible labels in a repo."""
     params = request.args.copy()
     path = 'repos/{0}/labels'.format(REPO_PATH)
-    return api_request('get', path, params=params)
+    return api_request('get', path, params=params, mime_type=JSON_MIME_HTML)
```
karlcow: this one doesn't seem to be about issues. ;)

karlcow:
@marimeireles  coragem! we are almost there. You do good.

webcompat/api/endpoints.py
marimeireles: This was wrong. I missed it.
Thanks Karl!

marimeireles:
Merci @karlcow !
Could you or @miketaylr help me with this circleci thing?

According to the 🌈internet🌈 it's possible to fix this bug allocating more memory to Nodejs using the `—- max-old-space-size` command [[1]](https://medium.com/@vuongtran/how-to-solve-process-out-of-memory-in-node-js-5f0de8f8464c)[[2]](https://stackoverflow.com/questions/54057671/node-10-15-0-unable-to-create-a-large-matrix).
The default memory of Node is 512 mb but I don't know how much we should extend it to work because I can't understand the circleci output.

I believe we'd have to do something like this?
`node --max-old-space-size=XXXX ./tests/functional/_intern.js "--reporters=runner" "--firefoxBinary=/usr/local/bin/firefox"`

I don't know why this error is happening, though. I believe it's just because the commit is big(?).

miketaylr:
> I believe we'd have to do something like this?

@marimeireles Yeah, that LGTM. Let's hit it with a big `--max-old-space-size=8192` hammer and see what happens.

>  I believe it's just because the commit is big(?).

The size of the commit shouldn't matter, it seems like intern is running out of memory trying to run the functional tests. So either there's a memory leak, or circleci is running in a very constrained environment. Let's see if bumping the memory helps!

marimeireles:
>it seems like intern is running out of memory trying to run the functional tests

Is that something we should worry about?

miketaylr:
> @miketaylr Is that something we should worry about?

heh, i don't know. if the intern tests run locally, we're probably fine. if the tests start to intermittently fail, we'll at least know where to start investigating. 

It's at least the documented suggestion by circlci support, so that's good: https://support.circleci.com/hc/en-us/articles/360009208393-How-can-I-increase-the-max-memory-for-Node-



> Also, I thought I couldn't do that, but to add that flag I just have to update the `config.yml`on the circleci folder? Sry if it's a silly question. I just -don't want to break anything.

Not a silly question.

in the config.yml, we invoke intern via 
```
npm run test:js -- --reporters="runner" --firefoxBinary=`which firefox`
```

you should be able to replace that with:

```
node --max-old-space-size=8192 ./tests/functional/_intern.js --reporters=runner" --firefoxBinary=`which firefox`
```

marimeireles:
>It's at least the documented suggestion by circlci support, so that's good:

Ok. Good to know.
Wish me luck. haha

miketaylr:
🤞 🍀

marimeireles:
I'm so happy :D
Okay @karlcow, it's on you.

tests/fixtures/api/issues/101.json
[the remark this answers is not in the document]
karlcow: I opened #2755 for this

tests/fixtures/api/issues/101/comments.fafc13ba597647ded279ea1b663135a6.json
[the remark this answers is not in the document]
karlcow: I opened #2755 for this.

PR #2737, opened 2018-12-13
Makes the right number of needstriage requests on webcompat webpage

marimeireles:
Github will fetch 30 issues by default as shown [here](https://api.github.com/repos/marimeireles/webcompat-tests/issues). Adding ` + '?per_page=5'` to the API request fixes it.


marimeireles:
I think you're right. Makes sense.

marimeireles:
When I changed the `untriaged.js` file it generated a new hash to make the requisition on test mode I had to create a new file on `tests/fixtures/api/category` to send the issues when requested.

miketaylr:
> @marimeireles @miketaylr What do you think?

Sorry for the late response, but yes, I agree!

PR #2739, opened 2018-12-13
Fixes #2738 - Adds a new mode to fetch fixtures

### LINKED ISSUE #2738: Create an Update mock outs mode that fetches new mock outs when needed
While working on #2727 I had to make a change on the fixture files.
Instead of changing these files manually the point here is to make a script that runs it automatically.

The goal of this issue is to:
* Create a flag on the terminal that allows the user to enter in this mode
* Fetch the new mock outs and update the old files

marimeireles:
To begin the mode that will download the new mock outs simply type:

`start:test:update-mock-outs` 

on your terminal. This mode will both fetch the new fixture files and enter into test mode.

r? @karlcow

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "prestart": "npm run build",
     "start": "source env/bin/activate || . env/bin/activate && python run.py",
     "start:test": "npm run build && source env/bin/activate || . env/bin/activate && python run.py -t",
+    "start:test:update-mock-outs": "npm run build && source env/bin/activate || . env/bin/activate && python run.py -tu",
     "virtualenv": "pip2 install virtualenv && virtualenv -p python2.7 env && source env/bin/activate || . env/bin/activate && npm run pip",
     "pip": "pip2 install -r config/requirements.txt",
     "config": "cp config/secrets.py.example config/secrets.py",
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -388,6 +388,7 @@ def mockable_response(func):
     """
     @wraps(func)
     def wrapped_func(*args, **kwargs):
+        import json
         if app.config['TESTING']:
             get_args = request.args.copy()
             if get_args:
@@ -397,7 +398,18 @@ def mockable_response(func):
                 file_path = FIXTURES_PATH + request.path + "." + checksum
             else:
                 file_path = FIXTURES_PATH + request.path
-            if not os.path.exists(file_path + '.json'):
+            if app.config['UPDATE_MOCK_OUTS']:
+                data = func(*args, **kwargs)
+                with open(file_path + '.json', 'w') as f:
+                    data_json = json.loads(data[0])
+                    if isinstance(data_json, str):
+                        data_json[0]["_fixture"] = True
+                    if isinstance(data_json, dict):
+                        data_json["_fixture"] = True
+                    data_json = json.dumps(data_json, indent=2)
+                    f.write(data_json)
+                return (data_json, 200, get_fixture_headers(data[0]))
+            elif not os.path.exists(file_path + '.json'):
                 print('Expected fixture file: ' + file_path + '.json')
                 return ('', 404)
             else:
--- a/run.py
+++ b/run.py
@@ -102,6 +102,8 @@ if __name__ == '__main__':
                         help='Run server in "test mode".')
     parser.add_argument('-l', '--listen', default='localhost',
                         help='Interface to listen on.')
+    parser.add_argument('-u', '--updatemockouts', action='store_true',
+                        help='Update test\'s mock outs.')
     args = parser.parse_args()
 
     if check_pip_deps():
@@ -118,5 +120,9 @@ if __name__ == '__main__':
             app.config['SESSION_COOKIE_HTTPONLY'] = False
             app.config['SESSION_COOKIE_SAMESITE'] = None
             app.config['TESTING'] = True
+            app.config['UPDATE_MOCK_OUTS'] = False
             print("Starting server in ~*TEST MODE*~")
+            if args.updatemockouts:
+                app.config['UPDATE_MOCK_OUTS'] = True
+                print('Remember: you have to access each one of the issues you want to fetch.')
             app.run(host=args.listen)


marimeireles:
Hey @karlcow I don't know if this is the best way to implement it, so, as always, if you have any advice on how to make it better, please share.

I'm having trouble with ci, but I have no idea why, when I run the tests on my computer it's okay, I pass in all of the tests, but on ci:
```
ERROR: Request to /dashboard/triage should be 200.
  File "/home/circleci/repo/tests/unit/test_urls.py", line 228, in test_dashboard_triage
    rv = self.app.get('/dashboard/triage')
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/werkzeug/test.py", line 830, in get
    return self.open(*args, **kw)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/testing.py", line 200, in open
    follow_redirects=follow_redirects
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/werkzeug/test.py", line 803, in open
    response = self.run_wsgi_app(environ, buffered=buffered)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/werkzeug/test.py", line 716, in run_wsgi_app
    rv = run_wsgi_app(self.application, environ, buffered=buffered)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/werkzeug/test.py", line 923, in run_wsgi_app
    app_rv = app(environ, start_response)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/circleci/repo/webcompat/views.py", line 458, in dashboard_triage
    needstriage_issues = get_milestone_list('needstriage', params)
  File "/home/circleci/repo/webcompat/helpers.py", line 567, in get_milestone_list
    milestone_name, other_params=params)
  File "/home/circleci/repo/webcompat/helpers.py", line 401, in wrapped_func
    if app.config['UPDATE_MOCK_OUTS']:
KeyError: 'UPDATE_MOCK_OUTS'
```
I don't understand how `KeyError: 'UPDATE_MOCK_OUTS'` could be related to `ERROR: Request to /dashboard/triage should be 200.` When I make a request the program enters on the following else:

https://github.com/webcompat/webcompat.com/blob/208d66c02c6c115368d6f7c620f6cbc73b662bf6/webcompat/helpers.py#L403-L406

That returns 200. Does that mean there's something wrong with the way I'm declaring `UPDATE_MOCK_OUTS`?

And last but not least, I just realized I should make some tests, right?
What should I cover with them? I have a vague idea on what to do but I'd appreciate if you gave me a more solid idea on what to implement, like topics, or something. (if you think the tests are relevant) 

Sorry for asking so many questions... And thanks!

webcompat/helpers.py
```diff
@@ -397,7 +398,18 @@ def wrapped_func(*args, **kwargs):
...
+            if app.config['UPDATE_MOCK_OUTS']:
+                data = func(*args, **kwargs)
+                with open(file_path + '.json', 'w') as f:
+                    data_json = json.loads(data[0])
+                    if isinstance(data_json, str):
+                        data_json[0]["_fixture"] = True
+                    if isinstance(data_json, dict):
+                        data_json["_fixture"] = True
+                    data_json = json.dumps(data_json, indent=2)
+                    f.write(data_json)
```
karlcow: Before trying to understand why UPDATE_MOCK_OUTS is failing, I want to understand what this code is trying to do. 

fixtures are preserved in the history of the project through a git commit in the project. Their parameters are carefully chosen to work and fail given a certain number of parameters. Sometimes they are like the real issue online, sometimes they are not. So it's important to not update them when running the app. They should be read but never written by the app. That would defeat the purpose of the fixture.

Loading them and saving them in a test environment will not change them in the git history. 

It seems that what you need is a tool to adjust some specific fixtures **once** then commit them in the project.

This should probably be a tooling script under `tools/` But I might have misunderstood what you are trying to do.

package.json
```diff
@@ -49,10 +49,10 @@
     "postcss-import": "^12.0.0",
     "postcss-reporter": "^6.0.0",
     "postcss-url": "^8.0.0",
-    "prettier": "1.15.2",
+    "prettier": "1.15.3",
```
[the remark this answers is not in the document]
marimeireles: you're right. I didn't see that. will fix

webcompat/helpers.py
marimeireles: >It seems that what you need is a tool to adjust some specific fixtures once then commit them in the project.

That's exactly what I want to do.

>I want to understand what this code is trying to do.

I needed to modify all of the mock outs we had to be able to remove the `markdown-it`. 
The way `markdown-it` works is the following: it receives a content formatted as markdown and then translates it to html. But, the new way I'm doing is making a request to Github where the content is already in html.
I can't pass the tests because the mock outs have this old markdown format and the program is expecting to deal with html now, so it crashes.
This code should enable the user to get into this "update mock outs mode" where the user can download the new version of the mock outs from the repository and it will be formatted and saved automatically.

>Sometimes they are like the real issue online, sometimes they are not. So it's important to not update them when running the app.

That's a problem. If the repo contains different versions of the issues we're using to make the tests, than I think it's useless.

>This should probably be a tooling script under `tools/`

I didn't know that. I asked Mike and he said it was okay to do that way, but I think I should probably have asked you :p
- sorry for that. If you think this is still useful I could try to rewrite this on `tools/`. But if you think it isn't... Then I could just do the easiest thing and substitute the old markdown content with html with a ctrl+c/ctrl_v

karlcow: >  This code should enable the user to get into this "update mock outs mode" where the user can download the new version of the mock outs from the repository and it will be formatted and saved automatically.
> Does this answer your question or would you like me to be more specific?

It does. So indeed this should be a script in tools as it is to manage the project and is not really part of the app. 

> That's a problem. If the repo contains different versions of the issues we're using to make the tests, than I think it's useless.

:) not necessary. That's the usual issue with mocking in general. Mocking is a way to isolate the tests from external factors (for example http requests to GitHub when testing). If GitHub is changing its API, some tests might still pass and we would not know. The bet/assumption we make when we are mocking is that the API is changing less often than when we run the tests.

That's the first part of mocking.

The second part of mocking is to be able to test again cases you want to fail. Sometimes you want to create a response that you are not likely to get but if you were getting it, you want to be able to handle it. Responses might evolve too. Think about HTTP headers  and things like pagination for example.

So indeed updating tests is often unrewarding. :) 

1. Make the tool you need in `tools/` to adjust the tests that's a good idea to include it, even if we run it only once every 6 months. Whatch out for fixtures which are not for functional tests. `nosetests` is **not** using the `python run.py -t` mode but is using some files in fixtures.
2. Commit the tool.
3. Run the tool to change the fixtures in your markdown-it removal branch (careful, not all fixtures are about functional tests)
4. Run the functional tests and see what is breaking. We might learn things.
5. Commit the new tests and the markdown-it removal in the same pull request.
6. If everything is working as expected, we should be able to merge :)

Thanks a lot @marimeireles 

> I didn't know that. I asked @miketaylr and he said it was okay to do that way, but I think I should probably have asked you :p

hehe. It's why it's often better to discuss options in the issues. So people can jump in with rationales for different options. :)

marimeireles: >Whatch out for fixtures which are not for functional tests. nosetests is not using the python run.py -t mode but is using some files in fixtures.

Okay. Thanks for the checklist :)

>hehe. It's why it's often better to discuss options in the issues. So people can jump in with rationales for different options. :)

You're right. I'll learn from this mistake haha

marimeireles:
@karlcow I was thinking... I thought of two different ways of putting this code under `tools`.

1. I'd create a `fetch-mock-outs.py` file that would set this flag to true:
https://github.com/webcompat/webcompat.com/blob/7526e7d28517738ede33ac9762607f860d0b133a/run.py#L126
The way it is implemented now you have to type -u on the terminal and the way I'm proposing you'd have to run `fetch-mock-outs.py`.

But Mike and I got to the conclusion that you wouldn't be happy about it. So Mike proposed another solution:

2. I create a `fetch-mock-outs.py` that receives as input the number of the issue you want to fetch. The program will make a request to the the api, fetch only that issue and then terminate.

Keep in mind that this new mode will only fetch new issues. So I don't think there will be an issue.

karlcow:
@marimeireles 

> I create a fetch-mock-outs.py that receives as input the number of the issue you want to fetch. The program will make a request to the the api, fetch only that issue and then terminate.

Option 2. Sounds like a good plan, except for the name because basically this is only fetching a response from an API. 

which makes me think in which ways the tool would be really useful?

A simple does that already :)  
```bash
curl 'https://api.github.com/repos/webcompat/web-bugs/issues/20000' 
```

or using [httpie](https://httpie.org/)

```bash
http -b 'https://api.github.com/repos/webcompat/web-bugs/issues/20000'
```

Or do you want the code to 

1. fetch one issue 
2. move the old mock  in a backup directory
3. replace the mock with the new response

I'm checking the current fixtures.

```bash
find -s tests/fixtures/api/issues | grep "/issues/[0-9]" | grep json | wc -l
```

We currently have only 18 JSON fixtures.

This is the list
```bash
find -s tests/fixtures/api/issues | grep "/issues/[0-9]" | grep json
```

```
tests/fixtures/api/issues/100/comments.fafc13ba597647ded279ea1b663135a6.json
tests/fixtures/api/issues/100/comments.json
tests/fixtures/api/issues/100.json
tests/fixtures/api/issues/101/comments.fafc13ba597647ded279ea1b663135a6.json
tests/fixtures/api/issues/101.json
tests/fixtures/api/issues/15000.json
tests/fixtures/api/issues/2.json
tests/fixtures/api/issues/200.json
tests/fixtures/api/issues/23/comments.fafc13ba597647ded279ea1b663135a6.json
tests/fixtures/api/issues/23.json
tests/fixtures/api/issues/252/comments.fafc13ba597647ded279ea1b663135a6.json
tests/fixtures/api/issues/252.json
tests/fixtures/api/issues/396/comments.fafc13ba597647ded279ea1b663135a6.json
tests/fixtures/api/issues/396.json
tests/fixtures/api/issues/70/edit.json
tests/fixtures/api/issues/70.json
tests/fixtures/api/issues/9/comments.fafc13ba597647ded279ea1b663135a6.json
tests/fixtures/api/issues/9.json
```

I wonder how many of these are already different than the current json. 
A quick script could help us check that.

1. fetch the online_json associated with the issue number
2. prettify the online_json and fixture_json (to get the same layout)
3. makes a diff in between those too. 
4. Flag all of these which are different
5. Understand why they are different and if it matters.

I also wonder which ones need to be updated for our current where the body as html matters.

marimeireles:
> A simple does that already :)

You're right.

> A quick script could help us check that.
>1. fetch the online_json associated with the issue number
>2. prettify the online_json and fixture_json (to get the same layout)
>3. makes a diff in between those too. 
>4. Flag all of these which are different
>5. Understand why they are different and if it matters.


So instead of creating this mode you think I should create a script that does that on `/tools` or should we just abandon this idea?

marimeireles:
Ok.
So we can close this, imo.
Thank you Karl and Mike for the discussion. :)

PR #2744, opened 2018-12-18
Fixes #1822 - Adds GitHub contact form for anonymous reporting

### LINKED ISSUE #1822: Provide a way to get in touch with anonymous reporters

karlcow:
This PR fixes issue #1822

## Proposed PR background

We identified that some people have a github username but don't want the too hungry webcompat.com to be allowed for permissions. This is a very low tech, soft way to give an opportunity for people to share their GitHub nickname.

This will hopefully work better than the previous pull request #2713 

r? @miketaylr

### DIFF
--- a/webcompat/static/js/lib/bugform.js
+++ b/webcompat/static/js/lib/bugform.js
@@ -18,6 +18,7 @@ function BugForm() {
   this.submitTypeInput = $("#submit_type:hidden");
   this.uploadLabel = $(".js-label-upload");
   this.urlParamRegExp = /url=([^&]+)/;
+  this.githubRegexp = /^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}$/;
 
   this.UPLOAD_LIMIT = 1024 * 1024 * 4;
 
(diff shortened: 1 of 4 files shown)


karlcow:
Let's see if CircleCI is happier.

karlcow:
Alleluiah! All checks have passed! @miketaylr

webcompat/static/js/lib/bugform.js
```diff
@@ -341,6 +351,26 @@ function BugForm() {
...
+    - no two consecutive hyphens
+  */
+  this.isValidGitHubUsername = function(contact) {
+    return this.githubRegexp.test(contact);
+  };
+
+  /* Check to see if the GitHub username has the right syntax. */
+  this.checkGitHubUsername = function() {
+    var contact = this.contactField.val();
+    if (this.isValidGitHubUsername(contact) || $.trim(contact) === "") {
```
miketaylr: (thanks for tweaking this)

PR #2745, opened 2018-12-18
Fixes #2729 - Remove labels request when they're unnused

### LINKED ISSUE #2729: Remove labels request at single issue pages
We don't need too make a label request to Github on single issue pages because we already have the information we need on the issue request.

So we can remove the label request.

Some context: #2726

marimeireles:
This fix checks if the user is logged in (because it's the only scenario where all of the labels need to be fetched). 

If the user is logged in, we fetch our 100+ labels, if it isn't we don't even show the possibility of editing the labels.

r? @miketaylr


marimeireles:
Is this a terrible solution? (It seems terrible... for some reason)

And also you think I should remove this "fetch all labels" request from closed issues? Or should it still be possible to edit an issue even after it's closed?

Thanks!

Issue #2774, opened 2019-01-24
Fix code blocks (with language identifiers) on webcompat.com


marimeireles:
The following is happening:

If you type something like:

```
```html
<div>
```

It'll break on webcompat.com (not upper cased, that was fixed):

<img width="361" alt="screen shot 2019-01-24 at 15 51 28" src="https://user-images.githubusercontent.com/17600982/51697863-fafc6300-1fef-11e9-9be9-694a95985f53.png">

If just type your code directly after the three back ticks you'll get the correct output on webcompat:
<img width="682" alt="screen shot 2019-01-24 at 15 51 33" src="https://user-images.githubusercontent.com/17600982/51697952-40209500-1ff0-11e9-8e6f-a5535b8da26c.png">

Like shown by Mike on this [issue](https://github.com/webcompat/webcompat-tests/issues/1747).

marimeireles:
What's happening here is that github is sending the classes they use on their website to ours directly. But since we don't have their .css it doesn't take any effect.

One thing we could do is download their lib. I found this one: https://github.com/primer/github-syntax-light and I'm going to make some tests with it.

marimeireles:
So this is what I got after downloading this lib:

<img width="687" alt="screen shot 2019-01-24 at 16 48 38" src="https://user-images.githubusercontent.com/17600982/51701127-ee7c0880-1ff7-11e9-85ec-8b0f3fd3a669.png">

The code with the word after the back ticks gets highlighted correctly, but doesn't get that little box around it. That box comes from the prism.js lib.
I think that there are two alternatives to fix this. 

First, we can just copy the class from prism.js and add `.highlight > pre` to it and we'll have the same result we have now.

```
 /* Code blocks */
 pre[class*="language-"], .highlight > pre{
   padding: 1em;
   margin: .5em 0;
   overflow: auto;
   border: 1px solid #dddddd;
   background-color: white;
 }
```
I think this is the only part of the prism.js code we're using, right? If yes, we could remove it and keep only the github-syntax-light lib.

Another thing we could do is use this lib: https://github.com/primer/primer/blob/master/modules/primer-markdown/README.md
Primer markdown is an official lib by Github that offers "Stylesheets for rendering GitHub Flavored Markdown and syntax highlighted code snippets". 
It would mimic all of the Github style perfectly, but this is a big change and might break parts of our code. I could test it if you think it's a good idea.

What do you think @miketaylr ?

miketaylr:
[...]

PR #2775, opened 2019-01-24
Fixes #2768 - Renames highlight class so it doesn't interact with the…

### LINKED ISSUE #2768: Code is now uppercased.
```css
.highlight {
	font-weight: 300;
	text-transform: uppercase;
}
```

See for example https://webcompat.com/issues/22453
This is another regression from the switch to markdown html sent by GitHub.

marimeireles:
Renames highlight class so it doesn't interact with the Github class.

This should make the the uppercase effect disappear on the codes sessions. 

But it's not supposed to fix the problem entirely for that there is this new issue #2774


webcompat/templates/contributors/report-bug.html
```diff
@@ -41,7 +41,7 @@ <h2 id="how" class="headline-1 highlight">How To Report An Issue?</h2>
...
             <li>Modify the browser information if you are reporting an issue for another browser.</li>
             <li>If possible, make sure you test in another browser before reporting. It may be that the web site simply has an issue no matter which browser is used. These are not web compatibility issues.</li>
-            <li>A screenshot of the page, or just the area of the page highlighting the issue, is one of the best ways to communicate the problem you experienced. Be mindful about screenshots containing private information about yourself or others (such as social network private pages or bank sites).</li>
+            <li>A screenshot of the page, or just the area of the page headlineing the issue, is one of the best ways to communicate the problem you experienced. Be mindful about screenshots containing private information about yourself or others (such as social network private pages or bank sites).</li>
```
miketaylr: We probably don't want this change. :)

webcompat/templates/contributors/reproduce-bug.html
```diff
@@ -59,28 +59,28 @@ <h2 id="next" class="headline-1 highlight">Possible Next Steps</h2>
           </p>
 
           <dl id="statuses">
-            <dt id="incomplete" class="bold highlight">
+            <dt id="incomplete" class="bold headline">
```
miketaylr: Ah, seeing how this is used elsewhere... can we use `heading` instead of `headline`?

marimeireles: You mean just replace headline for heading in all of the places?

webcompat/templates/home-page/browse-issues.html
```diff
@@ -10,7 +10,7 @@ <h2 class="headline-1 highlight">Latest Untriaged Issues</h2>
     </div>
 
     <div class="grid-row">
-      <p class="highlight grid-cell x3">
+      <p class="headline grid-cell x3">
```
miketaylr: Missed one!

webcompat/templates/home-page/hero.html
```diff
@@ -1,6 +1,6 @@
 <section class="hero-section grid color-first-bg">
   <article class="hero-text">
-      <h1 class="hero-claim highlight js-hero-title">Bug reporting <br> for the web.</h1>
+      <h1 class="hero-claim headline js-hero-title">Bug reporting <br> for the web.</h1>
```
miketaylr: Here too.

marimeireles:
@miketaylr changes were made.
r?

.python-version
```diff
@@ -0,0 +1 @@
+2.7.8
```
miketaylr: This thingy appeared again. :)

(Where does it come from?)

If it's semi-common, you could add it to our .gitignore file.

marimeireles: arr >:(
I'm sorry. I usually add it. Thanks Mike

Issue #1295, opened 2017-01-20
Implement Twitter Cards for webcompat.com issues


karlcow:
This morning I was seeing this tweet from a contributor to the project ❤️ 

![capture d ecran 2017-01-20 a 10 18 01](https://cloud.githubusercontent.com/assets/505230/22131884/d610d528-def9-11e6-82af-5b100f787251.png)

And I thought that we should probably take advantage of this twitter feature. It is quite easy to implement.

Explanations are at https://dev.twitter.com/cards/overview

And example from https://dev.twitter.com/cards/types/summary

```html
<meta name="twitter:card" content="summary" />
<meta name="twitter:site" content="@flickr" />
<meta name="twitter:title" content="Small Island Developing States Photo Submission" />
<meta name="twitter:description" content="View the album on Flickr." />
<meta name="twitter:image" content="https://farm6.staticflickr.com/5510/14338202952_93595258ff_z.jpg" />
```

There is a validator for it.
https://cards-dev.twitter.com/validator

zoepage:
@karlcow what would be the purpose of the cards?

karlcow:
The card would be unique to each issue. And have a screenshot when it exists and basic description of the issue.

miketaylr:
I like this idea. Makes tweets w/ links to webcompat issues much more compelling.

zoepage:
@karlcow I don't get the flow of it. Sorry.
So basically every issue would be a tweet or whenever someone tweets an issue at webcompat, we'd get it displayed on our site?

miketaylr:
> @karlcow I don't get the flow of it. Sorry.
> So basically every issue would be a tweet or whenever someone tweets an issue at webcompat, we'd get it displayed on our site?

It just means if someone tweets a tweet w/ a link to a `webcompat.com/issues/NNN` in it, it will have an embedded card w/ an image and some context. Currently it's just a link.

zoepage:
Thank you for making that clear @miketaylr 

👍

miketaylr:
I looked into this, I don't think it's currently possible to do this with client-side rendering since Twitter is fetching the meta tags from the html (and doesn't render JS, from what i've read).

zoepage:
😿 

I've removed the refactor label then. Is it ok, if we close the issue and move it to the backlog?

zoepage:
Let's close as this is for backlog anyway...

marimeireles:
I'm adding to my project so I don't forget about it :)

PR #2783, opened 2019-01-28
Fixes #1870 - Renders webcompat.com using python instead of js

### LINKED ISSUE #1870: [performance] issue data inside the HTML at first load
## Context

The page for an issue view is currently an empty HTML template, which is populated through JS calls to webcompat.com API endpoint.

```html
<title>Issue #1299 | webcompat.com</title>
```html

The choice was made for not having the impression to wait too much after GitHub by quickly delivering a page to see to the users while the JS is getting the data to populate the html page.

## Why this issue?

This issue is just here to **test some options** on making the code less dependent on JS for issues data and see if we can deliver right away the data to users. It might not results in any improvements for the code itself, but it will help determine if we made the right choices in terms of perceived performance for the user.

Benefits: 
* Caching/index-ability
* Full titles
* meta allowing cards
* parseable HTML through cli.

### PR DESCRIPTION
**Pull request by marimeireles**
This PR is supposed to keep track of my changes so @miketaylr and @karlcow can follow it.
It's not necessarily going to be merged.
It depends if it will improve webcompat's performance or if it will be beneficial in some other way.

### CODE REVIEW

**Comment by marimeireles** [comment]
@karlcow and/or @miketaylr 
Can you check the code I wrote? 
I know it's breaking the tests, but I'd like to know if you think this is the right way to do things and also if you think we should add this to our codebase. 
IMO we can obtain even better results from these changes. I only ran tests on localhost and I was able to load the issue page about 15% faster while rendering them with python. In localhost, per example, my requests to the API are instantaneous while in the real website they'd take much longer.

**Comment by karlcow** [comment]
Sure. I'll do that tomorrow. Thanks a lot for the work.

**Comment by karlcow** [comment]
just to keep a trace of how this is cool. This is the raw HTML sent from the server. (Not the modified DOM that you could see in the inspector)

capture d ecran 2019-02-06 a 10 07 12

aka `curl http://localhost:5000/issues/1754`

This is awesome. And indeed it doesn't make a difference, I even had the impression it was faster (from Japan. tune bias here).

The cool thing about this are the secondary thing we can do such as 

* [sending meta for twitter like things](https://github.com/webcompat/webcompat.com/issues/1295).
* Giving better `<title>` so this appears in bookmarks, tabs and search engines results.

File: webcompat/static/js/lib/issues.js [hunk 1]
```diff
@@ -145,41 +132,6 @@ issues.AsideView = Backbone.View.extend({
...
-      })
-      .wrap('<p class="is-hidden"></p>');
-
-    // this is probably really slow, but it's the safest way to not hide user data
-    issueDesc
-      .find("p:last-of-type em:contains(From webcompat.com)")
-      .parent()
-      .addClass("is-hidden");
-
-    if (this.mainView._isNSFW) {
```
karlcow: This is something, we would probably have to cater for in the server side rendering. To note that the request of the JSON gives us already the tag with the NSFW.

webcompat/views.py
```diff
@@ -275,14 +276,19 @@ def create_issue():
 @app.route('/issues/<int:number>')
 @cache_policy(private=True, uri_max_age=0, must_revalidate=True)
 def show_issue(number):
+    issue_body_json, issue_body_status_code, issue_body_headers = api.proxy_issue(number)
```
karlcow: This will help us also to treat the [PR and deleted issues](https://github.com/webcompat/webcompat.com/issues/1311) in a more elegant way.

webcompat/views.py
```diff
@@ -275,14 +276,19 @@ def create_issue():
...
+    issue_body_json = json.loads(issue_body_json)
+
+    issue_comments_json, issue_comments_status_code, issue_comments_headers = api.proxy_comments(number)
+    issue_comments_json = json.loads(issue_comments_json)
+
     """Route to display a single issue."""
```
karlcow: nit: The comment should be above, just after the `def`

webcompat/views.py
```diff
@@ -275,14 +276,19 @@ def create_issue():
...
     if g.user:
         get_user_info()
     if session.get('show_thanks'):
         flash(number, 'thanks')
         session.pop('show_thanks')
-    return render_template('issue.html', number=number)
-
+    return render_template('issue.html', number=number, issue_body=issue_body_json['body_html'], issue_comments_json=issue_comments_json)
```
karlcow: Also see my comment in https://github.com/webcompat/webcompat.com/pull/2783#issuecomment-460867838 about the twitter cards.

marimeireles: Cool! :)
I'm going to read the docs to learn how to do it.

webcompat/static/js/lib/issues.js
marimeireles: > This is something, we would probably have to cater for in the server side rendering

Would you like me to do that in this PR?
Or should I just add it back now?

karlcow: whichever is 
1. simpler for you
2. effective for the code

Probably aim for now for the JS if it's simple and straightforward and we can still open a second issue to improve it if it's faster on the server side.

webcompat/views.py
karlcow: very cool. But let's not do it in that PR. We can do it in the other issue aka #1295

marimeireles: Yeah. You're right. okay :)

webcompat/static/js/lib/issues.js
marimeireles: Okay. I'll explore a bit around this and decide if it's worth it to do it now.

webcompat/views.py
marimeireles: Sorry @karlcow, but I don't know exactly what you mean.
I read the issue and I saw you gave a few solutions:

> - We create a relevant generic body message the same way we do for our 404 message.
> - So all of these can be put in a specific milestone, called `ignore` and it's another way to identify them.

What do you think it could be a good solution here? Thanks.

karlcow: Sorry. This is just a side comment to say that server side rendering will help us to solve #1311 in a simpler way that all the acrobatic solutions we were trying to do. This will be straightforward (if we decide to go that way).

webcompat/views.py
marimeireles: Thanks.

webcompat/views.py
karlcow: So basically, you can ignore for now. And we can solve it later if we deploy this.

marimeireles: Okay! :)

marimeireles:
When I'm on test mode the variable fixture_path (which gives the path to the fixture path, lol) has a weird behavior, it points to the wrong place. It points to tests/fixtures/issues instead of tests/fixtures/api/issues and that's breaking the code.
The code runs normally on production or dev in the normal mode. It only doesn't work on test mode.

karlcow:
> When I'm on test mode the variable fixture_path (which gives the path to the fixture path, lol) has a weird behavior, it points to the wrong place. It points to tests/fixtures/issues instead of tests/fixtures/api/issues and that's breaking the code.

Does it do that after your changes or on master?

karlcow:
https://github.com/webcompat/webcompat.com/search?q=fixtures_path&unscoped_q=fixtures_path

marimeireles:
After my changes!
I just wrote here so I won't forget and keep working on it tomorrow.

marimeireles:
I've been investigating what's wrong with this weird `fixtures_path` bug and here is what I found:

Currently I'm making the api requests on `view.py` like this:

```
@app.route('/issues/<int:number>')
@cache_policy(private=True, uri_max_age=0, must_revalidate=True)
def show_issue(number):
    """Route to display a single issue."""
    issue_body_json = api.proxy_issue_direct(number)[0]
    issue_body_json = json.loads(issue_body_json)

    issue_comments_json = api.proxy_comments_direct(number)[0]
    issue_comments_json = json.loads(issue_comments_json)

    if g.user:
        get_user_info()
    if session.get('show_thanks'):
        flash(number, 'thanks')
        session.pop('show_thanks')
    return render_template('issue.html', number=number, issue_body=issue_body_json['body_html'], issue_comments_json=issue_comments_json)
```

On test mode our code depends on several single requests to work, that is, if you're on localhost:5000/issues/100 we'll generate several different requests `api/issues/100`, `api/issues/100/comments/`, `api/issues/100/labels`, etc. We use each one of these single requests on `@mockable_response` to make a fake request and load our mocks into the page. To do that we use a flask attribute called `request.path`:

https://github.com/webcompat/webcompat.com/blob/d2edc8ce4f303055e3ef7c82e64d730a4c4ca57e/webcompat/helpers.py#L397

The problem is that the way I'm doing I don't generate different requests, the only request that's generated the way I'm doing is the one in `views.py`, `@app.route('/issues/<int:number>')`. Therefore `request.path` doesn't work properly and it's not able to load the page because it can't make the correct requests to access our mocks.

To fix this issue I proposed a solution that is probably not the best one, but it works. 
Fixing this will probably mean a lot of refactoring since `request.path` isn't working anymore. We'd have to rewrite `@mockable_response` and probably change a lot of things on `endpoints.py` too. @karlcow  @miketaylr I think this is too big for me to go along by myself, can you help me figure out a good solution here?

marimeireles:
It's still breaking a lot of tests and I didn't investigate that yet, but the page loads normally.
screen shot 2019-02-10 at 23 31 33

I hope I can fix these tests easily, but I'm not sure.
screen shot 2019-02-10 at 23 25 57

tests/fixtures/api/issues/101/comments.json
```diff
@@ -0,0 +1,87 @@
+[
```
marimeireles: I had to add all of these `comments.json` files because I'm not able to generate the old hash code because the `get_args` I'm receiving in my code is different from the old one:
https://github.com/webcompat/webcompat.com/blob/d2edc8ce4f303055e3ef7c82e64d730a4c4ca57e/webcompat/helpers.py#L393-L397
But if you think it's necessary I could generates new hashs.

webcompat/helpers.py
```diff
@@ -408,6 +408,35 @@ def wrapped_func(*args, **kwargs):
...
+                file_path = FIXTURES_PATH + full_path
+            if not os.path.exists(file_path + '.json'):
+                print('Expected fixture file: ' + file_path + '.json')
+                return ('', 404)
+            else:
+                with open(file_path + '.json', 'r') as f:
+                    data = f.read()
+                    return (data, 200, get_fixture_headers(data))
+        return func(*args, **kwargs)
+    return wrapped_func
```
marimeireles: I had to add this function and keep the old `mockable_response` because there are still parts of the code rendered in js but I needed a new function to deal with the python rendered part.

The idea here is that `wrapped_func` now receives as an argument the path of the request that was made, that's why I added a string on the function on the `endpoints.py` ` proxy_comments_direct = mockable_response_direct(proxy_comments_impl, '/api/issues/{0}/comments')`. With that string I'm able to make the fake requests and get the correct mocks. :)

webcompat/api/endpoints.py
```diff
@@ -193,6 +199,15 @@ def proxy_comments(number):
                            mime_type=JSON_MIME_HTML)
 
 
+@api.route('/issues/<int:number>/comments', methods=['GET', 'POST'])
+@mockable_response
+def proxy_comments(number):
+    return proxy_comments_impl(number)
+
+
+proxy_comments_direct = mockable_response_direct(proxy_comments_impl, '/api/issues/{0}/comments')
```
marimeireles: This is the line called on `views.py` when we're making a request, per example: `issue_body_json = api.proxy_issue_direct(number)[0]`
What this line does is if we're on normal mode we'll just call the `proxy_comments_impl` and everything will run normally as the old times. But if we're on test mode we'll pass the string `'/api/issues/{0}/comments'` to the new wrapped_function on `helpers.py` and that function will now be able to find the mock outs without using `request.path`.

marimeireles:
Hope my review isn't too confusing. Thanks!

miketaylr:
> @karlcow @miketaylr I think this is too big for me to go along by myself, can you help me figure out a good solution here?

IMO, the objective here is to prove if we should seriously consider this server-rendered approach, not necessarily land it as you've written. I agree it probably needs some serious re-thinking in terms of architecture, so I wouldn't worry too much about your attempts to get it working here.

webcompat/api/endpoints.py
karlcow: This is weird. This line is in the module by itself in between two functions. 

if it's a specific function you should make it so and call this as a function. Not as a assigned variable in the module. 

This also seems to create a double indirection.

1. `api.proxy_issue_direct(number)[0]` calls …
2. `mockable_response_direct(proxy_issue_impl, '/api/issues/{0}')` calls…
3. `mockable_response_direct` calls…
3. `proxy_issue_impl` calls…
4. `api_request`

It's probably not the best way to do that **but** it shows that if we go this road we need to revise a bit the test strategy which is normal. 

Thanks a lot @marimeireles

webcompat/helpers.py
karlcow: ok basically if I understand it's code to make it work, but not code necessary if we revise the tests.

marimeireles: Yes, exactly. I know it's ugly. 

To make it pretty we would have to change how the test mode works. Currently it's completely dependent of `request.args` but it can't be at all because since we're not making real requests this attribute is useless.

miketaylr:
I think we should close this PR since we won't merge it as is. @marimeireles did great work justifying the further experimentation / move to a server-side rendering solution for the non-interactive bits. Realistically, this probably won't happen until the second half of the year. Let's come back to this. ❤️

Issue #2450, opened 2018-05-10
Triage dashboard displays incorrect information


softvision-sergiulogigan:
**Environment:**
Browser / Version: Chrome 66 (Desktop), Firefox Nightly 62.0a1 (Desktop)
Operating System: Windows 10

**Steps to Reproduce:**
1. Go to https://webcompat.com/dashboard/triage
2. Observe the "Older than 48h" information box.

**Expected Behavior:**
The "Older than 48h" information box displays the same number as the number of cards.

**Actual Behavior:**
The "Older than 48h" information box displays a different number of issues than the number of cards.

**Notes:**
1. Screen capture attached.
2. Reproducible both on Chrome and Firefox.

Watchers:
@softvision-sergiulogigan
@softvision-oana-arbuzov

sv;

![image](https://user-images.githubusercontent.com/16573771/39871308-0c884b42-546d-11e8-9f1c-2132869d5a33.png)

magsout:
Some Issues are not tagged by the back-end, so theses issues are always hidden

![capture d ecran 2018-05-10 a 15 05 07](https://user-images.githubusercontent.com/1997108/39872031-f73e413e-5466-11e8-9b5f-a7cd3cd8b783.png)

magsout:
Follow Up on discussions with @karlcow on this morning, we thought it was about issues without milestone

softvision-sergiulogigan:
Now we have 3 counted, and 5 displayed.

![image](https://user-images.githubusercontent.com/16573771/40310582-ee43e838-5d15-11e8-8e7e-20a3655fb756.png)

softvision-sergiulogigan:
@magsout any updates on this issue? Is it easily fixable?

magsout:
@softvision-sergiulogigan I tried to figure out why issues does'nt appears. 

I don't get it why yet ;p

But it seems nothing to do with label or milestone.. I have some work on the metrics dashboard, but I wiil try to spend some time on it

magsout:
Few examples of issues not tagged:

https://webcompat.com/issues/17349
https://webcompat.com/issues/17360
https://webcompat.com/issues/17365

magsout:
@karlcow  is it possible to see the object of these issues?

or maybe I missed something in the code:

```
  <article
    data-number="{{ issue['number'] }}"
    class="
      wc-Triage
      {% if issue['older'] %} priority {% endif %}
[...]

PR #2784, opened 2019-01-28
Fixes #2726 - Handle pagination when fetching labels from GitHub

### LINKED ISSUE #2726: Handle pagination when fetching labels from GitHub
In troubleshooting issue #2482, it seems like the problem may not be in the regex, but rather in the fact that we're only fetching the first 100 labels in our API call.

laghee:
This PR fixes issue #2726 

## Proposed PR background

Since we are not currently fetching all the existing labels with our GitHub repo call when constructing the label editor on an issue page, filtering behavior and selection is wonky. I added a pagination function to `get_issue_labels` and a test to make sure it works correctly. (I included stubs to expand into eventual testing of the other label functions as well.)

---

- [✅] I have read the [Pull Request + Code Style Guidelines](https://github.com/webcompat/webcompat.com/blob/master/docs/pr-coding-guidelines.md) thoroughly.

### DIFF
--- a/tests/unit/test_tools_labels.py
+++ b/tests/unit/test_tools_labels.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Tests for the label functions."""
+
+import unittest
+
+from tools.labels import get_issue_labels
+from tools.labels import USER_LABELS_URI
+
+LABELS_URI = USER_LABELS_URI
+
+
+class TestLabels(unittest.TestCase):
+    """Test for label operations."""
+
+    def setUp(self):
+        """Set up the tests."""
+        pass
+
+    def tearDown(self):
+        """Tear down the tests."""
+        pass
+
+    def test_get_issue_labels(self):
+        """Fetch all labels from GitHub."""
+        actual = get_issue_labels(LABELS_URI)
+        self.assertEqual(len(actual), 32)
+
+    def test_create_label(self):
+        """Create a label in the user repo."""
+        pass
+
+    def test_delete_label(self):
+        """Delete a label in the user repo."""
+        pass
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/tests/unit/test_tools_labels.py
+++ b/tests/unit/test_tools_labels.py
@@ -11,8 +11,6 @@ import unittest
 from tools.labels import get_issue_labels
 from tools.labels import USER_LABELS_URI
 
-LABELS_URI = USER_LABELS_URI
-
 
 class TestLabels(unittest.TestCase):
     """Test for label operations."""
@@ -27,7 +25,7 @@ class TestLabels(unittest.TestCase):
 
     def test_get_issue_labels(self):
         """Fetch all labels from GitHub."""
-        actual = get_issue_labels(LABELS_URI)
+        actual = get_issue_labels(USER_LABELS_URI)
         self.assertEqual(len(actual), 32)
 
     def test_create_label(self):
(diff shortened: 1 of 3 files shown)


laghee:
Hmm... OK, running lint locally (which seems weird since I didn't touch any js or css files), I'm not getting these errors. Will investigate further...

miketaylr:
I'm not super familiar with this tools/labels.py script. Is it the right place to make these changes? (I'm not sure what it does, tbqh).

What about https://github.com/webcompat/webcompat.com/blob/master/webcompat/api/endpoints.py#L213-L219, where the JS requests all the labels?

laghee:
@miketaylr You make an interesting point... 🤔 I picked this up mostly done from the end of AllHands, and I can't completely remember how I landed on `labels.py`. I know we had talked about server-side potentially being a better fix, but I'll need to go back and retrace my path to have a smarter answer.

laghee:
> I'm not super familiar with this tools/labels.py script. Is it the right place to make these changes? (I'm not sure what it does, tbqh).

Oh, 💩 . I think this script -- or these label tools, anyway... update the `issues.db`, maybe? I have no idea what I was thinking (other than python , obviously).

Sorry! 🙈  Regroup...

laghee:
OK, so I re-oriented myself, and now I have a question.

Might it be a good idea to cover more than the current labels issue with this fix (part of the problem in issue #2450 also seems to be with pagination -- whenever the total needstriage issues go over 100, an incomplete list is filtered and sent to the dashboard)?

I'm thinking of working the loop into a new helper method that can be called from `api_request`:
https://github.com/webcompat/webcompat.com/blob/4cad018bb0454720765bc4a9520bac5c21de149c/webcompat/helpers.py#L450
if that makes sense. Directly changing `api_request` would blow up the issues endpoint (which already has pagination worked out through the frontend JS), but just fixing the labels seems like it will end up being very non-DRY if other fixes are needed later.

I'm not nearly as familiar with the front end, though, so maybe it's better to fix one thing at a time and worry about scope later? 

@miketaylr @karlcow ?

karlcow:
@laghee @miketaylr (adding @marimeireles as she discussed about labels for performances too, aka requesting labels only when we actually open the menu, like GitHub does.)

ok the current request is

```python
@api.route('/issues/labels')
@mockable_response
def get_repo_labels():
    """XHR endpoint to get all possible labels in a repo."""
    params = request.args.copy()
    path = 'repos/{0}/labels'.format(REPO_PATH)
    return api_request('get', path, params=params)
```

Let's try to run ideas (bad and good).

## walk_pagination
we could imagine to have something like

```python
def api_request(method, path, params=None, data=None, mime_type=JSON_MIME, walk_pagination=False):
```
default is `False`.

when `walk_pagination=True`, we could

1. check if there is actually more data to fetch (link http header)
2. if yes, fetch the next data
3. add it to the previous response
4. send the results back.

Possible issues:
1. we blow everything up if we forget about the way this is working and we enter in a 10 pages data set.
2. the network breaks in between the two or more requests

## parameter

The request is currently `/api/issues/labels?per_page=100`
but let's say we push it to 200. This is a request to **our API**, we could then fetch two pages and re-assemble them into one JSON that we send back.

Possible issue:
1. The network breaks in between the two requests.

## local
We do not request milestones each time we use them because we consider them stable. We could perfectly through a CRON job (for example, once a day or once a hour), have a script creating a local JSON file with the list of all labels. The JSON is then local, cached, with the appropriate ETAGs, and can be requested by the JS to build the menu without having to make any requests to GitHub. Performance++

Possible issue:
Being unsettled because the labels are not sync on webcompat.com with what exist on GitHub. Though currently… we don't have the end of the list at all… so probably an improvement. 😛

marimeireles:
>local
>walk_pagination

Both seems like good ideas.

I couldn't think of anything else... But let me know if I can help with something!

karlcow:
I'm closing the PR here. 
Let's start with a fresh PR.
I documented on the issue what we need to do. 
https://github.com/webcompat/webcompat.com/issues/2726#issuecomment-477005645

Issue #2778, opened 2019-01-25
Make HTTP2 server push with Flask


marimeireles:
Mike said it'd be a good idea to use the same idea as this lib is using: https://github.com/jdaroesti/flask-http2-push to make the server push.

It seems pretty straightforward and I think it's a good option.

karlcow:
https://github.com/jdaroesti/flask-http2-push/blob/master/flask_http2_push.py

marimeireles:
Ok... Bad news, preload doesn't work on Firefox: https://caniuse.com/#feat=link-rel-preload and we [don't have a estimate](https://bugzilla.mozilla.org/show_bug.cgi?id=1394778#c65) for when will it be.
So this approach using preload doens't work... I'll try to find something else we can use.

marimeireles:
I didn't find anything...
I was thinking about implementing this for Chrome... and not using push on Firefox, what do you think @karlcow or @miketaylr ... I don't know who to ping here.

marimeireles:
I could push things directly from the server using `http_push` instead of preload, that'd push files to Firefox, but it's not worth it since it will push the files every time and it's much faster to just load the files from cache if it's not the first time the user access the website.

karlcow:
> I was thinking about implementing this for Chrome... and not using push on Firefox,

but that would mean user agent sniffing, and I would avoid this at all cost (for such a feature).

We are not also a huge site, so maybe improving performances with HTTP/2 was a cool idea to explore but not yet worth implementing given the lack of support as you demonstrated. You discovered, studied, found a roadblock. It's part of the process. We are not forced to push it to the implementing end. We can close and put this on the back log until there is a better interop in between browsers.

miketaylr:
> but that would mean user agent sniffing, and I would avoid this at all cost (for such a feature).

@karlcow would it really require UA sniffing? if we just send the push `Link` header with the request for the html, and the browser supports it, it will populate the preload/push cache and then there's a perf win. if it doesn't, nothing should happen (for Firefox, until https://bugzilla.mozilla.org/show_bug.cgi?id=1222633 is fixed).

miketaylr:
Also, I'm pretty sure we do support http2 push? 

https://bugzilla.mozilla.org/show_bug.cgi?id=1150812

https://tools.ietf.org/html/rfc7540#section-8.2.1
https://searchfox.org/mozilla-central/search?q=PUSH_PROMISE&path=

[...]

PR #2792, opened 2019-02-05
Issue #2778 - Using flask to push files

marimeireles:
These are the relevant changes to test the firehose lib reponsible for making push requests on Flask.
Can you please add them to Staging? So we can test if there is something wrong with my local nginx configuration.

Here is some more context to why we're doing this: https://github.com/webcompat/webcompat.com/issues/2778#issuecomment-460782693

r? @miketaylr 

Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,14 +53,13 @@ jobs:
           command: |
             . venv/bin/activate
             pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example
+            npm run lint
             npm run build
+            nosetests
             python run.py -t &
             sleep 5
-            nosetests
-            npm run lint
             node --max-old-space-size=8192 ./tests/functional/_intern.js --reporters="runner" --firefoxBinary=`which firefox`
             firefox --version
-
       - store_artifacts:
           path: test-reports
           destination: test-reports
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,11 +53,11 @@ jobs:
           command: |
             . venv/bin/activate
             pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example
+            npm run build
             python run.py -t &
             sleep 5
             nosetests
             npm run lint
-            npm run build
             node --max-old-space-size=8192 ./tests/functional/_intern.js --reporters="runner" --firefoxBinary=`which firefox`
             firefox --version
 
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -21,11 +21,13 @@ from flask import session
 from flask import url_for
 
 from dashboard import filter_needstriage
+from flask_firehose import push
 from form import AUTH_REPORT
 from form import get_form
 from form import PROXY_REPORT
 from helpers import add_csp
 from helpers import add_sec_headers
+from helpers import bust_cache
 from helpers import cache_policy
 from helpers import form_type
 from helpers import get_browser_name
@@ -153,6 +155,22 @@ def file_issue():
 @app.route('/', methods=['GET'])
 def index():
     """Set the main view where people come to report issues."""
+    push('/css/dist/webcompat.min.css', **{
+        'as': 'style',
+        'rel': 'preload'
+    })
+    push(bust_cache('/js/dist/webcompat.min.js'), **{
+        'as': 'script',
+        'rel': 'preload'
+    })
+    push('/img/svg/icons/svg-leaf_right.svg', **{
+        'as': 'img',
+        'rel': 'preload'
+    })
+    push('/img/svg/icons/svg-leaf_left.svg', **{
+        'as': 'img',
+        'rel': 'preload'
+    })
     ua_header = request.headers.get('User-Agent')
     bug_form = get_form({'user_agent': ua_header})
     # browser_name is used in topbar.html to show the right add-on link
@@ -167,6 +185,18 @@ def index():
 @cache_policy(private=True, uri_max_age=0, must_revalidate=True)
 def show_issues():
     """Route to display global issues view."""
+    push('/css/dist/webcompat.min.css', **{
+        'as': 'style',
+        'rel': 'preload'
+    })
+    push(bust_cache('/js/dist/webcompat.min.js'), **{
+        'as': 'script',
+        'rel': 'preload'
+    })
+    push(bust_cache('/js/dist/issues.min.js'), **{
+        'as': 'script',
+        'rel': 'preload'
+    })
     if g.user:
         get_user_info()
     categories = app.config['CATEGORIES']
@@ -206,6 +236,14 @@ def create_issue():
     Any deceptive requests will be ended as a 400.
     See https://tools.ietf.org/html/rfc7231#section-6.5.1
     """
+    push('/css/dist/webcompat.min.css', **{
+        'as': 'style',
+        'rel': 'preload'
+    })
+    push(bust_cache('/js/dist/webcompat.min.js'), **{
+        'as': 'script',
+        'rel': 'preload'
+    })
     # Starting a logger
     log = app.logger
     log.setLevel(logging.INFO)
@@ -276,6 +314,18 @@ def create_issue():
 @cache_policy(private=True, uri_max_age=0, must_revalidate=True)
 def show_issue(number):
     """Route to display a single issue."""
+    push('/css/dist/webcompat.min.css', **{
+        'as': 'style',
+        'rel': 'preload'
+    })
+    push(bust_cache('/js/dist/webcompat.min.js'), **{
+        'as': 'script',
+        'rel': 'preload'
+    })
+    push(bust_cache('/js/dist/issues.min.js'), **{
+        'as': 'script',
+        'rel': 'preload'
+    })
     if g.user:
         get_user_info()
     if session.get('show_thanks'):
--- a/config/requirements.txt
+++ b/config/requirements.txt
@@ -1,5 +1,6 @@
 blinker==1.4
 Flask==1.0.2
+Flask-Firehose==0.2.2
 Flask-Limiter==0.9.3
 Flask-SQLAlchemy==2.1
 Flask-WTF==0.14.2
--- a/webcompat/__init__.py
+++ b/webcompat/__init__.py
@@ -9,12 +9,17 @@
 import logging
 
 from flask import Flask
+from flask_firehose import Firehose
 from flask_github import GitHub
 from flask_limiter import Limiter
 from flask_limiter.util import get_remote_address
 
 # Define the application
 app = Flask(__name__, static_url_path='')
+Firehose(app)
+# Adds the HTTP/2 Server Push functionality to app
+# firehose = Firehose(connector=Custom_connector())
+# firehose.init_app(app)
 
 # Configurations
 app.config.from_object('config')


webcompat/views.py
```diff
@@ -153,6 +154,10 @@ def file_issue():
 @app.route('/', methods=['GET'])
 def index():
     """Set the main view where people come to report issues."""
+    push('/css/dist/webcompat.min.css', **{
+        'as': 'style',
+        'rel': 'preload'
+    })
```
karlcow: @marimeireles 
So if I look at the results locally. This does the right thing in sending the right headers.

```http
HTTP/1.0 200 OK
Content-Length: 51341
X-Frame-Options: DENY
Content-Security-Policy: default-src 'self'; object-src 'none'; connect-src 'self' https://api.github.com; font-src 'self' https://fonts.gstatic.com; img-src 'self' https://webcompat.com https://www.google-analytics.com https://*.githubusercontent.com data:; manifest-src 'self'; script-src 'self' https://www.google-analytics.com https://api.github.com 'nonce-63c75e1bc1745d52ccbe69eca1c683ba2795a918'; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; base-uri 'self'; frame-ancestors 'self'; report-uri /csp-report
Link: </css/dist/webcompat.min.css>; as=style; rel=preload,
Vary: Cookie
Set-Cookie: session=******; HttpOnly; Path=/
Server: Werkzeug/0.14.1 Python/2.7.10
Date: Wed, 06 Feb 2019 02:08:25 GMT
```

We can see `Link: </css/dist/webcompat.min.css>; as=style; rel=preload,` at the HTTP level.

We probably need to check if the trailing comma could create an issue too. 

On the other hand it does nothing to the HTML side of things, where the link stays the same.

```html
<link href="/css/dist/webcompat.min.css" type="text/css" rel="stylesheet">
```

So I wonder if the issue is not just that. Aka the HTML Link overriding the HTTP Link header.

MDN docs doesn't mention HTTP in [Link preload](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content), just the HTML part.

I don't see anything in the [processing model of preload spec](https://w3c.github.io/preload/#processing) except the first step.

> When the user agent that supports [RFC5988] creates a Document and processes `Link` headers that contain a preload link. 

Though it says

> The user agent SHOULD abort the current request if the `href` attribute of the link element of a preload link is changed, removed, or its value is set to an empty string.

but this is only for changing the `href`, not other values.

So I guess if the HTML parsing says something different, I wonder if it cancels the instructions from the HTTP `Link`

Checking [RFC8288](https://tools.ietf.org/html/rfc8288#appendix-A.1) which obsoleted RFC5988.

And [HTML is not clear either ](https://html.spec.whatwg.org/multipage/semantics.html#processing-link-headers) about possible conflicts.

yoohoo there is another spec talking about it [CSSOM](https://drafts.csswg.org/cssom/#requirements-on-user-agents-implementing-the-http-link-header).

miketaylr:
@karlcow can you (re-)deploy this branch to staging? I tried, but apparently it doesn't like me. The fabfile is asking for a pw, that I guess I don't have (or possibly it changed)?

webcompat/views.py
marimeireles: >1. remove from the HTML template to see if it's working

Good idea! :)
It's worth a try and it's a quick test. I'll do it and post the result here.

>2.Pass the values to the HTML template to see if it's working better.

You mean like this https://github.com/shreyanshk/flask-firehose#pushing-resources?

Thanks for the review Karl!

marimeireles: >  It's worth a try and it's a quick test. I'll do it and post the result here.

It didn't work. :(

karlcow: so just the `Link` without the code in the template didn't work?
What about option 2?

marimeireles: No. :/

I didn't understand what you meant on option 2. You meant I should do something like this `<link rel="stylesheet" href="{{ push('/static/css/main.css', as='style', rel='preload') }}">`? Adding the `preload` tag on the HTML?

karlcow:
> @karlcow can you (re-)deploy this branch to staging? I tried, but apparently it doesn't like me. The fabfile is asking for a pw, that I guess I don't have (or possibly it changed)?

Probably the issue we talk about last week. I will deploy this branch. It has not been deployed yet.

webcompat/views.py
marimeireles: It's fine!
Already tried that. Same result as adding on the backend. I even tried both at the same time :p
No positive results.

karlcow:
@miketaylr @marimeireles Firehose lib deployed on staging as of now.

marimeireles:
Thanks for the help Karl and Mike.
I'll finish this PR and ask for your review once it's finished.
I still have to add the other files that have to be pushed and the code that's going to prevent the browser to fetch resources if it already has it stored.

karlcow:
See @marimeireles' [comment](https://github.com/webcompat/webcompat.com/issues/2778#issuecomment-461225849)

.circleci/config.yml
```diff
@@ -53,11 +53,11 @@ jobs:
           command: |
             . venv/bin/activate
             pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example
+            npm run build
             python run.py -t &
             sleep 5
             nosetests
             npm run lint
-            npm run build
```
marimeireles: I had to change the place where npm run build is called because I'm using the path to some files that have the hash generated by: https://github.com/webcompat/webcompat.com/blob/0133d3dfc2070dda72979b199dff30af1074806c/webcompat/helpers.py#L46-L52
Hope it's not a problem. CircleCI was breaking because it tried to run python before generating the files I needed

karlcow: @marimeireles Is this working?

```yaml
      - run:
          name: run tests
          command: |
            # install
            . venv/bin/activate
            pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example
            # python unittests. Do not require a server
            nosetests
            # creating a build
            npm run lint
            npm run build
            # starting the test server
            python run.py -t &
            sleep 5
            # running the functional JS tests
            node --max-old-space-size=8192 ./tests/functional/_intern.js --reporters="runner" --firefoxBinary=`which firefox`
            firefox --version
```

marimeireles: I think it is.
Tests are running fine with this change.

marimeireles: Yes.
As long as `npm run build` runs before the `python run.py` thing I think it's going to work.

webcompat/helpers.py
```diff
@@ -643,3 +643,26 @@ def prepare_form(form_request):
...
+
+    def set_pushed(self, inset):
+        """Update client state after pushing more items at the end of request.
+
+        Parameters
+        ----------
+        inset : set
+            A set of URLs of pushed items.
+        """
+        # your code here
```
marimeireles: But I completely forgot to erase this. This is wrong. I'm going to fix it.

.circleci/config.yml
karlcow: ok it's more logical this way. Because nosetests doesn't need the server to be started.

marimeireles: Ok. I'm going to change it then.
Thank you for this, Karl. :)

marimeireles: But apparently I can't add your comments because they're in the command session.

karlcow: note to myself: the `cache_dict` used in `bust_cache` is not global but per app. So I guess we create multiple ones, one  for each app being launched.

marimeireles:
Does it look good for you @karlcow ?

.circleci/config.yml
```diff
@@ -53,11 +53,11 @@ jobs:
           command: |
             . venv/bin/activate
             pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example
-            python run.py -t &
-            sleep 5
-            nosetests
             npm run lint
             npm run build
+            nosetests
```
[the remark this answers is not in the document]
karlcow: @marimeireles did you have an error message? What was it?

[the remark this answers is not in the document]
marimeireles: @karlcow 
You can check the log here: https://circleci.com/gh/webcompat/webcompat.com/363?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
I've just updated the repo with the code you told me to use.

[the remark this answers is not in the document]
karlcow: Thanks a lot! I will check.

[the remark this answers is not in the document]
karlcow: ah! Understood!

```
  File "/home/circleci/repo/webcompat/helpers.py", line 532, in policy
    response = make_response(view(*args, **kwargs))
  File "/home/circleci/repo/webcompat/views.py", line 321, in show_issue
    push(bust_cache('/js/dist/webcompat.min.js'), **{
  File "/home/circleci/repo/webcompat/helpers.py", line 61, in bust_cache
    return file_path + '?' + get_checksum(STATIC_PATH + file_path)
  File "/home/circleci/repo/webcompat/helpers.py", line 57, in get_checksum
    checksum = md5_checksum(file_path)
  File "/home/circleci/repo/webcompat/helpers.py", line 66, in md5_checksum
    with open(file_path, 'rb') as fh:
IOError: [Errno 2] No such file or directory: '/home/circleci/repo/webcompat/static/js/dist/webcompat.min.js'
```

This is happening because the test relies now  on this 

```
push(bust_cache('/js/dist/webcompat.min.js'), **{
```

AND this 

```
return file_path + '?' + get_checksum(STATIC_PATH + file_path)
```

which indeed will create an issue if `'/js/dist/webcompat.min.js'` has not been generated.

I need to dive in the test also look if it might create issues when running on localhost.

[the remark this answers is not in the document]
karlcow: There's another one which is slightly different and to be checked. 

```
FAIL: Page title format for different URIs.
  File "/home/circleci/repo/tests/unit/test_rendering.py", line 66, in test_titles
    self.assertTrue(expected in rv.data)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/ctx.py", line 243, in __exit__
    self.pop(exc_value)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/flask/ctx.py", line 235, in pop
    % (rv, self)
AssertionError: Popped wrong app context.  (<flask.ctx.AppContext object at 0x7fd8d6222350> instead of <flask.ctx.AppContext object at 0x7fd8d61e2c10>)
```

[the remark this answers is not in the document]
karlcow: That said there is something interesting.

pulling out @marimeireles branch

```
→ git checkout -b marimeireles-issues/2778/1 master
Switched to a new branch 'marimeireles-issues/2778/1'
🐝 14:28:33 ~/code/webcompat.com
→ git pull https://github.com/marimeireles/webcompat.com.git issues/2778/1
remote: Enumerating objects: 13, done.
remote: Counting objects: 100% (13/13), done.
remote: Total 17 (delta 11), reused 13 (delta 11), pack-reused 4
Unpacking objects: 100% (17/17), done.
From https://github.com/marimeireles/webcompat.com
 * branch            issues/2778/1 -> FETCH_HEAD
Merge made by the 'recursive' strategy.
 .circleci/config.yml    |  5 ++---
 config/requirements.txt |  1 +
 webcompat/__init__.py   |  5 +++++
 webcompat/views.py      | 50 ++++++++++++++++++++++++++++++++++++++++++++++++++
 4 files changed, 58 insertions(+), 3 deletions(-)
```

Then checking the tests.

```
🐝 14:28:54 ~/code/webcompat.com
→ nosetests
.............................................................................................................
Ran 109 tests in 2.437s

OK
```

Everything runs without issue.

It should not fail on CircleCI too.

I'm not an expert in CircleCI config. but I noticed that we activate twice the environment?

Here
https://github.com/webcompat/webcompat.com/blob/master/.circleci/config.yml#L31

and here
https://github.com/webcompat/webcompat.com/blob/master/.circleci/config.yml#L54

But this should not matter for nosetests.

This branch is:

```
      # run tests!
      - run:
          name: run tests
          command: |
            . venv/bin/activate
            pep8 --ignore=E402 webcompat/ tests/ config/secrets.py.example
            nosetests
            npm run lint
            npm run build
            python run.py -t &
            sleep 5
            node --max-old-space-size=8192 ./tests/functional/_intern.js --reporters="runner" --firefoxBinary=`which firefox`
            firefox --version
```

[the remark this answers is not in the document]
karlcow: Gotcha!

Partly Flawed tests.
Partly the modification of @marimeireles 

And a good lesson to learn and improve the tests.

Currently when we run nosetests on a localhost, we usually do it without erasing the static files.

Let me do that and run the tests.

```
🐝 14:49:28 ~/code/webcompat.com
→ rm -rf webcompat/static/js/*.js
🐝 14:49:53 ~/code/webcompat.com
→ rm -rf webcompat/static/js/build/
🐝 14:50:01 ~/code/webcompat.com
→ rm -rf webcompat/static/js/dist/
🐝 14:50:07 ~/code/webcompat.com
→ rm -rf webcompat/**/*.min.css
🐝 14:50:14 ~/code/webcompat.com
→ rm -rf webcompat/static/css/dist/
🐝 14:50:20 ~/code/webcompat.com
→ nosetests
.............................E.................EEEEEFE.........................EEEEEEEEEE..E.................
```

ooopsy…

OK let's go back to master.

```
🐝 14:50:25 ~/code/webcompat.com
→ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'webcompat/master'.

```
Then let's make sure the static files are not here.

```
🐝 14:56:12 ~/code/webcompat.com
→ find . -name webcompat.min.js
```

ok nothing

then let's run the tests.

```
🐝 14:56:41 ~/code/webcompat.com
→ nosetests
.............................................................................................................
Ran 109 tests in 1.166s

OK
```

working! no flawed tests.

So the results of the modifications by @marimeireles is that it broke tests because it pushes a resource which is not available when doing nosetests. Which means we included a new dependency that should not exist.

There are probably different ways to solve this. 

For example to use the push feature only when we are not in LOCALHOST.
OR to mock the files.

But these could be the topic of a new issue.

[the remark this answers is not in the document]
karlcow: @marimeireles has a summary.

You can move back nosetests after `npm run build` in the meantime. And we will find a solution later for it where the unit tests will not be dependent of building the secondary resources.

THANKS!

karlcow:
The rest looks good to me.

thanks a lot. Once you put back the order of nosetest, I'll approve and merge.

.circleci/config.yml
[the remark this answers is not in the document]
marimeireles: Ok!
Thanks for the clarifying review @karlcow. 
Do you want me to file an issue?

miketaylr:
@karlcow are we good to merge here? thanks.

karlcow:
I will double check the config on prod and staging today. And we could probably merge it.

.circleci/config.yml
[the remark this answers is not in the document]
karlcow: @marimeireles filing an issue would be good. yup. Thanks. "Make sure unittest are not dependent of compiled assets"

PR #2800, opened 2019-02-13
Fixes #2793 - Comments are sent as markdown to Github

### LINKED ISSUE #2793: comments are sent as HTML not markdown
I have the feeling that webcompat.com is not sending anymore markdown raw text to github but HTML.

It has plenty of consequences on escaping. See comments for example.
https://github.com/webcompat/web-bugs/issues/24472#issuecomment-460929940

I have typed markdown but this is what GitHub has stored.

```
<p>when I first access the page all the buttons have the same color.</p>
<p><a href="https://webcompat.com/uploads/2019/2/675cb7a1-ef72-4d02-ad56-be195940d0e4.jpg" rel="nofollow"><img src="https://webcompat.com/uploads/2019/2/675cb7a1-ef72-4d02-ad56-be195940d0e4-thumb.jpg" alt="Screenshot Description"></a></p>
<p>for me hovering change the color of the button BUT moving away the color doesn't stick. So the issue is not really about the reload but more about keeping the color. (i do not have a touch screen.)</p>
<p>Let me simulate a touch screen. Still the same thing.</p>
<p>so on tap or click, in chrome, the button seems to retain the focus while firefox doesn't</p>
<p>btw, doing</p>
<pre><code class="language-css">      .btn {
        background-color: #4285f4;
      }

      .btn:hover {
        background-color: red;
      }

      .btn:focus {
        background-color: pink;

        /* The outline parameter suppresses the border
        color / outline when focused */
      }

      .btn:active {
        background-color: green;
      }
</code></pre>
<p>makes it more visible.</p>
<p>Firefox keeps the focus only with tab navigation on the button.</p>
<p>Something here is doing testing in between platform.<br>
<a href="https://gist.github.com/cvrebert/68659d0333a578d75372" rel="nofollow">https://gist.github.com/cvrebert/68659d0333a578d75372</a></p>
<p>There are differences it seems in between Firefox on Windows and Firefox on macOS</p>
<p>I'm pretty sure there must be a bug on mozilla bugzilla for this.</p>
<p><a href="https://github.com/masayuki-nakano" rel="nofollow">@masayuki-nakano</a> might know which one.</p>
```

@marimeireles do we convert to HTML the comment before sending it? If yes we need to not do that.

marimeireles:
Locally they're rendered correctly in the website in the moment that they're posted:
screen shot 2019-02-13 at 16 35 17

Checking the code that is sent to Github:
screen shot 2019-02-13 at 16 35 04
it's markdown!

r? @karlcow

### DIFF
--- a/webcompat/static/js/lib/issues.js
+++ b/webcompat/static/js/lib/issues.js
@@ -530,6 +530,7 @@ issues.MainView = Backbone.View.extend(
           commentLinkId: null
         });
         this.addComment(newComment);
+        newComment.attributes.body = textarea.val();
         // Now empty out the textarea.
         textarea.val("");
         // Push to GitHub


webcompat/static/js/lib/issues.js
```diff
@@ -530,6 +530,7 @@ issues.MainView = Backbone.View.extend(
           commentLinkId: null
         });
         this.addComment(newComment);
+        newComment.attributes.body = textarea.val();
```
karlcow: What about doing the opposite? The only reason we have html is because of this.

```js
 body: this.sanitizeMarkdown(md.render(textarea.val())),
```

Why not changing the rendering of the text in `addComment` instead of `addNewComment`, and just use the text version in `addNewComment`



So we really separate what is done for posting the comment to GitHub and what is done for displaying the comment.

karlcow: ah probably not working because addComment is used in other contexts which are not the new comment and where the comment is already HTML.

ok let's see.

karlcow:
```js
    addNewComment: function(event) {
      var form = $(".js-Comment-form");
      var textarea = $(".js-Comment-text");

      if (form[0].checkValidity()) {
        event.preventDefault();
        var newComment = new issues.Comment({
          avatarUrl: form.data("avatarUrl"),
          body: textarea.val(),
          commenter: form.data("username"),
          createdAt: moment(new Date().toISOString()).fromNow(),
          commentLinkId: null
        });
        // Displays the comment for the user after formatting it as HTML
        var displayComment = newComment;
        displayComment.attributes.body = this.sanitizeMarkdown(md.render(textarea.val()));
        this.addComment(displayComment);
        // Push the markdown text comment to GitHub
        newComment.save();
      }
},
```

Do you think this would work?

marimeireles: I've just tested it and it doesn't. It's sending the comments as HTML.
Which is **really** weird, I don't know why it's doing it. Your code should work...

marimeireles: Humm... maybe when you call `addComment` you're overwriting the real value of newComment for the displayComment value?

marimeireles: I printed the value of the comment inside `addComment` and before `newComment.save()` and I think that's what's happening.

karlcow: ah yes my bad because the assignment is not creating a new object but just a pointer.
We would need to clone it. ok sorry for this.

marimeireles: You think we need to change something? @karlcow

karlcow: nope let's go with what you did. Thanks a lot.

PR #2804, opened 2019-02-14
Fixes #2771 - Removes unnecessary favicons

### LINKED ISSUE #2771: How to optimize our use of favicons?
Currently we have 5 or 6 different favicons and instead of downloading only the right one Firefox and Chrome download a bunch of different favicons. That sums up to more than 50kb and a bunch of new requests.

There are small res. favicons to show on the top of the tab and bigger ones to show on the new tab session, or when you download the website to be the icon of the website. 

According to Mike this might be a good low hanging fruit and we could just erase a bunch of those.

I want to learn more about how they work to understand which ones I can erase and if it's possible to optimize it downloading only the ones we're going to use. Mike said this would be a good starting point: https://mathiasbynens.be/notes/touch-icons

marimeireles:
Removed a few favicons that weren't necessary and were being downloaded when user was on Safari.

r? @miketaylr


webcompat/templates/layout.html
```diff
@@ -22,13 +22,8 @@
   <link rel="apple-touch-icon" sizes="152x152" href="{{ url_for('static', filename='favicon/apple-touch-icon-152x152.png') }}">
   <link rel="mask-icon" href="{{ url_for('static', filename='favicon/safari-pinned-tab.svg') }}" color="#202020">
 
-  <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-160x160.png') }}" sizes="160x160">
-  <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-96x96.png') }}" sizes="96x96">
-  <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-16x16.png') }}" sizes="16x16">
   <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-32x32.png') }}" sizes="32x32">
-  <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-196x196.png') }}" sizes="196x196">
   <meta name="msapplication-TileColor" content="#202020">
```
miketaylr: I think if we're removing one `msapplication-` thing, might as well remove this meta too.

miketaylr:
I want to be sure we're doing this for a principled reason and not just because there's a lot of them. @marimeireles do we have measurements of the before/after to justify deleting them, I guess in Safari since that's the browser with the bug?

miketaylr:
That said, I do agree that we're maintaining a lot of images just for the possibility that someone bookmarks the site on their phone or ipad or whatever, and wants a pretty icon.

webcompat/templates/layout.html
marimeireles: ok

marimeireles:
>I want to be sure we're doing this for a principled reason and not just because there's a lot of them

No, but I can do it. I'll post my results in the issue https://github.com/webcompat/webcompat.com/issues/2771, better than here, right?

miketaylr:
@marimeireles can we do a little bit of research and find the minimal viable set of favicon images required to not look like crap on regular and hi-dpi monitors? I'm inclined to suggest we just burn it all down and revisit app icons once we add some app-like functionality (via service workers, or anything compelling that would inspire you to save the site to your home screen)

marimeireles:
>I'm inclined to suggest we just burn it all down and revisit app icons once we add some app-like functionality (via service workers, or anything compelling that would inspire you to save the site to your home screen)

I thought we already had the functionality.

miketaylr:
> I thought we already had the functionality.

No, we just have an app manifest. No service workers.

webcompat/templates/layout.html
```diff
@@ -5,30 +5,9 @@
...
-  <link rel="apple-touch-icon" sizes="57x57" href="{{ url_for('static', filename='favicon/apple-touch-icon-57x57.png') }}">
-  <link rel="apple-touch-icon" sizes="114x114" href="{{ url_for('static', filename='favicon/apple-touch-icon-114x114.png') }}">
-  <link rel="apple-touch-icon" sizes="72x72" href="{{ url_for('static', filename='favicon/apple-touch-icon-72x72.png') }}">
-  <link rel="apple-touch-icon" sizes="144x144" href="{{ url_for('static', filename='favicon/apple-touch-icon-144x144.png') }}">
-  <link rel="apple-touch-icon" sizes="60x60" href="{{ url_for('static', filename='favicon/apple-touch-icon-60x60.png') }}">
-  <link rel="apple-touch-icon" sizes="120x120" href="{{ url_for('static', filename='favicon/apple-touch-icon-120x120.png') }}">
-  <link rel="apple-touch-icon" sizes="76x76" href="{{ url_for('static', filename='favicon/apple-touch-icon-76x76.png') }}">
-  <link rel="apple-touch-icon" sizes="152x152" href="{{ url_for('static', filename='favicon/apple-touch-icon-152x152.png') }}">
   <link rel="mask-icon" href="{{ url_for('static', filename='favicon/safari-pinned-tab.svg') }}" color="#202020">
```
miketaylr: Can we get rid of this one too?

webcompat/templates/layout.html
```diff
@@ -5,30 +5,9 @@
...
-
-  <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-160x160.png') }}" sizes="160x160">
-  <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-96x96.png') }}" sizes="96x96">
   <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-16x16.png') }}" sizes="16x16">
```
miketaylr: Can we keep the 32 by 32 instead of the 16 by 16?

marimeireles: I read in some places that the 32 was only used by Safari... But it isn't true, is it?
The browser will just scale the 32 to 16 if needed.

webcompat/templates/layout.html
marimeireles: ok.

karlcow:
no issue for me.

PR #2805, opened 2019-02-14
Fixes #2746 - Renders title with HTML code correctly

### LINKED ISSUE #2746: HTML tags inside the "code" markdown tag is rendered in webcompat.com
SeeAlso: https://webcompat.com/issues/23056 or webcompat/web-bugs#23056

In Markdown (especially [GitHub Flavored Markdown](https://github.github.com/gfm/)) we can add codes under a backtick character. Here, webcompat.com seems to render the HTML code inside these characters to produce this issue. For example, writing "`All HTML (backtick)<select>(backtick) choices are not visible`" will render as "All HTML `<select>` choices are not visible" in GitHub and

a real HTML select

in webcompat.com.

marimeireles:
This PR is behaving exactly like pointed in this https://github.com/webcompat/webcompat.com/issues/2746#issuecomment-463410928.

### DIFF
--- a/webcompat/static/js/lib/models/issue.js
+++ b/webcompat/static/js/lib/models/issue.js
@@ -91,13 +91,11 @@ issues.Issue = Backbone.Model.extend({
 
   getDescription: function(body) {
     // Get the Description of the body
-    var el = body;
-    el = $("strong:contains('Description')", el);
-    if (el[0]) {
-      var description = el[0].nextSibling.textContent;
-      description = description.replace(": ", "");
-      return description != null ? description.slice(0, 74) : null;
-    }
+    var div = document.createElement("div");
+    div.innerHTML = body;
+    var regex = /Description:(.+)\n/;
+    var description = regex.exec(div.textContent);
+    return description != null ? description[1].slice(0, 74) : null;
   },
 
   getDomain: function(title) {


miketaylr:
Kind of surprising that so many tests failed here.

webcompat/static/js/lib/models/issue.js
```diff
@@ -91,13 +91,11 @@ issues.Issue = Backbone.Model.extend({
...
   getDescription: function(body) {
     // Get the Description of the body
-    var el = body;
-    el = $("strong:contains('Description')", el);
-    if (el[0]) {
-      var description = el[0].nextSibling.textContent;
-      description = description.replace(": ", "");
-      return description != null ? description.slice(0, 74) : null;
-    }
+    var regex = /<strong>Description<\/strong>:(.*)+\n<strong>Steps to Reproduce<\/strong>:/gm;
```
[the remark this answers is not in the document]
marimeireles: Ohh, good point about the old issues.
Let me check.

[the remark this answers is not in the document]
marimeireles: Yeah.. It just breaks and doesn't render anything because of the `el[1]` thing.

This PR was made so Karl could check the code I was writing. I wasn't intending to merge it. Sorry for not making it clear. I'll keep your ideas in my, though.
Thanks for the review @miketaylr :)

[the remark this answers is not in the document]
marimeireles: I was thinking... Probably we'll have to write an `if` to treat this old issue case?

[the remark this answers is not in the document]
karlcow: Check the old code with blame. I was handling the old issues case. :)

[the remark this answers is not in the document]
marimeireles: Ok!

marimeireles:
I tested and it also works with older issues.
What do you think of this solution @miketaylr ? Is it really evil somehow?

miketaylr:
>  What do you think of this solution @miketaylr ? Is it really evil somehow?

Nah, not evil. But I think we should try the parsing text solution proposed by @karlcow.

See my comments in https://github.com/webcompat/webcompat.com/issues/2746#issuecomment-465259267 for some direction on how to grab that text at the right time.

webcompat/static/js/lib/models/issue.js
```diff
@@ -91,13 +91,10 @@ issues.Issue = Backbone.Model.extend({
...
   getDescription: function(body) {
     // Get the Description of the body
-    var el = body;
-    el = $("strong:contains('Description')", el);
-    if (el[0]) {
-      var description = el[0].nextSibling.textContent;
-      description = description.replace(": ", "");
-      return description != null ? description.slice(0, 74) : null;
-    }
+    body = document.createRange().createContextualFragment(body).textContent;
```
miketaylr: (this is a little weird, even if we wanted to go down this route, we could just create a `DocumentFragment` directly: https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment

miketaylr:
Actually, I hadn't refreshed this issue... It seems fine to do it this way (my review related to the previous approach), if you want to pursue it. But let's create the `DocumentFragment` directly, rather than use the Selection APIs.

webcompat/static/js/lib/models/issue.js
```diff
@@ -91,13 +91,11 @@ issues.Issue = Backbone.Model.extend({
...
-    if (el[0]) {
-      var description = el[0].nextSibling.textContent;
-      description = description.replace(": ", "");
-      return description != null ? description.slice(0, 74) : null;
-    }
+    var div = document.createElement("div");
+    div.innerHTML = body;
+    var regex = /Description:(.*)+\n/gm;
+    var description = regex.exec(div.textContent);
+    return description != null ? description[1].slice(0, 74) : null;
```
miketaylr: Since you're editing this code, can we return a more useful string in the off-chance this doesn't match (rather than null), something like "FIXME - No description found"?

webcompat/static/js/lib/models/issue.js
```diff
@@ -91,13 +91,11 @@ issues.Issue = Backbone.Model.extend({
...
-    var el = body;
-    el = $("strong:contains('Description')", el);
-    if (el[0]) {
-      var description = el[0].nextSibling.textContent;
-      description = description.replace(": ", "");
-      return description != null ? description.slice(0, 74) : null;
-    }
+    var div = document.createElement("div");
+    div.innerHTML = body;
+    var regex = /Description:(.*)+\n/gm;
```
miketaylr: I don't think my questions about the `gm` flags were answered from the first review. I don't think we need them (unless I misunderstand something?) https://github.com/webcompat/webcompat.com/pull/2805#discussion_r257047185

webcompat/static/js/lib/models/issue.js
karlcow: @miketaylr Probably not, because this will break this part of the code. 

https://github.com/webcompat/webcompat.com/blob/6ba66d1667646fec75f583e81c570eb5ab201f48/webcompat/static/js/lib/models/issue.js#L110

We handled the `null` with the default title as we were before.

marimeireles:
> I don't think my questions about the `gm` flags were answered from the first review. I don't think we need them (unless I misunderstand something?) [#2805 (comment)](https://github.com/webcompat/webcompat.com/pull/2805#discussion_r257047185)

Sorry I missed that.
Your comment makes sense. I'll erase it.

webcompat/static/js/lib/models/issue.js
karlcow: And there is probably a typo in there. `+\n` instead of `\n+`

My original suggestion was around, but if we stop after the first `\n`, but we could just probably do:

`/Description:(.+)\n/`

marimeireles:
Thanks @karlcow! 🐄
What do you think @miketaylr, does it look good to merge?

>Since you're editing this code, can we return a more useful string in the off-chance this doesn't match (rather than null), something like "FIXME - No description found"?

I could open a new issue to try to work on it. "Returning something when no description is found".

miketaylr:
> I could open a new issue to try to work on it. "Returning something when no description is found".

I don't think we need a new issue to add a string here. Instead of returning `null` at the end of the ternary, just add the new string.

webcompat/static/js/lib/models/issue.js
miketaylr: Ah, OK. Makes sense.

Issue #2750, opened 2018-12-28
Enable text compression


marimeireles:
I was running some profilers and PageSpeed gave me this result:
![screen shot 2018-12-28 at 10 53 10](https://user-images.githubusercontent.com/17600982/50515972-1591f980-0a8f-11e9-81c1-63915fe486f3.png)
 
I've already wrote a little bit about how to do that on my [proposal](https://psychonautgirl.space/webcompat.html#Future-implementations-in-details). It might be a good thing to start from these sources.

If PageSpeed is telling the true than we can get huge improvements for mobile users. As big as the ones removing markdown-it will do.

This is somewhat related to #2747 because there are some alternatives of doing this on the server side.

marimeireles:
Our currently gzip settings are:

```nginx
##
# Gzip Settings
##

gzip on;
gzip_disable "msie6";

gzip_vary on;
gzip_proxied any;
gzip_comp_level 6;
gzip_buffers 16 8k;
gzip_http_version 1.0;

# Turn on gzip for all content types that should benefit from it.
gzip_types application/ecmascript;
gzip_types application/javascript;
gzip_types application/json;
gzip_types application/pdf;
gzip_types application/postscript;
gzip_types application/x-javascript;
gzip_types image/svg+xml;
gzip_types text/css;
gzip_types text/csv;
# "gzip_types text/html" is assumed.
gzip_types text/javascript;
gzip_types text/plain;
gzip_types text/xml;
```
We're not doing anything specific to the fonts.
According to my research there are many different ways to improve font loading and I'll use this issue to record my results. I'll do a few different tests to see what will work best for us. 

The #2751 was specifically to add gzip compression to the server.

marimeireles:
Running this [tool](https://github.com/Munter/subfont#readme) will probably help.

marimeireles:
I was inspecting the fonts we use in webcompat.com and I found that we make two requests to get sourcesanpro webfont. https://fonts.googleapis.com/css?family=Open+Sans:400,600|Source+Sans+Pro:300,400|PT+Mono
It's the same request duplicated:
<img width="1082" alt="screen shot 2019-02-17 at 21 57 53" src="https://user-images.githubusercontent.com/17600982/52922313-2c223780-32ff-11e9-8e4a-8ccc0202d314.png">
I think something similar is happening to the requests to get opensans:
[...]

PR #2808, opened 2019-02-19
Issues #2750 - Improves text rendering

marimeireles:
This PR is probably not ready yet.
I got some doubts and I thought it'd be easier to clarify them if I posted a PR with my code.

What I've done here was add a few minor improvements to how we render text in webcompat.com for desktop & mobile and some bigger changes to how we do it on mobile. You can check my explanations of why I chose to do this here https://github.com/webcompat/webcompat.com/issues/2750

I think the more appropriated person to ping is @miketaylr


webcompat/static/css/src/typo.css
```diff
@@ -1,7 +1,7 @@
 body {
   color: var(--color-second);
   font-family: "Open Sans", sans-serif;
-  font-size: 16px;
+  font-size: 18px;
```
marimeireles: I shouldn't have changed that. Gonna fix it.

marimeireles:
Mike wants to see if it's worth it to fetch the fonts from our server first (https://github.com/webcompat/webcompat.com/issues/2809)
I don't think one thing excludes the other. If it's faster to fetch fonts from our server we can still use the solution I proposed here.

@miketaylr you think this PR is a bad idea because we're adding a new lib and maybe it's an overhead for our website? Like more things to maintain.

package.json
```diff
@@ -20,6 +20,8 @@
     "cssrecipes-grid": "^1.0.0",
     "cssrecipes-reset": "^0.5.0",
     "cssrecipes-utils": "^0.6.2",
+    "fontfaceobserver": "^2.1.0",
```
miketaylr: We wouldn't need to add these here, unless they're dependencies for our build tool chain.

webcompat/static/js/lib/fonts.js
```diff
@@ -0,0 +1,27 @@
+if (/Mobi/.test(navigator.userAgent)) {
```
miketaylr: I'm generally not a fan of this kind of UA sniffing... :/

webcompat/static/js/lib/fonts.js
```diff
@@ -0,0 +1,27 @@
...
+
+  var openSans400 = new window.FontFaceObserver("Open Sans", {
+    weight: 400
+  });
+
+  var openSans600 = new window.FontFaceObserver("Open Sans", {
+    weight: 600
+  });
+
+  Promise.all([
```
miketaylr: What about browsers that don't support Promises? This will likely throw in IE, https://caniuse.com/#feat=promises

webcompat/templates/layout.html
```diff
@@ -9,7 +9,7 @@
   <link rel="mask-icon" href="{{ url_for('static', filename='favicon/safari-pinned-tab.svg') }}" color="#202020">
   <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-16x16.png') }}" sizes="16x16">
 
-  <link rel="preload" href="https://fonts.googleapis.com/css?family=Open+Sans:400,600|Source+Sans+Pro:300,400|PT+Mono" as="style">
+  <link rel="preload" href="https://fonts.googleapis.com/css?family=Open+Sans:400,600|Source+Sans+Pro:300,400" as="style">
```
miketaylr: What about the corresponding change in https://github.com/webcompat/webcompat.com/blob/d56cd783366d725bcf2cf560723a590faf41e08f/webcompat/static/css/lib/prism-gh-theme.css#L10?

webcompat/static/js/lib/fonts.js
marimeireles: Oh... :/
We'd have to use some kind of sniffing to find out if the user in on mobile. Do you think that's a bad idea?
I could remove it and make it run on Desktop & mobile.

package.json
marimeireles: Makes sense. I thought we'd have to use it but we didn't and I forgot to delete.
Thanks Mike!

webcompat/static/js/lib/fonts.js
marimeireles: Right I haven't thought about that.
Do you think [jQuery.when()](https://api.jquery.com/jquery.when/) is a good solution?

marimeireles:
Even if we decide not to use font face observer I think we should still keep the @font-face thing. The explanation of why can be found in the first point of this https://github.com/webcompat/webcompat.com/issues/2750#issuecomment-464926948.

webcompat/static/js/lib/fonts.js
miketaylr: Yeah, that would cover our bases here.

miketaylr:
> Even if we decide not to use font face observer I think we should still keep the @font-face thing.

@marimeireles can you clarify what you mean by this comment? If it's something that can get merged without the fontfaceobserver, and improves things, I'm likely in favor. :)

marimeireles:
@miketaylr 

The changes proposed here improves webcompat's rendering significantly in slower connections. According to some tests I ran here https://github.com/webcompat/webcompat.com/issues/2750#issuecomment-465244617 the page spends 25% of its loading time loading fonts, that means that during 25% of the time the user is looking for a page with no relevant information since we're not providing any text to them.
The problem is that we're just interested in using this FOUT solution in mobiles because for faster connections this enhancement doesn't make a lot of difference and looks very ugly. (I was trying to record my Nexus screen while it was loading but it didn't work, let me know if you want to see how it looks and I can try to do it with another smartphone!)

After learning more about its consequences I decided to not use UA sniffing here and I think Karl and Mike agree that this is a bad idea. Therefore either we use this solution both on mobile and desktop or we don't use it.

If we decide not to use FOUT I think we should still keep the other changes in this PR: (https://github.com/webcompat/webcompat.com/pull/2808/commits/b5f0638df4fdc9ac888ebae260f41e6c96f9acc4) and (https://github.com/webcompat/webcompat.com/pull/2808/commits/1b9bc9b200e82718faa227c79529eb0c02d96493), where I remove a font request to `PT+Mono` and add `font-face` configuration.

> If it's something that can get merged without the fontfaceobserver, and improves things, I'm likely in favor. :)

I'm going to make a PR with no `fontFaceObserver` things :)

Issue #2809, opened 2019-02-20
Test if fetching the fonts from our server is faster than fetching from Google


marimeireles:
This issue is somewhat related to https://github.com/webcompat/webcompat.com/issues/2750.

marimeireles:
* We don't have to add anything to our mime.types file because it already has the woffs formats by default.

I tested and found this tool really helpful to download the correct set of fonts: https://github.com/fontello/fontello/wiki/How-to-setup-server-to-serve-fonts

You just have to run the command:
```
./google-font-download \
    "'Open Sans:400'" "'Open Sans:600'" \
    "'Source Sans Pro:300'" "'Source Sans Pro:400'"
```

marimeireles:
Currently this are our stats for dealing with text in the new issue page:
(time in ms, median of 5 tests on Chrome)

CSS request time: 141
CSS size: 964b
Fonts request time: 112
Fonts size: 29.6kb
First paint: 1309
Fully loaded: 3587
First visual change: 1333

Time on mobile to load the font since first paint:
546ms in a internet speed of 500~600kbps
848ms in a internet speed of 200~300kbps

marimeireles:
I found a [utility](https://github.com/filamentgroup/glyphhanger#installing-pyftsubset) capable of subsetting the webfonts we use based on the chars that exists in webcompat.com. I was able to significantly decrease the size of the files!

<img width="971" alt="screen shot 2019-02-21 at 10 37 37" src="https://user-images.githubusercontent.com/17600982/53172817-0332c880-35c5-11e9-8356-c83d9357bcca.png">

@karlcow let me know if you want me to send you these subsetted files somehow or if you want to run the tool yourself.

karlcow:
I'll check that on Monday (JST).

karlcow:
(just for memory)
downloading font locally also means that there is a risk in the future that bugs in the fonts are not fixed, because it's… a local copy.

@marimeireles which kind of subsetting do you want to do on the fonts.

> based on the chars that exists in webcompat.com.

which chars exists?

We already have a static font directory. :) Do you need something else? Apart of the fonts inside.
https://github.com/webcompat/webcompat.com/tree/master/webcompat/static/font

marimeireles:
> downloading font locally also means that there is a risk in the future that bugs in the fonts are not fixed, because it's… a local copy.

Yes. That's a downside we have to consider. If hosting fonts doesn't make a lot of difference I don't think we should do it.

[...]

PR #2815, opened 2019-02-27
Issue #2809 - Add fonts to our server

marimeireles:
Hey @karlcow can you please review my code and if it's everything alright send these changes to Staging?

What I'm trying to do here is to measure how much we'd improve our font loading fetching from the server comparing to fetching from Google. 

(I already have the tests for fetching on Google.)


marimeireles:
Our linter tool has a bug:

screen shot 2019-02-27 at 14 25 59

If I add a space it tells me to remove the space if I remove the space it tells me to add a space :p
There is an issue filled here https://github.com/stylelint/stylelint/issues/3270, but I don't think it's fixed.

I've fixed it adding an ugly css.

karlcow:
So I'm not merging (for now) just pushing the branch to stagiing.

karlcow:
@marimeireles this branch is currently deployed on staging.

marimeireles:
Thanks @karlcow.

The way this is working now is not worth it. We spend a really long time stuck waiting for other process to happening. But one thing surprised me: we're able to download the fonts approximately as fast as we used to download them from Chrome.

screen shot 2019-02-28 at 16 12 47

I'd like to test how much we can improve by pushing the fonts. I researched all around the web and I don't think there is a way to do it other than UA sniffing what's the browser the person is using. What do you think @miketaylr? Can we do that? If we don't do it than I don't think we should serve our own fonts because they represent a significantly drop in performance. We could use the `link=preload` thing but it only works for newer versions of Chrome, so it's not worth it either.

marimeireles:
ping @miketaylr

marimeireles:
I'm going to undo the changes I did here. It's not worth it to host our fonts and I think UA sniffing is something we shouldn't use.
We can try to use this solution once [preload](https://bugzilla.mozilla.org/show_bug.cgi?id=1222633) is added to FIrefox. But I don't think it's worth it otherwise.

marimeireles:
@karlcow Can you please undo the changes this PR makes?
(Once you're back from PTO!)

karlcow:
@marimeireles not that the branch has been approved and not merged. So we can just safely close it. and redeploy staging.

Thanks a lot for exploring this path. It was cool to be working with you in the context of this project.

PR #2821, opened 2019-03-15
Fixes #2750 - Improves text rendering

### LINKED ISSUE #2750: Enable text compression

marimeireles:
This PR follows the previous discussion we had here: https://github.com/webcompat/webcompat.com/pull/2808
I decided to create a new version so we still keep a log from the discussion we had about trying to use the `fontfaceobserver`.

There are two changes in this PR.
1. I removed the PT+Mono webfont because Mike and I agreed it wasn't worth it to make a request for that font. @karlcow what do you think?
2. Added some configuration to our fonts using `@font-face`. The reason why I added it can be found in this https://github.com/webcompat/webcompat.com/issues/2750#issuecomment-464926948 but summarizing I think the `font-display: optional` option can have a huge impact for users in slow internet connections showing text for them much earlier + this is super easy to implement.

@miketaylr these are the changes I proposed here https://github.com/webcompat/webcompat.com/pull/2808#issuecomment-473409855.

### DIFF
--- a/webcompat/templates/layout.html
+++ b/webcompat/templates/layout.html
@@ -8,7 +8,7 @@
   <link rel="shortcut icon" href="{{ url_for('static', filename='favicon.ico') }}">
   <link rel="icon" type="image/png" href="{{ url_for('static', filename='favicon/favicon-32x32.png') }}" sizes="32x32">
 
-  <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,600|Source+Sans+Pro:300,400|PT+Mono" rel="stylesheet">
+  <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,600|Source+Sans+Pro:300,400" rel="stylesheet">
   <link href="{{ url_for('static', filename='css/dist/webcompat.min.css') }}" type="text/css" rel="stylesheet">
 </head>
 <body id="body-webcompat" data-username="{{ session.username }}">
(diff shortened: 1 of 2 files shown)


marimeireles:
wow this is weird. Why my PR isn't being tested?

karlcow:
> wow this is weird. Why my PR isn't being tested?

indeed.

webcompat/templates/layout.html
```diff
@@ -46,3 +46,4 @@
 {% block extrascripts %}{% endblock %}
 </body>
 </html>
+</html>
```
karlcow: Hmm this doesn't seem to be a legit thing 😜

marimeireles: hehe
Sorry for that!

marimeireles:
@karlcow 🐄

karlcow:
yoohoo @marimeireles Hello and welcome back.

PR #2827, opened 2019-03-27
Fixes #2759 - Adds more files to ignore on deploy

### LINKED ISSUE #2759: deploy.ignore needs to be updated
@miketaylr These are a list of things we could add to `deploy.ignore`

```
.circleci/
.editorconfig
.eslintrc.json
.github/
.gitignore
.prettierrc
.stylelintrc
config/secrets.py.example
CODE_OF_CONDUCT.md
package-lock.json
run.py
```

as well sorting alphabetically  the current one or any kind of logical order. :)

```
*.pyc
.git/
CHANGELOG.md
CONTRIBUTING.md
data/
docs/
env/
fabfile.py*
git_modules/
grunt-tasks/
Gruntfile.js
icon-licenses/
Makefile
node_modules/
package.json
README.md
screenshots/
selenium*
session.db
tests/
tmp/
uploads/
webcompat.sh
```

karlcow:
This PR fixes issue #2759

## Proposed PR background

We are uploading a couple of unnecessary thing to the server.

r? @miketaylr


config/deploy.ignore
```diff
@@ -1,23 +1,36 @@
...
+session.db
 tests/
-icon-licenses/
-CHANGELOG.md
-CONTRIBUTING.md
-webcompat.sh
-uploads/
 tmp/
+uploads/
+webcompat.sh
```
miketaylr: This could be removed, it doesn't exist anymore.

config/deploy.ignore
```diff
@@ -1,23 +1,36 @@
...
-node_modules/
 Gruntfile.js
+icon-licenses/
 Makefile
+node_modules/
+package-lock.json
 package.json
 README.md
-session.db
+run.py
```
miketaylr: What's the rationale for removing this? I've used it in the past to quickly see if I could get the app to run on staging, when debugging weird issues.

(I guess we can still start the app via `flask run` with `FLASK_APP` or similar?)

config/deploy.ignore
karlcow: ok I'll do remove this one from the deploy and double check it's not on the server too.

config/deploy.ignore
karlcow: Ah! I didn't you were using it locally. The server is not using it indeed in a normal operation. 

You could run indeed with

```
FLASK_APP=webcompat flask run
```

I wonder if we should get rid of run.py locally too to avoid confusion and be closer from what the server needs.

karlcow: ah wait…
https://github.com/webcompat/webcompat.com/blob/c58c8ee45cc327bd585114c86b103ae6dbaf318e/.circleci/config.yml#L59

karlcow: and 

https://github.com/webcompat/webcompat.com/blob/c58c8ee45cc327bd585114c86b103ae6dbaf318e/package.json#L75-L76

and 

https://github.com/webcompat/webcompat.com/blob/c58c8ee45cc327bd585114c86b103ae6dbaf318e/tests/functional/lib/setup.js#L25-L39

karlcow: Let me remove this from the deploy.ignore.

PR #2829, opened 2019-03-28
Fixes #2828 - Changes from pep8 to pycodestyle

### LINKED ISSUE #2828: Update from pep8 to pycodestyle
see https://github.com/webcompat/webcompat.com/pull/2825#issuecomment-477397243
for the context but basically pep8 has been renamed pycodestyle to avoid confusion with the real PEP8.

karlcow:
This PR fixes issue #2828

## Proposed PR background

pep8 has changed name and is deprecated. We need to switch to pycodestyle (new name for pep8). This created a version bump  with a couple of things to fix, choose. 

This adjusts the 

* circleci configuration
* the code itself
* The documentation

The tests are running on my local computer.

The biggest change in the code is probably 
https://github.com/webcompat/webcompat.com/commit/cba11936476c7dd0a6a8645713d64f34653a8dfe

that I have singled out. The rest is following basically the recommendation of pycodestyle.
It will also make it possible to test the code on #2825 project, where circleci requires pycodestyle https://github.com/webcompat/webcompat.com/pull/2825#issuecomment-477397243

r? @miketaylr

### DIFF
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -414,7 +414,7 @@ def extract_url(issue_body):
     URL in webcompat.com bugs follow this pattern:
     **URL**: https://example.com/foobar
     """
-    url_pattern = re.compile('\*\*URL\*\*\: (.+)\n')
+    url_pattern = re.compile(r'\*\*URL\*\*\: (.+)')
     url_match = re.search(url_pattern, issue_body)
     if url_match:
         url = url_match.group(1).strip()
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -501,19 +501,20 @@ def add_csp(response):
     This should be used in @app.after_request to ensure the header is
     added to all responses.
     """
-    response.headers['Content-Security-Policy'] = (
-        "default-src 'self'; " +
-        "object-src 'none'; " +
-        "connect-src 'self' https://api.github.com; " +
-        "font-src 'self' https://fonts.gstatic.com; " +
-        get_img_src_policy() +
-        "manifest-src 'self'; " +
-        "script-src 'self' https://www.google-analytics.com https://api.github.com 'nonce-{nonce}'; ".format(nonce=request.nonce) +  # noqa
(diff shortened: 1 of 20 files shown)


webcompat/helpers.py
```diff
@@ -414,7 +414,7 @@ def extract_url(issue_body):
     URL in webcompat.com bugs follow this pattern:
     **URL**: https://example.com/foobar
     """
-    url_pattern = re.compile('\*\*URL\*\*\: (.+)\n')
+    url_pattern = re.compile(r'\*\*URL\*\*\: (.+)')
```
karlcow: just explaining here the why. This was W605 invalid escape sequence ‘x’
The explanation is https://lintlyci.github.io/Flake8Rules/rules/W605.html

https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

> Both string and bytes literals may optionally be prefixed with a letter 'r' or 'R'; such strings are called raw strings and treat backslashes as literal characters. As a result, in string literals, '\U' and '\u' escapes in raw strings are not treated specially. Given that Python 2.x’s raw unicode literals behave differently than Python 3.x’s the 'ur' syntax is not supported.

webcompat/helpers.py
```diff
@@ -501,19 +501,20 @@ def add_csp(response):
...
+        get_img_src_policy(),
+        "manifest-src 'self'; ",
+        "script-src 'self' https://www.google-analytics.com https://api.github.com 'nonce-{nonce}'; ".format(nonce=request.nonce),  # noqa
+        "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; ",
+        "base-uri 'self'; ",
+        "frame-ancestors 'self'; ",
         "report-uri /csp-report"
-    )
+    ]
+    response.headers['Content-Security-Policy'] = (''.join(csp_params))
```
karlcow: I was not forced to do that as I chose to ignore W504, but this is more performant. So this is a win-win change.

karlcow:
@miketaylr could we merge this?

PR #2834, opened 2019-04-04
Fixes #1838 - Separates the python dependencies into prod and dev

### LINKED ISSUE #1838: Separate requirements.txt into dev-req.txt and production-req.txt
Currently we have a unique requirements.txt
Some of the modules are not useful once the application is deployed in production. Let's separate these in two files, so that when we deploy we only consider what we really need for running the application.

karlcow:
This PR fixes issue #1838

## Proposed PR background

This is the first step of properly separating the production and the development environment. Once this is done, we can think of ways to modify the fabfile so it deploys only the prod environment. 

r? @miketaylr

### DIFF
--- a/config/requirements-dev.txt
+++ b/config/requirements-dev.txt
@@ -1,0 +1,5 @@
+# This is for dev purpose
+-r requirements.txt
+mock==2.0.0
+nose==1.3.7
+pycodestyle==2.5.0
--- a/run.py
+++ b/run.py
@@ -16,7 +16,7 @@ IMPORT_ERROR = '''
 ==============================================
 It seems like you don't have all dependencies.
 Please re-run:
-    pip install -r config/requirements.txt
+    pip install -r config/requirements-dev.txt
 ==============================================
 '''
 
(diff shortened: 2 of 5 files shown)


karlcow:
This also requires to have #2829 merged BEFORE this one.

run.py
```diff
@@ -16,7 +16,7 @@
 ==============================================
 It seems like you don't have all dependencies.
 Please re-run:
-    pip install -r config/requirements.txt
+    pip install -r config/dev-requirements.txt
```
miketaylr: This should say `requirements-dev.txt`.

config/requirements-dev.txt
```diff
@@ -0,0 +1,5 @@
+# This is for dev purpose
+-r requirements.txt
```
miketaylr: cool, I didn't know you could do this.

https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format

run.py
karlcow: 🤦‍♂️ thanks.

PR #2843, opened 2019-04-15
Fixes #2839 - Changes checksum algorithm for fixtures. Updates fixtures.

### LINKED ISSUE #2839: fixtures files and checksum function need to be fixed.
This comes up in my attempts to make our code python 3 ready.
https://github.com/webcompat/webcompat.com/pull/2825#issuecomment-482034803

Basically the function which is computing the checksum for the fixture files has variability depending on the python version/environment. The checksum is based on a stringified dictionary, but dictionaries are not ordered by default, so the checksum is variable. 

We need to change this in a way that is more stable. Probably the URI would be more stable for the specific case. The pain in adjusting this is that we will need to rename a couple of files with new checksums. 

Probably needs to recheck #1717 too

karlcow:
This PR fixes issue #2839

## Proposed PR background

This changes the way we compute the checksum for fixture files.
Instead of being based on parameters (which was creating variability. see the comments in #2839), we compute the fixture checksum on the URL full path of the HTTP request.

r? @miketaylr


webcompat/helpers.py
```diff
@@ -383,25 +383,31 @@ def mockable_response(func):
     """Mock out API reponses with a decorator.
 
     This allows us to send back fixture files when in TESTING mode, rather
-    than making API requests over the network. See /api/endponts.py
+    than making API requests over the network. See /api/endpoints.py
```
miketaylr: endponts sounds cooler tho.

webcompat/helpers.py
```diff
@@ -383,25 +383,31 @@ def mockable_response(func):
...
+                checksum = hashlib.md5(full_path).hexdigest()
                 file_path = FIXTURES_PATH + request.path + "." + checksum
             else:
                 file_path = FIXTURES_PATH + request.path
-            if not os.path.exists(file_path + '.json'):
-                print('Expected fixture file: ' + file_path + '.json')
+            full_file_path = file_path + '.json'
+            if not os.path.exists(full_file_path):
+                print('Fixture expected at: {fix}'.format(fix=full_file_path))
+                print('by the http request: {req}'.format(req=request.url))
```
miketaylr: 👍 good addition.

Issue #2856, opened 2019-04-26
Consider adding `engine-gecko` label for Firefox browsers?


miketaylr:
A proposal -- curious about other people's input here.

I would like us to add a label -- the exact wording doesn't matter to me -- that captures that the browser is a Mozilla-produced product running on Gecko.

The use case is I want to track all issues on webcompat.com that affect Gecko-based Firefox browsers.

Currently, GitHub search doesn't allow logical OR searches. 

So instead of having something like `... label:browser-firefox OR label:firefox-mobile..`, etc., I need to do 6 queries. In addition to firefox and firefox-mobile, we also have `browser-focus-geckoview` and `browser-geckoview` and `browser-firefox-reality`, and `browser-fenix`.

Thoughts? @karlcow @adamopenweb @webcompat/core-developers ?

wisniewskit:
Would this be intended to include Gecko browsers like Waterfox (for the rare report we get in that regard), or just the Mozilla releases?

miketaylr:
> Would this be intended to include Gecko browsers like Waterfox (for the rare report we get in that regard), or just the Mozilla releases?

I thought about this, and I'm not 100% sure. If it's about UA sniffing, probably not. But bugs that stem from core Gecko platform differences would be good to capture.

tagawa:
The concept sounds good to me. My belief is that labels should be honest about what they represent, so a `browser-gecko` label would include all Gecko browsers, e.g. Waterfox. If it's just for Mozilla browsers then `browser-mozilla` seems a better label IMO.

Alternatively, existing labels seem to have a kind of hierarchical naming, e.g. `browser-firefox` and `browser-firefox-mobile`, so if it's for Mozilla-only Gecko browsers then how about `browser-mozilla-gecko` or `browser-gecko-mozilla`?

karlcow:
> Alternatively, existing labels seem to have a kind of hierarchical naming

There's a plan to get away from this 🤣 but I didn't get the time to punch it through. 
see #997 and #2531

karlcow:
* engine-gecko
* engine-blink
* engine-webkit

are the 3 main engines which still exist.

Maybe a bit of engine-presto for Opera Mini, but I'm not even sure. 

The most difficult part is how to identify them, because they usually don't advertise themselves. 
A google example is Firefox Focus on Android which has both a webkit and a geckoview.

I'm interested to understand the false positives we would get with this system.

miketaylr:
[...]

Issue #2531, opened 2018-06-28
Implement webhook to categorize browser-*, os-*, form-* with labels


karlcow:
The webhook needs to label for 
* [ ] browser-[name] or browser-fixme for unknown browsers
* [ ] os-{windows,macos,linux,android,ios}
* [ ] form-{mobile,desktop,tablet}
* [ ] version-nnn

miketaylr:
Per @karlcow's request, adding a note here that we should attempt to not break `engine-gecko` when we do this (possibly by re-writing how that is done).

Unrelated, we probably also want `form-vr` too!

PR #2858, opened 2019-04-29
Issue #2856 - Add 'engine-gecko' label for Mozilla browsers

miketaylr:
Not sure the test I added adds much value...

r? @karlcow

### DIFF
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -83,6 +83,11 @@ class TestWebhook(unittest.TestCase):
         **Tested Another Browser**: Yes
         """  # noqa
 
+        self.issue_body6 = u"""
+        **URL**: https://www.not-a-mozilla-browser.com/
+        <!-- @browser: Safari 99.0 -->
+        """
+
     def tearDown(self):
         """Tear down tests."""
         pass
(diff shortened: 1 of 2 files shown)


tests/unit/test_webhook.py
```diff
@@ -83,6 +83,11 @@ def setUp(self):
         **Tested Another Browser**: Yes
         """  # noqa
 
+        self.issue_body6 = u"""
+        **URL**: https://www.not-a-mozilla-browser.com/
```
karlcow: did you mean not gecko? ;)

```suggestion
        **URL**: https://not-gecko.example.com/
```

webcompat/webhooks/helpers.py
```diff
@@ -18,6 +18,14 @@
 from webcompat.helpers import proxy_request
 
 BROWSERS = ['blackberry', 'brave', 'chrome', 'edge', 'firefox', 'iceweasel', 'ie', 'lynx', 'myie', 'opera', 'puffin', 'qq', 'safari', 'samsung', 'seamonkey', 'uc', 'vivaldi']  # noqa
+MOZILLA_BROWSERS = ['browser-fenix',
+                    'browser-firefox',
```
karlcow: This will fail when we separate firefox from the hierarchical structure.
[browser-firefox-ios](https://github.com/webcompat/web-bugs/issues?q=is%3Aissue+label%3Abrowser-firefox-ios) will become `browser-firefox` + `os-ios`.

Maybe we need to add a note on #2531 so we remember that to not break `engine-*`

tests/unit/test_webhook.py
miketaylr: good catch, :)

webcompat/webhooks/helpers.py
miketaylr: Done.

miketaylr:
OK, nit addressed. I'm going to manually add the engine-gecko label to existing bugs now, and I can double-check on ones that got missed between now and whenever this gets deployed. Thanks.

karlcow:
Thanks @miketaylr Merged!

PR #2874, opened 2019-05-17
Fixes #2873 - Add AB testing plumbing

### LINKED ISSUE #2873: Add support for AB testing in webcompat.com
As part of our (Open Innovation) experiments to test some UX changes in the reporting flow, it would be useful to be able to use some sort of AB testing capabilities in webcompat.com.

johngian:
This PR fixes issue #2873 

## Proposed PR background

Adds lightweight AB testing capabilities in webcompat.com (both server side and client side).

### DIFF
--- a/config/environment.py
+++ b/config/environment.py
@@ -51,6 +51,8 @@ if LOCALHOST:
     UPLOADS_DEFAULT_DEST = BASE_DIR + '/uploads/'
     UPLOADS_DEFAULT_URL = 'http://localhost:5000/uploads/'
 
+# Enable/disable AB testing
+AB_TESTING = os.environ.get('AB_TESTING') == 'True'
 
 # BUG STATUS
 # The id will be initialized when the app is started.
--- a/tests/unit/test_helpers.py
+++ b/tests/unit/test_helpers.py
@@ -12,6 +12,9 @@ import unittest
 import flask
 
 import webcompat
+from webcompat.helpers import ab_active
+from webcompat.helpers import ab_redirect
+from webcompat.helpers import ab_view
 from webcompat.helpers import form_type
 from webcompat.helpers import format_link_header
 from webcompat.helpers import get_browser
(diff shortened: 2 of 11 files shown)


tests/unit/test_helpers.py
```diff
@@ -316,6 +319,80 @@ def test_json_object(self):
...
+        with webcompat.app.test_request_context(
+                '/?foo=bar',
+                method='GET'):
+
+            def _bar():
+                return 'bar'
+
+            @ab_view('foo', 'bar', _bar)
+            def _foo():
+                return "foo"
```
miketaylr: nit: we prefer single quotes for consistency, here and below in a couple places

webcompat/static/js/lib/ab.js
```diff
@@ -0,0 +1,11 @@
+/*global Mozilla experiment:true*/
+/*eslint no-undef: "error"*/
+
+var experiment = new Mozilla.TrafficCop({
+  id: "issue-form-experiment",
+  variations: {
+    "experiment=issueFormExperiment": 0
```
miketaylr: 0 means nobody will get the variation, correct?
(presumably we would change this to a meaningful value once we have the form variations built and ready to test)

config/environment.py
```diff
@@ -51,6 +51,8 @@
     UPLOADS_DEFAULT_DEST = BASE_DIR + '/uploads/'
     UPLOADS_DEFAULT_URL = 'http://localhost:5000/uploads/'
 
+# Enable/disable AB testing
+AB_TESTING = os.environ.get('AB_TESTING') == 'True'
```
miketaylr: I'm not sure when Flask loads this, presumably we need to restart the Flask app for this to take effect.

webcompat/static/js/lib/ab.js
miketaylr: Reading https://github.com/mozilla/trafficcop#traffic-cop is helpful here.

tests/unit/test_helpers.py
```diff
@@ -316,6 +319,80 @@ def test_json_object(self):
         # A JSON value, which is not an object
         self.assertFalse(is_json_object(json.loads('null')))
 
+    def test_ab_active_true(self):
+        """Check if we return the right view based on AB params."""
```
karlcow: All the tests docstrings are the same. It would be good to make them more explicit of what the test is checking.

tests/unit/test_helpers.py
```diff
@@ -316,6 +319,80 @@ def test_json_object(self):
...
+        cookie = 'foo=bar; Path=/'
```
karlcow: Could we avoid the foo, foobar, bar everywhere? It makes the tests harder to read. Replace them with a value which looks more real(?).

webcompat/static/js/lib/ab.js
johngian: Yes, I initialized it with 0 just as a placeholder for the experiments.

config/environment.py
johngian: I am not sure how the app is deployed but I would assume that it needs to be restarted.

johngian:
> There is something I'm not clear with the way this is working. All the tests implies it relies on a GET request with a parameter. While it's possible for people to directly file a bug through direct access to the Web site. We also use HTTP POST with a JSON payload to populate the form when it comes through the web extension. Do you think it will keep working as-is?
> 
> Thanks for the contribution!

Hey @karlcow, the way I implemented it, AB testing is currently requiring the user to actually visit the website (since it's cookie-based). The current functionality will continue to work as it is, except in case the user sends a GET/POST request using the experiment's query params eg:

```
curl -d '{"key1":"value1", "key2":"value2"}' -H "Content-Type: application/json" -X POST "https://httpbin.org/post?exp=var-v1"
```

config/environment.py
karlcow: yes. It needs to be restarted.

karlcow:
@johngian thanks for the modifications. Appreciated.

> Hey @karlcow, the way I implemented it, AB testing is currently requiring the user to actually visit the website (since it's cookie-based). The current functionality will continue to work as it is, except in case the user sends a GET/POST request using the experiment's query params or customizing the cookie headers eg:

Let's deploy on staging and we can test manually if it does what it is supposed to do for most of our reports coming through a JSON payload.

To clarify what I meant in the previous comment. 

The report site issue button (most of our reports) in Firefox Nightly is 
1. sending a HTTP POST with a JSON payload to webcompat.com/issues/new 
2. In returns generate a prepopulated form web page with the JSON data in the payload.
3. The user can then complete.
4. And finally push the submit button making an HTTP post to the server to create an actual bug report.

So I wonder how the current AB code will work in this context. I'm curious.
Because if I understand the https://github.com/mozilla/trafficcop#example-flow the user has first to visit the site to get the cookie of the experiment, but this is not the case for reports submitted through the report site button. 

The AB testing I guess would apply mainly to a small portion of the population who are already using the site directly. I can try to draw stats about this.

karlcow:
Another thought should Core team (softvision, us) be part of the AB testing (they file a lot of bugs) and it kind of distort the AB testing.

miketaylr:
> Another thought should Core team (softvision, us) be part of the AB testing (they file a lot of bugs) and it kind of distort the AB testing.

Yeah, we should filter out our reports before attempting any analysis.

karlcow:
I will deploy on staging.

PR #2880, opened 2019-05-23
Revert "Fixes #2873 - Add AB testing plumbing"

### LINKED ISSUE #2873: Add support for AB testing in webcompat.com

karlcow:
Reverts webcompat/webcompat.com#2874


karlcow:
Let me revert master to its state before AB. No need to deploy AB testing in master without first having a way to really test it.

miketaylr:
So is the plan to re-merge once we have another UI we can test on staging?

Once that's on staging, we can change `extensions.webcompat-reporter.newIssueEndpoint` to https://staging.webcompat.com and go from there.

miketaylr:
(@karlcow i'll let you merge this one)

karlcow:
> So is the plan to re-merge once we have another UI we can test on staging?

yes and something that will work with on HTTP post.

PR #2908, opened 2019-08-16
Fixes #2905 - Adjust documentation and command for python 3.7

### LINKED ISSUE #2905: Adjust documentation for Python 3 development
This is related to #1332 

We need to rewrite a bit of the documentation so it makes sense in the context of #2348 landing on master.

karlcow:
This PR fixes issue #2905

## Proposed PR background

Adjust the documentations and commands for taking into account python3.7


docs/dev-env-setup.md
```diff
@@ -18,50 +21,30 @@
...
 
-# installing
-curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash -
-sudo apt-get install -y nodejs
-sudo apt-get install -y build-essential
-```
+#### Install nodejs
 
-### Simple setup
+@@magsout@@
```
miketaylr: I'm lazy, so I use homebrew to install node.

package.json
```diff
@@ -74,7 +74,7 @@
     "prestart": "npm run build",
     "start": "source env/bin/activate || . env/bin/activate && python run.py",
     "start:test": "npm run build && source env/bin/activate || . env/bin/activate && python run.py -t",
-    "virtualenv": "pip install virtualenv && virtualenv -p python env && source env/bin/activate || . env/bin/activate && npm run pip",
+    "virtualenv": "python3 -m venv env && source env/bin/activate || . env/bin/activate && npm run pip",
     "pip": "pip install -r config/requirements-dev.txt",
```
miketaylr: Do we want to point to `pip3` here? Or just keep as pip?

We need to update the docs in the detailed section as well, because they recommend `pip2`: https://github.com/webcompat/webcompat.com/blob/e1609766663673f56d547240dd7a893adec45fcc/docs/dev-env-setup.md#installing-pip

miketaylr:
I tried to follow https://github.com/webcompat/webcompat.com/blob/e1609766663673f56d547240dd7a893adec45fcc/docs/dev-env-setup.md#installing-pip and ran into an issue (from a fresh clone):

If I run `python run.py` or `npm run start`, I get the following:
```
  File "run.py", line 41, in <module>
    raise ImportError('{0}\n\n{1}'.format(e, IMPORT_ERROR))
ImportError: No module named flask


Please re-run:
    pip install -r config/requirements-dev.txt
```

If I type `python3 run.py` all is well. Can we update that as well?

miketaylr:
(i tried to push a commit, but goofed. so here's a patch with the install nodejs bits:

```
From 70cc965b393a1bfda6081731a52bb90daef6b6ad Mon Sep 17 00:00:00 2001
From: Mike Taylor <miket@mozilla.com>
Date: Mon, 19 Aug 2019 10:30:22 -0500
Subject: [PATCH] Issue #2905 - Add instructions for NodeJS installation

---
 docs/dev-env-setup.md | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/docs/dev-env-setup.md b/docs/dev-env-setup.md
index 2cb5622b..3a52be90 100644
--- a/docs/dev-env-setup.md
+++ b/docs/dev-env-setup.md
@@ -35,11 +35,11 @@ Go to https://github.com/webcompat/webcompat.com/ and click on the Fork button (
 
 #### Install Python 3.7
 
-This project requires Python 3. This is not available per default on most platforms. You will need to install it on your computer. The [download section](https://www.python.org/downloads/) of Python website has binaries for most platforms. Choose the 3.7.3 or more version minimum.
+This project requires Python 3. This is not available per default on most platforms. You will need to install it on your computer. The [download section](https://www.python.org/downloads/) of the Python website has binaries for most platforms. Choose the 3.7.3 or more version minimum.
 
 #### Install nodejs
 
-@@magsout@@
+This project requires NodeJS for building the application and running certain tests. The NodeJS website provides installers in the [downloads section](https://nodejs.org/en/download/). Choose the [current LTS version](https://nodejs.org/en/about/releases/) if you don't have NodeJS installed, or have an older version.
 
 #### Initializing project source code
 
-- 
2.21.0

```

package.json
karlcow: so once the `virtualenv` has been started the version of pip which is available is pip3.

```
→ which pip
/Users/karl/code/webcompat.com/env/bin/pip
→ pip -V
pip 19.2.2 from /Users/karl/code/webcompat.com/env/lib/python3.7/site-packages/pip (python 3.7)
```

but yes the detailed version needs to be updated too.

karlcow:
> I tried to follow https://github.com/webcompat/webcompat.com/blob/e1609766663673f56d547240dd7a893adec45fcc/docs/dev-env-setup.md#installing-pip and ran into an issue (from a fresh clone):

yes because this part has not been yet updated :)

package.json
karlcow: updated the detailed section.

karlcow:
> (i tried to push a commit, but goofed. so here's a patch with the install nodejs bits:)

sweet. I learned to do a new thing.

```bash
# check the stats about the modifications.
git apply --stat ~/Desktop/mike.patch 
# check that everything is correct
git apply --check ~/Desktop/mike.patch 
# apply the patch with the signature of the person who applied the patch.
git am --signoff  ~/Desktop/mike.patch 
# Check if the log is ok.
git log
# Pushed it back
git push karlcow 2905/1
```

and indeed that worked.
https://github.com/webcompat/webcompat.com/pull/2908/commits

❤️

Issue #2791, opened 2019-02-04
GitHub username text is unclear


martinthomson:
This text...

> Sharing your GitHub username—without logging in—could help us with diagnosis. This will be publicly visible.

... combined with the two buttons "Report Anonymously" and "Report via GitHub" is confusing.

The without logging in might be read to imply that you are not going to be required to authorize an application or log in even if you choose the "Report via GitHub" option.  That is not the case.

Also, "This will be publicly visible" might be read to imply that none of the other information will be.  It's all public info (#2690).

My suggestion:

> Your GitHub username (optional):
> [ @github-username ]
> [ Report ] [ Login and Report ]

With the latter only becoming accessible when a GitHub username is entered (I think that this is already the case, but I didn't double-check).

karlcow:
@martinthomson thanks for the comment.

Currently the flow is this (and sure can be improved as we are testing for the last couple of weeks this new feature aka github username)

1. Report Anonymously
2. Report Anonymously + GitHub username (which will add the username to the report, without the person having to login)
3. Report via GitHub (login process)

The difference in between 2 and 3 is that people do not have to give webcompat so many permissions.  See #655 This was the intent to cater for people who are uncomfortable with this big list of permissions.

Also the GitHub username field is only visible if the user is **not** already logged in. The flow you are proposing is not exactly doing the same thing (if I understood).

martinthomson:
Thanks @karlcow,

I'm not suggesting a change in process.  I think that these are the right reporting flows, given the constraints as I understand them.  All I'm doing here is trying to suggest using different words on the form to better communicate what is going to happen.

karlcow:
This relates to #2762 too.

karlcow:
@martinthomson 

We changed things a bit. The username is not hotlinked anymore. 
see https://webcompat.com/issues/29931

![Capture d’écran 2019-04-25 à 12 58 04](https://user-images.githubusercontent.com/505230/56708994-d7b19300-6759-11e9-874b-5627f547f43d.png)
![Capture d’écran 2019-04-25 à 12 58 23](https://user-images.githubusercontent.com/505230/56708995-d7b19300-6759-11e9-832c-87745b96f386.png)

And the wording has been tweaked.

![Capture d’écran 2019-04-25 à 13 01 29](https://user-images.githubusercontent.com/505230/56709067-4abb0980-675a-11e9-85e8-80743eeb1a08.png)

Let's see if that improves thing a bit. 

[...]

Issue #2762, opened 2019-01-16
Consider not @-linking github username field


miketaylr:
```
<sergiu> heya!
karlcow, looks like we have some people who abuses the GitHub handle field: https://github.com/webcompat/web-bugs/issues/24074#issuecomment-454776680
7:58 AM some time ago I got another complaint from a mentioned user, saying that he/she did not log that issue in the first place, and doesn't kno how the github handle ended up in our bug report
8:00 AM did not pay too much attention at the time
8:01 AM 
— sergiu remembered that karl is on STO...
8:02 AM 
<oana> https://github.com/webcompat/web-bugs/issues/23442
8:02 AM this is another one 
```

https://github.com/webcompat/web-bugs/issues/23442
https://github.com/webcompat/web-bugs/issues/24074

This would be less spammy if we wrote `foo` instead of `@foo`.

softvision-sergiulogigan:
> This would be less spammy if we wrote `foo` instead of `@foo`.

i think this would not be enough, as we would most probably ping them at triage time...

softvision-sergiulogigan:
I don't think those GitHub handles are typed randomly. maybe ppl are confusing GH handles with twitter ones?
maybe... we should also all a twitter field?
maybe some validation too on those fields?

karlcow:
That's not cool if we become a source of spam for users. That was the risk.

https://github.com/webcompat/web-bugs/issues/24074
https://twitter.com/lava
It doesn't seem to be anything related.

https://github.com/webcompat/web-bugs/issues/23442
https://twitter.com/sunny
not related either? maybe?

karlcow:
Not sure it will deter people doing it. 
A couple of cases:

1. mistyping a username
2. goofing around college-humor style
3. intended miscconduct for annoying someone else
4. Just not understanding what this box is for

karlcow:
Probably people don't understand what this is about? 

Paper trail.

1. Issue #655 about permissions model.
2. We need to make stats about this new feature in a couple of weeks (was it successful, how many abuse?)
3. Kill it because potential abuse.
4. ❤️ let's hope for the best

karlcow:
An anonymous person complained about the feature. *sigh*. It doesn't help for having an interesting discussions. Also in #2791 

I want to do stats on the feature since we deployed it. And then I guess we can kill it. :)

karlcow:
[...]

PR #2914, opened 2019-08-22
Fixes #2911 - All pop-ups and slider

### LINKED ISSUE #2911: Update modals
For serving information around:
- [x] What web compatibility is
- [x] Privacy Policy

munteanu210:
This PR fixes issue #2911 

## Proposed PR background

The PR contains the following:
1) All pop-ups and slider
2) Add the characters limit for "describe the issue"
3) Add the question mark icon
---


webcompat/form.py
```diff
@@ -380,6 +382,18 @@ def domain_name(url):
     return domain
 
 
+def switch_problem_subtype(type):
```
karlcow: [`type`](https://docs.python.org/3/library/functions.html#type) is a reserved keyword in Python. Choose another word.

webcompat/form.py
```diff
@@ -428,15 +444,20 @@ def build_formdata(form_object):
     if extra_labels:
         metadata_keys.append('extra_labels')
 
+    clean = re.compile('<.*?>')
```
karlcow: What is this regex trying to solve? regex and html are not good friends.

webcompat/form.py
```diff
@@ -418,6 +432,8 @@ def build_formdata(form_object):
     domain = domain_name(normalized_url)
     problem_summary = get_problem_summary(form_object.get('problem_category'))
 
+    switch_problem_subtype(form_object.get('problem_category'))
```
karlcow: This is a function which doesn't assign anything. What was the intent?

webcompat/form.py
```diff
@@ -380,6 +382,18 @@ def domain_name(url):
     return domain
 
 
+def switch_problem_subtype(type):
+    """Return the correct problem subtype object in base of the selected
+    problem type.
```
karlcow: nit: function doc first line should be on one line only.

```
"""Return the correct problem subtype object in base of the selected problem type."""
```

webcompat/form.py
```diff
@@ -80,44 +80,44 @@
 ]
 
 problem_choices = [
-    ('detection_bug', '<div class="icon-container"><img src="../../img/svg/icons/svg-problem-mobile-vs-desktop.svg" /></div> Desktop site instead of mobile site'),
-    ('site_bug', '<div class="icon-container"><img src="../../img/svg/icons/svg-problem-no-use.svg" /></div> Site is not usable'),
-    ('layout_bug', '<div class="icon-container"><img src="../../img/svg/icons/svg-problem-broken-design.svg" /></div> Design is broken'),
-    ('video_bug', '<div class="icon-container"><img src="../../img/svg/icons/svg-problem-no-video.svg" /></div> Video or audio doesn\'t play'),
-    ('unknown_bug', '<div class="icon-container"><img src="../../img/svg/icons/svg-problem-question.svg" /></div> Something else')
+    ('detection_bug', '<div class="icon-container"><img src="../../img/svg/icons/svg-problem-mobile-vs-desktop.svg" /></div> Desktop site instead of mobile site'),  # noqa
```
karlcow: The links should be absolute with regards to the root of the website, aka `<img src="/` etc.

karlcow: Why  HTML is being used in there?

webcompat/form.py
```diff
@@ -143,8 +143,10 @@
 browser_test_label = 'Did you test in another browser?'
 textarea_label = 'Please describe what happened, including any steps you took before you saw the problem'  # noqa
 
-contact_message = 'There is a mistake in the username.'  # noqa
-contact_label = 'Sharing your GitHub username—without logging in—could help us with diagnosis. This will be publicly visible.'  # noqa
+contact_message = 'There is a mistake in the username.'
+contact_label = 'Optional GitHub username'
```
karlcow: This seems to reverse a previous decision where users were thinking it was confusing. Is there additional information about it?

webcompat/form.py
karlcow: * If these images are supposed to be meaningful, they lack of a meaningful `alt` attributes.
* If they are not, it would be necessary to add an `alt=''` to them, if accessibility is handled by another mean.

webcompat/form.py
```diff
@@ -295,7 +295,9 @@ def get_problem_summary(category):
     else:
         # Return the usual message in lowercase
         # because it is not at the start of the summary.
-        return get_radio_button_label(category, problem_choices).lower()
+        problem_type_stripped = re.sub(clean, '', get_radio_button_label(
```
karlcow: see my previous comment about using clean. What the code is trying to do here?
Why not working directly on `get_radio_button_label` function  if the issue is there in the first place?

webcompat/form.py
```diff
@@ -428,15 +444,20 @@ def build_formdata(form_object):
     if extra_labels:
         metadata_keys.append('extra_labels')
 
+    clean = re.compile('<.*?>')
+    problem_type_stripped = re.sub(clean, '', get_radio_button_label(
+        form_object.get('problem_category'), problem_choices))
+    browser_tested_on = form_object.get('browser_test')
+    if browser_tested_on is None:
+        browser_tested_on = "Unknown"
```
karlcow: we use single quotes in the code, except for function comment.

webcompat/form.py
```diff
@@ -428,15 +444,20 @@ def build_formdata(form_object):
     if extra_labels:
         metadata_keys.append('extra_labels')
 
+    clean = re.compile('<.*?>')
+    problem_type_stripped = re.sub(clean, '', get_radio_button_label(
+        form_object.get('problem_category'), problem_choices))
+    browser_tested_on = form_object.get('browser_test')
```
karlcow: The 3 lines here can be reduced with:
```suggestion
    browser_tested_on = form_object.get('browser_test', 'Unknown')
```

webcompat/form.py
karlcow: [no unit tests](https://github.com/webcompat/webcompat.com/blob/f3cce74a422c51d492ef9a1772108001ad5aa105/tests/unit/test_form.py#L4) have been created for this function.

webcompat/form.py
karlcow: There is [no new additional tests](https://github.com/webcompat/webcompat.com/blob/f3cce74a422c51d492ef9a1772108001ad5aa105/tests/unit/test_form.py#L178-L196) testing the changed logic of `build_form_data`

webcompat/static/img/svg/icons/svg-question-mark.svg
```diff
@@ -0,0 +1,9 @@
+<svg width="16px" height="26px" viewBox="0 0 16 26" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>question-mark</title>
+    <desc>Created with Sketch.</desc>
```
karlcow: nit: Probably not neccessary  to have this. :)

webcompat/form.py
munteanu210: So as [these options](https://www.screencast.com/t/kdoYgXDiKhV) contain HTML, this regex is eliminating all the tags, leaving only the text for the options. @johngian if you know a better option, I would appreciate your help.

webcompat/form.py
munteanu210: You are right. I created this function to extract the problem sub-type. Then I integrated all the additional fields into existing fields (sub-type went to description), and there is no need for this function any more. I will eliminate the code where this function is defined, and the code where it is called.

webcompat/form.py
munteanu210: As I wrote above, this function needs to be removed, which eliminates the need for this update.

webcompat/form.py
munteanu210: [WTForms](https://wtforms.readthedocs.io/en/stable/fields.html) was used to generate the fields. For every option (in our case - the radio button), there is the possibility to specify only one label and a key. The options are defined in python. I am referring to these options on the [old site](https://www.screencast.com/t/DqOTQl9t) and the [new site](https://www.screencast.com/t/mBcEyDKQr97X). On the old site, there is only a text value and a key that is not visible. On the new site, each option has a text value, a key, and an image. Ideally, each option that comes from python should have these values. WTForms only supports two values (label and key), and that's why I created the HTML inside the Label. In this way, I am not sending only a text, but an image and a text altogether, in HTML format.

webcompat/form.py
munteanu210: Unfortunately, I don't know about the previous conversations around this text. The update comes from [the design](https://www.screencast.com/t/F0iQFCU77). As far as I know the design and the copy was approved, so I followed the design when implementing this. We may want to run this by Rosana and Christos, and see what they say, and then make a decision if we keep the copy as it is, or change it.

webcompat/form.py
karlcow: [Linking to the code](https://github.com/webcompat/webcompat.com/blob/d995b67769fb0b5bb4bed3a529dea2f2a315a43e/webcompat/form.py#L82-L121) to have a resilient trace of what we are talking about

webcompat/form.py
karlcow: You might want to use [Custom widgets](https://wtforms.readthedocs.io/en/stable/widgets.html#custom-widgets) then

webcompat/form.py
karlcow: For previous discussions, See 

* #1822
* #2791 
* #2762

webcompat/form.py
munteanu210: This function will be removed (as per this [conversation](https://www.screencast.com/t/QG8mxGSg)), that's why there is no need to create a unit test.

munteanu210:
@miketaylr, as you know, the frontend-redesign-dev branch contains [these commits](https://www.screencast.com/t/uuR0Fzxp6). My pull request was intended to add more of the front end functionality on top of what was done already. @johngian could you suggest the right approach on how we should structure this PR and the next ones in a way that allows the team to review the PRs with the initial parts, as @miketaylr is asking?

I wasn't aware of the A/B testing needs, and didn't know we have to keep two versions of the codes. I will let @johngian come up with a plan, and we will go ahead with it.

miketaylr:
> I wasn't aware of the A/B testing needs, and didn't know we have to keep two versions of the codes. I will let @johngian come up with a plan, and we will go ahead with it.

Sounds good to me!

> @miketaylr, as you know, the frontend-redesign-dev branch contains [these commits](https://www.screencast.com/t/uuR0Fzxp6). My pull request was intended to add more of the front end functionality on top of what was done already. @johngian could you suggest the right approach on how we should structure this PR and the next ones in a way that allows the team to review the PRs with the initial parts, as @miketaylr is asking?

Yeah, it would be nice if we could review all the new pieces in a PR -- the current situation is too difficult. Thanks for your work here, it looks great.

webcompat/form.py
munteanu210: @karlcow I see, makes sense. @johngian I may need your help with this update.

miketaylr:
Also, it probably makes sense to just merge this PR w/o reviewing the JS and CSS, once you've implemented the changes requested for the Python stuff, @munteanu210.

We can tackle the frontend stuff in a larger PR containing all the new changes from master once nemo gets back.

miketaylr:
I think this is superseded by #2923.

Issue #2921, opened 2019-09-10
Allow both current and experimental version of bug form to work in parallel


johngian:

miketaylr:
I think this is done now.

PR #2923, opened 2019-09-10
Fixes #2922 #2921 - Initial redesigned bug form integration

### LINKED ISSUE #2922: Integrate frontend from A/B testing experimental form with webcompat.com

johngian:
This PR fixes issue #2922 #2921 

## Proposed PR background

This is an attempt to start the integration of the frontend changes that @munteanu210 is currently working on. The rationale behind this PR after discussing it with @miketaylr is to:

* Make sure that the current webcompat.com functionality is not broken
*  Ensure that we don't break current tests
* Allow running new frontend designs in parallel with current webcompat.com state (A/B)
* Try to avoid nit-picking the new work since its experimental. The project changes suggested in this PR have a minor impact to the current webcompat.com codebase on purpose.
* Have a more sane way forward by opening issues for fixes to something actionable instead of building frontend code that piles up.

I understand that its a massive PR and shouldn't have been like that in the first place but I tried to cleanup the commits, make them more atomic, rebase on master. Some comments about the PR:

* For all the code (`bugform.js`, `form.html`) that was based on current functionality, I created a copy in a new commit and started committing changes on top of that so

  1. Its easier to isolate old and new functionality.
  2. We don't break current functionality
  3. Each easier to review diffs, rather than a huge blob

* All the new images/icons/svgs introduced are in separate commits
  * f724e4fbf21088b6715799b842bb29d4ef9d71ef
  * d0dbd3db4a87eb8fbdfe4094b82550dab512c971
* For easier review these are the only changes that might have impact to the current codebase
  * b34f136d4ef58dfdc91468cb2845d2207a5ec4d4
  * 1c3f77d804105227c91e0dd7e81ebe110e148cd1
 
I understand that the process is not ideal but I hope I made the review a little bit easier. Let me know what you think.

cc @miketaylr @karlcow 

---


karlcow:
> Try to avoid nit-picking the new work since its experimental. The project changes suggested in this PR have a minor impact to the current webcompat.com codebase on purpose.

Understood for this. 

This comes with a requirement though. We should not merge this branch ever in master. 
If the experiment is successful, we would then to have to recreate a PR with proper syntax, commits, and nit-picking 😉.

@johngian which type of review do you expect and what would be helpful for you?

webcompat/form.py
```diff
@@ -111,12 +206,72 @@ class IssueForm(FlaskForm):
...
-def get_form(form_data):
+class PrefixedRadioField(RadioField):
+    """Prefix radio field label with an image."""
+    def __init__(self, *args, **kwargs):
+        prefixed_choices = kwargs.pop('choices')
+        template = '<div class={cls}><img src={src}/></div> {text}'
+        choices = []
+
+        for slug, img, text in prefixed_choices:
+            cls = 'icon-container'
```
karlcow: This `cls` might trigger some bells for some syntax checkers, highlighters.
https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments

> Always use cls for the first argument to class methods.

This is not a reserved keyword, but this is used in different contexts.

webcompat/form.py
```diff
@@ -111,12 +206,72 @@ class IssueForm(FlaskForm):
...
+            src = url_for('static', filename=filename)
+            label = Markup(template.format(src=src, cls=cls, text=text))
+            choice = (slug, label)
+            choices.append(choice)
+
+        kwargs['choices'] = choices
+        super(PrefixedRadioField, self).__init__(*args, **kwargs)
+
+
+class IssueFormRedesigned(IssueForm):
```
karlcow: nit: Probably will deserve another title in the future. #tofixlater

webcompat/form.py
```diff
@@ -111,12 +206,72 @@ class IssueForm(FlaskForm):
...
+    """Re-designed version of IssueForm:
```
karlcow: nit: docstring doesn't follow our standard. #tofixlater

Though it's probably our fault to not enforce it through circleci check.

webcompat/form.py
```diff
@@ -111,12 +206,72 @@ class IssueForm(FlaskForm):
...
+    Define form fields and validation for our bug reporting form.
+    """
+
+    steps = NEW_ISSUE_STEPS
+
+    browser = StringField(u'Browser', [Optional()])
+    device = StringField(u'Device', [Optional()])
+    os = StringField(u'Operating System', [Optional()])
```
karlcow: `u` in `u'…'` should be dropped. Python 3 code.

webcompat/form.py
```diff
@@ -111,12 +206,72 @@ class IssueForm(FlaskForm):
...
+        [InputRequired(message=radio_message)],
+        choices=video_bug_choices
+    )
+    browsers = PrefixedRadioField(
+        [InputRequired(message=radio_message)],
+        choices=browser_choices
+    )
+
+
+def get_form(form_data, form=IssueForm):
```
karlcow: nit: In the future, this is probably a welcome improvement. #tofixlater
There could be a separate PR which basically give structure to the code for things without AB testing in a way that make it easier to configure. 

Thanks!

webcompat/helpers.py
```diff
@@ -716,3 +716,7 @@ def ab_init(response):
             response.set_cookie(exp_id, var, max_age=max_age)
 
     return response
+
+
+# Register `ab_active` as jinja function
+app.jinja_env.globals.update(ab_active=ab_active)
```
karlcow: I'm not a big fan of this hanging out in the bottom of a module function-less. That seems not at the right place. Was there a reason why it was not up there with cache_dict and other things.

webcompat/views.py
```diff
@@ -175,7 +177,9 @@ def index():
         'rel': 'preload'
     })
     ua_header = request.headers.get('User-Agent')
+
```
karlcow: That doesn't need to be here. 😁

webcompat/views.py
```diff
@@ -175,7 +177,9 @@ def index():
...
     bug_form = get_form({'user_agent': ua_header})
+
```
karlcow: That doesn't need to be here. 😁

config/__init__.py
```diff
@@ -205,7 +205,8 @@ def update_status_config(milestones_content):
 AB_EXPERIMENTS = {
     'exp': {
         'variations': {
-            'form-v1': (0, 0),
+            'form-v1': (0, 100),
```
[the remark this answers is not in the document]
miketaylr: Does this mean we serve v2 (the new one?) 100% of the time?

webcompat/form.py
```diff
@@ -15,11 +15,13 @@
 import re
 import urllib.parse
 
-from flask import g
+from flask import g, url_for
```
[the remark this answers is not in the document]
miketaylr: This one seems simple to fix now. 😅

webcompat/form.py
miketaylr: maybe something like `StepByStepForm` or `FormWizard` i dunno

webcompat/helpers.py
miketaylr: @johngian can we follow the way we do this for `bust_cache`? It seems like it's equivalent, and would be nice to be consistent.

https://flask.palletsprojects.com/en/1.1.x/templating/#registering-filters

webcompat/templates/new-issue.html
```diff
@@ -1,14 +1,33 @@
 {%- extends "layout.html" -%}
+
+{% block extracss %}
+{%- if ab_active('exp') == 'form-v2' %}
+<link href="{{ url_for('static', filename='css/src/issue-wizard.css') }}" type="text/css" rel="stylesheet">
```
miketaylr: I like the name issue-wizard.

config/__init__.py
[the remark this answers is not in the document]
johngian: I just changed this one for convenience when starting testing in a live deployment. `form-v1` is current state, `form-v2` is the redesign which is served at 0%.

webcompat/form.py
johngian: Noted for the issues to be filled for future reference.

webcompat/helpers.py
johngian: Sure thing. I just followed the flask documentation and looked like a pretty common way to configure global functions.

johngian:
Hey @miketaylr @karlcow I think I addressed most of the comments and I kept a couple of notes to open issues for future reference.

Regarding the last comment:

* I opened some issues to track the work already done. I am pretty sure we need to file more issues moving forward for the rest of the pending work
* I kept the `issue wizard` naming convention for the new files introduced

Let me know what you think

config/__init__.py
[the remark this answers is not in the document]
karlcow: thanks for the explanation

karlcow:
I need to talk with @magsout to give a bit of context on this project.

PR #2942, opened 2019-09-29
Fixes #2912 - Implement login functionality

### LINKED ISSUE #2912: Implement login functionality
- [x] login with GitHub to submit the report
- [x] submit the report anonymously
- [x] provide an optional GitHub username while report anonymously

munteanu210:
This PR fixes issue #2912

## Proposed PR background

This pull request contains the [redesign](https://www.screencast.com/t/CN912Jvqp) for the login with GitHub and report anonymously on the bug form.

### DIFF
--- a/webcompat/templates/home-page/issue-wizard-form.html
+++ b/webcompat/templates/home-page/issue-wizard-form.html
@@ -248,35 +248,7 @@
     <div class="button-control row">
       <div class="input-control">
         <div class="col center">
-          <button class="button next-step step-8 issue-btn disabled" data-nextstep="9">Continue</button>
-        </div>
-      </div>
-    </div>
-  </div>
-
-  <div class="step-container step9 col" data-nextstep="10" data-activate-stepper="7">
-    <div class="row">
-      <div class="input-control half">
-        <div class="description-text top">
-          Fantastic, thank you. <br>
-          We’re almost done. <strong>We took a screenshot of the website</strong> when you began this report. <strong>Are you OK with publicly sharing the image we captured?</strong>
-        </div>
-      </div>
-      <div class="input-control half col">
-        <div class="form-text form-element js-Form-group problem-screenshot">
-          <img class="screenshot" src="{{ url_for('static', filename='img/sample-screenshot.jpg') }}" title="Generated screenshot" alt="Generated screenshot" />
-          <div class="shadow"></div>
-        </div>
-      </div>
-    </div>
-    <div class="button-control row">
-      <div class="input-control">
-        <div class="col center">
-          <div class="row">
-            <button class="button next-step step-9 issue-btn" data-nextstep="11">Yes, continue</button>
-            <button class="button next-step step-9 issue-btn red" data-nextstep="10" data-hidestep="9">No, delete this image</button>
-          </div>
-          <a class="next-step step-9" data-nextstep="10" data-hidestep="9" href="#">Upload a different image</a>
+          <button class="button next-step step-8 issue-btn disabled" data-nextstep="10">Continue</button>
         </div>
       </div>
     </div>
@@ -286,7 +258,7 @@
     <div class="row">
       <div class="input-control half flex-left">
         <div class="description-text top up-message deleted-screenshot">
-          <strong>We deleted the automatic screenshot.</strong><br>
+          <strong>We deleted the screenshot.</strong><br>
           Do you have a different screenshot you would like to share?<br>
           <br>
           <strong>Uploaded images are publicly viewable.</strong>
--- a/webcompat/templates/home-page/issue-wizard-form.html
+++ b/webcompat/templates/home-page/issue-wizard-form.html
@@ -196,18 +196,18 @@
     <div class="row half centered other-browser form-element js-Form-groupform-element js-Form-group with-validation-icons low">
       <div class="column full-width">
         <div class="label">
-          {{ form.other_browser.label(class_='form-label') }}
+          Browser tested
         </div>
-        {{ form.other_browser(class_='form-field text-field required',
+        {{ form.browser_test(class_='form-field text-field required',
             placeholder='Browser',
-            required=True, type='text', pattern='') }}
+            required=True, type='text') }}
       </div>
     </div>
     <div class="button-control row">
       <div class="input-control">
         <div class="col center">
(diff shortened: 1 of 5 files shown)


webcompat/templates/home-page/issue-wizard-form.html
```diff
@@ -351,13 +352,43 @@ <h3 class="page-heading">Filing a web compatibility issue</h3>
             Good job! You can&nbsp;<strong>report via GitHub</strong>&nbsp;or&nbsp;<strong>report anonymously</strong>.
           </div>
           <div class="row">
-            <button class="button button-primary right js-Button" id="submitgithub" type="submit" name="github-auth-report">
+            <button class="button button-secondary right js-Button" id="submitgithub" type="submit" name="github-auth-report">
+              <img class="" src="{{ url_for('static', filename='img/svg/icons/svg-github-white.svg') }}" title="Report via GitHub" alt="Report via GitHub" />
               {% if session.username %}Report as {{ session.username }}{% else %}Report via GitHub{% endif %}
             </button>
-            <button class="button button-secondary js-Button" id="submitanon" type="submit" name="github-proxy-report">
+            <button class="button button-primary" id="open-username" type="submit">
```
miketaylr: I'm a little confused here.Removing the `name` here will break the reporting backend:

https://github.com/webcompat/webcompat.com/blob/04e3e5d5116ab8fffb7c09aaacb79c33ff4940f9/webcompat/issues.py#L27-L33

But I see below:

```
<button class="button button-primary js-Button" id="submitanon" type="submit" name="github-proxy-report">
                Send report
</button>
```

Is this correct? I think if you click "Report Anonymously" it will break, and I'm not sure what "Send report" is for, but it will report anonymously.

webcompat/templates/home-page/issue-wizard-form.html
```diff
@@ -196,18 +196,18 @@ <h3 class="page-heading">Filing a web compatibility issue</h3>
     <div class="row half centered other-browser form-element js-Form-groupform-element js-Form-group with-validation-icons low">
       <div class="column full-width">
         <div class="label">
-          {{ form.other_browser.label(class_='form-label') }}
+          Browser tested
```
miketaylr: Is there any reason to move the label out of form.py here?

miketaylr:
Just thinking about the failing test here... we probably need a mechanism to run tests in both "a" and "b" modes. Right now, I believe it's serving "b" 100% of the time, so as markup changes, tests assuming an "a" DOM being tested against a "b" DOM will diverge. 

This is where it gets painful... possibly #2944 can help us solve this problem, if we export the right env var before running the test suite.

miketaylr:
I filed https://github.com/webcompat/webcompat.com/issues/2947 to tackle the test thing.

webcompat/templates/home-page/issue-wizard-form.html
munteanu210: @miketaylr form.other_browser.label is also used in the A version of the form, so by changing it we change it for both versions. As far as I know, with WTForms we cannot have multiple labels for the same field, therefore the static label.

miketaylr: Makes sense, thanks. I guess the other option would be to do something silly like `other_browser_b`. But I don't think we consume the label on the backend, so it doesn't matter much.

webcompat/templates/home-page/issue-wizard-form.html
munteanu210: @miketaylr, the Report Anonymously button doesn't submit the form. It reveals an input where the user can type the optional username and only then the form gets submitted (the username field can be left empty). The "real" button that submits the form anonymously is the one below ("Send report").

miketaylr: OK, thanks!

webcompat/templates/layout.html
```diff
@@ -13,7 +13,9 @@
   {% block extracss %}{% endblock %}
 
 </head>
-<body id="body-webcompat" data-username="{{ session.username }}">
+<body id="body-webcompat"
+  class="{%- if request.url_rule.endpoint == 'create_issue' and ab_active('exp') == 'form-v2' %}no-top-padding{% endif %}"
```
johngian: Keep in mind that this doesn't only change the new form. It adds a class to every page pretty much. I would prefer if this wasn't fixed here but be contained in the codebase of the experiment.

munteanu210:
@miketaylr I will address your initial comment below:

This is a test that checks the submit buttons on the old form. Initially it must be disabled, which is verified by this test. Here is [the code of the test](https://www.screencast.com/t/gjzCthXMU6E). I don't know why this test failed, so as both buttons [are disabled](https://www.screencast.com/t/VQ0J0Hohu2).

In addition, I did not reach this part of the form. I don't know if it was a test execution error or something was broken there, but it's something that I did not touch for sure.

webcompat/templates/layout.html
```diff
@@ -35,8 +37,9 @@
 <script src="{{ url_for('static', filename='js/lib/navbar.js') }}"></script>
 <script src="{{ url_for('static', filename='js/lib/autogrow-textfield.js') }}"></script>
 <script src="{{ url_for('static', filename='js/lib/bugform-validation.js') }}"></script>
+{%- if ab_active('exp') == 'form-v1' %}
```
johngian: This was omitted here on purpose. I would rather override only things for `form-v2` version

johngian: Also by reverting this the current website tests are not broken.

johngian: Also keep in mind that not all the population would end up in an experiment branch (eg. private browsing etc)

johngian: Also people that work on the project are exempt from the experiments.

johngian: This might work better: https://github.com/johngian/webcompat.com/commit/975be048eff3d8c00d293d5628d2787b1a32bd1a

Also keeps the tests happy

munteanu210: @johngian thank you for your suggestions and for the [proposed commit](https://github.com/johngian/webcompat.com/commit/975be048eff3d8c00d293d5628d2787b1a32bd1a). I committed your update and all tests passed successfully.

webcompat/templates/layout.html
miketaylr: I filed https://github.com/webcompat/webcompat.com/issues/2955 to address this.

Issue #2947, opened 2019-10-07
Be able to run tests in a certain version of A/B tests


miketaylr:
I think https://github.com/webcompat/webcompat.com/issues/2944 will largely solve this, but the idea is before we run tests, we set "A" to 100%. Otherwise, depending on what the populations are set to, the tests may or may not fail.

miketaylr:
I think https://github.com/webcompat/webcompat.com/pull/2951 will fix this entirely -- if we don't have the relevant environment variables set, the application will default to v1 100% of the time so there's no need to do anything else.

miketaylr:
OK, let's close here.

PR #2951, opened 2019-10-07
Fixes #2944, #2947 - Read experiment variation values from envvars

### LINKED ISSUE #2940: Have a way to identify reports filed from the experimental form
A couple of options:

* use the `label` GET param
* HTML comment
* apply a label at report time

Something else?

### LINKED ISSUE #2944: Allow A/B testing settings to be set via environment variable
This way we wouldn't need to deploy every time we want to tweak the branches, we would just restart uwsgi.

cc @karlcow @johngian

miketaylr:
This feels maybe slightly hacky -- we can't safely store a tuple in an environment variable, so it build one up from something like `0 20`. Open to other ideas.

And add tests.

r? @johngian @karlcow

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -119,6 +119,29 @@ def update_status_config(milestones_content):
     return STATUSES
 
 
+def get_variation(variation_key, variations_dict, defaults_dict):
+    """Convert a string to a tuple of integers.
+
+    If the passed variation_key doesn't follow this pattern '0 100', it will
+    return default values defined in defaults_dict.
+
+    This is currently used for defining the variation data of the A/B
+    experiment regarding the multi-steps form.
+    """
+    try:
+        # We want to create a tuple of integers from a string containing
+        # integers. Anything else should throw.
+        rv = tuple(int(x) for x in variations_dict.get(variation_key)
+                                                  .strip().split())
+        if (len(rv) != 2):
+            raise ValueError('The format is incorrect. Expected "{int} {int}"')
+    except Exception as e:
+        print('Something went wrong with AB test configuration: {0}'.format(e))
+        print('Falling back to default values.')
+        rv = defaults_dict.get(variation_key)
+    return rv
+
+
 THREADS_PER_PAGE = 8
 
 # ~3 months-ish expires for static junk
@@ -200,14 +223,29 @@ EXTRA_LABELS = [
     'type-webvr',
 ]
 
+# Get AB experiment variation values from the environement.
+AB_VARIATIONS = {
+    'FORM_V1_VARIATION': os.environ.get('FORM_V1_VARIATION'),
+    'FORM_V2_VARIATION': os.environ.get('FORM_V2_VARIATION'),
+}
+# We define default values here, as a fallback.
+# By default, v1 will be served 100% of the time.
+AB_DEFAULTS = {
+    'FORM_V1_VARIATION': (0, 100),
+    'FORM_V2_VARIATION': (100, 100),
+}
+EXP_MAX_AGE = int(os.environ.get('EXP_MAX_AGE', 0))
+
 # AB testing config
 AB_EXPERIMENTS = {
     'exp': {
         'variations': {
-            'form-v1': (0, 100),
-            'form-v2': (100, 100)
+            'form-v1': get_variation('FORM_V1_VARIATION', AB_VARIATIONS,
+                                     AB_DEFAULTS),
+            'form-v2': get_variation('FORM_V2_VARIATION', AB_VARIATIONS,
+                                     AB_DEFAULTS),
         },
-        'max-age': None
+        'max-age': EXP_MAX_AGE
     }
 }
 
--- a/docs/ab-testing.md
+++ b/docs/ab-testing.md
@@ -1,10 +1,23 @@
 # AB testing
 ## Configuration
-You can enable/disable AB testing support by configuring the `config.AB_EXPERIMENT` dictionary.
-This will configure flask to set cookies to responses. For example:
+You can enable/disable AB testing support by configuring the following environment variables:
+
+`FORM_V1_VARIATION`: expects a string of the format '{int} {int}'
+`FORM_V2_VARIATION`: expects a string of the format '{int} {int}'
+`EXP_MAX_AGE`: expects a number which defines when the experiment cookie expires
+
+These are read by the `config.AB_EXPERIMENT` dictionary in `./config/__init__.py` and converted to a tuple.
+This will configure flask to set cookies to responses. For example, setting the variables to the following:
 
 ```
+export FORM_V1_VARIATION = "0 20"
+export FORM_V2_VARIATION = "20 100"
+export EXP_MAX_AGE = 604800
+```
 
+This will result in the `AB_EXPERIMENTS` dictionary with the following values:
+
+```
 AB_EXPERIMENTS = {
     'exp-1': {
         'variations': {
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -11,6 +11,7 @@ import os
 import unittest
 
 from config import update_status_config
+from config import get_variation
 from webcompat import webcompat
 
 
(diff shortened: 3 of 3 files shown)


miketaylr:
Oops, wrong bug number. :) Will fix the commits.

tests/unit/test_config.py
```diff
@@ -53,6 +54,19 @@ def test_update_status_config(self):
...
+        """
+        self.assertRaises(Exception, get_variation_from_env, 'a b')
+        self.assertRaises(Exception, get_variation_from_env, 'None')
+        self.assertRaises(Exception, get_variation_from_env, '')
+        self.assertRaises(Exception, get_variation_from_env, 2)
+        self.assertRaises(Exception, get_variation_from_env, '1 2 3')
+        self.assertRaises(Exception, get_variation_from_env, '1 ')
+        self.assertTupleEqual((1, 2), get_variation_from_env('1 2'))
+        self.assertTupleEqual((1, 2), get_variation_from_env(' 1 2 '))
+
```
karlcow: ❤️ I tested '0010    02' 
And this is working too.

tests/unit/test_config.py
```diff
@@ -53,6 +54,19 @@ def test_update_status_config(self):
         actual = update_status_config(milestones_json)
         self.assertEqual(actual, expected)
 
+    def test_get_variation_from_env(self):
+        """Make sure it doesn't blow up when it shouldn't, and does when
+        it should.
+        """
```
karlcow: nit: Probably a different docstrings ;) I know my docstrings are boring. :p 

```suggestion
"""
Check the conversion from string to tuples of two int.
"""
```

docs/ab-testing.md
```diff
@@ -1,10 +1,23 @@
...
+`FORM_V1_VARIATION`: expects a string of the format '{int} {int}'
+`FORM_V2_VARIATION`: expects a string of the format '{int} {int}'
+`EXP_MAX_AGE`: expects a number which defines when the experiment cookie expires
+
+These are read by the `config.AB_EXPERIMENT` dictionary in `__init__.py` and converted to a tuple.
+This will configure flask to set cookies to responses. For example, setting the variables to the following:
+
+```
+export FORM_V1_VARIATION = "0 20"
+export FORM_V2_VARIATION = "20 100"
```
karlcow: Maybe not related to this PR, but maybe @johngian can explain. Why do we need to define the two alternatives when one seems to be the differential of the other one.

johngian:
I was thinking of a different approach that
a) Would work for other experiments as well
b) Has less hardcoded values

can be something like that: https://github.com/johngian/webcompat.com/commit/0662bf14e81fd1b8b4a08e484fc97663e420a80f

The reasons that I came up with this is that if something goes very bad and we need to override the A/B testing cookies (in our case `exp`), this would require a code deployment. Where, if we manage the whole A/B config definition we can just change the value there.

docs/ab-testing.md
johngian: Just easier to implement, more explicit. We can change it in the future if want more A/B experiments and need a more flexible config.

config/__init__.py
```diff
@@ -200,14 +219,20 @@ def update_status_config(milestones_content):
     'type-webvr',
 ]
 
+# Get AB experiment variation values from the environement.
+# By default, it will serve v1 100% of the time.
+FORM_V1_VARIATION = os.environ.get('FORM_V1_VARIATION') or '0 100'
+FORM_V2_VARIATION = os.environ.get('FORM_V2_VARIATION') or '100 100'
+EXP_MAX_AGE = os.environ.get('EXP_MAX_AGE', None)
```
johngian: max-age is defined as an integer so if we want to read this from the env var we should cast it to int (env vars are strings)

miketaylr:
> @miketaylr I haven't really thought about that since its more a once-off thing (writing the config i mean) and if you are **really** careful and pass it through a JSON linter you are probably good to go. It feels like writing a script to generate a JSON for the config its too much for our case.

thanks -- i think let's just go with this approach and i'll file a follow-up to simplify things via JSON parsing (more simple implementation, slightly higher risk). i'll work on addressing feedback.

config/__init__.py
```diff
@@ -119,6 +119,32 @@ def update_status_config(milestones_content):
     return STATUSES
 
 
+def get_variation_from_env(envvar_value):
+    """Return a tuple with variation data from the specific environment
+    variable key.
+
+    Raise an exception if the passed in envvar doesn't look something
+    like: '0 100'
+    """Convert a string to a tuple of integers.
```
miketaylr: Oops, looks like GitHub's accept suggestion thingy got confused. 💣

config/__init__.py
```diff
@@ -200,14 +219,20 @@ def update_status_config(milestones_content):
...
+FORM_V2_VARIATION = os.environ.get('FORM_V2_VARIATION') or '100 100'
+EXP_MAX_AGE = os.environ.get('EXP_MAX_AGE', None)
+
 # AB testing config
 AB_EXPERIMENTS = {
     'exp': {
         'variations': {
-            'form-v1': (0, 100),
-            'form-v2': (100, 100)
+            'form-v1': get_variation_from_env(FORM_V1_VARIATION),
```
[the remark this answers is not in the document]
miketaylr: Looking again, I don't think this will break because we provide some default values `FORM_V1_VARIATION = os.environ.get('FORM_V1_VARIATION') or '0 100'` -- this should serve v1 100% of the time. We just need to add some logging here if it raises.

config/__init__.py
karlcow: yeah the suggestion works for one line only.

karlcow:
> > @miketaylr I haven't really thought about that since its more a once-off thing (writing the config i mean) and if you are **really** careful and pass it through a JSON linter you are probably good to go. It feels like writing a script to generate a JSON for the config its too much for our case.
> 
> thanks -- i think let's just go with this approach and i'll file a follow-up to simplify things via JSON parsing (more simple implementation, slightly higher risk). i'll work on addressing feedback.

the issue with a JSON file is that we are back to square one. As something which is dependent of a deploy (If I got what was proposed)

miketaylr:
> the issue with a JSON file is that we are back to square one. As something which is dependent of a deploy (If I got what was proposed)

I think the suggestion is to use a JSON *string*, not a file. So you export a var that contains JSON, rather than read it from disk. But I guarantee I will miquote the JSON and everything will fail. 😄

config/__init__.py
```diff
@@ -119,6 +119,25 @@ def update_status_config(milestones_content):
...
+    variable key.
+
+    Raise an exception if the passed in envvar doesn't look something
+    like: '0 100'
+    """
+    try:
+        # We want to create a tuple of integers from a string containing
+        # integers. Anything else should throw.
+        rv = tuple(int(x) for x in envvar_value.strip().split(' '))
+        if (len(rv) is not 2):
```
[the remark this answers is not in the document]
miketaylr: (note, it should be `!=` because we want to raise if the length is not 2, but i assume that's just a typo)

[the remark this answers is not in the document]
karlcow: yup

miketaylr:
I think I addressed all feedback, r? @karlcow 

Right now, if `get_variation_from_env` raises because something is wacky, it will print to `stdout`. So we'll have to make sure we don't see any weird messages when restarting the app after updating values.

config/__init__.py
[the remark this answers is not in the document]
karlcow: It will break take this scenario:

1. environment variable defined as `export FORM_V1_VARIATION='0 a'`
2. `FORM_V1_VARIATION = os.environ.get('FORM_V1_VARIATION', '0 100')` is valid and `FORM_V1_VARIATION` will be `0 a`
3. The code goes to `AB_EXPERIMENTS` which contains `'form-v1': get_variation_from_env(FORM_V1_VARIATION)`
4. This will raise an exception. There is no default value at this stage.

The default value applies only if environment variable doesn't exist.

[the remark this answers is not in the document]
miketaylr: Ah. Right.

miketaylr:
OK, not 100% sure I love this implementation, but it works and the tests pass. 

(It doesn't feel very DRY, but I'm not sure I care. It would have been nice to be able to read the `VARIATIONS` and `DEFAULTS` dicts from the config during testing, rather than have to pass them in to be testable, but I don't think it's possible to refer to the config object inside of `get_variation` because at the point it gets called, the config object hasn't even been created.)

r? @karlcow

miketaylr:
(before merging, we should probably squash this into a single commit -- i'm happy to do so)

miketaylr:
This will also fix #2947 because by default the variables won't be set, and it will only serve v1 of the form (which tests rely on).

karlcow:
ok squashed committed. Let's run CircleCi and wait a bit.

karlcow:
merged

PR #2957, opened 2019-10-09
Fixes #2940 - Add a label to v2 form issues

### LINKED ISSUE #2940: Have a way to identify reports filed from the experimental form

miketaylr:
It's hard-coded, but shortlived. We can also use `label` param if we want to, but this way seems easiest. 

r? @karlcow 

https://github.com/webcompat/webcompat-tests/issues/1953 is a sample issue (filed from chrome!!111), which has the label added, as well as the `<!-- @extra_labels: form-v2-experiment -->` comment.

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -263,6 +263,8 @@ def create_issue():
 
         if ab_active('exp') == 'form-v2':
             bug_form = get_form(form_data, form=FormWizard)
+            # TODO: remove this when the experiment has ended
+            form_data['extra_labels'].append('form-v2-experiment')
         else:
             bug_form = get_form(form_data)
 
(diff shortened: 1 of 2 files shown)


webcompat/views.py
```diff
@@ -263,6 +263,8 @@ def create_issue():
 
         if ab_active('exp') == 'form-v2':
             bug_form = get_form(form_data, form=FormWizard)
+            # TODO: remove this when the experiment has ended
+            form_data['extra_labels'].append('form-v2-experiment')
```
karlcow: I guess in one way or the other this full code will disappear anyway be in the case of rejecting or adopting the new form.

Issue #2958, opened 2019-10-09
A/B experiment cookie is not populated on first request


johngian:
Steps to reproduce:
* After cleaning cookies for 127.0.0.1:5000, visit: 127.0.0.1/5000/issues/new
* Check cookies
  * No A/B experiment cookie exists
* Click refresh
  * A/B experiment cookie gets populated

Expected behaviour:
* Exp cookies should be populated on first request

PR #2959, opened 2019-10-09
Issue #2958 - Fix `ab_active` to populate cookies on first request

johngian:
This PR fixes issue #2958 

## Proposed PR background

There was an issue on the way that we check if an experiment is active. `ab_active` helper instead of checking the ``g.current_experiments`` dict that we populate before each request, was checking the actual cookies. That means that on the first request `ab_active` was failing to give the proper result because cookies where not populated.

Changes introduced in the commit:

* Fix `ab_active` behaviour
* Change tests to work for the new behaviour
* Minor cleanup/docstring changes on tests

I tested in 2 temp firefox containers and it looks like experiment cookies are loaded on first request.

### DIFF
--- a/tests/unit/test_helpers.py
+++ b/tests/unit/test_helpers.py
@@ -321,8 +321,8 @@ class TestHelpers(unittest.TestCase):
         # A JSON value, which is not an object
         self.assertFalse(is_json_object(json.loads('null')))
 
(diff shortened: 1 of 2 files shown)


johngian:
@karlcow @miketaylr r?

PR #2990, opened 2019-10-21
Fixes #2989 - Create static bundle for form-v2 experiment.

### LINKED ISSUE #2989: Form wizard experiment doesn't work with production assets [New UI]
In development we used a flag to load the static assets properly. In production we load the static files as defined in the Gruntfile. That means that static assets in production don't load the same way as they do in development.

johngian:
This PR fixes issue #2989 

## Proposed PR background

Currently production serves the bundled minified assets. This means that the logic we had to serve different JS for the experiments worked only for the dev environment. This PR introduces a new entry in the JS bundles that we conditionally load.

### DIFF
--- a/grunt-tasks/concat.js
+++ b/grunt-tasks/concat.js
@@ -26,6 +26,25 @@ module.exports = function(grunt) {
       ],
       dest: "<%= jsDistPath %>/<%= pkg.name %>.js"
     },
+    distFormV2: {
+      src: [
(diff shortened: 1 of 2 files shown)


miketaylr:
@johngian seems like jinja is getting sad at the nested else / if

johngian:
@miketaylr circleci looks happy now.

PR #2991, opened 2019-10-21
Fixes #2975 -  "Briefly describe the issue" fix & validation

### LINKED ISSUE #2975: "Briefly describe the issue" does not accept any value [New UI]
On Step 2 (Issue), I select "Something else" from the options.
I enter some text in the "Briefly describe the issue:" text field  which appears.

No matter what I enter, the text field is not validated and the "Confirm" button remains disabled.

Screenshot 2019-10-11 at 16 49 27

munteanu210:
This PR fixes issue #2975

## Proposed PR background

I added validation for the field. When this field is empty, the "Continue" button is disabled. When you enter something, the field is validated, the check symbol appears and the button is activated to proceed to the next step.

### DIFF
--- a/webcompat/static/js/lib/bugform-validation.js
+++ b/webcompat/static/js/lib/bugform-validation.js
@@ -24,6 +24,10 @@ function Validation() {
       var val = descField.val();
       return $.trim(val) !== "";
     },
+    isIssueValid: function(issueField) {
+      var val = issueField.val();
+      return $.trim(val) !== "";
+    },
     isUrlValid: function(urlField) {
       var val = urlField.val();
       return $.trim(val) !== "" && isReportableURL(val);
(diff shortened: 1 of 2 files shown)


webcompat/static/js/lib/bugform-validation.js
```diff
@@ -24,6 +24,10 @@ function Validation() {
       var val = descField.val();
       return $.trim(val) !== "";
     },
+    isIssueValid: function(issueField) {
```
miketaylr: This seems like it's doing the same as `isDescriptionValid` -- in the future we might consider making them call the same method. But let's merge here to not block.

miketaylr: https://github.com/webcompat/webcompat.com/issues/2992

PR #3010, opened 2019-10-22
Fixes #2974 - Browser selection to say Yes/No as on current bug form

### LINKED ISSUE #2974: Selecting "Other" from browser selection presents an interesting text field [New UI]
On step 4 (testing), when I select Other, I should be presented with a text field labelled "Browser tested" featuring the placeholder text "Browser"

Screenshot 2019-10-11 at 16 33 40

What I get is an non-interactive text field with Yes/No values:
 
Screenshot 2019-10-11 at 16 34 27

munteanu210:
This PR fixes issue #2974

## Proposed PR background

On the current bug form, there is only a Yes / No answer for the question whether the user tested in another browser. There was no browser selection. On the new bug form, the initial intention was to let the user select the browser type and even insert a browser name.

The back end does not accept values other than Yes / No. As described in issue #2974, the decision was to send the value "Yes" to the server when selecting a browser, and to send the value "No" when the user selects "I have only tested on this browser", and this is exactly what this PR is fixing.

### DIFF
--- a/webcompat/static/js/lib/issue-wizard-bugform.js
+++ b/webcompat/static/js/lib/issue-wizard-bugform.js
@@ -53,7 +53,7 @@ BugForm.prototype.onDOMReadyInit = function() {
   this.step4Container = $(".step-container.step4");
   this.step5Container = $(".step-container.step5");
   this.step6Container = $(".step-container.step6");
-  this.step6Radio = $(".step-container.step6 input");
+  this.step6Radio = $(".step-container.step6 .input-control input");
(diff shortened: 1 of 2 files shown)


munteanu210:
@miketaylr and @karlcow Could you kindly review this PR, then merge & deploy it on staging if everything looks good? Thank you!

miketaylr:
> The back end does not accept values other than Yes / No

Note that this could be changed, it's not a permanent limitation.

PR #3011, opened 2019-10-22
Fixes #2983 - Upload a different image keeps the original image unless a new one is chosen

### LINKED ISSUE #2983: Unexpected behaviour of "Upload a different image" [New UI]
On step 6 (Screenshot), if I upload an image and then click to **Upload a different image**, the screenshot is deleted and I am taken back up to the very top of the page.

Untitled5

My expectation as a user would be to be presented with another Open file dialogue to select a different image without going through the process of deleting the current image.

munteanu210:
This PR fixes issue #2983 and particularly this [comment](https://github.com/webcompat/webcompat.com/issues/2983#issuecomment-544926336).

## Proposed PR background

Right now "Upload a different image" has the same effect as "Delete this image". In both cases, it deletes the image. 

The desired effect of "Upload a different image" is that it opens up the "Select file" dialogue without deleting the image first. If a new image is selected than this will replace the previous image. If the "Select file" dialogue is closed than the previous image remains.

### DIFF
--- a/webcompat/templates/home-page/issue-wizard-form.html
+++ b/webcompat/templates/home-page/issue-wizard-form.html
@@ -311,7 +311,7 @@
             <button class="button next-step step-10 issue-btn disabled" data-nextstep="11">Confirm</button>
             <button class="button js-remove-upload is-hidden issue-btn red">Delete this image</button>
           </div>
-          <a class="js-remove-upload is-hidden spaced-link" href="#">Upload a different image</a>
+          <label for="image" class="is-hidden screenshot-select-trigger spaced-link" href="#">Upload a different image</label>
         </div>
       </div>
     </div>
--- a/webcompat/static/css/src/issue-wizard.css
+++ b/webcompat/static/css/src/issue-wizard.css
@@ -419,6 +419,12 @@
   color: var(--link-color);
 }
 
+.input-control .screenshot-select-trigger {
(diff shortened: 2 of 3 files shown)


munteanu210:
@miketaylr and @karlcow Could you kindly review this PR, then merge & deploy it on staging if everything looks good? Thank you!

webcompat/templates/home-page/issue-wizard-form.html
```diff
@@ -311,7 +311,7 @@ <h3 class="page-heading">Filing a web compatibility issue</h3>
             <button class="button next-step step-10 issue-btn disabled" data-nextstep="11">Confirm</button>
             <button class="button js-remove-upload is-hidden issue-btn red">Delete this image</button>
           </div>
-          <a class="js-remove-upload is-hidden spaced-link" href="#">Upload a different image</a>
```
miketaylr: We're not running tests on this version of the form (which will come back to bite us in the near future), but `js-foo` classes are used by tests, e.g., https://github.com/webcompat/webcompat.com/blob/d56cd783366d725bcf2cf560723a590faf41e08f/tests/functional/image-uploads-non-auth.js#L142

Just something to document here for whoever comes back and writes tests in the future.

miketaylr:
@munteanu210 I've deployed the latest master to staging if you'd like to verify things!

PR #3012, opened 2019-10-22
Fixes #2985 - Add postMessage support for report site issue data

### LINKED ISSUE #2985: Add postMessage support for report site issue data (other than screenshots) to bugform.js
This issue tracks the work necessary for https://github.com/mozilla/webcompat-team-okrs/issues/107

Context:

In Firefox and Firefox for Android, when the user clicks "Report Site Issue", a few things happen:
1. some metadata about the page is collected (screenshot, console logs, value of some prefs, detect some libs, etc)
2. the metadata is `POST`ed to `webcompat.com/issues/new` in a new tab
3. the screenshot blob is sent to the webcompat form via `postMessage`

https://searchfox.org/mozilla-central/source/browser/extensions/report-site-issue/background.js#23 is where to start tracing the code. 

We would like to port Report Site Issue to Fenix (like, beyond the simple link it is today, see https://github.com/mozilla/webcompat-team-okrs/issues/110), but currently it does not support Experimental Web Extension APIs, which is how [`loadURIWithPostData`](https://searchfox.org/mozilla-central/source/browser/extensions/report-site-issue/experimentalAPIs/tabExtras.js#200) is implemented. And there are no web extension APIs to open a tab and `POST`, rather than `GET`. 

One way around this issue is to optionally send *all* the data via `postMessage`, along with the screenshot. We don't want to remove the POST abilities -- as it allows for HTTP client possibilities, but we should make sure we don't end up sending the data twice. Not 100% sure what that will look like right now, but I'm sure someone smart will think of something. 😄

ksy36:
These changes are to make the bug form accept metadata via postMessage from the report site issue addon in addition to screenshot blob. A message is expected in the following format:

```
{
    screenshot: blob, // image blob
    message: metadata // JSON-stringified metadata 
};
```

r? @miketaylr  and @karlcow


webcompat/views.py
```diff
@@ -263,8 +264,6 @@ def create_issue():
 
         if ab_active('exp') == 'form-v2':
             bug_form = get_form(form_data, form=FormWizard)
-            # TODO: remove this when the experiment has ended
-            form_data['extra_labels'].append('form-v2-experiment')
         else:
```
ksy36: I've moved this to the front end

webcompat/static/js/lib/bugform-prefill.js
```diff
@@ -0,0 +1,57 @@
...
+    Object.keys(config).forEach(function(key) {
+      if (data.hasOwnProperty(key)) {
+        var field = config[key];
+        field.element.val(prepareValue(field, data[key]));
+      }
+    });
+  };
+
+  var setAnalytics = function(campaign, source) {
+    if (!window.setAnalyticsData) return;
```
ksy36: This is to account for the inline script that is inserted based on "utm_campaign" and "utm_source" existence in request on prefill stage https://github.com/webcompat/webcompat.com/blob/master/webcompat/views.py#L273

webcompat/static/js/lib/bugform-prefill.js
```diff
@@ -0,0 +1,57 @@
...
+    extra_labels: {
+      element: $("#extra_labels"),
+      stringify: true,
+      label: true
+    }
+  };
+
+  var prepareValue = function(field, value) {
+    if (field.label && isExperiment) {
+      value.push("form-v2-experiment");
```
miketaylr: I like this.

ksy36:
I was testing it the following way: 

1. When a new tab is open, don't send postData [here](https://searchfox.org/mozilla-central/source/browser/extensions/report-site-issue/experimentalAPIs/tabExtras.js#262).

2.  Passed an object instead of an image blob [here](https://searchfox.org/mozilla-central/source/browser/extensions/report-site-issue/background.js#212)
```
        const data = {
          screenshot: blob,
          message: ${json}
        };
        postMessage(data, "${url.origin}");
```
So webcompat.com only receives a GET request in this case (with no parameters). And when the page is open it receives postMessage with metadata and screenshot

ksy36:
That makes me think that I need to account for two paths of handling data that comes in postMessage: 

If it's an image blob (as it is now), don't fill the form and only upload image. 
If it's an object, then fill the form and upload image. 

That way it won't break for users that don't get updated version of the addon yet

webcompat/helpers.py
```diff
@@ -722,3 +722,15 @@ def ab_init(response):
             response.set_cookie(exp_id, var, max_age=max_age)
 
     return response
+
+
+def get_extra_labels(form):
+    """If extra_labels exists in current session, use it,
+    otherwise use the value coming from form."""
```
karlcow: nit: The first line is a short summary of the function.
Then  two returns and the explanation if necessary.

I would probably do something along.

```python
"""Extract extra_labels.

If extra_labels exists in current session, use it,
otherwise use the value coming from form.
"""

File: tests/unit/test_helpers.py [hunk 5]
```diff
@@ -499,6 +500,16 @@ def test_ab_init_active(self):
             ab_init(response)
             response.set_cookie.assert_not_called()
 
+    def test_get_extra_labels(self):
+        """Test extra_label extraction from form object"""
```
karlcow: nit: missing the period at the end of the line.

```suggestion
        """Test extra_label extraction from form object."""
```

tests/functional/reporting-postMessage.js
```diff
@@ -0,0 +1,52 @@
...
+const { assert } = intern.getPlugin("chai");
+const { registerSuite } = intern.getInterface("object");
+const FunctionalHelpers = require("./lib/helpers.js");
+
+const url = intern.config.siteRoot + "/issues/new";
+
+// This string is executed by calls to `execute()` in various tests
+// it postMessages a small green test square.
+const POSTMESSAGE_TEST =
+  'postMessage({message: {"url":"http://www.test.com","utm_source":"desktop-reporter","utm_campaign":"report-site-issue-button","src":"desktop-reporter","details":{"gfx.webrender.all":false,"gfx.webrender.blob-images":true,"gfx.webrender.enabled":false,"image.mem.shared":true,"buildID":"20191016225400","channel":"default","consoleLog":[],"hasTouchScreen":false,"mixed active content blocked":false,"mixed passive content blocked":false,"tracking content blocked":"false","hasMarfeel":true},"extra_labels":["type-marfeel"]}}, "http://localhost:5000")';
```
karlcow: `test.com` is a real domain name, please replace with `example.com` which is a reserved domain specifically for documentation, see [RFC 6761](https://tools.ietf.org/html/rfc6761#section-6.2).

tests/functional/reporting-postMessage.js
```diff
@@ -0,0 +1,52 @@
...
+    "postMessaged object"() {
+      return (
+        FunctionalHelpers.openPage(this, url, ".js-report-buttons")
+          // send data object through postMessage
+          .execute(POSTMESSAGE_TEST)
+          .sleep(1000)
+          .findByCssSelector("#url")
+          .getProperty("value")
+          .then(function(value) {
+            assert.include(value, "http://www.test.com");
```
karlcow: See comment about domains and [RFC 6761](https://tools.ietf.org/html/rfc6761#section-6.2)

tests/unit/test_helpers.py
```diff
@@ -499,6 +500,16 @@ def test_ab_init_active(self):
             ab_init(response)
             response.set_cookie.assert_not_called()
 
+    def test_get_extra_labels(self):
```
karlcow: ❤️ for the tests!

webcompat/views.py
```diff
@@ -281,7 +280,7 @@ def create_issue():
             abort(400)
         # Adding parameters to the form
         form = request.form.copy()
-        extra_labels = session.pop('extra_labels', None)
```
karlcow: I haven't tested yet but what is happening with 

`/issues/new?label=browser-fenix`

karlcow:
ah also @ksy36 forgotten, the format of commits is not [in line with our guidelines](https://github.com/webcompat/webcompat.com/blob/master/docs/pr-coding-guidelines.md#add-commit-message-including-issue-number). :)

webcompat/views.py
ksy36: This part hasn't changed much, so the label will be saved to the current session here 
https://github.com/webcompat/webcompat.com/blob/master/webcompat/views.py#L271 

And then they're retrieved in the newly added method `get_extra_labels` (where the session  will be checked first, and then the form).

tests/functional/reporting-postMessage.js
ksy36: Ah thanks, I didn't know, I've changed that

ksy36:
sorry, didn't notice the commit format is different :(

Thanks for the feedback, I've addressed the comments here https://github.com/webcompat/webcompat.com/pull/3012/commits/4cafa8f775e006eee52e9c1b34f116f387a354cc

webcompat/static/js/lib/bugform.js
```diff
@@ -155,19 +155,40 @@ BugForm.prototype.onReceiveMessage = function(event) {
...
-    // the work of only accepting blobs, which should simplify things.
-    if (event.data instanceof Blob) {
-      // convertToDataURI sends the resulting string to the upload
-      // callback.
-      this.convertToDataURI(event.data, this.showUploadPreview.bind(this));
+    if (
+      event.data.hasOwnProperty("screenshot") ||
+      event.data.hasOwnProperty("message")
+    ) {
+      this.handleScreenshot(event.data.screenshot);
```
miketaylr: I'm not sure what the actual risk is, but in theory `event.data` could have a missing `screenshot` member, but have a `message` one. In that case, would `this.handleScreenshot(event.data.screenshot);` throw?

(Hard to say for sure, since right now Firefox doesn't send this object in this format currently (but I think you have local patches that do this).

miketaylr: edit: maybe the `||` is intended to be `&&`?

karlcow: @ksy36 this is the last question I guess before merging.

ksy36: Ahh thanks, I'll change this

ksy36:
@miketaylr Yeah, it's safe to land as is. As for sending details (in addition to screenshot) via postMessage I've created a bug for the desktop reporter https://bugzilla.mozilla.org/show_bug.cgi?id=1593684

webcompat/static/js/lib/bugform.js
miketaylr: > Ahh thanks, I'll change this

cool! Feel free to merge when you're ready.

miketaylr:
Note: had a conversation with @ksy36 about this, and she wants to hold off landing until she fixes a bug she found doing related work. We'll land this and the bugfix PR when it's ready.

webcompat/helpers.py
```diff
@@ -728,12 +728,19 @@ def get_extra_labels(form):
...
-    if not extra_labels and 'extra_labels' in form:
+    if not extra_labels and 'extra_labels' in form and form['extra_labels']:
         extra_labels = json.loads(form['extra_labels'])
 
+    if ab_active('exp') == 'form-v2':
+        if isinstance(extra_labels, list):
+            extra_labels.append('form-v2-experiment')
+        else:
+            extra_labels = ['form-v2-experiment']
+
```
ksy36: Changed my initial approach for adding experiment label since I've discovered a bug where the 'form-v2-experiment' label didn't get added if other labels exist.

ksy36:
@karlcow , @miketaylr some more changes to review, sorry :)

webcompat/helpers.py
karlcow: I will check that later today.

tests/unit/test_helpers.py
```diff
@@ -499,6 +500,48 @@ def test_ab_init_active(self):
...
+                {'extra_labels': '["type-marfeel", "browser-fenix"]'}),
+                ['type-marfeel', 'browser-fenix']
+            )
+            self.assertEqual(get_extra_labels({'extra_labels': '[]'}), [])
+            self.assertEqual(get_extra_labels({}), None)
+            self.assertEqual(get_extra_labels({'extra_labels': ''}), None)
+
+    def test_get_extra_labels_for_experiment(self):
+        """Test extra_labels extraction from form object if
+        experiment is active."""
```
karlcow: nit: one line for the docstring.

tests/unit/test_helpers.py
```diff
@@ -499,6 +500,48 @@ def test_ab_init_active(self):
...
+
+            webcompat.app.config['AB_EXPERIMENTS'] = {
+                'exp': {
+                    'variations': {
+                        'form-v2': (0, 100)
+                    },
+                    'max-age': 86400
+                }
+            }
+            webcompat.app.preprocess_request()
```
karlcow: Maybe let's add a comment above this, that we need `webcompat.app.preprocess_request()` because the `g.current_experiments = ab_current_experiments()` is defined in `before_request`, which is executed in tests only if `webcompat.app.preprocess_request()` is explicitly called.

webcompat/helpers.py
```diff
@@ -722,3 +722,25 @@ def ab_init(response):
...
+    """Extract extra_labels.
+
+    If extra_labels param exists in current session, use it,
+    otherwise use the value coming from form. Add form-v2-experiment
+    label if the experiment cookie exists.
+    """
+
+    extra_labels = session.pop('extra_labels', None)
+
+    if not extra_labels and 'extra_labels' in form and form['extra_labels']:
```
karlcow: @ksy36 This part of the code is slightly hard to read and I'm not sure it does what is expected.

```python
>>> form = {'a': 1, 'b': 2, 'c': 3}
>>> extra_labels = None
>>> not extra_labels and 'b' in form and form['b']
2
>>> not None
True
>>> 'b' in form
True
>>> form['b']
2
>>> not None and True and 2
2
```

ksy36: So for this: 

if `extra_labels` is not in the session, I check whether 'extra_labels' property exists in form and that it's not an empty string, otherwise if passing an empty string to `json.loads` I get this error `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

basically this test case covers this:
`self.assertEqual(get_extra_labels({'extra_labels': ''}), None)`

ksy36: but yeah it's hard to read, maybe I'll split this into separate functions

ksy36: I've refactored this and added some more tests here https://github.com/webcompat/webcompat.com/pull/3012/commits/aabf44faf6285337e5a357103df47c249e42f5c5

tests/unit/test_helpers.py
ksy36: thanks, added the comment

tests/unit/test_helpers.py
ksy36: I had it in one line, but the linter complained that it's too long

miketaylr: You can fix that with a magic `  # noqa` comment, but what would you recommend for a better docstring here @karlcow?

karlcow: yes :)

```
"""Test  extra_labels  extraction when A/B experiment is active.

And if you need to add more prose. Jump one line and add more
stuff. :)
"""
```

see https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

(A couple of hours later… doh… forgotten to push comment)

webcompat/helpers.py
karlcow: See https://github.com/webcompat/webcompat.com/pull/3012#commitcomment-35858847
and https://github.com/webcompat/webcompat.com/pull/3012#commitcomment-35858866

ksy36:
Thanks for taking time to look into this @karlcow . This looks much shorter :)
It still errors out though, if `extra_labels` is an empty string, e.g. 

```python
>>> form = {'extra_labels': ''}
>>> json.loads(form['extra_labels'])
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
so perhaps, I could do:
```python
 >>> json.loads(form['extra_labels'] or '[]')
[]
```

So, it'll be 
``` python
    if not extra_labels:
        extra_labels = json.loads(form.get('extra_labels', '[]') or '[]')
```
then my test passes

miketaylr:
TIL that dict.get will return an empty string, rather than the default value. Using `or []` like @ksy36 suggests seems like a good path forward.

ksy36:
Added that change, could you please take one more look @karlcow ?

karlcow:
> TIL that dict.get will return an empty string, rather than the default value. Using `or []` like @ksy36 suggests seems like a good path forward.

I'm looking a last time at the full code right now.

karlcow:
@ksy36 once you have resolved the conflict we can merge.

ksy36:
cool, thanks @karlcow

Issue #3005, opened 2019-10-22
User cannot proceed without uploading an image [New UI - Mobile]


mcroud:
Currently a user has no option to proceed without uploading an image: 

<img width="470" alt="Screenshot 2019-10-22 at 13 05 43" src="https://user-images.githubusercontent.com/42435754/67284037-b6ddb680-f4cc-11e9-9677-c656d1734a98.png">

This should be entirely optional, the yellow button should be clickable and say "Continue without"

<img width="390" alt="Screenshot 2019-10-22 at 13 07 15" src="https://user-images.githubusercontent.com/42435754/67284148-f4424400-f4cc-11e9-987b-32d1cb0517a3.png">

karlcow:
yes this is a major issue.

mcroud:
Thought this might come in handy!
I appreciate that the Miro flows concentrate on the "Report site issue..." route which may disguise the "optional" aspect of this step.

![webcompat com Mobile Designs](https://user-images.githubusercontent.com/42435754/67408936-b834df00-f5b1-11e9-8e6d-a801bfe4ca96.jpg)

miketaylr:
I just tested on staging, and it allows me to continue w/o a screenshot, but if I add a screenshot and delete it I'm stuck.

miketaylr:
This is fixed now.

PR #3017, opened 2019-10-23
Issue #3005 - Screenshot submit button shouldn't be disabled by default

johngian:
This PR fixes issue #3005 

## Proposed PR background

Submit button for the screenshot field shouldn't be disabled by default


johngian:
@miketaylr r?

miketaylr:
will deploy to staging.

PR #3028, opened 2019-10-24
Fixes #3009 - Adds a route for deployed version Git SHA

### LINKED ISSUE #3009: Adds a route with the git SHA of the latest deployed version
@johngian showed me this tool https://whatsdeployed.io/. It allows you to see what version of the repo is deployed to what env. Seems useful.

karlcow:
This PR fixes issue #3009

## Proposed PR background

This adds a route to the project to get the Git SHA of the current deployed version.
To note that this is basically a static request to a value in a file which is set by deployment scripts.
and which is initialized when the project is restarted. There is no point reading it for each request.

### DIFF
--- a/config/environment.py
+++ b/config/environment.py
@@ -113,3 +113,12 @@ Contact: mailto:miket@mozilla.com
 # AB setup
 # Comma separated list of user IDs to exempt from experiments
 AB_EXEMPT_USERS = os.environ.get('AB_EXEMPT_USERS', '').split(',')
+
+# SHA reference for the current deployed version
+# The Git SHA is written in data/ at deployment time.
+# This is used in views.py for /.well-known/deployed-version route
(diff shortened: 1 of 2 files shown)


config/environment.py
```diff
@@ -113,3 +113,12 @@
 # AB setup
 # Comma separated list of user IDs to exempt from experiments
 AB_EXEMPT_USERS = os.environ.get('AB_EXEMPT_USERS', '').split(',')
+
+# SHA reference for the current deployed version
+# The Git SHA is written in data/ at deployment time.
```
miketaylr: Is there a missing commit that writes out `sha.txt` into `data/` maybe?

miketaylr: edit: nope, it's just in another repo. :)

PR #3032, opened 2019-10-24
Fixes #3031 - Update privacy policy around experiments

### LINKED ISSUE #3031: Update privacy policy to document experiment cookies
This should probably block.

miketaylr:


webcompat/templates/privacy.html
```diff
@@ -68,6 +68,12 @@ <h3 class="headline-2">Cookies</h3>
             If you log into our site using your GitHub user account, webcompat.com will use cookies to record your session information. The cookie will expire when the session ends, i.e., when you click "logout".
         </p>
 
+        <p>
+            In order to measure possible improvements to the site, webcompat.com may use a cookie to serve different versions of the site to different users for the duration of an experiment. The experiment cookie will stay on your device until you delete it, or until it expires.
```
denschub: The cookie is not really needed for measurements, but for serving the experimental version, right? Measurements are covered by the Analytics section below. Can we make this something along the lines of

> To experiment with different improvements to the site, webcompat.com may use a cookie to serve alternative versions of the site to different users for the duration of an experiment. The experiment cookie will stay on your device until you delete it, or until it expires.

I'd probably also like to have a comment about the cookie only containing the site version to be served, not any PII. So, maybe add an additional

> The cookie contains only the experiment version you should receive and does not provide any personally identifiable information.

to the end?

miketaylr: Thank you for the suggestions!

miketaylr:
I've incorporated your feedback (with the one small change of `....does not record or provide any personally identifiable....`. Thanks!

PR #3064, opened 2019-10-28
Fixes #3063 - Move search bar handling into shared navbar.js

### LINKED ISSUE #3063: Clicking "Search" when on /issues/new from header does nothing
I guess we broke this with the redesign, but only tested it on the homepage (where it currently is working). 

https://github.com/webcompat/webcompat.com/blob/master/tests/functional/search-auth.js#L141-L171

miketaylr:
And add some missing tests.


webcompat/static/js/lib/homepage.js
```diff
@@ -1,41 +0,0 @@
...
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-function HomePage() {
-  var searchBar = $(".js-SearchBar");
-  var searchBarOpen = $(".js-SearchBarOpen");
-  var searchBarClose = $(".js-SearchBarClose");
-
-  this.init = function() {
-    var htmlClass =
```
miketaylr: I dug into the history, and this is unused (has been for a few years).

miketaylr:
r? @karlcow 

No new code here, just moving things around. But adding some new tests.

Issue #3118, opened 2020-01-02
[World is on Fire] Have a contingency plan when web-bugs repo disappears.


karlcow:
There is right now an incident.  The web-bugs repo is gone.

<img width="1013" alt="webcompat-web-bugs-repo" src="https://user-images.githubusercontent.com/505230/71700023-bd11dc80-2e05-11ea-8bd8-797032625f81.png">

We need a contingency plan and tools.

1. Missing a web page saying that the site is down currently.
2. Having a warning as early as possible on https://twitter.com/webcompat
    Ex: https://twitter.com/webcompat/status/1212880610089299968 
3. Have a backup of everything (at least comments and issues) so we can send a static version instead of the nothingness we currently have. #2449 
4. Consolidate our accounts information

<img width="977" alt="Capture d’écran 2020-01-03 à 08 51 17" src="https://user-images.githubusercontent.com/505230/71700130-3d384200-2e06-11ea-9bad-b1a36a7cb7c1.png">

karlcow:
As for the January 2020 incident.
* Detected by @ksy36 
* Reported by @miketaylr to GitHub and the internal mailing-list of Mozilla webcompat team
* Tweeted by @karlcow on webcompat account.

* 7:15am Only webcompat-bot issues are gone from the repo
* 8:33am The full repo is gone.

miketaylr:
We could revive some of the code from https://github.com/webcompat/webcompat.com/issues/2733 for a landing "page is down" page.

miketaylr:
I've sent some messages via LinkedIn (one to @clarkbw... hopefully GitHub @support can reverse this soon, and help us understand why it happened without any kind of heads up).

karlcow:
When the form can't communicate with github web-bugs. 
At the end of the process for users this is what happens. Probably another issue to solve in case of non communications with GitHub.

<img width="831" alt="Capture d’écran 2020-01-03 à 10 23 32" src="https://user-images.githubusercontent.com/505230/71702519-3d8b0a00-2e13-11ea-86e0-143238fda01c.png">

karlcow:
@miketaylr and I created an incident report with more details that we are updating once the details are coming in. 

@miketaylr also created a PR #3119 to have a landing page for the route /issues/new It was deployed today on staging and production.

miketaylr:
[...]

PR #3131, opened 2020-01-06
Fixes #3122 - Removes code related to images thumbnail

### LINKED ISSUE #3122: Remove inline screenshot images
Just link to them instead. See #3118

### LINKED ISSUE #3130: Tests involving the creation of labels have a dependency on topsites.db
When the webhook tests are running, the results depend on the topsites DB being populated or not. While trying to solve another issue and having a populated db, one test broke. 

This issue "fixes" only the breaking test, but it doesn't solve the dependency on the DB. We might want to check that later.

karlcow:
This PR fixes issue #3122 
This closes #3122, closes #3130 
## Proposed PR background

It removes the code related to thumbnail generation, following up on the events in #3118

### DIFF
--- a/webcompat/api/uploads.py
+++ b/webcompat/api/uploads.py
@@ -36,11 +36,8 @@ class BaseUpload(object):
         self.month = str(today.month)
         self.file_id = str(uuid.uuid4())
 
-    def get_file_path(self, year, month, file_id, ext, thumb=False):
-        thumb_string = ''
-        if thumb:
-            thumb_string = '-thumb'
-        file_name = file_id + thumb_string + '.' + ext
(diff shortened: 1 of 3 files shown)


webcompat/api/uploads.py
```diff
@@ -126,16 +118,11 @@ def save(self):
             save_parameters['save_all'] = True
         # unpacking save_parameters
         self.image_object.save(file_dest, **save_parameters)
-        # Creating the thumbnail
-        size = (1024, 1024)
-        self.image_object.thumbnail(size, Image.HAMMING)
```
miketaylr: goodbye, you beautiful ham

home-carousel-1

PR #3133, opened 2020-01-06
Fixes #3132 - Control anonymous reporting via config option

### LINKED ISSUE #3132: Make anonymous reporting configurable
In #3121 and other places I was pretty hacky about turning this off (just commenting it out, or ripping things out). 

I think we could have a config option in https://github.com/webcompat/webcompat.com/blob/master/config/__init__.py, something like `ENABLE_ANONYMOUS_REPORTS` and set it to `False`.

Then we can update tests to handle both scenarios, and check the config option before returning `400` here: https://github.com/webcompat/webcompat.com/blob/master/webcompat/issues.py#L27-L34.

I guess we could also control the button's state and text in the template too.

miketaylr:
r? @karlcow

### DIFF
--- a/config/environment.py
+++ b/config/environment.py
@@ -33,6 +33,7 @@ if PRODUCTION:
     SECRET_KEY = os.environ.get('PROD_SECRET_KEY')
     UPLOADS_DEFAULT_DEST = os.environ.get('PROD_UPLOADS_DEFAULT_DEST')
     UPLOADS_DEFAULT_URL = os.environ.get('PROD_UPLOADS_DEFAULT_URL')
+    ANON_REPORTING_ENABLED = os.environ.get('PROD_ANON_REPORTING_ENABLED')
 
 if STAGING:
     GITHUB_CLIENT_ID = os.environ.get('STAGING_GITHUB_CLIENT_ID')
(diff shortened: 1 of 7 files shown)


config/environment.py
```diff
@@ -57,6 +59,7 @@
     OAUTH_TOKEN = os.environ.get('OAUTH_TOKEN')
     SECRET_KEY = os.environ.get('SECRET_KEY') or 'NO SECRETS'
     HOOK_SECRET_KEY = os.environ.get('HOOK_SECRET_KEY') or 'SECRETS'
+    ANON_REPORTING_ENABLED = False
```
miketaylr: Maybe in the future we want to allow localhost anonymous reporting, but for now just set everything to False.

webcompat/issues.py
```diff
@@ -13,6 +13,7 @@
 
 from flask import abort
 
+from webcompat import app
```
miketaylr: oops, leftovers.

config/environment.py
```diff
@@ -33,6 +33,7 @@
     SECRET_KEY = os.environ.get('PROD_SECRET_KEY')
     UPLOADS_DEFAULT_DEST = os.environ.get('PROD_UPLOADS_DEFAULT_DEST')
     UPLOADS_DEFAULT_URL = os.environ.get('PROD_UPLOADS_DEFAULT_URL')
+    ANON_REPORTING_ENABLED = os.environ.get('PROD_ANON_REPORTING_ENABLED')
```
karlcow: What is happening if the value is not defined in environment?

Issue #3146, opened 2020-01-10
Make sure new private issues are assigned to the "unmoderated" milestone


miketaylr:
Depends on https://github.com/webcompat/webcompat.com/issues/3139 landing first.

https://github.com/webcompat/webcompat-tests-private/milestones already exists (as well as the corresponding milestone in the actual private repo).

We could do this in 2 ways:

1. Just assign the right milestone in the initial issue payload in `unmoderated_issue()` in issues.py

2. Write a webhook for new issues in the private repo to assign milestones for us.

We probably need to re-use the milestones checking machinery?

miketaylr:
@karlcow assuming https://github.com/webcompat/webcompat.com/pull/3147 is in the right direction, could you pick up this issue?

karlcow:
So I thought maybe we could try to use the fancy new GitHub Actions stuff.

And it didn't work. It [failed with the message](https://github.com/webcompat/webcompat-tests-private/commit/e99f787f9921e6c91de7deca55c7b4b76d676a2d/checks?check_suite_id=396879071):

> - Workflows can't be executed on this repository. Please check your payment method or billing status.

So searching a [bit the documentation](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions).

> GitHub Actions is available with GitHub Free, GitHub Pro, GitHub Team, GitHub Enterprise Cloud, and GitHub One. GitHub Actions is not available for public or private repositories owned by accounts using legacy per-repository plans. For more information, see "GitHub's products."

Let's check the [GitHub Free stuff](https://help.github.com/en/github/getting-started-with-github/githubs-products#github-free).

There is also the [Github Team Open Source](https://help.github.com/en/github/getting-started-with-github/githubs-products#github-team-for-open-source)

And I'm not sure [about the billing](https://help.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions).  So I guess we need to fall back on our traditional webhook.

karlcow:
The second method doesn't have a dependency on #3139 . It is working on just monitoring events.

karlcow:
Again here we have the issue of milestone number vs names.
https://github.com/webcompat/webcompat-tests-private/milestone/1


Here a non-tested possibility with webhook. Not very dry, not very clean. We could abstract the code so the previous covers the two cases **BUT** because the future of this code is bleak. I would rather not try to abstract if it's to do reconstructive surgery after. 



```diff
[...]

PR #3147, opened 2020-01-10
Fixes #3139 - Implement first stages of private reporting flow

### LINKED ISSUE #3139: Implement new anonymous issue creation flow
It may make sense to split this bug up, but keeping together for now because it's all related, and maybe awkward to implement in multiple PRs (maybe not?)

When a user goes to file an anonymous bug: 

1. Create a (public) issue in the web-bugs repository that contains the boilerplate “thanks, we will moderate, feel free to bookmark and check-back” text
2. Grab the Issue Number of the newly created public and create a private issue in web-bugs-private which associates the public issue number with it somehow. One idea, just store it in the issue body, for example we could re-use our: <!-- @public_url: foo > convention. Or we create a private -> public mapping in a database table.
3. <s>Make sure this new issue is an a “Unmoderated” milestone, if we’re taking the private GitHub repo shortcut.</s>: See https://github.com/webcompat/webcompat.com/issues/3146

miketaylr:
r? @karlcow 

The next one to tackle is probably #3146.

### DIFF
--- a/tests/unit/test_form.py
+++ b/tests/unit/test_form.py
@@ -147,26 +147,6 @@ class TestForm(unittest.TestCase):
         expected = '<!-- @browser: Firefox 59.0 -->\n<!-- @ua_header: Mozilla/5.0...Firefox 59.0 -->\n<!-- @reported_with: desktop-reporter -->\n<!-- @extra_labels: type-webrender-enabled -->\n'  # noqa
         self.assertEqual(actual, expected)
 
-    def test_get_metadata_with_extra_metadata_arg(self):
-        """HTML comments need the right values depending on the keys."""
-        metadata_keys = ('sky', 'earth')
-        extra_data = ('seatbelt', 'on!')
-        form_object = {'blah': 'goo', 'hello': 'moshi', 'sky': 'blue'}
-        actual = form.get_metadata(metadata_keys, form_object, extra_data)
-        expected = '<!-- @sky: blue -->\n<!-- @earth: None -->\n<!-- @seatbelt: on! -->\n'  # noqa
-        self.assertEqual(actual, expected)
-        form_object = MultiDict([
-            ('reported_with', 'desktop-reporter'),
-            ('url', 'http://localhost:5000/issues/new'),
-            ('extra_labels', ['type-webrender-enabled']),
-            ('ua_header', 'Mozilla/5.0...Firefox 59.0'),
-            ('browser', 'Firefox 59.0')])
-        metadata_keys = ['browser', 'ua_header', 'reported_with',
-                         'extra_labels']
-        actual = form.get_metadata(metadata_keys, form_object, extra_data)
-        expected = '<!-- @browser: Firefox 59.0 -->\n<!-- @ua_header: Mozilla/5.0...Firefox 59.0 -->\n<!-- @reported_with: desktop-reporter -->\n<!-- @extra_labels: type-webrender-enabled -->\n<!-- @seatbelt: on! -->\n'  # noqa
-        self.assertEqual(actual, expected)
-
     def test_get_metadata_browser_as_extra(self):
         """Test that we can handle a browser-foo inside of EXTRA_LABELS."""
         form_object = MultiDict([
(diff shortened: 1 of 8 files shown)


tests/unit/test_issues.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Tests for issue creation."""
```
miketaylr: These aren't great tests, because these methods aren't super testable.

webcompat/issues.py
```diff
@@ -26,10 +50,12 @@ def report_issue(form, proxy=False):
     submit_type = form.get('submit_type')
     if proxy and submit_type == 'github-proxy-report':
         # Return a Response object.
+        path = 'repos/{0}'.format(PRIVATE_REPO_URI)
         response = proxy_request('post',
                                  path,
-                                 data=json.dumps(build_formdata(form)))
+                                 data=json.dumps(unmoderated_issue()))
         json_response = response.json()
+        report_private_issue(form, json_response.get("html_url"))
```
miketaylr: This is slower than before, because we have to wait for 2 GitHub responses. I've filed https://github.com/webcompat/webcompat.com/issues/3148 to improve this, but I don't think we need to block anon reporting on it.

karlcow: So there's something I have difficulties to understand here. Why `report_private_issue`. `proxy_request` is already here. it's just the parameters which are changing.

I mean the idea to have function names for each tasks could clearly identified each scenario, but in this case, we probably need to give a name for each of them. Here it seems suddenly unbalanced.

webcompat/templates/form.html
```diff
@@ -130,7 +130,7 @@ <h1 class="headline-1">Report Site Issue</h1>
         <button
           class="button button-secondary {{ 'js-Button' if anon_reporting_enabled else '' }}"
           id="submitanon"
-          type="button"
+          type={{ 'submit' if anon_reporting_enabled else 'button' }}
```
karlcow: I don't think I understand this part of the code here. if `anon_reporting_enabled` is `False` should all the markup go away instead of just changing the type?

webcompat/issues.py
```diff
@@ -13,10 +13,34 @@
...
+
+
+def report_private_issue(form, public_url):
+    """Report the anonymous users issue to our private repo.
+
+    This also allows us to pass in public_url metadata, to be
+    embedded in the issue body.
+    """
+    path = 'repos/{0}'.format(PRIVATE_REPO_URI)
+    formdata = build_formdata(form, ('public_url', public_url))
```
karlcow: hmm… not sure the code should extend the logic here. Probably there should be another function for adding stuff to the form object, which is just a dictionary.

So if there was a 

```python
form = add_metadata(form, {'key': 'value'})
```

or even more directly:

```python
form['public_url'] = public_url
```

then the modification to `build_formdata` is minimal.

```diff
-    metadata_keys = ['browser', 'ua_header', 'reported_with']
+    metadata_keys = ['browser', 'ua_header', 'reported_with', 'public_url']
```

webcompat/form.py
```diff
@@ -445,7 +447,7 @@ def domain_name(url):
     return domain
 
 
-def build_formdata(form_object):
+def build_formdata(form_object, extra_metadata=None):
```
karlcow: Probably it would be better to have something like `add_metadata` to modify the `form` dictionary. Specifically when `build_formdata` has already mechanism for adding metadata into the form.

webcompat/form.py
```diff
@@ -383,6 +383,8 @@ def get_metadata(metadata_keys, form_object):
     metadata = [(md[0], normalize_metadata(md[1])) for md in metadata]
     if extra_labels:
         metadata.append(('extra_labels', ', '.join(extra_labels)))
+    if extra_metadata:
+        metadata.append(extra_metadata)
```
karlcow: See my previous comment about build_formdata

webcompat/issues.py
```diff
@@ -13,10 +13,34 @@
...
 from webcompat import github
 from webcompat.form import build_formdata
 from webcompat.helpers import proxy_request
-from webcompat.helpers import REPO_URI
+
+REPO_URI = app.config['ISSUES_REPO_URI']
+PRIVATE_REPO_URI = app.config['PRIVATE_REPO_URI']
+
+
+def unmoderated_issue():
```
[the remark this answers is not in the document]
miketaylr: Correct, I think it's fine to fix this in a follow-up issue (before we deploy). My intent was to get something sort of working as a base to make forward progress. We don't have the templates ready yet, so I didn't want to invest too much energy here:

https://github.com/webcompat/webcompat.com/issues/3150

webcompat/issues.py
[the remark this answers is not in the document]
miketaylr: > json_response.get("html_url") will not necessary create a URL, if the answer is different than 201. so data will be maybe saved in private repo uri, without an origin uri.

Good point, I can update this.

webcompat/templates/form.html
miketaylr: Yeah, my bad. This is a confusing ride-along commit in the context of this PR. I just noticed I forgot to include it as part of #3132

Changing the button `type` is just an extra-careful measure, it will already be `disabled`, and the text will say it's disabled, and the backend will reject any issues that somehow slip through. If you feel strongly about a different proposal, can you file a follow-up issue? I don't have strong feelings about the way I've done it here.

webcompat/issues.py
miketaylr: I'll try that approach, I agree this part felt a bit gross after it evolved.

webcompat/issues.py
miketaylr: No specific reason, I'll add a `report_public_issue` method to restore balance to the force.

webcompat/issues.py
miketaylr: I've implemented the direct solution, but I might create the add_metadata way in a bit -- it seems useful perhaps. Thanks for the suggestions, I was missing the obvious simple solution. :)))

miketaylr:
OK, @karlcow, ready for re-review. Thanks for the suggestions so far!

tests/unit/test_form.py
```diff
@@ -195,6 +195,28 @@ def test_build_formdata(self):
         expected = {'body': '<!-- @browser: None -->\n<!-- @ua_header: None -->\n<!-- @reported_with: None -->\n\n**URL**: http://\u611b\n\n**Browser / Version**: None\n**Operating System**: None\n**Tested Another Browser**: Unknown \n\n**Problem type**: Unknown\n**Description**: None\n**Steps to Reproduce**:\nNone\n\n\n\n_From [webcompat.com](https://webcompat.com/) with \u2764\ufe0f_', 'title': '\u611b - unknown'}  # noqa
         self.assertEqual(actual, expected)
 
+    def test_build_formdata_with_add_metadata(self):
+        """The data body sent to GitHub API, with additional metadata."""
+        # we just need to test that nothing breaks
+        # even if the data are empty
+        form_object = {'foo': 'bar'}
+        form_object = form.add_metadata(form_object, {'public_url':
+                                        'https://public.example.com'})
```
karlcow: so there might be an issue where it would be worth to test. Imagine the url is 

`https://public.example.com/life-->is-hard` or maybe even `https://public.example.com/life--is-hard`

`--` signifies the end of a comment. 

hmmm just tested and it seems that both webcompat and github manages things.

webcompat/issues.py
```diff
@@ -13,10 +13,47 @@
...
+    proxy_request('post', path, data=json.dumps(formdata))
+    return None
+
+
+def report_public_issue(form):
+    """Report the issue publicly.
+
+    Returns a requests.Response object.
+    """
+    path = 'repos/{0}'.format(PRIVATE_REPO_URI)
```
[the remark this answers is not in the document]
miketaylr: oops lol

miketaylr:
OK, pushed a fix!

karlcow:
@miketaylr  Thanks for the code, the pasta 🍝 and everything.

PR #3153, opened 2020-01-14
Fixes #3146 - Assigns unmoderated milestone to private repo issues

### LINKED ISSUE #3146: Make sure new private issues are assigned to the "unmoderated" milestone

karlcow:
This PR fixes issue #3146

## Proposed PR background

This just assigns the right milestones to the issue posted on the private repo.

r? @miketaylr @ksy36

### DIFF
--- a/webcompat/issues.py
+++ b/webcompat/issues.py
@@ -21,6 +21,7 @@ from webcompat.helpers import proxy_request
 
 REPO_URI = app.config['ISSUES_REPO_URI']
 PRIVATE_REPO_URI = app.config['PRIVATE_REPO_URI']
+PRIVATE_REPO_MILESTONE = app.config['PRIVATE_REPO_MILESTONE']
 
 
 def unmoderated_issue():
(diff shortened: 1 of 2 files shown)


webcompat/issues.py
```diff
@@ -41,8 +42,11 @@ def report_private_issue(form, public_url):
     Returns None (so we don't accidentally leak data).
     """
     path = 'repos/{0}'.format(PRIVATE_REPO_URI)
+    milestone = PRIVATE_REPO_MILESTONE
     form = add_metadata(form, {'public_url': public_url})
     formdata = build_formdata(form)
+    # add the milestone number to set
+    formdata['milestone'] = milestone
```
miketaylr: I don't think this is enough to set the milestone alone:

https://github.com/webcompat/webcompat.com/blob/588b1d1c97eccf8423c834efc272aeacc080006e/webcompat/form.py#L480

https://github.com/webcompat/webcompat.com/blob/588b1d1c97eccf8423c834efc272aeacc080006e/webcompat/form.py#L558

We need to send a `milestone` key in the POST data to GitHub, so somewhere we need to do something like:

```
if 'milestone' in formdata:
    rv.update({`milestone`: milestone})
```

(or however you prefer to add a k/v to an existing dict.

miketaylr: oh wait, i mistunderstood. you're adding this *after* calling `build_formdata`. i read that as *before*.

So we're good here!

PR #3154, opened 2020-01-14
Fixes #3136 - Adds inline images but inside details/summary markup

### LINKED ISSUE #3136: Add back inline screenshot hidden inside <details> by default
See #3122 for relevant code to tweak.

karlcow:
This PR fixes issue #3136

## Proposed PR background

We had removed the inline images, to avoid too much into your face screenshot. Putting them behind a details/summary element is hiding it. 

Also we had to use directly the `img` element instead of the markdown markup. Because markdown is being escaped inside the `details` element.

### DIFF
--- a/webcompat/static/js/lib/issue-wizard-bugform.js
+++ b/webcompat/static/js/lib/issue-wizard-bugform.js
@@ -1373,7 +1373,12 @@ BugForm.prototype.getDataURIFromPreviewEl = function() {
 */
 BugForm.prototype.addImageURL = function(response) {
   var img_url = response.url;
-  var imageURL = ["[Screenshot](", img_url, ")"].join("");
+  var imageURL = [
+    "<details><summary>View the screenshot</summary>",
+    "<img alt='Screenshot' src='",
+    img_url,
+    "'></details>"
+  ].join("");
 
   this.stepsToReproduceField.val(function(idx, value) {
     return value + "\n" + imageURL;
--- a/tests/functional/image-uploads-non-auth.js
+++ b/tests/functional/image-uploads-non-auth.js
@@ -111,7 +111,7 @@ registerSuite("Image Uploads (non-auth)", {
           .then(function(val) {
             assert.notInclude(
               val,
-              "[Screenshot(http://localhost:5000/uploads/",
+              "<img alt='Screenshot' src='http://localhost:5000/uploads/",
               "The data URI was not uploaded before form submission."
             );
           })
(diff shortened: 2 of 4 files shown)


magsout:
@karlcow do you forgot to update this line https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L769? or it's intentional?

webcompat/static/js/lib/issue-wizard-bugform.js
```diff
@@ -1373,7 +1373,12 @@ BugForm.prototype.getDataURIFromPreviewEl = function() {
 */
 BugForm.prototype.addImageURL = function(response) {
   var img_url = response.url;
-  var imageURL = ["[Screenshot](", img_url, ")"].join("");
+  var imageURL = [
+    "<details><summary>View the screenshot</summary>",
+    "<img alt='Screenshot' src='",
```
magsout: typo?

` "Screenshot ` "<img alt='Screenshot' src=",`

magsout: maybe is better using template string, more readable, what do you think?
`<details><summary>View the screenshot</summary>Screenshot</details>`

karlcow: ah no the typo is elsewhere 😝 

```diff
-       "></details>"
+       "'></details>"
```

karlcow: template string seems to be a good idea.

miketaylr: The one reason to not use a template strings (yet) would be cross-browser support: https://caniuse.com/#feat=template-literals.

magsout: @miketaylr ah, yes.. true. even on webcompat? I mean about the stats of browsers version of webcompat user? But yes, maybe not a good idea after all.

karlcow:
> @karlcow do you forgot to update this line https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/js/lib/bugform.js#L769? or it's intentional?

@magsout Are you sure?
https://github.com/webcompat/webcompat.com/pull/3154/files#diff-ce241059c28a13e03bcd0829da8d4eb7L769

webcompat/static/js/lib/issue-wizard-bugform.js
karlcow: ok modification done. fixed the quote issue in the 3 instances where it appeared.

karlcow:
@magsout we are good for a re-check.

webcompat/static/js/lib/issue-wizard-bugform.js
miketaylr: > even on webcompat? I mean about the stats of browsers version of webcompat user?

I just checked, and IE users are 1.87% of our users (wow!). Not super high, but if we can avoid an error for them until they disappear it seems good.

Issue #3141, opened 2020-01-09
Create method to PATCH existing issue with failed moderation template


miketaylr:
Using the GitHub API, we send a PATCH request to the existing public issue and replace the body with the “this is unacceptable” template. Also:

1. Add labels from private issue. Add an “status-moderation-rejected” label.
2. Close the public issue as WONTFIX.

miketaylr:
This is probably a duplicate of #3140?

Depends on #3150.

karlcow:
> Add labels from private issue.

do we need those?

karlcow:
See https://github.com/webcompat/webcompat.com/pull/3167/commits/e4aa9d22eae8ab99541c659b7e5c0fccbba0c0b1

Issue #3140, opened 2020-01-09
PATCH public issue with accepted content and title


miketaylr:
If an issue is moderated as acceptable we need to do the following (regardless of how or where we stored bug reports from the previous section):

1. Using the GitHub API, we send a PATCH request to the existing public issue and replace the body with the stored private details.
2. Make sure the labels match. Probably good to add a “status-moderation-accepted” label until we decide it's useless.

miketaylr:
Depends on #3150.

karlcow:
@ksy36 @miketaylr This is an RFC.

In fact there is no dependency on #3150. 
Probably #3141 is exactly the same than this here.

But there are probably two missing  milestones: 

* `accepted`
* `not-acceptable`

Let's think this

When the issue on `PRIVATE_REPO_URI` is assigned the milestone `accepted`, then a webhook will start the publishing process. 

1. monitor milestone changes
2. When `milestone == 'accepted'`
3. Read the issue content on `PRIVATE_REPO_URI`. The full content should already be formatted. 
4. With `webcompat-bot`, post to `ISSUES_REPO_URI` the body with an additional `status-moderation-accepted`.
5. Close (and lock?) the issue on `PRIVATE_REPO_URI`

This might work.

Where I'm a bit worried is the  `not-acceptable` milestone and human errors, which has the same mechanism but instead post a template saying no. 


**Let's factor in, the human error.** 


1. if it's a good issue, selecting by mistake `not acceptable`. It will do no harm. The person just need to reopen the issue and change to accepted.
2. if it's a bad issue, selecting by mistake `accepted` has bad consequences.

Probably we should avoid the `not acceptable` milestone and use another trigger for closing issues. To `close a bad issue`, maybe we should just do that **close the issue**. Then the webhook will monitor the status change and will patch the PUBLIC REPO to close the issue.

miketaylr:
>  Probably #3141 is exactly the same than this here.

Yeah, likely they can both be closed at the same time.

> When the issue on PRIVATE_REPO_URI is assigned the milestone accepted, then a webhook will start the publishing process.

I was thinking about this a little differently, but also... pretty close to your idea. 

The idea is basically captured in #3144 (which can also be closed as a dupe of this bug). My thinking was that the private and public repos have the same milestones, with the exception of an additional `unmoderated` milestone in the private repo.

One of 3 things is then possible:

[...]

PR #3156, opened 2020-01-15
Fixes #3150 - Send the "moderation in process" message to the public issue

### LINKED ISSUE #3137: Create copy for 2 moderation scenarios
We need text to cover 2 scenarios:

A pre-moderation message: something like, thanks for reporting an issue you can check back here once it's been moderated and made public. 

A post-moderation rejection message: this report doesn't follow our acceptable use guidelines. 

Note we don't have acceptable use guidelines, so it probably has to be more generic, or we adopt or hack Mozilla's: https://www.mozilla.org/en-US/about/legal/acceptable-use/

### LINKED ISSUE #3150: Send the "moderation in process" message to the public issue
@karlcow has a suggestion here: https://github.com/webcompat/webcompat.com/pull/3147#discussion_r365616901

It will return the text from #3137

ksy36:
r? @karlcow or @miketaylr 

closes #3137 
closes #3150

### DIFF
--- a/tests/unit/test_issues.py
+++ b/tests/unit/test_issues.py
@@ -9,12 +9,16 @@
 
 import unittest
 import requests
+import json
 
 from mock import patch
+from mock import ANY
 from werkzeug import MultiDict
 
 from webcompat.issues import report_issue
 from webcompat.issues import report_private_issue
+from webcompat.issues import report_public_issue
+from webcompat.issues import moderation_template
 
 
 class TestIssue(unittest.TestCase):
@@ -60,3 +64,10 @@ class TestIssue(unittest.TestCase):
             ('username', ''), ])
         rv = report_private_issue(form, 'http://public.example.com')
         self.assertIsNone(rv)
+
+    @patch.object(requests, 'post')
+    def test_report_public_issue_returns_moderation_template(self, mockpost):
+        """Test that we get a moderation template back from a public issue report."""   # noqa
+        report_public_issue()
+        data = json.dumps(moderation_template())
+        mockpost.assert_called_with(ANY, data=data, headers=ANY, params=ANY)
--- a/webcompat/issues.py
+++ b/webcompat/issues.py
@@ -23,12 +23,17 @@ REPO_URI = app.config['ISSUES_REPO_URI']
 PRIVATE_REPO_URI = app.config['PRIVATE_REPO_URI']
 
 
-def unmoderated_issue():
+def moderation_template():
     """Gets the placeholder data to send for unmoderated issues."""
-    # TODO: Replace this with something meaningful.
-    # See https://github.com/webcompat/webcompat.com/issues/3137
-    summary = 'Placeholder in-moderation title.'
-    body = 'Placeholder in-moderation body.'
+
+    summary = 'In the moderation queue.'
+    body = """<p>This issue has been put in the moderation queue. A human
+will review if the message meets our current
+<a href="https://www.mozilla.org/en-US/about/legal/acceptable-use/">
+acceptable use</a> guidelines.
+It will probably take a couple of days depending on the backlog.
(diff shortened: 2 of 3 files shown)


tests/unit/test_issues.py
```diff
@@ -9,12 +9,16 @@
 
 import unittest
 import requests
+import json
 
 from mock import patch
+from mock import ANY
```
karlcow: Note to self about [ANY](https://docs.python.org/3/library/unittest.mock.html#any).

webcompat/issues.py
```diff
@@ -23,12 +23,17 @@
...
 
-def unmoderated_issue():
+def moderation_template():
     """Gets the placeholder data to send for unmoderated issues."""
-    # TODO: Replace this with something meaningful.
-    # See https://github.com/webcompat/webcompat.com/issues/3137
-    summary = 'Placeholder in-moderation title.'
-    body = 'Placeholder in-moderation body.'
+
+    summary = """In the moderation queue."""
```
karlcow: nit:
* The `"""` are not required here as it is only one line.  
* Also we use (for now) single quote styles aka `'bla'` or `'''bla'''` except for docstring. We might change one day, but for now it's what we did.

webcompat/issues.py
```diff
@@ -23,12 +23,17 @@
 PRIVATE_REPO_URI = app.config['PRIVATE_REPO_URI']
 
 
-def unmoderated_issue():
+def moderation_template():
```
miketaylr: Do we want this to take an argument, and it will return the "in moderation" vs "you're a bad person" body content?

(For public issue, it will always return "in moderation", but then we could re-use the method if something fails moderation later on)

karlcow: I will modify it probably when doing #3141 (aka #3140)

webcompat/issues.py
ksy36: thanks! fixed this

webcompat/issues.py
ksy36: yeah, sounds good to me

ksy36:
could you please take another look @karlcow

webcompat/issues.py
```diff
@@ -23,12 +23,17 @@
...
     """Gets the placeholder data to send for unmoderated issues."""
-    # TODO: Replace this with something meaningful.
-    # See https://github.com/webcompat/webcompat.com/issues/3137
-    summary = 'Placeholder in-moderation title.'
-    body = 'Placeholder in-moderation body.'
+
+    summary = 'In the moderation queue.'
+    body = """<p>This issue has been put in the moderation queue. A human
```
karlcow:
```suggestion
    body = '''<p>This issue has been put in the moderation queue. A human
```

webcompat/issues.py
```diff
@@ -23,12 +23,17 @@
...
+will review if the message meets our current
+<a href="https://www.mozilla.org/en-US/about/legal/acceptable-use/">
+acceptable use</a> guidelines.
+It will probably take a couple of days depending on the backlog.
+Once it has been reviewed, the content will be either made public
+or deleted.</p>"""
```
karlcow:
```suggestion
or deleted.</p>'''
```

ksy36: Ahh so I wasn't sure if you're referring to this part as well @karlcow. 
Didn't find any occurrences of `'''some text'''` being used with other than docstring. 

There is `"""some text"""` is used if `form.py` for filling out issue details:

https://github.com/webcompat/webcompat.com/blob/master/webcompat/form.py#L338
https://github.com/webcompat/webcompat.com/blob/master/webcompat/form.py#L526

karlcow: indeed we seem to have inconsistencies. One day we will probably need to bit the bullet and just fix it with black. Maybe worth adding as topic to Berlin.

PR #3181, opened 2020-02-05
Fixes #3180 - Use full not (relative) for terms page.

### LINKED ISSUE #3180: acceptable-use hyperlink should not be relative
If you click on it from github, it will 404:

e.g. https://github.com/webcompat/webcompat-tests/issues/2361

miketaylr:
Maybe this is awkward because we're hard-coding production links, even on staging, but I can't imagine why we would have a different set of terms between the two.

The intent here is to prevent following a hyperlink from GitHub (or email notification) and having that 404 on GitHub's side.

### DIFF
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -652,7 +652,7 @@ class TestWebhook(unittest.TestCase):
     def test_prepare_rejected_issue(self):
         """Test we prepare the right payload for the rejected issue."""
         expected = {'body': "<p>The content of this issue doesn't meet our\n"
-                    '<a href="/acceptable-use">acceptable use</a>\n'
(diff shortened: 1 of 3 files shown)


ksy36:
Looks good to me, perhaps we should change this link to point to webcompat.com instead of mozilla.org? (think it was added before the creation of terms page)

https://github.com/webcompat/webcompat.com/blob/63ebeb05b4ee26481515f0c4099ff0800d32ccfc/webcompat/issues.py#L31-L37

miketaylr:
> Looks good to me, perhaps we should change this link to point to webcompat.com instead of mozilla.org? (think it was added before the creation of terms page)

Good point, our terms page points to that Mozilla page. I'll change it!

ksy36:
Cool thanks!

Issue #3178, opened 2020-02-05
500 in webhook on staging


miketaylr:
Trying to sort out STR:

`public_number = get_public_issue_number(issue_info['public_url'])#012KeyError: 'public_url'`

```
Feb  5 17:09:36 staging-server webcompat-staging: [2020-02-05 17:09:36,463] ERROR in app: Exception on /webhooks/labeler [POST]#012Traceback (most recent call last):#012  File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 2292, in wsgi_app#012    response = self.full_dispatch_request()#012  File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1815, in full_dispatch_request#012    rv = self.handle_user_exception(e)#012  File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1718, in handle_user_exception#012    reraise(exc_type, exc_value, tb)#012  File "/usr/local/lib/python3.7/dist-packages/flask/_compat.py", line 35, in reraise#012    raise value#012  File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1813, in full_dispatch_request#012    rv = self.dispatch_request()#012  File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1799, in dispatch_request#012    return self.view_functions[rule.endpoint](**req.view_args)#012  File "./webcompat/webhooks/__init__.py", line 38, in hooklistener#012    response = process_issue_action(issue_info)#012  File "./webcompat/webhooks/helpers.py", line 278, in process_issue_action#012    response = private_issue_rejected(issue_info)#012  File "./webcompat/webhooks/helpers.py", line 374, in private_issue_rejected#012    public_number = get_public_issue_number(issue_info['public_url'])#012KeyError: 'public_url'
[...]

PR #3182, opened 2020-02-06
Issue #3178 - Adds public_url for action: closed

karlcow:
This PR fixes issue #3178 

## Proposed PR background

we are now making sure the public_url is present in the data we receive.


karlcow:
ok @miketaylr @ksy36 here we are. Thanks for spotting it and opening the issue.

PR #3190, opened 2020-02-07
Fixes #3188 - Account for invalid domains before checking for is_darknet_domain

### LINKED ISSUE #3188: "Internal Server Error(500)" message displayed when filing an issue using an invalid URL on staging.webcompat.com
**Steps to reproduce**:

1. Access staging.webcompat.com.
2. File an issue using an invalid URL e.g. "https:www.test.com".
3. Report the issue anonymously.
4. Observe the behavior.

**Expected result**:
The issue must be successfully logged.

**Actual result**:
"Yikes! Internal Server Error (500)" message is displayed.

**Notes**:
1. Using the "https://www.test.com" valid URL files the issue without errors.
2. Screenshot attached.

image

ksy36:
If a domain with an incorrect scheme gets reported (e.g. `https:www.test.com`) , urlsplit returns `None` as a hostname:
https://github.com/webcompat/webcompat.com/blob/4fcbd519fdef4ec07ba8236b3e42624e4f34e982/webcompat/views.py#L302-L303

In this case `is_darknet_domain` method fails with `AttributeError: 'NoneType' object has no attribute 'endswith'` 

To fix this I've added `normalize_url(form['url'])` to the above method

r? @miketaylr

### DIFF
--- a/tests/unit/test_helpers.py
+++ b/tests/unit/test_helpers.py
@@ -602,6 +602,7 @@ class TestHelpers(unittest.TestCase):
         self.assertTrue(is_darknet_domain('www.gjobqjj7wyczbqie.onion'))
         self.assertFalse(is_darknet_domain('example.com'))
         self.assertFalse(is_darknet_domain('gjobqjj7wyczbqie.onion.com'))
+        self.assertFalse(is_darknet_domain(None))
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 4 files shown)


miketaylr:
@ksy36 ^^

ksy36:
Added some tests @miketaylr

miketaylr:
Thank you!

PR #3200, opened 2020-02-12
fixes #3197 - Adds fencing on accepted issues being closed in private

### LINKED ISSUE #3197: Closing a private issue after it's been moved to "accepted" sets the moderation template to rejected
I don't know if this is expected behavior...

STR: 
1. file anon issue
2. moderate as accepted
3. close issue

Expected: nothing changes
Actual: the public issue is set to rejected

It seems useful to be able to close accepted issues, so we don't have to keep looking at them. I guess the workaround is to filter on milestone, but the default view of "is:issue is:open" seems easier to work with.

https://github.com/webcompat/web-bugs/issues/48038 is an example where I accidentally discovered this. The history looks messy :P

karlcow:
This PR fixes issue #3197

## Proposed PR background

it adds a couple of tests for the case where an accepted issue in the private repo is being closed.
it fences the editing of the public repo on not having accepted status.
it modifies the issue_info model

### DIFF
--- a/webcompat/webhooks/helpers.py
+++ b/webcompat/webhooks/helpers.py
@@ -164,6 +164,9 @@ def get_issue_info(payload):
     # If labels on the original issue, we need them.
     original_labels = [label['name'] for label in labels]
     issue_info['original_labels'] = original_labels
+    # webhook with a milestone already set
+    if issue.get('milestone'):
+        issue_info['milestone'] = issue['milestone']['title']
     # webhook with a milestoned action
     if payload.get('milestone'):
         issue_info['milestoned_with'] = payload.get('milestone')['title']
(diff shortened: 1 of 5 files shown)


webcompat/webhooks/helpers.py
```diff
@@ -272,7 +275,9 @@ def process_issue_action(issue_info):
         else:
             msg_log('private:moving to public failed', issue_number)
             return ('ooops', 400, {'Content-Type': 'text/plain'})
-    elif (scope == 'private' and issue_info['state'] == 'closed'):
+    elif (scope == 'private' and
+          issue_info['state'] == 'closed' and
+          not issue_info['milestone'] == 'accepted'):
```
miketaylr: 👍

Issue #3106, opened 2019-12-09
[formv2] What is step9?


miketaylr:
It's in the JS, but there's no associated DOM element:

https://github.com/webcompat/webcompat.com/search?q=step9&unscoped_q=step9

miketaylr:
Maybe this is it... https://github.com/webcompat/webcompat.com/blob/f279404888772e029add202bc6d6e3f86c574fbe/webcompat/templates/issue-wizard-form.html#L254

Given that it appears between Steps 8 and 10, and `data-nextstep="10"`

miketaylr:
No, I'm just confused. Apparently there's 2 classes `step8` and `step-8`. 🤷‍♂

ksy36:
This was addressed in https://github.com/webcompat/webcompat.com/pull/3246

PR #3246, opened 2020-03-19
Fixes #3229 - Make each step in the wizard an individual module

### LINKED ISSUE #3229: Make it so each step in the wizard is an individual module
In this issue we need to refactor existing code in a way that each step becomes an individual module. An ability to navigate through steps should persist without any additional functionality (validation, image uploading, etc.).

ksy36:
I've made it so each step an individual module. It's possible to navigate through steps without any additional functionality (validation, image uploading, submitting the form, etc.), as this will be addressed in separate issues. 

r? @miketaylr

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,7 @@
 {
   "parserOptions": {
-    "ecmaVersion": 6
+    "ecmaVersion": 6,
+    "sourceType": "module"
   },
   "env": {
     "browser": true
@@ -20,6 +21,7 @@
     "console": true,
     "define": true,
     "ga": true,
+    "intern": true,
     "markdownitEmoji": true,
     "markdownitSanitizer": true,
     "md": true,
@@ -46,6 +48,7 @@
     "no-prototype-builtins": "off",
     "no-use-before-define": 2,
     "one-var": [2, "never"],
-    "prettier/prettier": "error"
+    "prettier/prettier": "error",
+    "no-unused-vars": "off"
   }
 }
--- a/package.json
+++ b/package.json
@@ -25,6 +25,9 @@
     "yargs": "^15.0.1"
   },
   "devDependencies": {
+    "@babel/core": "^7.8.7",
+    "@babel/preset-env": "^7.8.7",
+    "@babel/register": "^7.8.6",
     "eslint": "^6.1.0",
     "eslint-config-prettier": "^6.0.0",
     "eslint-plugin-prettier": "^3.0.0",
@@ -49,8 +52,9 @@
     "postcss-import": "^12.0.0",
     "postcss-reporter": "^6.0.0",
     "postcss-url": "^8.0.0",
-    "stylelint": "^13.1.0",
     "prettier": "1.19.0",
+    "sinon": "^9.0.1",
+    "stylelint": "^13.1.0",
     "stylelint-config-standard": "^20.0.0",
     "stylelint-order": "^4.0.0",
     "svg-sprite-generator": "0.0.7",
@@ -80,6 +84,7 @@
     "project-update": "pip install --upgrade pip && npm update",
     "test": "npm run test:js && npm run test:python",
     "test:js": "node ./tests/functional/_intern.js",
+    "test:unit-js": "intern config=./webcompat/static/js/lib/wizard/tests/intern.json",
     "test:python": "pip install -e . && pytest"
   },
   "husky": {
--- a/webcompat/static/css/src/issue-wizard.css
+++ b/webcompat/static/css/src/issue-wizard.css
@@ -225,7 +225,7 @@
   opacity: 0;
 }
 
-.step-container.step10 {
+.step-container.step9 {
   animation-name: slideupbig;
 }
 
--- a/webcompat/static/js/lib/wizard/notify.js
+++ b/webcompat/static/js/lib/wizard/notify.js
@@ -1,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const channels = {};
+
+export default {
+  subscribe: function(channel, cb) {
+    if (!channel) throw Error("Please provide channel name to subscribe");
+    if (!cb)
+      throw Error(
+        "Please provide callback function to be called once there is a publish to a channel"
+      );
+
+    if (!channels.hasOwnProperty(channel)) {
+      channels[channel] = [];
+    }
+
+    const index = channels[channel].push(cb) - 1;
+
+    return {
+      unsubscribe: function() {
+        delete channels[channel][index];
+      }
+    };
+  },
+  publish: function(channel, data) {
+    if (!channels.hasOwnProperty(channel) || !channels[channel].length) {
+      return;
+    }
+
+    channels[channel].forEach(cb => cb(data || {}));
+  }
+};
--- a/webcompat/static/js/lib/wizard/step1.js
+++ b/webcompat/static/js/lib/wizard/step1.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* Allows the user to enter a URl and navigate to the next step */
+
 import notify from "./notify.js";
 
 const urlField = $("#url");
@@ -20,7 +22,7 @@ const showStep = (id, data) => notify.publish("showStep", { id, data });
 
 const makeAStep = event => {
   event.preventDefault();
-  showStep(1, { url: extractPrettyUrl(urlField.val()) });
+  showStep(2, { url: extractPrettyUrl(urlField.val()) });
 };
 
 urlField.on("blur input", checkUrl);
--- a/webcompat/static/js/lib/wizard/step1.js
+++ b/webcompat/static/js/lib/wizard/step1.js
@@ -1,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import notify from "./notify.js";
+
+const urlField = $("#url");
+const nextStepButton = $(".next-step.step-1");
+
+//@todo remove this
+nextStepButton.removeClass("disabled");
+
+const extractPrettyUrl = url => {
+  const pathArray = $.trim(url).split("/");
+  return pathArray[2];
+};
+
+const checkUrl = () => nextStepButton.removeClass("disabled");
+const showStep = (id, data) => notify.publish("showStep", { id, data });
+
+const makeAStep = event => {
+  event.preventDefault();
+  showStep(1, { url: extractPrettyUrl(urlField.val()) });
+};
+
+urlField.on("blur input", checkUrl);
+nextStepButton.on("click", makeAStep);
+
+export default {};
--- a/webcompat/static/js/lib/wizard/step2.js
+++ b/webcompat/static/js/lib/wizard/step2.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* Allows the user to select a category of the problem they're experiencing */
+
 import utils from "./utils.js";
 import notify from "./notify.js";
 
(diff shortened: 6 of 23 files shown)


.eslintrc.json
```diff
@@ -46,6 +48,7 @@
     "no-prototype-builtins": "off",
     "no-use-before-define": 2,
     "one-var": [2, "never"],
-    "prettier/prettier": "error"
+    "prettier/prettier": "error",
+    "no-unused-vars": "off"
```
ksy36: Turned this off temporarily as
`/* exported variableName */ ` doesn't work when `parserOptions.sourceType` is "module", so eslint was complaining about unused variables 

https://eslint.org/docs/rules/no-unused-vars#exported

miketaylr:
(There seems to be some bustage with the Wizard tests.)

package.json
```diff
@@ -80,6 +84,7 @@
     "project-update": "pip install --upgrade pip && npm update",
     "test": "npm run test:js && npm run test:python",
     "test:js": "node ./tests/functional/_intern.js",
+    "test:unit-js": "intern config=./webcompat/static/js/lib/wizard/tests/intern.json",
```
miketaylr: We have another intern config file (for the functional tests) at `tests/functional/_intern.js` -- but this one is relative to the wizard lib sources. Is that a convention that folks follow, or should we consider moving those tests to the top level tests directory? I defer to you!

(to be clear, i'm not suggesting we try to combine these intern configs)

webcompat/static/css/src/issue-wizard.css
```diff
@@ -225,7 +225,7 @@
   opacity: 0;
 }
 
-.step-container.step10 {
+.step-container.step9 {
```
miketaylr: This reminds me of https://github.com/webcompat/webcompat.com/issues/3106 -- should it be closed?

webcompat/static/js/lib/wizard/notify.js
```diff
@@ -0,0 +1,34 @@
...
+        delete channels[channel][index];
+      }
+    };
+  },
+  publish: function(channel, data) {
+    if (!channels.hasOwnProperty(channel) || !channels[channel].length) {
+      return;
+    }
+
+    channels[channel].forEach(cb => cb(data || {}));
```
miketaylr: Excited that we can use modern JS once this lands!

webcompat/static/js/lib/wizard/step1.js
```diff
@@ -0,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import notify from "./notify.js";
+
+const urlField = $("#url");
+const nextStepButton = $(".next-step.step-1");
+
+//@todo remove this
```
miketaylr: (Why do you need to remove this, and the other `@todo`s?)

webcompat/static/js/lib/wizard/step1.js
```diff
@@ -0,0 +1,29 @@
...
+
+const makeAStep = event => {
+  event.preventDefault();
+  showStep(1, { url: extractPrettyUrl(urlField.val()) });
+};
+
+urlField.on("blur input", checkUrl);
+nextStepButton.on("click", makeAStep);
+
+export default {};
```
miketaylr: What's the point of exporting an empty object?

(asking because i don't know 😓 )

webcompat/static/js/lib/wizard/step1.js
```diff
@@ -0,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
miketaylr: So, maybe an annoying request... but could we write a small comment block at the top of each step module that explains what it does?

```
/* step1.js - allows the user to enter a URl and navigate to the next step */

(or something less crappy, just something to give some context as we revisit these in the future)

File: webcompat/static/js/lib/wizard/tests/notify.test.js [hunk 8]
```diff
@@ -0,0 +1,69 @@
+import notify from "../notify.js";
```
miketaylr: 🎉 nice work on these tests!

webcompat/static/js/lib/wizard/utils.js
```diff
@@ -0,0 +1,32 @@
...
+    setTimeout(() => {
+      const topOfElement =
+        window.pageYOffset +
+        element.getBoundingClientRect().top -
+        HEADER_HEIGHT;
+      window.scroll({ top: topOfElement, behavior: "smooth" });
+    }, delay);
+
+    //@todo figure out why this is needed
+    //this.isSubproblem ? 450 : 250
```
miketaylr: (since it's commented out, is it really needed 🤔)

webcompat/static/js/lib/wizard/utils.js
```diff
@@ -0,0 +1,32 @@
...
+
+const utils = {
+  scrollToElement: (element, delay = 250) => {
+    // Delay "scroll to element" effect in order to let the animation finish, otherwise the scroll point isn't correct
+    setTimeout(() => {
+      const topOfElement =
+        window.pageYOffset +
+        element.getBoundingClientRect().top -
+        HEADER_HEIGHT;
+      window.scroll({ top: topOfElement, behavior: "smooth" });
```
miketaylr: Do we know what happens in Safari? It looks like maybe `ScrollToOptions` is not supported?

https://developer.mozilla.org/en-US/docs/Web/API/Window/scroll

webcompat/templates/new-issue.html
```diff
@@ -25,7 +25,7 @@
 
 {%- block extrascripts -%}
 {% if ab_active('exp') == 'form-v2' %}
-<script src="{{ url_for('static', filename='js/lib/issue-wizard-bugform.js') }}"></script>
+<script type="module" src="{{ url_for('static', filename='js/lib/wizard/app.js') }}"></script>
```
miketaylr: Do we have a bug filed for the `nomodule` babel bundle version?

(This makes me think we should possibly be merging into a branch until that's ready (rather than master))

miketaylr:
> It's possible to navigate through steps without any additional functionality (validation, image uploading, submitting the form, etc.), as this will be addressed in separate issues.

Can you clarify what this means? Does this mean if we land as-is, master is "broken" (or the additional functionality no longer works)? If so, should we land this in a feature branch?

ksy36:
> > It's possible to navigate through steps without any additional functionality (validation, image uploading, submitting the form, etc.), as this will be addressed in separate issues.
> 
> Can you clarify what this means? Does this mean if we land as-is, master is "broken" (or the additional functionality no longer works)? If so, should we land this in a feature branch?

I should have probably mentioned that there is a project I've created https://github.com/webcompat/webcompat.com/projects/9. I've divided the work into smaller tasks, so it's easier to iterate and create code reviews. 

And yes this PR is meant to land in `formv2-refactor` branch. The next issue to work on the url validation, and so on. I'm planning on moving of most the existing code while trying to improve it and adding tests for each step where possible.  


> > (There seems to be some bustage with the Wizard tests.)
> 
> @ksy36 is this expected?

So yeah the tests will fail until this is done

webcompat/static/css/src/issue-wizard.css
ksy36: yeah, I'll close it. It's actually in the project :)

webcompat/static/js/lib/wizard/utils.js
ksy36: I've moved this from existing script and decided to leave it since not quite sure why it was used, planning on revisiting this..

webcompat/templates/new-issue.html
ksy36: I haven't filed it yet, was planning on adding it and build related issues once I sort out the functionality

package.json
ksy36: yeah, I wasn't sure about this :) but find it more convenient to have unit tests where the code is and tend to see that in other projects as well

miketaylr:
> And yes this PR is meant to land in `formv2-refactor` branch.

Great! Apologies for probably reading that incorrectly (and thanks for the patience with a confused manager). >_<

webcompat/static/js/lib/wizard/utils.js
ksy36: It's jumpy :| This was also moved from the existing script. I can add a separate issue for that

webcompat/static/js/lib/wizard/step1.js
ksy36: So I haven't moved any logic for disabling/enabling/validation yet and I'll revisit this in the next issues. Right now it's enabling the next button for easier steps navigation testing

webcompat/static/js/lib/wizard/step1.js
ksy36: This is a good point, I've importing it here, https://github.com/webcompat/webcompat.com/blob/1c9516ca8bfcaf1f93b774de56939a9ae5b02234/webcompat/static/js/lib/wizard/steps.js#L1  but not really using it since this step doesn't have any public methods (visible by default and not intended to be hidden). 

Without exporting anything, there is an error `SyntaxError: import not found: default`. 
I can probably change this and not import step1 since it's not interacting with the stepper

webcompat/static/js/lib/wizard/step1.js
ksy36: Yeah, good point :)

package.json
miketaylr: Cool, let's keep it for now -- easy enough to move later on if we feel strongly about it. At some point we'll want to update the testing docs so people can know where it is, etc.

webcompat/static/js/lib/wizard/step1.js
miketaylr: Got it!

webcompat/static/js/lib/wizard/step1.js
miketaylr: Cool thanks. Since this is a WIP branch, no need to block on this right now.

webcompat/static/js/lib/wizard/utils.js
miketaylr: Cool, thanks. We're bug-compatible! 😄

webcompat/static/js/lib/wizard/step1.js
```diff
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* Allows the user to enter a URl and navigate to the next step */
```
miketaylr: minor nit:

```suggestion
/* Allows the user to enter a URL and navigate to the next step */
```

ksy36:
Thanks 👍

webcompat/static/js/lib/wizard/steps.js
```diff
@@ -0,0 +1,27 @@
+import step1 from "./step1.js";
+import step2 from "./step2.js";
+import step3 from "./step3.js";
+import step4 from "./step4.js";
+import step5 from "./step5.js";
+import step6 from "./step6.js";
+import step7 from "./step7.js";
+import step8 from "./step8.js";
+import step9 from "./step9.js";
+import step10 from "./step10.js";
```
karlcow: Do we run the risk here of having to rename everything the day we need to remove, add steps or change the order? In historical discussions, we would need to do the mental translation so `step1-2020 is our step3-2021`. z

maybe we could find a system which based on names of what the step is doing instead of a number. If we we want to highlight the continuity in between the module, we could maybe mention in a comment the step orders.

For example step1 seems to be about url input, so maybe we should call it instead 

```js
import step_url from "./step_url.js" 
```

or what ever naming convention is the good way in JavaScript.

webcompat/static/js/lib/wizard/step2.js
```diff
@@ -0,0 +1,72 @@
...
+
+const container = $(".step-container.step2");
+const detectionBug = $("#problem_category-0");
+const siteBug = $("#problem_category-1");
+const layoutBug = $("#problem_category-2");
+const videoBug = $("#problem_category-3");
+const unknownBug = $("#problem_category-4");
+const sitePrettyUrl = $("#website-url");
+const otherProblem = $(".other-problem");
+const nextStepButton = $(".next-step.step-2");
```
karlcow: There's a similar comment here about numbering here in the html, and we see it that we conveniently translate it in JS as what is it really.

`layoutBug` is more telling than `#problem_category-2`. :) HTML may need a bit of rewriting.

webcompat/static/js/lib/wizard/step7.js
```diff
@@ -0,0 +1,28 @@
...
+import notify from "./notify.js";
+
+const container = $(".step-container.step7");
+const nextStepButton = container.find(".next-step.step-7");
+
+const handleNext = event => {
+  event.preventDefault();
+  notify.publish("showStep", { id: 8 });
+};
+
```
karlcow: Yes here i wonder what is step id: 8

webcompat/static/js/lib/wizard/steps.js
ksy36: Yeah, good point :) I've noticed sometimes when I open a file, can't really tell which step this is until I find it in html 😳  I'll rename them while working on the next issue

PR #3266, opened 2020-03-31
Fixes #3231 - Add validation for step 5

### LINKED ISSUE #3231: Add validation for step 5
Add validation for step 5 and enable "Confirm" if both fields are not empty
Screen Shot 2020-03-06 at 10 27 55 AM

See also: https://github.com/webcompat/webcompat.com/issues/3042

ksy36:
r? @miketaylr

### DIFF
--- a/webcompat/templates/issue-wizard-form.html
+++ b/webcompat/templates/issue-wizard-form.html
@@ -165,12 +165,10 @@
           <div class="row mobile-col">
             <div class="form-element js-Form-group low with-validation-icons half">
               {{ form.browser.label(class_='form-label') }}
-              <small class="os-browser-error">Browser required</small>
               {{ form.browser(class_='form-field text-field') }}
             </div>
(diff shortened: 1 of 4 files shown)


webcompat/templates/issue-wizard-form.html
```diff
@@ -165,12 +165,10 @@ <h3 class="page-heading">Filing a web compatibility issue</h3>
           <div class="row mobile-col">
             <div class="form-element js-Form-group low with-validation-icons half">
               {{ form.browser.label(class_='form-label') }}
-              <small class="os-browser-error">Browser required</small>
```
miketaylr: Do we want to remove the corresponding styles for `.os-browser-error`? 

https://github.com/webcompat/webcompat.com/search?l=CSS&q=os-browser-error

ksy36: Ah right, thanks! I've removed them

Issue #3113, opened 2019-12-18
improve config/environment.py - avoid import *


karlcow:
`import *` is usually not a recommended practice. 
This is not an urgent issue.

In https://github.com/webcompat/webcompat.com/blob/daaaa99fa93f944f39286391b0613526f660e2d2/config/__init__.py
Replace

```python
from config.environment import *  # noqa
```

by 

```python
import config.environment as cfg
```

And fixes the references for the used variables.

for example

```python
if not PRODUCTION:
```

would become

```python
if not cfg.PRODUCTION
```

apsychogirl:
Hi, I would like to take this up.

karlcow:
@apsychogirl would you like to tackle this one? now that you finished #3199

karlcow:
See the comment in https://github.com/webcompat/webcompat.com/pull/3273#issuecomment-628346573
and https://github.com/webcompat/webcompat.com/pull/3273/files#r424834237

> yes the change is a bit bigger in architecture than I was planning for at the beginning of this issue and probably harder to achieve. The project is suffering from his past.
> 
> Probably we should drop this issue #3113 and we could create a new issue with a scarier title 👺 like "Webcompat app configuration architecture overhaul".

karlcow:
I will close this issue and eventually we could create a new one.

Some ideas to explore.

* [How and why have a properly “Configuration Handling File” using Flask?](https://itnext.io/how-and-why-have-a-properly-configuration-handling-file-using-flask-1fd925c88f4c)
* [Configuration Handling](https://flask.palletsprojects.com/en/1.1.x/config/)
* [Flask configuration files | Learning Flask Ep. 12](https://pythonise.com/series/learning-flask/flask-configuration-files)

We would probably benefit of the pattern:


```python
class Config(object):
    DEBUG = False
    TESTING = False

class ProductionConfig(Config):
    pass

class DevelopmentConfig(Config):
    DEBUG = True

class TestingConfig(Config):
    TESTING = True
```

PR #3238, opened 2020-03-09
Issue #3113 - Improved config/environment.py - by avoiding import *

apsychogirl:
Fixes - #3113

PR #3273, opened 2020-04-04
Issue #3113 - Improved config/environment.py - by avoiding import *

apsychogirl:
Fixes - #3113

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -17,7 +17,7 @@ import urllib.parse
 
 import requests
 
-from config.environment import *  # noqa
+import config.environment as cfg
 
 MILESTONE_ERROR = """It failed with {msg}!
 We will read from data/milestones.json.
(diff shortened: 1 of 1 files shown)


miketaylr:
Linking back to https://github.com/webcompat/webcompat.com/pull/3238 to have a history trail.

@apsychogirl I saw you came into Matrix for help with tests -- sorry for missing you there. If you want to email miket@mozilla.com we can try to debug that way.

config/__init__.py
```diff
@@ -17,7 +17,7 @@
 
 import requests
 
-from config.environment import *  # noqa
+import config.environment as cfg
```
miketaylr, line 20: With this change, I think the errors we're seeing are because now we're not importing all the values from environment.py into the app's config (which happens via `app.config.from_object('config')`). 

I'm not sure how we would pull this off without redefining all the environment vars here...

`BASE_DIR = cfg.BASE_DIR` 🤔 

I'm not sure we actually want that. I'm curious if @karlcow has some other suggestions (though he is out sick for a few weeks).

miketaylr: One possibility would be to create a package for environment and split out 3 modules for staging, production and localhost, https://docs.python.org/3/tutorial/modules.html#importing-from-a-package, defining them inside `__all__`.

We would also want to move the other vars from environment.py to inside config/__init__.py (it seems like there's a few things in there that aren't really environment related).

But that seems like a lot of work for little benefit.

karlcow:
@apsychogirl do you want to finish this or do you want me to take it?

apsychogirl:
> @apsychogirl do you want to finish this or do you want me to take it?

I would definitely like to complete this, just needed a few pointers to proceed further.

karlcow:
> > @apsychogirl do you want to finish this or do you want me to take it?
> 
> I would definitely like to complete this, just needed a few pointers to proceed further.

ok I will give you today ropes that will help you to make progress. :)

config/__init__.py
karlcow: yes the change is a bit bigger in architecture than I was planning for at the beginning of this issue and probably harder to achieve. The project is suffering from his past. 

Probably we should drop this issue #3113 and we could create a new issue with a scarier title 👺 like "Webcompat app configuration architecture overhaul".

karlcow:
@apsychogirl see my comment in https://github.com/webcompat/webcompat.com/pull/3273/files#r424834237

I will close the PR and the issue. :) Sorry about that and thanks for the contribution, it helped figure out the required changes.

PR #3280, opened 2020-04-08
Fixes #2198 - Pass in targetOrigin from intern config for tests

### LINKED ISSUE #2198: Remove hardcoded instances of "http://localhost:5000/" from functional tests
They should instead pull that in from intern `siteRoot` config property.

miketaylr:
(Rather than hard-code)

r? @magsout

### DIFF
--- a/tests/functional/image-uploads-non-auth.js
+++ b/tests/functional/image-uploads-non-auth.js
@@ -7,11 +7,9 @@ const intern = require("intern").default;
 const { assert } = intern.getPlugin("chai");
 const { registerSuite } = intern.getInterface("object");
 const FunctionalHelpers = require("./lib/helpers.js");
-const url = intern.config.functionalBaseUrl + "issues/new";
-const targetOrigin = intern.config.functionalBaseUrl.slice(
-  0,
-  intern.config.functionalBaseUrl.length - 1
-);
+const baseUrl = intern.config.functionalBaseUrl;
+const url = baseUrl + "issues/new";
+const targetOrigin = baseUrl.slice(0, baseUrl.length - 1);
 
 // This string is executed by calls to `execute()` in various tests
 // it postMessages a small green test square.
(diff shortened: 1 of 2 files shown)


tests/functional/image-uploads-non-auth.js
```diff
@@ -8,6 +8,10 @@ const { assert } = intern.getPlugin("chai");
 const { registerSuite } = intern.getInterface("object");
 const FunctionalHelpers = require("./lib/helpers.js");
 const url = intern.config.functionalBaseUrl + "issues/new";
+const targetOrigin = intern.config.functionalBaseUrl.slice(
```
miketaylr, line 11: this is super verbose!

magsout: like verbose const ;)

tests/functional/image-uploads-non-auth.js
```diff
@@ -7,7 +7,9 @@ const intern = require("intern").default;
 const { assert } = intern.getPlugin("chai");
 const { registerSuite } = intern.getInterface("object");
 const FunctionalHelpers = require("./lib/helpers.js");
-const url = intern.config.functionalBaseUrl + "issues/new";
+const baseUrl = intern.config.functionalBaseUrl;
+const url = baseUrl + "issues/new";
+const targetOrigin = baseUrl.slice(0, baseUrl.length - 1);
```
magsout, line 12: hum, why? to remove the last 0 or maybe an `/` ?

```
"http://localhost:5000".slice(0, "http://localhost:5000".length -1)

-> "http://localhost:500"
```

magsout: `"functionalBaseUrl": "http://localhost:5000",` : https://github.com/webcompat/webcompat.com/blob/master/intern.json#L5

tests/functional/image-uploads-non-auth.js
miketaylr: heh

tests/functional/image-uploads-non-auth.js
miketaylr: yeah, so this is kind of weird, but when you access the value intern adds a terminating `/` -> `http://localhost:5000/`, and postMessage won't match on that. 

(maybe i should file a bug with intern...)

PR #2395, opened 2018-04-16
Fixes #2042: Send details params content via a hidden input.

miketaylr:
The user won't see the pref values until the report has been created. This should be less confusing.

screen shot 2018-04-16 at 1 37 53 pm

PR #3290, opened 2020-04-16
Fixes #3289 - Add support for nested ul in browser details

### LINKED ISSUE #3289: Display GPU browser info as nested UL
Right now it will look like this:

`GPUs: [{'active': True, 'description': '', 'deviceID': '0x67ef', 'vendorID': '0x1002', 'driverVersion': ''}]`

Not the end of the world, but we could make this a nested UL to be more human friendly.

miketaylr:
(This will be used by the GPU data, if it exists)

I feel like Karl will have some opinions on this one, and it's cosmetic, so it can wait for him to come back.

### DIFF
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -85,10 +85,16 @@ def md5_checksum(file_path):
 
 
 def get_str_value(val):
-    """Map values from JSON to python."""
+    """Map values from JSON to a more human readable format."""
     details_map = {False: 'false', True: 'true', None: 'null'}
     if isinstance(val, (bool, type(None))):
         return details_map[val]
+    if isinstance(val, list):
+        # if val is a list, we expect there to be a single object
+        # inside, so we build a nested <ul> from that object's data
+        rv = ''.join(['<li>{k}: {v}</li>'.format(k=k, v=get_str_value(v))
+                     for k, v in list(val[0].items())])
+        return '<ul>{rv}</ul>'.format(rv=rv)
     if isinstance(val, str):
         return val
     return str(val)
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -84,17 +84,42 @@ def md5_checksum(file_path):
         return m.hexdigest()
 
 
-def get_str_value(val):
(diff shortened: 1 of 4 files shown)


webcompat/helpers.py
```diff
@@ -85,10 +85,16 @@ def md5_checksum(file_path):
 
 
 def get_str_value(val):
```
karlcow, line 87: This was created by #2395. And probably the name was not great at the start, but now it seems to have drifted a bit. We should probably change that to reflect the new semantics

And this is used by 
https://github.com/webcompat/webcompat.com/blob/9187af92432fc4046c9dcca39be5b6f009fabaea/webcompat/form.py#L296-L309

karlcow: What I'm not sure is that are we trying to do twice the same thing at two different places. It looks very similar and confusing. 

Maybe our tests were not complete at the start.

`get_details` is being tested here

https://github.com/webcompat/webcompat.com/blob/9187af92432fc4046c9dcca39be5b6f009fabaea/tests/unit/test_form.py#L220-L227

miketaylr: Yeah, it is a bit confusing. We're essentially going one more level deep so we handle a value that's more complex than a simple string.

miketaylr:
OK, @karlcow. I refactored / renamed things a bit -- hopefully less confusing.

r? @karlcow

webcompat/helpers.py
```diff
@@ -84,11 +84,42 @@ def md5_checksum(file_path):
         return m.hexdigest()
 
 
-def get_str_value(val):
-    """Map values from JSON to python."""
+def get_list_items(val_dict):
+    """Return list items (<li>) from the passed in list ([])."""
+    rv = ''.join(['<li>{k}: {v}</li>'.format(k=k, v=get_serialized_value(v))
+                 for k, v in list(val_dict.items())])
```
karlcow, line 90: The code was inherited from probably the conversion from python 2 to python 3. IN this specific case `val_dict.items()` is a `<class 'dict_items'>` and is an iterable, which makes the `list()` wrapper unecessary

PR #3300, opened 2020-05-08
Fixes #3258 - Change steps class names from numerical values to their actual names

### LINKED ISSUE #3258: Change steps class names from numerical values to their actual names
We need to change steps class names from numerical names to their actual names. For example here `step1` should be converted to `step-url`
https://github.com/webcompat/webcompat.com/blob/cc9e12f89424c025584db42f85015fcc6f9d63d4/webcompat/templates/issue-wizard-form.html#L13-L14

https://github.com/webcompat/webcompat.com/blob/cc9e12f89424c025584db42f85015fcc6f9d63d4/webcompat/static/js/lib/wizard/steps/index.js#L12-L23

ksy36:
A bit of renaming to the class names 

r? @miketaylr

### DIFF
--- a/tests/functional/reporting-auth.js
+++ b/tests/functional/reporting-auth.js
@@ -29,7 +29,7 @@ registerSuite("Reporting (auth)", {
           .type("http://example.com")
           .end()
           // Click on "Confirm"
-          .findByCssSelector(".button.step-1")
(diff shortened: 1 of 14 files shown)


ksy36:
thanks!

PR #3301, opened 2020-05-10
Fixes #3297 - Adjust changelog for dependabot

### LINKED ISSUE #3297: changelog.py script unable to handle commit messages from @dependabot
```
  File "tools/changelog.py", line 139, in <module>
    sys.exit(main())
  File "tools/changelog.py", line 119, in main
    loglines = create_changelog(json_response)
  File "tools/changelog.py", line 59, in create_changelog
    title = normalize_title(issue['title'])
  File "tools/changelog.py", line 87, in normalize_title
    title = log_line.format(msg=m.groupdict())
AttributeError: 'NoneType' object has no attribute 'groupdict'
```

Probably from https://github.com/webcompat/webcompat.com/pull/3296 (we switched to a new dependency bot thingy):

`Bump grunt from 1.0.4 to 1.1.0`

apsychogirl:
Fixes #3297 
>Modified changelog.py script to handle commit messages from @dependabot

### DIFF
--- a/tests/unit/test_tools_changelog.py
+++ b/tests/unit/test_tools_changelog.py
@@ -55,9 +55,16 @@ class TestChangelog(unittest.TestCase):
             'Fixes #101 — Everything you can imagine is real.')
         expected = 'Fixes #101 - Everything you can imagine is real.'
         self.assertEqual(actual, expected)
-        # GreenKeeper title style
-        actual = changelog.normalize_title('Greenkeeper style 🚀')
-        expected = 'NPM update - Greenkeeper style.'
+        
+        # # GreenKeeper title style
+        # actual = changelog.normalize_title('Greenkeeper style 🚀')
+        # expected = 'NPM update - Greenkeeper style.'
+        # self.assertEqual(actual, expected)
+
+        # Dependabot-review title style 
+        actual = changelog.normalize_title(
+            'Bump grunt from 1.0.4 to 1.1.0') 
+        expected = 'NPM update - grunt from 1.0.4 to 1.1.0.'
         self.assertEqual(actual, expected)
 
 
--- a/tests/unit/test_tools_changelog.py
+++ b/tests/unit/test_tools_changelog.py
@@ -55,14 +55,9 @@ class TestChangelog(unittest.TestCase):
             'Fixes #101 — Everything you can imagine is real.')
         expected = 'Fixes #101 - Everything you can imagine is real.'
         self.assertEqual(actual, expected)
-        # # GreenKeeper title style
-        # actual = changelog.normalize_title('Greenkeeper style 🚀')
-        # expected = 'NPM update - Greenkeeper style.'
-        # self.assertEqual(actual, expected)
-
         # Dependabot-review title style
         actual = changelog.normalize_title('Bump grunt from 1.0.4 to 1.1.0')
-        expected = 'NPM update - grunt from 1.0.4 to 1.1.0.'
+        expected = 'NPM update - Upgrade grunt from 1.0.4 to 1.1.0.'
         self.assertEqual(actual, expected)
 
 
--- a/tests/unit/test_tools_changelog.py
+++ b/tests/unit/test_tools_changelog.py
@@ -55,15 +55,13 @@ class TestChangelog(unittest.TestCase):
             'Fixes #101 — Everything you can imagine is real.')
         expected = 'Fixes #101 - Everything you can imagine is real.'
         self.assertEqual(actual, expected)
-        
         # # GreenKeeper title style
         # actual = changelog.normalize_title('Greenkeeper style 🚀')
         # expected = 'NPM update - Greenkeeper style.'
         # self.assertEqual(actual, expected)
 
-        # Dependabot-review title style 
-        actual = changelog.normalize_title(
-            'Bump grunt from 1.0.4 to 1.1.0') 
+        # Dependabot-review title style
+        actual = changelog.normalize_title('Bump grunt from 1.0.4 to 1.1.0')
         expected = 'NPM update - grunt from 1.0.4 to 1.1.0.'
         self.assertEqual(actual, expected)
 
--- a/tools/changelog.py
+++ b/tools/changelog.py
@@ -76,9 +76,18 @@ def normalize_title(title):
     See the test suite for the potential matches.
     GitHub sends us a unicode string.
     """
-    if '🚀' in title:
-        title = title.replace('🚀', '')
-        title = title.strip()
+    # if '🚀' in title:
+    #     title = title.replace('🚀', '')
+    #     title = title.strip()
+    #     title = 'NPM update - {title}.'.format(title=title)
+
+    if 'Bump' in title:
+        """
+        index [0] -> Bump always for dependabot
+        index [1:] -> name of package and version info
+        """
+        title = title.split()[1:]
+        title = ' '.join(title)
         title = 'NPM update - {title}.'.format(title=title)
     else:
         regex = r"[^ ]?\#(?P<number>\d+)[^\w]+(?P<prose>.*)"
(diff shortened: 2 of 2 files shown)


tools/changelog.py
```diff
@@ -76,9 +76,18 @@ def normalize_title(title):
     See the test suite for the potential matches.
     GitHub sends us a unicode string.
     """
-    if '🚀' in title:
-        title = title.replace('🚀', '')
-        title = title.strip()
+    # if '🚀' in title:
```
miketaylr, line 79: Greenkeeper is going away next month (see https://greenkeeper.io/), so we can probably just delete this, rather than comment it out. Same for the corresponding tests.

tests/unit/test_tools_changelog.py
```diff
@@ -55,9 +55,14 @@ def test_normalize_title(self):
...
-        actual = changelog.normalize_title('Greenkeeper style 🚀')
-        expected = 'NPM update - Greenkeeper style.'
+        # # GreenKeeper title style
+        # actual = changelog.normalize_title('Greenkeeper style 🚀')
+        # expected = 'NPM update - Greenkeeper style.'
+        # self.assertEqual(actual, expected)
+
+        # Dependabot-review title style
+        actual = changelog.normalize_title('Bump grunt from 1.0.4 to 1.1.0')
+        expected = 'NPM update - grunt from 1.0.4 to 1.1.0.'
```
miketaylr, line 65: This title reads a bit odd after removing the word `Bump`, maybe expected could be: `NPM update - bump grunt from 1.0.4 to 1.1.0`?

tests/unit/test_tools_changelog.py
```diff
@@ -55,9 +55,14 @@ def test_normalize_title(self):
             'Fixes #101 — Everything you can imagine is real.')
         expected = 'Fixes #101 - Everything you can imagine is real.'
         self.assertEqual(actual, expected)
-        # GreenKeeper title style
-        actual = changelog.normalize_title('Greenkeeper style 🚀')
-        expected = 'NPM update - Greenkeeper style.'
+        # # GreenKeeper title style
+        # actual = changelog.normalize_title('Greenkeeper style 🚀')
+        # expected = 'NPM update - Greenkeeper style.'
+        # self.assertEqual(actual, expected)
```
karlcow, line 61: These will not be necessary anymore. If the code is not useful, do not hesitate to delete it.

tests/unit/test_tools_changelog.py
karlcow: 👍 agreed with @miketaylr 
Though honestly the bump is not necessary very telling too. 

I would favor a `'NPM update - upgrade grunt from 1.0.4 to 1.1.0.'`
on the other hand, the proposal of mike is requiring less work. :)

tools/changelog.py
karlcow: Yeah. Let's delete all greenkeeper related code.

tools/changelog.py
```diff
@@ -76,9 +76,18 @@ def normalize_title(title):
...
+    #     title = title.replace('🚀', '')
+    #     title = title.strip()
+    #     title = 'NPM update - {title}.'.format(title=title)
+
+    if 'Bump' in title:
```
karlcow, line 84: ok this is the core of the code. While 🚀 was probably a unique character I have the feeling it will not be the case for the `Bump` word, aka there could be something else called Bump. So probably the code needs to be a bit more specific. Maybe the code should target the beginning of the string with [startswith](https://docs.python.org/3/library/stdtypes.html#str.startswith)

tools/changelog.py
```diff
@@ -76,9 +76,18 @@ def normalize_title(title):
...
+        """
+        index [0] -> Bump always for dependabot
+        index [1:] -> name of package and version info
+        """
+        title = title.split()[1:]
+        title = ' '.join(title)
```
karlcow, line 90: so there are a couple of ways to do that. You found one. Here's another one using [partition](https://docs.python.org/3/library/stdtypes.html#str.partition)

```python
>>> # the current code
>>> title = 'Bump grunt from 1.0.4 to 1.1.0'
>>> ' '.join(title.split()[1:])
'grunt from 1.0.4 to 1.1.0'
>>> # another solution using partition
>>> title = 'Bump grunt from 1.0.4 to 1.1.0'
# partition returns a tuple with the 3 parts of the left partition
>>> title.partition(' ')
('Bump', ' ', 'grunt from 1.0.4 to 1.1.0')
>>> # so straightforward this will return what we want in one line.
>>> title.partition(' ')[2]
'grunt from 1.0.4 to 1.1.0'
```

@apsychogirl but maybe we don't need to remove the `bump` word depending on what you decide for the test in our previous comments

tests/unit/test_tools_changelog.py
apsychogirl: > +1 agreed with @miketaylr
> Though honestly the bump is not necessary very telling too.
> 
> I would favor a `'NPM update - upgrade grunt from 1.0.4 to 1.1.0.'`
> on the other hand, the proposal of mike is requiring less work. :)

Yoss! I guess upgrade is a better term to use instead of bump.

tools/changelog.py
apsychogirl: > Greenkeeper is going away next month (see https://greenkeeper.io/), so we can probably just delete this, rather than comment it out. Same for the corresponding tests.

Roger that.

tests/unit/test_tools_changelog.py
apsychogirl: Or should we follow the old trend
"Update grunt to latest version"?

karlcow: having the version numbers can help figure out the regressions.

tools/changelog.py
apsychogirl: > so there are a couple of ways to do that. You found one. Here's another one using [partition](https://docs.python.org/3/library/stdtypes.html#str.partition)
> 
> ```python
> >>> # the current code
> >>> title = 'Bump grunt from 1.0.4 to 1.1.0'
> >>> ' '.join(title.split()[1:])
> 'grunt from 1.0.4 to 1.1.0'
> >>> # another solution using partition
> >>> title = 'Bump grunt from 1.0.4 to 1.1.0'
> # partition returns a tuple with the 3 parts of the left partition
> >>> title.partition(' ')
> ('Bump', ' ', 'grunt from 1.0.4 to 1.1.0')
> >>> # so straightforward this will return what we want in one line.
> >>> title.partition(' ')[2]
> 'grunt from 1.0.4 to 1.1.0'
> ```
> 
> @apsychogirl but maybe we don't need to remove the `bump` word depending on what you decide for the test in our previous comments

How about if we follow the same trend as already used in changelog.md

`>>> title="Bump grunt from 1.0.4 to 1.1.0"
>>> title=title.partition('grunt')
>>> title = 'NPM update - Update {title} to latest version.'.format(title=title[1])
>>> print(title)
NPM update - Update grunt to latest version.`

tools/changelog.py
apsychogirl: > ok this is the core of the code. While rocket was probably a unique character I have the feeling it will not be the case for the `Bump` word, aka there could be something else called Bump. So probably the code needs to be a bit more specific. Maybe the code should target the beginning of the string with [startswith](https://docs.python.org/3/library/stdtypes.html#str.startswith)

True, I will change that.

tools/changelog.py
karlcow: 1. note that this example used grunt, but it could be anything, next time it will not be grunt. :) `.partition('grunt')` will not work. 
2. If the keyword is already known and extracted, then `partition` is not needed at all. :)

karlcow:
Thanks a lot @apsychogirl 
I rebased a bit.

karlcow:
Closed by https://github.com/webcompat/webcompat.com/commit/ad693004dcc017fc56b7a2e85f287948db0075a2

miketaylr:
Awesome, thanks!

PR #3307, opened 2020-05-14
Issue #3038 - Correct focus for text fields

### LINKED ISSUE #3038: Text fields do not have correct focus and valid styling [form v2]
The styling demo presented focus and validated styling for text fields:

Screenshot 2019-10-25 at 11 27 37

Untitled5

The correct styles are the following:

**Valid**
`border: 1px solid #22b573`

**Focus**
`box-shadow: 0 0 0 3px  #56638c`

soniasingla:
Fixes #3038 

Correct focus and styling for text-fields

### DIFF
--- a/webcompat/static/css/src/issue-wizard.css
+++ b/webcompat/static/css/src/issue-wizard.css
@@ -446,6 +446,7 @@
   font-size: 14px;
   font-weight: 200;
   padding: 4px 32px 5px 8px;
+  transition: box-shadow .3s;
 }
 
 .with-validation-icons.is-validated::after,
@@ -652,6 +653,14 @@
   height: 28px;
   right: -11px;
   width: 28px;
+}
+
+.issue-form .is-validated .text-field {
+  border-color: var(--issue-selected-outline);
+}
+
+.issue-form .text-field:focus {
+  box-shadow: 0 0 0 3px var(--issue-type-bg);
 }
 
 #browsers li {


ksy36:
Hi @soniasingla , thanks for looking into this! We also need to add styles for the valid state of the input as part of the https://github.com/webcompat/webcompat.com/issues/3038  (green border).

Also we need a transition effect for the focus box-shadow as in the demo https://cdpn.io/Octavector/debug/5c04c472d928b6a34521febf2789a6d2 (you can just use the same styles as in the demo).

webcompat/static/css/src/issue-wizard.css
```diff
@@ -1041,6 +1047,10 @@
     border-color: var(--wizard-step-error);
   }
 
+  .issue-form .text-field:focus {
+    box-shadow: 0 0 0 3px #56638c;
+  }
+
```
ksy36, line 1053: Since this rule located under `@media (min-width: 665px)`, it will currently apply only to desktop, so it should be moved up (similarly to where `.issue-form .is-validated .text-field` is ).  

Also probably better to use `var(--issue-type-bg)` as a box-shadow color

webcompat/static/css/src/issue-wizard.css
```diff
@@ -1027,6 +1032,7 @@
     font-size: 90%;
     font-weight: 600;
     padding: 11px 36px 11px 8px;
+    transition: box-shadow .3s;
```
ksy36, line 1035: This should be moved out of `@media (min-width: 665px)` as well in order to work for both mobile and desktop
```css
.issue-form .text-field {
    transition: box-shadow .3s;
}
```

webcompat/static/css/src/issue-wizard.css
```diff
@@ -654,6 +655,10 @@
   width: 28px;
 }
 
+.issue-form .is-validated .text-field {
+  border: 1px solid var(--issue-selected-outline);
+}
```
ksy36, line 660: I think here we can only change border-color (since the border is already 1px solid )

webcompat/static/css/src/issue-wizard.css
```diff
@@ -496,6 +496,7 @@
   font-weight: 200;
   margin-bottom: 17px;
   padding: 5px 32px 7px 8px;
+  transition: box-shadow .3s;
```
ksy36, line 499: This transition on `.low .text-field` is not needed I think, as long as it's specified in `.issue-form .text-field`

PR #3294, opened 2020-04-22
Fixes #3281 - Convert issue page to be server rendered (part 1)

miketaylr:
Right now I'm going to leave this as a draft. I decided to not tackle existing comments here just to keep things small. I'll file a follow-up on that and tackle it once this lands.

PR #3308, opened 2020-05-15
Fixes #3295 - Fetch and render existing issue comments as HTML

### LINKED ISSUE #3295: Convert existing comments to be statically rendered.
Follow up to https://github.com/webcompat/webcompat.com/issues/3281

This may or may not be desirable, since it will require us to block on another request. Perhaps dynamic is the way to go here. 🤔 

But if we want to create an archive, we will at least need a method that allows for static comment rendering, so it could be optional.

miketaylr:
TODO:

* [x] write some tests for the modified behavior of `get_response_headers`
* [x] remove the entire client-side part of adding a new comment, and handle the XHR bits related to that
* [x] add more tests
* [x] fix the tests i broke
* [x] fix the NSFW handling stuff that's broken now
* [x] fix busted unit tests
* [x] add some kind of ajax spinner when submitting a comment
* [x] remove client-side markdown stuff


webcompat/static/js/lib/issues.js
```diff
@@ -367,11 +367,18 @@ issues.MainView = Backbone.View.extend(
...
+            {
+              type: "GET",
+              dataType: "html",
+            }
+          )
           .done(
-            _.bind(function (response) {
-              this.addExistingComments();
+            _.bind(function (response, textStatus, jqXHR) {
+              console.log(response, textStatus, jqXHR);
```
miketaylr, line 380: oops

miketaylr:
So I decided to just do away with the CommentsView and Collection and Model, etc. Instead we'll go old-school, and after doing an XHR post for new comments, we just append the HTML the server sends back to us.

The cool thing about this is it means we no longer need the Markdown client stuff, so we can delete a ton of code here.

webcompat/static/js/lib/issues.js
```diff
@@ -395,29 +400,9 @@ issues.MainView = Backbone.View.extend(
       }
     },
 
-    getRemainingComments: function (count) {
```
miketaylr, line 398: I should probably leave this in here, but it will need to be re-written (because `this.comments.fetchPage` no longer exists)... not sure.

webcompat/templates/issue/issue-comment-list.html
```diff
@@ -0,0 +1,23 @@
+{#  Note: this needs to be kept in sync with issue-comment-list.jst #}
```
miketaylr, line 1: This is no longer true.

webcompat/api/helpers.py
```diff
@@ -0,0 +1,36 @@
...
+
+from webcompat.helpers import get_response_headers
+
+HTML_MIME = 'text/html'
+
+
+def get_html_comments(response):
+    """Helper method to create a response from our comment template.
+
+    The template exects a list of dict objects, so create a list
```
miketaylr, line 22: nit: s/exects/expects/

miketaylr:
OK, I think this is ready for review now.

r? @karlcow 
r? @ksy36

miketaylr:
🥳 Screen Shot 2020-05-18 at 6 03 36 PM

webcompat/api/endpoints.py
```diff
@@ -170,22 +169,32 @@ def get_search_results(query_string=None, params=None):
...
-                           mime_type=JSON_MIME_HTML)
+        new_comment = api_request('post', path, params=params,
+                                  data=get_comment_data(request.data),
+                                  mime_type=JSON_MIME_HTML)
+        return get_html_comments(new_comment)
     else:
-        path = 'repos/{0}/{1}/comments'.format(ISSUES_PATH, number)
-        return api_request('get', path, params=params,
-                           mime_type=JSON_MIME_HTML)
+        # TODO: handle the (rare) case for more than 1 page of comments
```
miketaylr, line 186: I will file a bug for this, but I think we're fine to deploy without it (since the most comments we have on an issue is 41 right now, well below the 100 this supports).

webcompat/templates/issue/issue-comment-list.html
```diff
@@ -0,0 +1,24 @@
+{% autoescape false %}{% for response in comments %}
```
miketaylr, line 1: I'm not sure we want to disable autoescaping... I don't recall why I added this, but removing it doesn't do anything strange AFAICT.

webcompat/static/js/lib/issues.js
```diff
@@ -269,215 +159,211 @@ issues.ImageUploadView = Backbone.View.extend({
...
+              _.bind(function (response) {
+                $(".js-Issue-commentList").html(response);
+                this.onAfterCommentsRendered();
+                // If there's a #hash pointing to a comment (or elsewhere)
+                // scrollTo it.
+                if (location.hash !== "") {
+                  var _id = $(location.hash);
+                  window.scrollTo(0, _id.offset().top);
+                }
+                if (response[0].lastPageNumber > 1) {
```
miketaylr, line 279: mike: remove this.

tests/functional/comments-auth.js
```diff
@@ -41,35 +41,6 @@ registerSuite("Comments (auth)", {
...
 
-    "Posting a comment"() {
-      var originalCommentsLength;
-      var allCommentsLength;
-      return (
-        FunctionalHelpers.openPage(this, url("issues/100"), ".js-image-upload")
-          .findAllByCssSelector(".js-Issue-comment")
-          .then((elms) => {
-            originalCommentsLength = elms.length;
-          })
```
ksy36, line 52: not very familiar with this part of the app, but wonder why this test is no longer needed?

miketaylr: Good question! It's probably still needed, but the way this passed before was that it never really posted a new comment, but it added a new comment to the DOM. So we were just testing that that succeeded. In this current PR, new comments are only added to the UI after we get a 201 back from GitHub. It might be possible to mock this out... I'll have to think (and maybe file a follow-up bug).

miketaylr: Actually, we don't need to do much to get this working. I'll push another commit. 😄

tests/unit/test_issues.py
```diff
@@ -35,47 +36,50 @@ def tearDown(self):
...
-            ('description', 'sup'),
-            ('details', ''),
-            ('os', 'Mac OS X 13'),
-            ('problem_category', 'unknown_bug'),
-            ('submit_type', 'github-proxy-report'),
-            ('url', 'http://3139.example.com'),
-            ('username', ''), ])
-        rv = report_issue(form, True)
-        self.assertEqual(rv.get('number'), 2)
+        with app.test_request_context():
```
karlcow, line 39: What makes the new context `app.test_request_context()` required? Just curious.

webcompat/api/endpoints.py
```diff
@@ -15,10 +15,13 @@
 from flask import abort
 from flask import Blueprint
 from flask import g
+from flask import make_response
 from flask import request
+from flask import render_template
```
karlcow, line 20: nit: 

```suggestion
from flask import render_template
from flask import request
```

webcompat/api/endpoints.py
```diff
@@ -170,22 +169,32 @@ def get_search_results(query_string=None, params=None):
...
 
 @api_bp.route('/issues/<int:number>/comments', methods=['GET', 'POST'])
-@mockable_response
 def proxy_comments(number):
-    """XHR endpoint to get issues comments from GitHub.
+    """XHR endpoint to get issues comments from GitHub, or to add
+    a new comment.
 
     Either as an authed user, or as one of our proxy bots.
     """
```
karlcow, line 177: nit: first docstring line should be only one line.

```suggestion
    """XHR endpoint for GitHub issue comments.
    
    * GET an issue comments 
    * POST a comment on an issue (only as an authorized GitHub user)
    """
```

webcompat/api/endpoints.py
```diff
@@ -170,22 +169,32 @@ def get_search_results(query_string=None, params=None):
...
     params = request.args.copy()
+    path = 'repos/{0}/{1}/comments'.format(ISSUES_PATH, number)
```
karlcow, line 179: 😱 
pro optimization 😉 Thanks!

webcompat/api/endpoints.py
karlcow: Maybe it could be a question of a load more buttons. or something.

webcompat/helpers.py
```diff
@@ -187,19 +187,24 @@ def get_os(user_agent_string=None):
...
+    # handle the case where we get the data directly
+    headers = {}
+    if isinstance(response, requests.models.Response):
+        headers = response.headers
+    # or the case where we proxy an already fetched reponse
+    if isinstance(response, tuple):
+        headers = response[2]
+    new_headers = {'etag': headers.get('etag'),
+                   'cache-control': headers.get('cache-control'),
+                   'content-type': mime_type}
```
karlcow, line 204: if by chance the two `if` fails. We end up with  `headers.get('etag')` on `{}`   aka `'etag': None`. 
Maybe it's working or does it 💣

Maybe ok? 😆 

```python
>>> Response('foo', 200, {'etag': None}).headers
Headers([('etag', 'None'), ('Content-Type', 'text/html; charset=utf-8'), ('Content-Length', '3')])
```

tests/unit/test_api_helpers.py
```diff
@@ -0,0 +1,52 @@
...
+    """Class for testing API helpers."""
+
+    def setUp(self):
+        """Set up the tests."""
+        webcompat.app.config['TESTING'] = True
+        self.app = webcompat.app.test_client()
+
+    def tearDown(self):
+        """Tear down the tests."""
+        pass
```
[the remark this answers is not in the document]
miketaylr: yeah, a sample template of what we're aiming for would be useful. i'll read the linked article!

webcompat/api/helpers.py
```diff
@@ -0,0 +1,36 @@
...
+HTML_MIME = 'text/html'
+
+
+def get_html_comments(response):
+    """Helper method to create a response from our comment template.
+
+    The template expects a list of dict objects, so create a list
+    in the instance where we have a single object.
+    """
+    comment_json, comment_status = response[0:2]
```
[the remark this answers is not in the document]
miketaylr: frig, twice in a row... https://github.com/webcompat/webcompat.com/pull/3294#discussion_r417082700

🙈

tests/unit/test_helpers.py
```diff
@@ -12,8 +12,9 @@
 import unittest
 
 import flask
-from werkzeug.http import parse_cookie
 from flask import session
+from werkzeug.http import parse_cookie
+from requests.models import Response
```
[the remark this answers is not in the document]
miketaylr: :) oops

tests/unit/test_api_helpers.py
```diff
@@ -0,0 +1,52 @@
...
+            {"user": {"login": "miketaylr","avatar_url": "https://avatars1.githubusercontent.com/u/67283?v=4"},"created_at": "2016-07-28T14:41:43Z","body_html": "<p>one</p>"},
+            {"user": {"login": "miketaylr","avatar_url": "https://avatars1.githubusercontent.com/u/67283?v=4"},"created_at": "2016-07-28T14:42:43Z","body_html": "<p>two</p>"}]""",  # noqa
+            200, {'content-type': 'application/json'})
+        with webcompat.app.app_context():
+            rv = get_html_comments(response)
+            assert b'<p>one</p>' in rv.data
+            assert b'<p>two</p>' in rv.data
+            assert rv.status_code == 200
+            assert rv.content_type == 'text/html'
+
```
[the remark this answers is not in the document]
miketaylr: 304 actually fails... 😈 

I think it's because in the normal case, GitHub sends back an empty response body (`b''`) but the browser pulls the data from its HTTP cache so the templates don't throw. Maybe the right thing to do is just read from a fixture file to mimic a read from the HTTP cache. I'll do that.

tests/unit/test_issues.py
miketaylr: So I *think* (but am not 100% certain because the universe is a big place) the reason is because I moved the `@mockable_response` decorator to `proxy_request`, which `report_public_issue`, `report_private_issue`, and `report_issue` all eventually call.

Inside of that decorator, we inspect the `request` object: https://github.com/webcompat/webcompat.com/blob/ad693004dcc017fc56b7a2e85f287948db0075a2/webcompat/helpers.py#L401-L402

And in order to look at a `request` you have to have a request context available: https://flask.palletsprojects.com/en/1.1.x/reqcontext/

tests/unit/test_api_helpers.py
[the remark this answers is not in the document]
miketaylr: OK, I re-wrote how we handle this, with a `try... except`. If we fail to parse due to 304, we should still survive.

webcompat/api/helpers.py
```diff
@@ -0,0 +1,36 @@
...
+    if comment_status != 304:
```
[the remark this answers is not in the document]
miketaylr: In theory it's possible for us to get a 401, 404, or 500 from github. In those instances, the xhr request will throw and we'll end up in the `fail` condition and send a flash message "There was an error retrieving issue comments. Please reload to try again.". 

(the only way I can see us getting a 404 is if the API somehow lied to us -- we only fetch comments after the Issues API told us there is more than 0 comments. entirely possible, but probably more rare than a 500).

webcompat/api/helpers.py
```diff
@@ -0,0 +1,36 @@
...
+        comment_json = json.loads(comment_json)
+        if isinstance(comment_json, dict):
+            comment_json = [comment_json]
+    return (
+        make_response(
+            render_template('issue/issue-comment-list.html',
+                            comments=comment_json),
+            comment_status,
+            get_response_headers(response, HTML_MIME))
+    )
```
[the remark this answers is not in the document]
miketaylr: will do, thanks

webcompat/helpers.py
miketaylr: :))~ yeah, I guess the worst case scenario here is there's no cache-hit.

(I hope)

miketaylr:
OK @karlcow, I believe I addressed everything!

Thanks to you and @ksy36 for the review -- this is in better shape as a result.

miketaylr:
yay!

@ksy36 ok, you can rebase your form-v2 branch, promise i won't touch build configs for a while ^__^

ksy36:
sounds good, thanks!

PR #3312, opened 2020-05-22
fixes #3310 - improves tests documentation

### LINKED ISSUE #3310: Create a "model" test file using pytest
@karlcow do you want to port over any unit test file from unittest/nose to pytest? It would be nice to have something to model after, and then we can update docs as well (pointing to the flask + other useful articles).

karlcow:
* Adds documentation for pytests
* Fixes outdated links to documentation
* Fixes the table of content
* Adds a template for Python tests using pytest
* Adds information about data fixtures

This PR fixes issue #3310

### DIFF
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -1,13 +1,20 @@
+
+
 * [Running Tests](#running-tests)
+  * [Backend Unit Tests](#backend-unit-tests)
   * [Functional Tests](#functional-tests)
-  * [Functional Tests using Fixture Data](#functional-tests-using-fixture-data)
 * [Writing Tests](#writing-tests)
   * [Python Unit Tests](#python-unit-tests)
   * [JS Functional Tests](#js-functional-tests)
+  * [Adding Data Fixtures](#adding-data-fixtures)
 
 ## Running Tests
 
-We assume you have already installed the development python modules with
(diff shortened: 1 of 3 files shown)


karlcow:
@miketaylr I kind of did it in the opposite way ;) I changed the documentation, but if you want I can add to this pull request at least one converted file. So we could link the diff to the documentation. WDYT?

karlcow:
Let's make it a draft to not do accidental merge.

miketaylr:
> I kind of did it in the opposite way ;)

I am shocked 😱 

> I changed the documentation, but if you want I can add to this pull request at least one converted file. So we could link the diff to the documentation. WDYT?

These are great improvements. r+++++

I think with just the skeleton example you linked, it's enough to learn from. But converting one additional file wouldn't hurt, if you're up to it. Basically, I'll leave it up to you. :)

karlcow:
@miketaylr what about two files converted. :)

That was an interesting learning process and an opportunity to tweak a couple of things for the tests.

Using some nice features of pytests here and there.

@ksy36 might be interested by the modifications I have done to the tests for console log file. 
And i'm happy to answer any questions on the why?

I think it will be fun to convert the other tests. It's not always easy because we have to change the mindset a bit. But in the end this is fun.

karlcow:
ok all green.

miketaylr:
awesome, thank you!

PR #3320, opened 2020-05-28
Issue #3317 - Add correct border for invalid form element state on Mo…

### LINKED ISSUE #3317: [New UI] Add correct border for invalid form element state works on desktop, not on Mobile
Follow up from issue #2980 

Currently, the border is applied only on Desktop, not on mobile. It will be great to have it on mobile.

cc : @ksy36 :smile:

soniasingla:
Fixes #3317 

This pull request applies the correct border for invalid form elemental state on both, Desktop and Mobile.

CC : @ksy36 :)

### DIFF
--- a/webcompat/static/css/src/issue-wizard.css
+++ b/webcompat/static/css/src/issue-wizard.css
@@ -498,6 +498,10 @@
   padding: 5px 32px 7px 8px;
 }
 
+.issue-form .is-error .text-field {
+  border-color: var(--wizard-step-error);
+}
+
 .low .text-field::placeholder {
   color: var(--input-placeholder);
 }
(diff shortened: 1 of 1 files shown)


webcompat/static/css/src/issue-wizard.css
```diff
@@ -498,6 +498,10 @@
   padding: 5px 32px 7px 8px;
 }
 
+.issue-form .is-error .text-field {
+  border-color: var(--wizard-step-error);
```
ksy36, line 502: So we also need to remove the old rule: 
https://github.com/webcompat/webcompat.com/blob/master/webcompat/static/css/src/issue-wizard.css#L1040
Since here you have it as a default style, it will apply both to mobile and desktop, so having it within `@media (min-width: 665px)` is unnecessary.

Given that this branch doesn't have the old rule, you probably need to rebase this branch with master.

soniasingla:
hey @ksy36 , ready for another review :)

Issue #3205, opened 2020-02-12
Python unittest coverage


karlcow:
I was testing the test coverage of our code for python.
To experiment.

```bash
cd webcompat.com/
. env/bin/activate
pip install coverage
```

Then for the coverage

```bash
coverage run --source webcompat,tools -m nose tests 
```

which gives us an error (this to explore) while nosestests is ok.

```
...E....................................................................................................................................................
======================================================================
ERROR: Patching the issue is working only with certain circumstances.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/mock.py", line 1101, in _dot_lookup
    return getattr(thing, comp)
AttributeError: 'Blueprint' object has no attribute 'endpoints'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/karl/code/webcompat.com/tests/unit/test_api_urls.py", line 175, in test_api_patch_issue
    with patch('webcompat.api.endpoints.proxy_request') as github_data:
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/mock.py", line 1257, in __enter__
    self.target = self.getter()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/mock.py", line 1427, in <lambda>
    getter = lambda: _importer(target)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/mock.py", line 1114, in _importer
    thing = _dot_lookup(thing, comp, import_path)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/mock.py", line 1104, in _dot_lookup
    return getattr(thing, comp)
AttributeError: 'Blueprint' object has no attribute 'endpoints'

----------------------------------------------------------------------
Ran 152 tests in 1.650s

FAILED (errors=1)
```

Here the nosetests results

```
nosetests
........................................................................................................................................................
----------------------------------------------------------------------
Ran 152 tests in 1.120s

OK
```

Once we ran coverage, we can display a report:

```bash
coverage report -m 
```

which is not too bad.

```
                                 
Name                                   Stmts   Miss  Cover   Missing
--------------------------------------------------------------------
tools/__init__.py                          0      0   100%
tools/changelog.py                        78     33    58%   50-52, 68-69, 94-134, 139
[...]

PR #3322, opened 2020-05-29
Fixes #3321 - Converts to pytest and improves coverage

### LINKED ISSUE #3321: [test] webcompat/issues.py needs more tests coverage
coverage is missing currently 

```
Name                                   Stmts   Miss  Cover   Missing
--------------------------------------------------------------------
webcompat/issues.py                       48      4    92%   99-104
```

https://github.com/webcompat/webcompat.com/blob/54a27f0a1b0968195e31d3c0c447593d5011af18/webcompat/issues.py#L99-L104

Probably two steps.

* [x] convert to pytest (see [this diff as an example](https://github.com/webcompat/webcompat.com/pull/3312/commits/d6c87dfee3cd23b004b137e1fbd4df5010ebca35) and [documentation](https://github.com/webcompat/webcompat.com/blob/master/docs/tests.md#python-unit-tests)
* [x] add the remaining missing tests scenario

the file to edit is 
https://github.com/webcompat/webcompat.com/blob/master/tests/unit/test_issues.py

karlcow:
This PR fixes issue #3321 

## Proposed PR background

This took a bit more time than I was expecting but I have figured out a couple of things. 
This also should lead to more test organization. I need to open an issue so that we put the mock_api_response outside of test_api_urls.  (probably in conftest.py if I understood well pytest) 

Anyway.

It basically 
1. Converts the test for test_issues.py to pytest
2. Adds missing tests for cases which were not covered in the tests (thanks to coverage for finding them.)
3. Modifies mock_api_response to add `.json()` method for the mocks.

Now we have 100% coverage for the issues.py module. See #3205 for the work to do still.
 
```
(env) ~/code/webcompat.com % COLUMNS=50 coverage report -m                                            
Name                                   Stmts   Miss  Cover   Missing
--------------------------------------------------------------------
…
webcompat/issues.py                       48      0   100%
…
--------------------------------------------------------------------
TOTAL                                   1897    343    82%
```

ok let's see if we stay green or if we broke something.

### DIFF
--- a/tests/unit/test_issues.py
+++ b/tests/unit/test_issues.py
@@ -39,12 +39,12 @@ FORM_AUTH['submit_type'] = 'github-auth-report'
 
 
 @pytest.fixture
-def client():
+def setup():
     """Test client for requests."""
     webcompat.app.config['TESTING'] = True
 
 
-def test_report_with_auth_on(client):
+def test_report_with_auth_on(setup):
     """Test that we can still work with github-auth-report."""
     with patch('webcompat.issues.github.post') as json_resp:
         json_resp.return_value = {'number': 2}
@@ -53,7 +53,7 @@ def test_report_with_auth_on(client):
         assert json_data.get('number') == 2
 
 
-def test_report_issue_returns_number(client):
+def test_report_issue_returns_number(setup):
     """Test issue posting report the right json."""
     with patch('webcompat.issues.proxy_request') as github_data:
         github_data.return_value = mock_api_response({
@@ -68,13 +68,13 @@ def test_report_issue_returns_number(client):
             assert json_data.get('number') == 2
 
 
-def test_report_issue_fails_400_for_unknown_type(client):
+def test_report_issue_fails_400_for_unknown_type(setup):
     """Test request fails with a 400 because type unknown."""
     with pytest.raises(BadRequest):
         rv = report_issue(FORM, proxy=True)
 
 
-def test_report_issue_fails_400_for_proxy(client):
+def test_report_issue_fails_400_for_proxy(setup):
     """Test github-proxy-report fails with a 400."""
     with patch('webcompat.issues.proxy_request') as github_data:
         github_data.return_value = mock_api_response({
@@ -85,7 +85,7 @@ def test_report_issue_fails_400_for_proxy(client):
             rv = report_issue(FORM_ANON, proxy=True)
 
 
-def test_report_private_issue_returns_nothing(client):
+def test_report_private_issue_returns_nothing(setup):
     """Test that we get nothing back from a private issue report."""
     with patch('webcompat.issues.proxy_request') as github_data:
         github_data.return_value = mock_api_response({
@@ -96,7 +96,7 @@ def test_report_private_issue_returns_nothing(client):
         assert rv is None
 
 
-def test_report_public_issue_returns_moderation_template(client):
+def test_report_public_issue_returns_moderation_template(setup):
     """Test the data in report_public_issue contains the right data."""
     with patch('webcompat.issues.proxy_request') as req:
         req.return_value = mock_api_response({
@@ -113,7 +113,7 @@ def test_report_public_issue_returns_moderation_template(client):
         assert ['action-needsmoderation'] == post_data['labels']
 
 
-def test_moderation_template(client):
+def test_moderation_template(setup):
     """Check the moderation template structure.
 
     - must be a dictionary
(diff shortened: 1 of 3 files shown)


tests/unit/test_issues.py
```diff
@@ -6,109 +6,145 @@
...
+FORM_AUTH = FORM.copy()
+FORM_AUTH['submit_type'] = 'github-auth-report'
+
+
+@pytest.fixture
+def client():
+    """Test client for requests."""
+    webcompat.app.config['TESTING'] = True
+    with webcompat.app.test_client() as client:
+        yield
```
miketaylr, line 46: [In this example](https://flask.palletsprojects.com/en/1.1.x/testing/#the-testing-skeleton), it `yield`s the `client`. Is there a specific reason we don't do that here?

I see none of these tests actually use the `test_client`, so I guess all this fixture is doing is setting testmode. Probably better to yield the `client` I think.

tests/unit/test_issues.py
```diff
@@ -6,109 +6,145 @@
...
+
+
+def test_report_with_auth_on(client):
+    """Test that we can still work with github-auth-report."""
+    with patch('webcompat.issues.github.post') as json_resp:
+        json_resp.return_value = {'number': 2}
+        json_data = report_issue(FORM_AUTH, False)
```
miketaylr, line 53: Can we pass in the second arg as a kwarg? I think `report_issue(FORM_AUTH, proxy=False)` is easier to understand than `report_issue(FORM_AUTH, False)`:

https://github.com/webcompat/webcompat.com/blob/7ada3d5f34b848d88d60843dc8b824533c357df8/webcompat/issues.py#L88

tests/unit/test_issues.py
```diff
@@ -6,109 +6,145 @@
...
+    """Test issue posting report the right json."""
+    with patch('webcompat.issues.proxy_request') as github_data:
+        github_data.return_value = mock_api_response({
+            'status_code': 201,
+            'content': '[{"number":2}]',
+        })
+        with patch('webcompat.issues.report_private_issue') as silent:
+            # We avoid making a call to report_private_issue
+            silent.return_value = None
+            json_data = report_issue(FORM_ANON, True)
```
miketaylr, line 68: (same comment here for `report_issue`)

tests/unit/test_issues.py
```diff
@@ -6,109 +6,145 @@
...
+            assert type(json_data) is dict
+            assert json_data.get('number') == 2
+
+
+def test_report_issue_fails_400_for_unknown_type(client):
+    """Test request fails with a 400 because type unknown."""
+    with pytest.raises(BadRequest):
+        rv = report_issue(FORM, True)
```
miketaylr, line 76: (same comment here for `report_issue`)

tests/unit/test_issues.py
```diff
@@ -6,109 +6,145 @@
...
+
+def test_report_issue_fails_400_for_proxy(client):
+    """Test github-proxy-report fails with a 400."""
+    with patch('webcompat.issues.proxy_request') as github_data:
+        github_data.return_value = mock_api_response({
+            'status_code': 400,
+            'content': '[{"number":2}]',
+        })
+        with pytest.raises(BadRequest):
+            rv = report_issue(FORM_ANON, True)
```
miketaylr, line 87: (same comment here for `report_issue`)

tests/unit/test_issues.py
karlcow: Good point I will remove it entirely.

tests/unit/test_issues.py
karlcow: will do for the four of them. Better being explicit indeed.

tests/unit/test_issues.py
karlcow: see https://github.com/webcompat/webcompat.com/pull/3322#discussion_r433003788

tests/unit/test_issues.py
karlcow: see https://github.com/webcompat/webcompat.com/pull/3322#discussion_r433003788

tests/unit/test_issues.py
karlcow: see https://github.com/webcompat/webcompat.com/pull/3322#discussion_r433003788

karlcow:
ok rebased. and pushed.

tests/unit/test_issues.py
```diff
@@ -6,109 +6,143 @@
...
+    ('username', ''), ])
+
+FORM_ANON = FORM.copy()
+FORM_ANON['submit_type'] = 'github-proxy-report'
+FORM_AUTH = FORM.copy()
+FORM_AUTH['submit_type'] = 'github-auth-report'
+
+
+@pytest.fixture
+def client():
```
miketaylr, line 42: nit: maybe better to rename this method to `setup` or something now?

karlcow: good idea. done.

miketaylr:
Thanks Karl, very nice improvements.

PR #3325, opened 2020-05-30
Issue #3319 - Removes the form building from home-page

### LINKED ISSUE #3319: Remove the form building from home page
In the spirit of removing unnecessary call to unused functions for this context.

The home page is making a call to `get_form` which is not useful anymore.

https://github.com/webcompat/webcompat.com/blob/b3e5e1ffb58dc4cc372e4dad94eb390675a9c2d3/webcompat/views.py#L180

and communicating the data to a template which does nothing with it.
https://github.com/webcompat/webcompat.com/blob/b3e5e1ffb58dc4cc372e4dad94eb390675a9c2d3/webcompat/views.py#L187

we just need to remove these.

Capture d’écran 2020-05-28 à 12 16 38

soniasingla:
Fixes #3319 

cc : @karlcow @miketaylr

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -177,14 +177,13 @@ def index():
         'rel': 'preload'
     })
     ua_header = request.headers.get('User-Agent')
-    bug_form = get_form({'user_agent': ua_header})
 
     # browser_name is used in topbar.html to show the right add-on link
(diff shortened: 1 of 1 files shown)


karlcow:
test failing, but not sure it's a real fail. Let's try to re-run the tests.

PR #3328, opened 2020-06-01
Fixes #2900 - Converts the sources to use webpack instead of grunt

### LINKED ISSUE #2900: Create webpack config
After discussion with @magsout, we would like to add a simple webpack config and make sure nothing breaks. Once it's working, we can gradually add the corresponding functionality from grunt listed here https://github.com/webcompat/webcompat.com/issues/1352

ksy36:
In this PR I've created development and production configs for webpack and changed some js files to work with webpack. Mainly the changes include:

1.  import jst file template directly instead of using `wcTmpl`
2. import Backbone models and mixins instead of using global variable (for example `issues` or `issueList`)

I've changed output folder for the files included to the bundle to be `static/dist` (instead of having separate folders for js and css), and the output looks like this

Screen Shot 2020-06-01 at 12 13 20 PM

`vendor.js` bundle contains files from `js/vendor` and the rest of js files are for the corresponding pages (similarly how it is done with grunt in concat.js)

r? @miketaylr


webcompat/static/js/lib/editor.js
```diff
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-var issues = issues || {}; // eslint-disable-line no-use-before-define
+import $ from "jquery";
```
miketaylr, line 4: nit: can we add a newline before this?

webcompat/static/js/lib/flash-message.js
```diff
@@ -12,6 +12,9 @@ eventBus.trigger('flash:error', {message: 'hi', timeout: 1000});
 `opts.timeout` (optional) is the length of time before fading the message out.
                           Default is 3 seconds
 */
+import Backbone from "Backbone";
```
miketaylr, line 15: nit: can we add a newline?

webcompat/static/js/lib/models/issue.js
```diff
@@ -146,10 +147,11 @@ issues.Issue = Backbone.Model.extend({
   },
 });
 
-issueList.Issue = issues.Issue.extend({});
+// issueList.Issue = issue.extend({});
```
miketaylr, line 150: nit: can we delete this comment?

webcompat/static/js/lib/models/label-list.js
```diff
@@ -22,7 +25,7 @@ var issues = issues || {}; // eslint-disable-line no-use-before-define
  * new issues.LabelList({url:'/path/to/labels.json'});
  */
 
-issues.LabelList = Backbone.Model.extend({
+const LabelList = Backbone.Model.extend({
```
miketaylr, line 28: <s>Question: in issue.js, we do `export const Issue = Backbone.Model.extend` and `export const IssueCollection = Backbone.Collection.extend`. 

So why don't we `export` `LabelList`?</s>

I answered by scrolling down. 😄 

We export it later: `export { LabelList };`

webcompat/static/js/lib/wizard/validation.js
```diff
@@ -10,6 +10,8 @@ export const isUrlValid = (value) => {
 };
 
 export const isEmpty = (value) => {
+  if (!value) return true;
```
miketaylr, line 13: (this seems unrelated to webpack, was this just a drive-by improvement?)

webcompat/templates/new-issue.html
```diff
@@ -1,9 +1,5 @@
 {%- extends "layout.html" -%}
 
-{% block extracss %}
-<link href="{{ url_for('static', filename='css/src/issue-wizard.css') }}" type="text/css" rel="stylesheet">
```
miketaylr, line 4: Is this unused?

tests/functional/image-uploads-non-auth.js
```diff
@@ -168,7 +168,7 @@ registerSuite("Image Uploads (non-auth)", {
           .execute(() => {
             // work around for chrome "Other element would receive the click"
             // error
-            $(".js-remove-upload")[0].click();
+            document.querySelector(".js-remove-upload").click();
```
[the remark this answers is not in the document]
magsout: @miketaylr you can define jquery globally in webpack configuration.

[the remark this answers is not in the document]
ksy36: Yeah, it's no longer available globally on the window object (e.g. window.$ or window.Jquery), so I decided to use DOM methods instead. 

However jst templates still use jquery and underscore without explicitly importing it, so I made it "global" for the bundle for now with https://webpack.js.org/plugins/provide-plugin/#usage-jquery, which exposes a corresponding module once "_" or "$" is found in the code with no explicit import
https://github.com/webcompat/webcompat.com/blob/4bb999e2c08cd9629723912576e7174ed2c90dfb/webpack/webpack.common.js#L58-L65

It's also possible to make jquery available on the window object with this plugin as @magsout mentioned, but I don't think we need it, since it would be beneficial only for tests 
```js
new webpack.ProvidePlugin({
  'window.jQuery': 'jquery'
});
```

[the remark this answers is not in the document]
miketaylr: > @miketaylr you can define jquery globally in webpack configuration.

Cool good to know.

> I don't think we need it

Yeah, it's overkill for just selecting a DOM element.

magsout:
super nice job @ksy36

miketaylr:
@magsout one thing we don't have yet in this webpack config is the cssnext stuff: 

https://github.com/webcompat/webcompat.com/blob/master/grunt-tasks/cssnext.js#L13

I'll be honest, I'm not 100% sure what it's doing. 😅 (I assume it's like a polyfill for new CSS features for older browsers).

Is there a webpack-y way to do this that you recommend, assuming we still want / need it?

miketaylr:
Or asked another way, what happens if we removed this? Do things break in older browsers?

miketaylr:
(is https://github.com/csstools/postcss-preset-env/blob/master/INSTALL.md#webpack the right thing?)

webcompat/templates/new-issue.html
ksy36: It has moved to https://github.com/webcompat/webcompat.com/blob/4bb999e2c08cd9629723912576e7174ed2c90dfb/webcompat/static/js/lib/webcompat.js , since css files need to be included in js in order for webpack to pick it up

miketaylr: Aha, thanks. Let's pretend I already noticed that, since I reviewed it. 🙈

magsout:
@miketaylr cssnext compiles new css stuffs on caniuse-db relay on our configuration on what browsers we want to support. cssnext is deprecated so we have to migrate to `postcss-prev-env`.

magsout:
@miketaylr @ksy36 I can do it if you want/need

ksy36:
Thanks @magsout and @miketaylr, I'll try adding `postcss-prev-env` and test it out

ksy36:
Added `postcss-prev-env` and it's working good :) Also added image minification which currently minifies images referenced in the css. I'll create a separate PR for minifing images that are referenced in html since it requires some changes on the backend

webpack/webpack.common.js
```diff
@@ -35,7 +36,27 @@ module.exports = {
...
+        use: [
+          MiniCssExtractPlugin.loader,
+          "css-loader",
+          {
+            loader: "postcss-loader",
+            options: {
+              ident: "postcss",
+              plugins: () => [
+                postcssPresetEnv({
+                  browsers: [
```
miketaylr, line 48: We could probably revisit these versions... Maybe Firefox 52+, IE11, Safari 9, Chrome...75?80?. Not sure about Opera. Does that mean Chromium Opera or Presto opera?

ksy36: Good idea. I've looked at the documentation https://github.com/browserslist/browserslist#best-practices and there is "[defaults](https://browserl.ist)" list that we can use and possibly add things to it. 

So it can be `defaults, Firefox > 52, Safari >=9, Chrome >=75`
https://browserl.ist/?q=defaults%2C+Firefox+%3E+52%2C+Safari+%3E%3D9%2C+Chrome+%3E%3D75
And for Opera there is 64 and 63 already in the list, 63 has 0.01% usage.. so maybe that's enough?
And to your question on Chromium Opera or Presto, I think it depends on version number (I guess Presto has stopped at 12.18)

miketaylr: LGTM!

miketaylr:
> Nice work! Note to self: we need to update the build scripts before we can deploy now.

Well, before we merge this branch into master.

ksy36:
Thanks!

magsout:
👍🎊🎉💪

karlcow:
🌳 !

PR #3331, opened 2020-06-07
Fixes #3330 - separate bundle for modern browsers

### LINKED ISSUE #3330: Add a separate bundle for modern browsers

ksy36:
r? @miketaylr 

In this PR I've changed webpack production build script to create two versions of the same script, one for modern browsers and another for browsers that do not support es6.
The files are included in html the following way:
```js
  <script type="module" src="file.js"></script>
  <script nomodule src="file.es5.js"></script>
```
Browsers that do not support `type="module"` will download nomodule version, and browsers that do support it, will download the modern build with less transpiled code and ignore nomodule version.

This is based on the idea from these two blog posts:
https://philipwalton.com/articles/deploying-es2015-code-in-production-today/
https://web.dev/serve-modern-code-to-modern-browsers/

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,14 @@
 {
-  "presets": ["@babel/preset-env"]
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "bugfixes": true,
+        "modules": false,
+        "targets": {
+          "esmodules": true
+        }
+      }
+    ]
+  ]
 }
--- a/webcompat/templates/console-logs.html
+++ b/webcompat/templates/console-logs.html
@@ -27,6 +27,9 @@
 {% endfor %}
 </div>
 
-<script src="{{ url_for('static', filename='dist/console-logs.js')|bust_cache }}"></script>
+<script type="module" src="{{ url_for('static', filename='dist/console-logs.js')|bust_cache }}"></script>
+{%- if config.PRODUCTION or config.STAGING -%}
+<script nomodule src="{{ url_for('static', filename='dist/console-logs.es5.js')|bust_cache }}"></script>
+{%- endif %}
 </body>
 </html>
--- a/webpack/webpack.common.js
+++ b/webpack/webpack.common.js
@@ -28,16 +28,6 @@ module.exports = {
         ],
       },
       {
-        test: /\.js$/,
-        exclude: [
-          /node_modules/,
-          path.resolve(__dirname, "../webcompat/static/js/vendor"),
-        ],
-        use: {
-          loader: "babel-loader",
-        },
-      },
-      {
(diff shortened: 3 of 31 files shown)


.babelrc
```diff
@@ -1,3 +1,14 @@
 {
-  "presets": ["@babel/preset-env"]
+  "presets": [
+    [
+      "@babel/preset-env",
```
ksy36, line 4: This is default babel configuration for dev and production for targeting "modern" browsers (versions that support ES Modules )
https://babeljs.io/docs/en/babel-preset-env#targetsesmodules

webpack/webpack.common.js
```diff
@@ -21,19 +21,12 @@ module.exports = {
           },
           {
             loader: "ejs-loader",
+            options: {
+              variable: "data",
```
miketaylr, line 25: I don't 100% understand the need for this change (and the associated changes), but I can guess what it does based on reading https://github.com/difelice/ejs-loader#options

webpack/webpack.prod.js
```diff
@@ -4,12 +4,43 @@ const OptimizeCSSAssetsPlugin = require("optimize-css-assets-webpack-plugin");
 const commonConfig = require("./webpack.common.js");
 const ImageminPlugin = require("imagemin-webpack-plugin").default;
 
-module.exports = () =>
+const isEs5 = (env) => env && env.es5;
+
+// If it's a es5 build, use babel env preset
```
miketaylr, line 9: 💯 points for the comment

webpack/webpack.common.js
miketaylr: OK, my guess wasn't really accurate, the commit message gives a better hint: `Add data variable to ejs loader for jst templates, to avoid errors in strict mode`

miketaylr: better explanation from @ksy36:

> ejs loader adds `with()` for variables that are passed to the template, which errors out in strict mode

Now i get it. 👍

webcompat/templates/console-logs.html
```diff
@@ -27,6 +27,9 @@
 {% endfor %}
 </div>
 
-<script src="{{ url_for('static', filename='dist/console-logs.js')|bust_cache }}"></script>
+<script type="module" src="{{ url_for('static', filename='dist/console-logs.js')|bust_cache }}"></script>
+{%- if config.PRODUCTION or config.STAGING -%}
+<script nomodule src="{{ url_for('static', filename='dist/console-logs.es5.js')|bust_cache }}"></script>
```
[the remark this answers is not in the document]
ksy36: lol

ksy36:
Thanks!

PR #3335, opened 2020-06-09
Fixes #3334 - Fix for "Thanks for reporting an issue" not showing 

### LINKED ISSUE #3334: "Thanks for reporting an issue" is missing in the refactored version
After submitting an issue, "Thanks for reporting an issue!" flash message is missing at the top of the page

ksy36:
Seems that `document.currentScript` is null with the way webpack loads `flashed-messages.js`, and also currentScript it's not available for es6 modules https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript . I've changed the script a bit, so it'll work with both modern and legacy bundles

r? @miketaylr

### DIFF
--- a/webcompat/static/js/lib/flashed-messages.js
+++ b/webcompat/static/js/lib/flashed-messages.js
@@ -7,13 +7,19 @@
 // in flash-message.js. See layout.html for how category and message data
 // attributes are set.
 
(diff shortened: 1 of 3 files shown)


ksy36:
Thanks!!

Issue #3354, opened 2020-06-16
`get_checksum` should not be able to break the app


miketaylr:
The root cause of #3352 is that `get_checksum` throws because it's trying to find a file that doesn't exist (due to a simple typo).

We try to handle a `KeyError` but don't handle another big pile of exceptions:

```python
Jun 16 19:27:59 staging-server webcompat-staging: Traceback (most recent call last):
Jun 16 19:27:59 staging-server webcompat-staging:   File "./webcompat/templates/__init__.py", line 33, in get_checksum
Jun 16 19:27:59 staging-server webcompat-staging:     checksum = cache_dict[file_path]
Jun 16 19:27:59 staging-server webcompat-staging: KeyError: '/[snip]/staging.webcompat.com/webcompat/static/dist/flashed-messages.e5.js'
Jun 16 19:27:59 staging-server webcompat-staging: #012During handling of the above exception, another exception occurred:
Jun 16 19:27:59 staging-server webcompat-staging: Traceback (most recent call last):
Jun 16 19:27:59 staging-server webcompat-staging:   File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 2292, in wsgi_app
Jun 16 19:27:59 staging-server webcompat-staging:     response = self.full_dispatch_request()
Jun 16 19:27:59 staging-server webcompat-staging:   File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1815, in full_dispatch_request
Jun 16 19:27:59 staging-server webcompat-staging:     rv = self.handle_user_exception(e)
Jun 16 19:27:59 staging-server webcompat-staging:   File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1718, in handle_user_exception
Jun 16 19:27:59 staging-server webcompat-staging:     reraise(exc_type, exc_value, tb)
Jun 16 19:27:59 staging-server webcompat-staging:   File "/usr/local/lib/python3.7/dist-packages/flask/_compat.py", line 35, in reraise
Jun 16 19:27:59 staging-server webcompat-staging:     raise value
Jun 16 19:27:59 staging-server webcompat-staging:   File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1813, in full_dispatch_request
Jun 16 19:27:59 staging-server webcompat-staging:     rv = self.dispatch_request()
[...]

PR #3355, opened 2020-06-17
Issue #3354 - Handles FileNotFoundError for assets

karlcow:
This PR fixes issue #3354

## Proposed PR background

Basically in case the actual files where missing, we got a FileNotFound spiraling up and returning a 500.

This now will help a bit more by failing gracefully. The resource still does not exist but we will be able to see it in the network panel of devtools. The uris will carry a `?missing_file`?

### DIFF
--- a/tests/unit/test_template.py
+++ b/tests/unit/test_template.py
@@ -6,7 +6,7 @@
 
 """Tests for webcompat/templates/__init__.py."""
 
-import unittest
+import pytest
 
 from webcompat.templates import format_date
 from webcompat.templates import format_milestone_class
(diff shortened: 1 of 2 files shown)


karlcow:
@miketaylr do you remember if there was a reason to put the get_checksum inside? it's slightly weird.

karlcow:
This was done a long time ago
https://github.com/webcompat/webcompat.com/commit/133cffe4806ef57104b78ec8f773b501192f6e21

:)

miketaylr:
> do you remember if there was a reason to put the get_checksum inside? it's slightly weird.

you might have left out a word. inside _____?

karlcow:
@miketaylr do you remember if there was a rationale for doing this nesting instead of two flat functions? 

```python
def bust_cache(file_path):
    def get_checksum(file_path):
         …
         return
    return
```

karlcow:
Let's see the test coverage.

```
coverage report -m webcompat/templates/__init__.py 
Name                              Stmts   Miss  Cover   Missing
---------------------------------------------------------------
webcompat/templates/__init__.py      75     19    75%   32-37, 42, 47-61, 96
```

We can do better. 

Let me complete this PR. Do not merge for now.

karlcow:
I have switched to draft to not break anything.

karlcow:
```
(env) ~/code/webcompat.com-karlcow % coverage report -m webcompat/templates/__init__.py 
Name                              Stmts   Miss  Cover   Missing
---------------------------------------------------------------
webcompat/templates/__init__.py      79      1    99%   99
```

ok better the 1% missing is because we have one case which is not tested because the function doesn't make sense. I will open another issue for it.

karlcow:
> ok better the 1% missing is because we have one case which is not tested because the function doesn't make sense. I will open another issue for it.

opened in #3356

miketaylr:
> @miketaylr do you remember if there was a rationale for doing this nesting instead of two flat functions?

Ah, no -- no memory. Nest functions seems nicer and more pythonic.

miketaylr:
Thanks @karlcow

Issue #3316, opened 2020-05-27
Check-marks are displayed misaligned on "Browser/Operating System" fields


softvision-oana-arbuzov:
**URL**: https://webcompat.com/issues/new

**Browser / Version**: Firefox Preview Nightly 200526 (🦎 78.0a1-20200522094316), Chrome 81.0.4044.138
**Operating System**: Huawei P20 Lite (Android 8.0.0) - 1080 x 2280 pixels, 19:9 ratio (~432 ppi density), Samsung Galaxy S6 Edge (Android 7.0) - 1440 x 2560 pixels (~577 ppi pixel density)

**Description**: Items are misaligned

**Steps to Reproduce**:
1. Navigate to https://webcompat.com/issues/new
2. Fill in and choose required data until reaching "Environment" section.
3. Tap "different device or browser" link.
4. Observe check-marks alignment.

**Expected Behavior:**
Check-marks are correctly aligned, on the right side of "Browser/Operating System" fields.

**Actual Behavior:**
Check-marks are displayed misaligned over the "Browser/Operating System" fields.

**Notes:**
1. Screenshot attached.
![Screenshot_3](https://user-images.githubusercontent.com/12184325/83002492-cc443180-a015-11ea-9cee-1e4c4e74c679.png)
2. Also reproducible on Firefox Nightly RDM.

Affected area:
```html
<div class="row mobile-col">
	<div class="form-element js-Form-group low with-validation-icons half is-validated js-no-error">
		<label class="form-label" for="browser">Browser</label>
		<small class="os-browser-error">Browser required</small>
		<input class="form-field text-field" id="browser" name="browser" type="text" value="Firefox 78.0">
	</div>
    <div class="form-element js-Form-group low with-validation-icons half is-validated js-no-error">
			<label class="form-label" for="os">Operating System</label>
			<small class="os-browser-error">OS required</small>
			<input class="form-field text-field" id="os" name="os" type="text" value="Windows 10">
	</div>
</div>
```

Watchers:
@softvision-oana-arbuzov
@cipriansv

karlcow:
This can be "fixed" by 

```css

/* webcompat.css | https://webcompat.com/dist/webcompat.css */

.low.is-error::before, .low.is-validated::after {
  /* bottom: 32px; */
  bottom: 25px;
}

```

but it has still the same issue on using position absolute for something which should be just flexible depending on the fonts and the context.

why not instead of `:after` using a `:valid` selector on input.
[...]

PR #3363, opened 2020-06-26
Issue #3316 - fixed misaligned check-marks

magsout:
This PR fixes issue #3316 

see sample https://share.getcloudapp.com/ApuAxkWO


magsout:
I added a wrapper around the input to have a constant height. We can add checkmarks in the center using absolute and a transform.

karlcow:
I let @ksy36 review the JS part.
Thanks a lot @magsout

PR #3371, opened 2020-07-02
fix #3360 - creates controled values list for reported_with

### LINKED ISSUE #3359: constraints for reported_with are not defined
Right now:

`reported_with` is a free form field.

https://github.com/webcompat/webcompat.com/blob/d6781a0829ffb68afd7e4286531d505f338cf2eb/webcompat/form.py#L297

### LINKED ISSUE #3360: Some issues with empty <!-- @reported_with: -->
https://github.com/webcompat/web-bugs/issues/52582 is one example:

Screen Shot 2020-06-24 at 10 54 18 AM

I would have expected this to be impossible, since there should be a fallback if there's no `src` param. Perhaps we need to add one to the Fenix/AC web extension?

karlcow:
fix #3360 
fix #3359

## Proposed PR background

```
issue #3360 - adds list of reported_with values
e48e2e7

This makes a controled list of accepted values for the parameter reported_with which is added to the body of the bug report.

The reports which are out of this list will be either web (default) or unknown.

Ultimately we will move this list and a couple of others outside of the committed code but still loaded by the config file. So we can have more flexibility when it's time to add or retire a value without having to merge a pull request.

issue #3360 - adds a test for reported_with
037fd68

We were not testing that we were getting the value we needed when it exists.

Issue #3360 - Adds control for reported_with
22c964c

This makes a couple of modifications

1. It uses a list of controled values coming from the config file.
2. it makes sure the extracted value is part of the controled list
3. It also records when the value is not known aka not part of the controled list. This is important to not screw the stats for the Web values AND to understand when there is a new client.

It creates the tests associated with the function.

Probably we need to convert test_form.py fully to pytest, but that should be done in another PR to not muddy this one.

```

### DIFF
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -294,7 +294,7 @@ def get_form(form_data, form=IssueForm):
     bug_form.details.data = json.dumps(form_data.get('details'), indent=2)
     bug_form.extra_labels = form_data.get('extra_labels', None)
     bug_form.os.data = get_os(ua_header)
-    bug_form.reported_with.data = form_data.get('src', 'web')
+    bug_form.reported_with.data = extract_report_source(form_data)
     bug_form.ua_header.data = ua_header
     bug_form.url.data = form_data.get('url', None)
     return bug_form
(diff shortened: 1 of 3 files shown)


karlcow:
psstt 😉 @miketaylr

webcompat/form.py
```diff
@@ -443,6 +443,28 @@ def add_metadata(form, metadata_dict):
...
+    """Extract the source of report.
+
+    This also controls if the src argument is part of a known list.
+    """
+    reported_with = 'unknown'
+    try:
+        reported_with = form_data['src']
+        if not is_valid_source(reported_with):
+            reported_with = 'unknown'
+    except KeyError:
```
miketaylr, line 456: Just writing this down to make sure I understand it:

if we have a `src` key in the `form_data` dict, and it's not in our `REPORTED_WITH` control list, we set `reported_with` to `unknown`. But if the `src` key isn't even there, we assume it's a `web` report (because that doesn't set one).

LGTM.

PR #3379, opened 2020-07-08
Fixes #3291 - Make issue moderation slightly more efficient

### LINKED ISSUE #3291: Make issue moderation more efficient
Thinking of ways to optimize issue moderation. Right now the workflow is basically:

1. look at the issue, make sure it's not a violation of our TOS
2. move it to the accepted milestone
3. close the issue
(and then probably go triage the issue)

That's 2 clicks and it gets annoying, because you have to wait on GitHub's UI. Kind of slow sometimes.

Just by moderating some issues, it's obvious that they are acceptable, but invalid. It makes me think we could speed up certain interactions.

A couple of ideas that wouldn't be hard to implement:

1. a browser addon that adds a button somewhere that moves it to "accepted" and closes the issue in one click (via the API in the background). it could even open a new tab with the public issue right after, if that was part of the workflow.

2. a command line tool to do something similar (seems like a not great idea, unless it was connected to your browsing session. having to enter in URLs wouldn't be an improvement)

3. add some common milestones like "accepted: invalid" or "accepted: incomplete". Once you change the milestone, we have the app close the issue, and put in some template comment on the public issue, and close that issue.

I think the 3rd option would be pretty simple.

miketaylr:
@karlcow this is still WIP, I'd like to do 2 things:

1. close an accepted issue automatically (what these patches do)
2. create a few milestones like "accepted - invalid" and "accepted - incomplete" that will make an issue public, but close it and stick it in the corresponding milestone (not yet tackled)

Do you have any suggestions for testing part 1? I'm not sure if it's useful to mock the entire github interaction "github will tell me it's closed => assert is closed". Thoughts?

### DIFF
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -407,7 +407,7 @@ class TestWebhook(unittest.TestCase):
         json_event, signature = event_data('new_event_valid.json')
         payload = json.loads(json_event)
         issue_info = helpers.get_issue_info(payload)
-        with patch('webcompat.webhooks.helpers.proxy_request') as proxy:
+        with patch('webcompat.webhooks.helpers.make_request') as proxy:
             proxy.return_value.status_code = 200
             response = helpers.tag_new_public_issue(issue_info)
             self.assertEqual(response.status_code, 200)
@@ -660,36 +660,30 @@ class TestWebhook(unittest.TestCase):
             self.assertEqual(rv.status_code, 200)
             self.assertEqual(rv.content_type, 'text/plain')
 
-    @patch('webcompat.webhooks.helpers.proxy_request')
+    @patch('webcompat.webhooks.helpers.make_request')
     def test_arguments_private_issue_moderation(self, mock_proxy):
         """Test accepted issue request with the right arguments."""
         with webcompat.app.test_client() as c:
             mock_proxy.return_value.status_code = 200
             rv = helpers.private_issue_moderation(self.issue_info2)
-            args, kwargs = mock_proxy.call_args
-            mock_proxy.assert_called_with(
-                path='repos/webcompat/webcompat-tests/issues/1',
-                method='patch',
-                data=ANY,
-                headers=ANY)
-            self.assertIn(
-                'www.netflix.com - test private issue accepted',
-                kwargs['data'])
+            method, url_path, data = mock_proxy.call_args[0]
+            assert url_path == 'repos/webcompat/webcompat-tests/issues/1'
+            assert method == 'patch'
+            assert data == ANY
+            assert 'www.netflix.com - test private issue accepted' in data['title']  # noqa
 
     def test_get_public_issue_number(self):
         """Test the extraction of the issue number from the public_url."""
         public_url = 'https://github.com/webcompat/webcompat-tests/issues/1'
         self.assertEqual(helpers.get_public_issue_number(public_url), '1')
 
-    @patch('webcompat.webhooks.helpers.proxy_request')
+    @patch('webcompat.webhooks.helpers.make_request')
     def test_arguments_private_issue_rejected(self, mock_proxy):
         """Test rejected issue request with the right arguments."""
         with webcompat.app.test_client() as c:
             mock_proxy.return_value.status_code = 200
             rv = helpers.private_issue_rejected(self.issue_info3)
-            args, kwargs = mock_proxy.call_args
-            post_data = json.loads(kwargs['data'])
-            url_path = kwargs['path']
+            method, url_path, data = mock_proxy.call_args[0]
             issue_number = url_path.replace(
                 'repos/webcompat/webcompat-tests/issues/', '')
             # testing that the path is having a number
@@ -697,17 +691,15 @@ class TestWebhook(unittest.TestCase):
             # testing the path starts with the right string
             self.assertTrue(
                 url_path.startswith('repos/webcompat/webcompat-tests/issues/'))
-            self.assertEqual('Issue rejected.', post_data['title'])
+            self.assertEqual('Issue rejected.', data['title'])
             self.assertIn('Its original content has been deleted',
-                          post_data['body'])
-            self.assertEqual(['status-notacceptable'], post_data['labels'])
-            self.assertEqual('closed', post_data['state'])
-            self.assertIn('milestone', post_data)
-            mock_proxy.assert_called_with(
-                path='repos/webcompat/webcompat-tests/issues/1',
-                method='patch',
-                data=ANY,
-                headers=ANY)
+                          data['body'])
+            self.assertEqual(['status-notacceptable'], data['labels'])
+            self.assertEqual('closed', data['state'])
+            self.assertIn('milestone', data)
+            assert url_path == 'repos/webcompat/webcompat-tests/issues/1'
+            assert method == 'patch'
+            assert data == ANY
 
     def test_prepare_rejected_issue(self):
         """Test we prepare the right payload for the rejected issue."""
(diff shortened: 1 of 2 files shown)


tests/unit/test_webhook.py
```diff
@@ -660,54 +660,48 @@ def test_private_issue_moderated_ok(self, mock_proxy):
...
-                data=ANY,
-                headers=ANY)
+            method, url_path, data = mock_proxy.call_args[0]
+            assert url_path == 'repos/webcompat/webcompat-tests/issues/1'
+            assert method == 'patch'
+            assert data == ANY
             self.assertIn(
-                'www.netflix.com - test private issue accepted',
-                kwargs['data'])
+                'www.netflix.com - test private issue accepted', data['title'])
```
karlcow, line 674: As you switch to pytest for the other ones, you could do this here too.

```diff
-            self.assertIn(
-                'www.netflix.com - test private issue accepted', data['title'])
+            assert 'www.netflix.com - test private issue accepted' in data['title']
```

tests/unit/test_webhook.py
```diff
@@ -660,54 +660,48 @@ def test_private_issue_moderated_ok(self, mock_proxy):
...
+    @patch('webcompat.webhooks.helpers.make_request')
     def test_arguments_private_issue_rejected(self, mock_proxy):
         """Test rejected issue request with the right arguments."""
         with webcompat.app.test_client() as c:
             mock_proxy.return_value.status_code = 200
             rv = helpers.private_issue_rejected(self.issue_info3)
-            args, kwargs = mock_proxy.call_args
-            post_data = json.loads(kwargs['data'])
-            url_path = kwargs['path']
+            method, url_path, data = mock_proxy.call_args[0]
```
karlcow, line 687: better!

tests/unit/test_webhook.py
```diff
@@ -722,18 +716,18 @@ def test_prepare_rejected_issue(self):
...
             mock_proxy.return_value.status_code = 200
             rv = helpers.comment_public_uri(self.issue_info4)
-            request_args = mock_proxy.call_args[1]
-            assert request_args['method'] == 'post'
-            assert request_args['path'] == dest
-            assert request_args['data'] == payload
+            method, path, payload = mock_proxy.call_args[0]
+            assert method == 'post'
+            assert path == dest
+            assert payload == payload
```
karlcow, line 730: Are you sure? 😉

miketaylr: ha, good eye. 👁️

karlcow:
@miketaylr 

> Do you have any suggestions for testing part 1? I'm not sure if it's useful to mock the entire github interaction "github will tell me it's closed => assert is closed". Thoughts?

Asserting that GitHub does what it should be doing is not useful indeed. 
But testing what the answer should be when the code makes certain actions is. 

Thinking about business transaction. Our code currently is probably very tied to the request nature of things instead of the business transation and it's why sometimes we are struggling with what we should test.

Basically we should think in terms of 

* Moderation Actions
* Response from our Proxy Adapter (aka the thing which dialog with GitHub).

We send a message to the adapter, and in return we expect a type of response. This is the business logic we want to test.

The other part is the testing of the Adapter, but it's practically unrelated to the current case of this PR. 

So basically is what we are `make_request` the right abstraction. Let's say just for the sake of it to think about it in our mind here, what would it mean to name `def send_action(issue_number, action_name):`.

Then now we know that when we say `send_action(1000, 'close')`

And maybe this is still too close to what we do, maybe the code should be `Issue.close()`

test would be (pseudo-code):


```python
def test_issue_closing_for_invalid():
    # test setup
    # issue.status is 'open'
    issue = Issue(1000)
    # test
    issue.moderation('invalid')
    assert issue.category == 'invalid'
    assert issue.status == 'closed'
```

or something similar


But it's probably too much for this PR. a lot of refactoring. 

## Summary

1. I would not mock all GH interactions
2. But I would try to test my interactions with the webcompat interface (webcompat is the proxy not GitHub)

miketaylr:
> So basically is what we are `make_request` the right abstraction.

Yeah, right now it's just a low-level wrapper to a lower-level request. Will think about this.

miketaylr:
I like the Issue() class interface idea -- we could use this webhook module as an experimental implementation that might lead to some useful refactoring elsewhere in the app?

To accomplish what I want to do to fix #3291 we need the ability to:

* close an issue
* set a milestone

Not much. To cover what we currently do, we need the ability to:

* set and get issue details (basically consume all info from `get_issue_info`: 

https://github.com/webcompat/webcompat.com/blob/fbaba42888d046e109008a56a18c41f7ba5f90a4/webcompat/webhooks/helpers.py#L146
* get public issue number
* add a comment on private issue
* set labels on public issue (maybe?, maybe better standalone?)
* set issue moderation state (maybe? maybe better standalone?)

miketaylr:
(Update: I probably won't have time to work on this for another week or so, but plan to make progress at some point!)

miketaylr:
(weird..... https://app.circleci.com/pipelines/github/webcompat/webcompat.com/1439/workflows/3a70bbfb-e1dd-4f50-8c1a-5d9ae8667a43/jobs/1429 is what the circle ci Details link points to, which is not my commit:

Screen Shot 2020-07-21 at 3 48 49 PM

miketaylr:
I was looking at using the `@dataclass` decorator (stealing ideas from https://github.com/webcompat/webcompat.com/pull/3406/files#diff-2c52b18d30123a278e9929d3c3a6930b), but there doesn't seem to be a nice way to initialize a dataclass from a dictionary or JSON object. That would be cool to just create the class from the GitHub Issue response.

I did find https://github.com/Fatal1ty/mashumaro#dataclassjsonmixinfrom_jsondata-unionstr-bytes-bytearray-decoder-optionaldecoder-dict_params-optionalmapping-decoder_kwargs which provides this, but unsure we want to add another dependency just for that.

The alternative is to just live with passing in all the properties as arguments:

`Issue(one, two, three, four, five, six, seven, eight, nine, ten)`... etc. Maybe that's fine. I guess I'll start there and see how terrible it turns out to be.

miketaylr:
Or, I don't really know Python that well. This seems to be fine:

```
from dataclasses import dataclass
@dataclass
class Foo:
    hi: str
    yo: str

>>> f = Foo(**{'hi': 'there', 'yo': 'dude'})
>>> f
Foo(hi='there', yo='dude')
```

So I can just make sure all the args in the dict match the dataclass and unpack those.

karlcow:
Is it because you want to pass directly the json/dict response from webcompat api to the dataclass? That would create a lot of unecessary variables. just for syntactic sugar. But you can `__post_init__` the values that you have declared.

karlcow:
> So I can just make sure all the args in the dict match the dataclass and unpack those.

Ah! I thought you wanted to create the attributes dynamically. ok. so yes. your `**d` is the good solution.

karlcow:
There are indeed serious issues with GitHub/CircleCi here…
Capture d’écran 2020-07-23 à 10 05 30

Details pointing to https://app.circleci.com/pipelines/github/webcompat/webcompat.com/1439/workflows/3a70bbfb-e1dd-4f50-8c1a-5d9ae8667a43/jobs/1429

if I go to https://app.circleci.com/pipelines/github/webcompat/webcompat.com?branch=pull%2F3366 it exists.

but if I go to https://app.circleci.com/pipelines/github/webcompat/webcompat.com?branch=pull%2F3379… it doesn't ?

Capture d’écran 2020-07-23 à 10 08 49

karlcow:
I created #3411 for the CircleCi failure.

miketaylr:
Oops, I pushed to the wrong branch. I'll just open anew PR.

PR #3384, opened 2020-07-14
Fixes #3380 - Makes Wizard Form the default (python)

### LINKED ISSUE #3380: Remove Python Code legacy which was used for the old form
* [x] Find code which is only used for the old form
* [x] Remove it
* [x] Double check if there are relevant tests

karlcow:
This PR fixes issue #3380 

## Proposed PR background

```
(env) ~/code/webcompat.com-karlcow % git log 23f2665a..73dcff1f                                
commit 73dcff1fb95acf013328de6461a4ac911070c9ce (HEAD -> 3380/1, origin/3380/1)
Author: Karl Dubost <karl+github@la-grange.net>
Date:   Tue Jul 14 11:37:29 2020 +0900

    Issue #3380 - Makes WizardForm the default form
    
    So this is doing a couple of things.
    
    - It removes the duplicate problem_choices list of 2 items tuples.
    - it removes the IssueForm class
    - it transfers the IssueForm attributes into FormWizard
    - it reorders in some broad categories the fields  for the form
    - it switches getform to FormWizard. (it was not the case!)
    - it adjusts the tests

commit 304d11a7c23ac845ef192cb3d00671ec81ee888b
Author: Karl Dubost <karl+github@la-grange.net>
Date:   Tue Jul 14 11:32:58 2020 +0900

    Issue #3380 - accepts list with 3 and 2 items tuples
    
    Originally the list of tuples were 2 items long.
    but the wizard is using in some cases 3 items long list (because of icons) for radio fields.
    
    This leds to the duplication of the list of problem_choices with 2 items and 3 items. So instead of duplicating code, let's handle it in the function itself. There might be something more elegant to do. We can discover that later.

commit 09853fb7abf48f77866840c70d4a3a928b53ae7c
Author: Karl Dubost <karl+github@la-grange.net>
Date:   Tue Jul 14 11:28:52 2020 +0900

    Issue #3380 - simplifies PrefixedRadioField
    
    This optimizes a bit the code to make it easier to read. It doesn't change the global logic

commit 23f2665af5333ade44833f9c9a71b115f2d9ed52
Author: Karl Dubost <karl+github@la-grange.net>
Date:   Tue Jul 14 11:25:59 2020 +0900

    issue #3380 - Cleans up code
    
    These are breadcrumbs of dead code and useless redirection of variables.
    
    Let's remove them before modifying the Form.

```

### DIFF
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -173,22 +173,20 @@ contact_label = 'Sharing your GitHub username—without logging in—could help 
 
 
 class PrefixedRadioField(RadioField):
-    """Prefix radio field label with an image."""
+    """Prefix radio field label with an image.
+
+    This renders the radio elements with a specific html markup.
+    """
     def __init__(self, *args, **kwargs):
         prefixed_choices = kwargs.pop('choices')
-        template = '<div class={css_class}><img src={src}/></div> {text}'
         choices = []
-
-        css_class = 'icon-container'
         for slug, img, text in prefixed_choices:
-            filename = 'img/svg/icons/{img}'.format(img=img)
+            filename = f'img/svg/icons/{img}'
             src = url_for('static', filename=filename)
-            label = Markup(template.format(
-                src=src, css_class=css_class, text=text)
-            )
+            t = f'<div class="icon-container"><img src="{src}"/></div> {text}'
+            label = Markup(t)
             choice = (slug, label)
             choices.append(choice)
-
         kwargs['choices'] = choices
         super().__init__(*args, **kwargs)
 
--- a/webcompat/form.py
+++ b/webcompat/form.py
@@ -22,7 +22,6 @@ from flask_wtf.file import FileAllowed
 from flask_wtf.file import FileField
 from markupsafe import Markup
 from wtforms import HiddenField
-from wtforms import Label
 from wtforms import RadioField
 from wtforms import StringField
 from wtforms import TextAreaField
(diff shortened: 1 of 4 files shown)


webcompat/form.py
```diff
@@ -329,6 +329,8 @@ def build_details(details):
 
 def get_radio_button_label(field_value, label_list):
     """Return human-readable label for problem choices form value."""
+    if len(label_list[0]) == 3:
```
miketaylr, line 322: Here are we checking the length of the first item of `label_list` and not `label_list` itself? 

Can we update `test_radio_button_label` to test a label_list with 2 and 3 members? Right now it looks like it only tests that one scenario.

webcompat/form.py
```diff
@@ -270,9 +253,16 @@ class FormWizard(IssueForm):
         [InputRequired(message=radio_message)],
         choices=browser_choices
     )
+    # Bots Trap
```
miketaylr, line 256: I think this comment is in the wrong spot. A few lines above is `username`, which is the dummy field. `username_pattern` is used as validation for the `contact` field (which is the optional non-auth github username field). Maybe renaming it to something like `github_username_pattern` would make it less confusing.

webcompat/form.py
karlcow: oh good catch, I have forgotten. Thanks.

webcompat/form.py
karlcow: nice catch I had done it correctly at first. And messed up when reordering.

webcompat/form.py
karlcow: done.

webcompat/form.py
karlcow: done

miketaylr:
(we should be sure to test carefully on staging here, just to make sure we don't break reporting :))

karlcow:
> (we should be sure to test carefully on staging here, just to make sure we don't break reporting :))

Definitely.

Issue #3046, opened 2019-10-25
Allow browser multiple selection


softvision-oana-arbuzov:
URL: https://staging.webcompat.com/issues/new 
Browser / Version: Firefox Nightly 71.0a1 (2019-10-24)
Operating System: Windows 10 Pro

**Steps to Reproduce:**
    1. Navigate to https://staging.webcompat.com/issues/new
    2. Fill in the requires information until reaching “Browser” selection area.
    3. Observe behavior after selecting one of the browsers in the list.
       
**Expected Behavior:**
Multiple selection is allowed.

**Actual Behavior:**
Single selection is allowed.

**Note**
    1. Screenshot attached.
    2. Also reproducible on Chrome 78.0.3904.70
	
Affected area:
```html
<ul id="browsers">
	<li>
		<input id="browsers-0" name="browsers" type="radio" value="chrome">
			<label for="browsers-0">
				<div class="icon-container">
					<img src="/img/svg/icons/svg-chrome.svg/">
				</div> Chrome
			</label>
	</li>
	<li>
		<input id="browsers-1" name="browsers" type="radio" value="edge">
			<label for="browsers-1">
				<div class="icon-container">
					<img src="/img/svg/icons/svg-edge.svg/">
				</div> Edge
		</label>
	</li>
<!--cut out-->
</ul>
```

Proposed design:
![image](https://user-images.githubusercontent.com/12184325/67571699-47bac900-f73d-11e9-9d99-a5216b036f40.png)

**Watchers:**
@softvision-sergiulogigan
@softvision-oana-arbuzov
@cipriansv

sv;

miketaylr:
I think this is an interesting suggestion. Something we could add if we decide the form is sticking around.

PR #3385, opened 2020-07-14
Fix message requesting multiple browser selection

tagawa:
This PR avoids confusion caused by issue #3046 

## Proposed PR background

The issue form asks for multiple browsers to be selected but only one selection can be made. This change indicates that one browser should be selected.


karlcow:
@tagawa Could you open an issue and then change the commit messages to link to that issue. :) Thanks

tagawa:
@karlcow Not sure I understand – I linked to the existing issue #3046 in the PR.

tagawa:
@karlcow Oh, maybe you mean you want a dedicated issue for the for message itself, so it can be closed leaving #3046 open? I can do that. Seems a bit bureaucratic though.

karlcow:
It's because the scope of #3046 is different. 
#3046 is about adding the multi-browser feature.
Fixing #3046 will mean removing your commit :)

tagawa:
Thanks for the explanation – I'll follow your instructions.

PR #3390, opened 2020-07-14
Fixes #3378 - IE 11 fixes

### LINKED ISSUE #3378: webcompat form broken in IE11
**URL**: https://webcompat.com/issues/new

**Browser / Version**: IE 11.0
**Operating System**: Windows 10
**Tested Another Browser**: Yes Chrome

**Problem type**: Design is broken
**Description**: Items are overlapped
**Steps to Reproduce**:
The whole page is broken. Problem is quite obvious

<details>
<summary>Browser Configuration</summary>
<ul>
  <li>None</li>
</ul>
</details>

_From [webcompat.com](https://webcompat.com/) with ❤️_

ksy36:
This PR includes changes to make the site usable in IE 11 and also clean up of the old form js files

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,5 @@
 {
+  "sourceType": "unambiguous",
   "presets": [
     [
       "@babel/preset-env",
@@ -7,7 +8,9 @@
         "modules": false,
         "targets": {
           "esmodules": true
-        }
+        },
+        "corejs": "3",
+        "useBuiltIns": "usage"
       }
     ]
   ]
--- a/webcompat/static/js/lib/wizard/ui-utils.js
+++ b/webcompat/static/js/lib/wizard/ui-utils.js
@@ -17,11 +17,8 @@ export const scrollToElement = (element, delay = 250) => {
   setTimeout(() => {
     const topOfElement =
       window.pageYOffset + element.getBoundingClientRect().top - HEADER_HEIGHT;
-    window.scroll({ top: topOfElement, behavior: "smooth" });
+    $("html, body").animate({ scrollTop: topOfElement });
   }, delay);
-
-  //@todo figure out why this is needed
-  //this.isSubproblem ? 450 : 250
 };
 
 export const showContainer = (container, animationName = "slidedown") => {
--- a/webpack/webpack.common.js
+++ b/webpack/webpack.common.js
@@ -2,7 +2,6 @@ const path = require("path");
 const webpack = require("webpack");
 const entries = require("./entries.js");
 const MiniCssExtractPlugin = require("mini-css-extract-plugin");
-const postcssPresetEnv = require("postcss-preset-env");
 
 module.exports = {
   entry: entries,
@@ -29,26 +28,7 @@ module.exports = {
       },
       {
         test: /\.css$/,
-        use: [
-          MiniCssExtractPlugin.loader,
-          "css-loader",
-          {
-            loader: "postcss-loader",
-            options: {
-              ident: "postcss",
-              plugins: () => [
-                postcssPresetEnv({
-                  browsers: [
-                    "defaults",
-                    "Firefox > 52",
-                    "Safari >=9",
-                    "Chrome >=75",
-                  ],
-                }),
-              ],
-            },
-          },
-        ],
+        use: [MiniCssExtractPlugin.loader, "css-loader", "postcss-loader"],
       },
       {
         test: /\.(png|svg|jpg|gif)$/,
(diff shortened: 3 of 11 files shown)


webcompat/static/js/lib/wizard/ui-utils.js
```diff
@@ -17,11 +17,8 @@ export const scrollToElement = (element, delay = 250) => {
   setTimeout(() => {
     const topOfElement =
       window.pageYOffset + element.getBoundingClientRect().top - HEADER_HEIGHT;
-    window.scroll({ top: topOfElement, behavior: "smooth" });
+    $("html, body").animate({ scrollTop: topOfElement });
```
ksy36, line 20: Added it to make scrolling smooth in IE. So it makes it smooth in Safari as well :)

webcompat/static/js/lib/wizard/ui-utils.js
```diff
@@ -17,11 +17,8 @@ export const scrollToElement = (element, delay = 250) => {
...
   }, delay);
-
-  //@todo figure out why this is needed
-  //this.isSubproblem ? 450 : 250
```
ksy36, line 24: I've accounted for this todo above, by putting showContainer in a callback of `show` function, that way we don't need a bigger timeout for animating the subcategory

webpack/webpack.prod.js
```diff
@@ -15,8 +15,17 @@ const getOptions = (env) => {
...
 
   return {
+    sourceType: "unambiguous",
     babelrc: false,
-    presets: ["@babel/env"],
+    presets: [
+      [
+        "@babel/env",
+        {
+          corejs: "3",
```
ksy36, line 24: I've noticed even though we have babel transpiling modern code, we still need polyfills for functions that do not supported in IE 11 (Array.includes, string.startsWith). I've used babel-polyfill in the past for polyfilling, but it's deprecated now in favor of core-js. With this config polyfills will be added automatically when a feature is unsupported in target environment (in this case target env is legacy browsers)

Explanation is here: https://stackoverflow.com/a/56505264

.babelrc
```diff
@@ -1,4 +1,5 @@
 {
+  "sourceType": "unambiguous",
```
ksy36, line 2: Similarly to the legacy babel config, polyfills will be added here too, if needed (in case a certain feature is not supported in "modern" browsers)

miketaylr: (just linking myself to https://babeljs.io/docs/en/options#sourcetype `unambiguous` here)

webcompat/static/js/lib/wizard/ui-utils.js
miketaylr: thanks, i was wondering what https://github.com/webcompat/webcompat.com/pull/3390/files#diff-1d0db961f9ba9046d10b32edb5d31394R16 was for.

webpack/webpack.common.js
```diff
@@ -57,7 +37,10 @@ module.exports = {
     ],
   },
   resolve: {
-    modules: [path.resolve(__dirname, "../webcompat/static/js/vendor")],
+    modules: [
+      path.resolve(__dirname, "../webcompat/static/js/vendor"),
+      "node_modules",
```
miketaylr, line 42: This just means you can look inside `node_modules` for a module to include in the webpack bundle, correct?

webpack/webpack.prod.js
miketaylr: Thanks for the comment! Could you include a shorter version of this comment inside the JS file directly?

webcompat/static/js/lib/bugform-prefill.js
```diff
@@ -1,52 +0,0 @@
-/* exported prefillForm */
```
miketaylr, line 1: Do you think you could split out all the old form deletion stuff into a separate commit? 

It's sort of a pain, but it would be nice to not have these 2 unrelated changes coupled together. If you want to crate an issue for that, it could be referenced in https://github.com/webcompat/webcompat.com/projects/12 as well.

webpack/webpack.common.js
ksy36: yeah, so we could just import a package from node_modules, without specifying the full path.
it'll be just `import "package_name"`

webcompat/static/js/lib/bugform-prefill.js
ksy36: Yeah, good point, I've moved them to the issue Karl created https://github.com/webcompat/webcompat.com/pull/3394

ksy36:
ready for another look @miketaylr  :)

ksy36:
Thanks!

PR #3393, opened 2020-07-15
fix #3392 - Removes html and css for original form

### LINKED ISSUE #3392: Remove HTML and CSS code associated with the old form
We switched to the wizard form and we need to double check that some of the CSS code needs to be removed

karlcow:
This PR fixes issue #3392

## Proposed PR background

We switch to the wizard form. This PR removes the HTML and CSS code related to the previous form 

I hope I didn't miss anything.


ksy36:
Just a drive by comment, the wizard styles actually depend on css in the `form.css`. I think also some other form elements across the site using it. So some of the code should stay. 

I've started refactoring it, but got carried away by the fixes for IE11. So if you like, I can do the css part @karlcow

miketaylr:
> I've started refactoring it, but got carried away by the fixes for IE11. So if you like, I can do the css part @karlcow

(It sounds like maybe we should not merge this PR yet)

karlcow:
> Just a drive by comment, the wizard styles actually depend on css in the `form.css`. I think also some other form elements across the site using it. So some of the code should stay.

Ah cool! I tested yesterday and didn't see any failures, but I was not sure, it's why I was asking for review.

karlcow:
@ksy36 

Forgotten to reply to this.
> I've started refactoring it, but got carried away by the fixes for IE11. So if you like, I can do the css part @karlcow

YES no issue.

ksy36:
@karlcow  I've created a new PR https://github.com/webcompat/webcompat.com/pull/3416 with css refactoring

karlcow:
@ksy36 very cool.

karlcow:
ok so I guess we can now close this one.

PR #3394, opened 2020-07-15
Fixes #3391 - Remove js files for the old form and adjust eslint config

### LINKED ISSUE #3391: Remove JS code associated with the old form
We switched to the wizard form and we need to double check that some of the JS code needs to be removed

ksy36:
in this PR I've removed js files for the old form and removed some global variables from eslint config that are no longer needed


.eslintrc.json
```diff
@@ -14,28 +14,15 @@
     "prettier"
   ],
   "globals": {
-    "jQuery": true,
```
miketaylr, line 17: This is awesome, love having fewer globals.

PR #3399, opened 2020-07-17
Fixes #3398 - Add analytics event and vurtual page view tracking

### LINKED ISSUE #3398: Add analytics to measure where users drop off

ksy36:
I've added two types of analytics:

1) Events tracking 
we used to have this before, but with different labels

2) Virtual page views 
each step is a "virtual page", this type allows to set goal (in our case issue submission) and get funnel report.

An example of a funnel report:
image

These two types pretty much track the same actions, but the funnel report is easier to analyze. As far as I understand, it's not possible to create a funnel report with events, so I've added virtual page views for now.

r? @miketaylr

### DIFF
--- a/webcompat/static/js/lib/issue-wizard-popup.js
+++ b/webcompat/static/js/lib/issue-wizard-popup.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import $ from "jquery";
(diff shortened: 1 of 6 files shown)


miketaylr:
> but the funnel report is easier to analyze

Yeah, this looks super interesting. Revenue will go through the roof! 💰

ksy36:
Thanks!

Issue #3029, opened 2019-10-24
Mobile viewport submit button layout visual issue


miketaylr:
Depending on the viewport width, this button looks wonky. The first is on my Pixel 3 phone, the second is resizing the browser. 

![Screenshot_20191024-104944](https://user-images.githubusercontent.com/67283/67502879-3e056880-f64c-11e9-81bf-6392676ca854.png)

![Screen Shot 2019-10-24 at 10 52 44 AM](https://user-images.githubusercontent.com/67283/67502977-760cab80-f64c-11e9-9c02-ce33dca0de29.png)

soniasingla:
@miketaylr this issue no longer exists :)

karlcow:
The issue still exists. When at this stage. It's impossible to scroll down and then access the content of the button.
<img width="776" alt="Capture d’écran 2020-07-20 à 13 39 34" src="https://user-images.githubusercontent.com/505230/87900501-804ebf80-ca8f-11ea-86ef-a82cac4fd05e.png">


The wrapping box seems to have a size determined by flexbox, but when the animation expand the box, the box is not being resized related to the content.

karlcow:
@magsout do you want to give it a shot?

ksy36:
This will most likely be fixed in https://github.com/webcompat/webcompat.com/pull/3404#discussion_r458434068

soniasingla:
This issue has been fixed in this [PR](https://github.com/webcompat/webcompat.com/pull/3404) :)

CC : @ksy36 @karlcow

karlcow:
Thanks!

PR #3404, opened 2020-07-18
Fixes #3387 - Prevent clipping of focus style on inputs

### LINKED ISSUE #3387: Add correct focus styles for "Briefly describe the issue"
"Briefly describe the issue" input field is missing left and right border:

Screen Shot 2020-07-14 at 1 05 53 AM

soniasingla:
This PR fixes issue #3387

### DIFF
--- a/webcompat/static/css/src/issue-wizard.css
+++ b/webcompat/static/css/src/issue-wizard.css
@@ -308,7 +308,6 @@
   animation-name: slideupandheight;
   max-height: 0;
   opacity: 0;
-  overflow: hidden;
   padding: 10px;
   width: 100%;
 }
--- a/webcompat/static/css/src/issue-wizard-animations.css
+++ b/webcompat/static/css/src/issue-wizard-animations.css
@@ -111,7 +111,7 @@
     opacity: 0;
   }
   100% {
-    max-height: 300px;
+    max-height: 340px;
     opacity: 1;
     padding-top: 45px;
   }


miketaylr:
(don't forget to ping someone for review :))

I've tested this locally and it seems to do what we want:

Before:
Screen Shot 2020-07-21 at 1 21 22 PM

After:
Screen Shot 2020-07-21 at 1 21 43 PM

So that's cool. :)

webcompat/static/css/src/issue-wizard.css
```diff
@@ -308,7 +308,7 @@
   animation-name: slideupandheight;
   max-height: 0;
   opacity: 0;
-  overflow: hidden;
+  overflow: visible;
```
miketaylr, line 311: Setting `overflow` to `visible` is the same as just removing the `overflow` property (since `visible` is the default).

Can we just remove it?

cc @ksy36 in case she has any idea why that might be a bad idea. In my testing, it didn't matter (at various viewport sizes).

ksy36: Seems to work good, except this button is sticking out now:

Screen Shot 2020-07-21 at 6 39 58 PM

Which is actually https://github.com/webcompat/webcompat.com/issues/3029

We can change `max-height` to be 340px in `slidedownusername` keyframe (instead of 300) and that will fix it @soniasingla  :)

soniasingla:
@miketaylr @ksy36 it's ready for the another review 🤗

ksy36:
Thanks @soniasingla 👍

PR #3406, opened 2020-07-20
Fix #3326 - Adds a static issue generator

### LINKED ISSUE #3326: Create module for rendering issues with comments as a static HTML file
It should be possible to create a method that takes an issue number as an argument and produces an entire issue page as static HTML.

karlcow:
This PR fixes issue #3326

## Proposed PR background

This creates the pipes for archiving issues as html


tools/archive/model.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""WebCompat Issue Model for archives."""
+
+from dataclasses import dataclass
```
miketaylr, line 9: Pretty cool, https://docs.python.org/3/library/dataclasses.html -- I hadn't learned about this yet.

karlcow: @miketaylr you may want to read this too. 
* https://realpython.com/python-data-classes/
* https://www.python.org/dev/peps/pep-0557/

miketaylr: Thanks, the realpython article is very nice.

karlcow:
Do we want lockbot comments to be part of the archived issue. Probably not neccessary. Probably there should be in the contextual message at the top an invitation for opening a new issue. 

> "This issue has been automatically locked since there has not been any recent activity after it was closed. Please open a new issue at https://webcompat.com/issues/new if you are experiencing a similar problem."

https://api.github.com/repos/webcompat/web-bugs/issues/40000/comments

miketaylr:
> Do we want lockbot comments to be part of the archived issue. Probably not neccessary.

Agree. It's not harmful for it to be there. But you could probably just have that contextual information in the "this page is archived" message.

tests/unit/test_tools_archive.py
```diff
@@ -5,17 +5,15 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 """Tests for webcompat.com archive tools.
 
-TODO: test for json to html
```
miketaylr: This a cool way to develop locally, and plan out commits ahead of time (something I struggle with a lot).

karlcow: I also starts to do things like this https://github.com/webcompat/webcompat.com/blob/034d7d7c68bf3246f3d4afd60a75839a8683d3bc/tests/unit/test_tools_archive.py#L140-L143

aka creating a test that is marked as skipped. 

```
(env) ~/code/webcompat.com-karlcow % COLUMNS=50 pytest -r s tests/unit/test_tools_archive.py
============== test session starts ===============
platform darwin -- Python 3.8.5, pytest-5.4.3, py-1.8.1, pluggy-0.13.1
rootdir: /Users/karl/code/webcompat.com-karlcow, inifile: pytest.ini
plugins: mock-3.2.0
collected 13 items                               

tests/unit/test_tools_archive.py ......... [ 69%]
.sss                                       [100%]

============ short test summary info =============
SKIPPED [1] tests/unit/test_tools_archive.py:136: TODO: HTTP ERROR log for comments fetching
SKIPPED [1] tests/unit/test_tools_archive.py:142: TODO: JSON for ALL comments
SKIPPED [1] tests/unit/test_tools_archive.py:148: TODO: Test comment as list of Comment
========= 10 passed, 3 skipped in 0.24s ==========

```

karlcow:
@miketaylr We are ready for a first review. 

- About the logic and architecture of the code. 
- And see if we have not forgotten anything fundamentals. 

For the templates design and layout, we can do a second run. 

Also probably once things have been reviewed I'll probably recommit things with cleaner commits, except if you prefer me to do that first and I will do it tomorrow . 

Thanks in advance for any comments.

tests/unit/test_tools_archive.py
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""Tests for webcompat.com archive tools.
+
+TODO: test that an issue can be on a different repo
```
miketaylr, line 8: I wonder what this TODO means... 🤔

tests/unit/test_tools_archive.py
```diff
@@ -0,0 +1,171 @@
...
+import requests
+from requests.exceptions import HTTPError
+
+from tools.archive import model
+
+
+def get_fixture(filename):
+    """Return the data fixture."""
+    current_root = pathlib.Path('.').resolve()
+    fixture_root = current_root / 'tests' / 'fixtures' / 'tools'
```
miketaylr, line 26: This looks magical -- it implements the `/` operator to create (navigate) child paths.

https://docs.python.org/3/library/pathlib.html#operators

tests/unit/test_tools_archive.py
```diff
@@ -0,0 +1,171 @@
...
+    fixture_path = fixture_root / filename
+    return fixture_path.read_text()
```
miketaylr, line 28: For my future reference: https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text

tests/unit/test_tools_archive.py
```diff
@@ -0,0 +1,171 @@
...
+import pathlib
+
+import pytest
+import requests
+from requests.exceptions import HTTPError
+
+from tools.archive import model
+
+
+def get_fixture(filename):
```
miketaylr, line 23: This is a cool method -- we could re-use this in other places where we do manually file traversal and file IO.

tools/archive/model.py
```diff
@@ -0,0 +1,193 @@
...
+            return comments
+
+
+@dataclass
+class ArchivedIssue(Issue):
+    """WebCompat Archived Issue Model.
+
+    This provide a couple of additional methods
+    to make it possible to archive the issue as html or any format
+    dimmed important. It inherits from a larger Issue model
```
miketaylr, line 127: nits: 
s/provide/provides/
s/dimmed/deemed/

tests/unit/test_tools_archive.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def test_comments_fetch_no_comments(issue_100):
+    """Test the comments fetching without comments available."""
+    # Before fetching the comments
+    issue_100.fetch_comments()
+    assert len(issue_100.comments) == 0
+
+
+def test_make_request(mocker):
```
miketaylr, line 138: Where does `mocker` come from?

Ah,

https://pypi.org/project/pytest-mock/
`This plugin provides a mocker fixture which is a thin-wrapper around the patching API provided by the mock package: Besides undoing the mocking automatically after the end of the test, it also provides other nice utilities such as spy and stub, and uses pytest introspection when comparing calls.`

Seems cool.

tests/unit/test_tools_archive.py
```diff
@@ -0,0 +1,171 @@
...
+def test_to_datetime():
+    """convert date string to datetime object."""
+    expected_date = datetime.datetime(2020, 8, 14, 12, 34, 56,
+                                      tzinfo=datetime.timezone.utc)
+    assert model.to_datetime('2020-08-14T12:34:56Z') == expected_date
+
+
+def test_issue_is_locked(issue_100):
+    """test if an issue is locked."""
+    assert issue_100.locked
```
miketaylr, line 171: Isn't this just testing that the fixture data has `"locked": true`? Is that a useful test (forgive me if I am missing something)? 

(I guess it does test that the `locked` property is set on the dataclass, but we don't really test all the props similarly (number, title, body, etc.))

tools/archive/templates/archive/issue.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Issue {{number}}</title>
+    <link href="https://webcompat.com/dist/webcompat.css" type="text/css" rel="stylesheet">
+    <style>table, td, th { border: 1px black solid; border-collapse: collapse;}</style>
```
miketaylr, line 8: I needed to remind myself if our CSP would allow this, https://github.com/webcompat/webcompat.com/blob/9b39a457bab1453ed9015a8cd5033c1e6837818f/webcompat/helpers.py#L546

And it does!

tools/archive/templates/archive/issue.html
```diff
@@ -0,0 +1,41 @@
...
+  </head>
+  <body id="body-webcompat">
```
miketaylr, line 10: nit: I know the template will evolve, but here we're doing a dash and in the other classes we're using an underscore. Probably good to be consistent with however we currently do CSS for the rest of the site.

tests/unit/test_tools_archive.py
karlcow: @miketaylr 

* https://realpython.com/python-pathlib/
* https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/
* https://rednafi.github.io/digressions/python/2020/04/13/python-pathlib.html

tools/archive/templates/archive/issue.html
karlcow: Maybe I'm missing what you meant, but I used `body-webcompat` because

https://github.com/webcompat/webcompat.com/blob/a0becc68ccc5f30ca67383ec2adb712639d502d6/webcompat/templates/layout.html#L16

karlcow: Ah no understood.
For the other classes in this template!

tests/unit/test_tools_archive.py
karlcow: Yes it's a remaining of a strategy that I was planning to use, but I have not implemented. Some of our closed issues are not locked.

tests/unit/test_tools_archive.py
karlcow: yup. It's added to the requirements-dev.txt

tests/unit/test_tools_archive.py
karlcow: yes we have plenty of functions in the project for testing that we should group in a utils module.

tests/unit/test_tools_archive.py
karlcow: I think it's probably because I was wondering if I would have a dependency inside this on hardcoding the repo path. As I was planning to initiliaze the issue from the URL but I changed my mind, and decided to initialize the issue from a payload. So the fetching of the issue itself is outside of the model and already versatile.

karlcow:
@miketaylr I addressed a couple of the comments. 
There will be a bit more work about it. 

Maybe the two biggest issues to discuss are:

1. Templates and content
2. Markdown conversion

tools/archive/templates/archive/issue.html
miketaylr: Yep, sorry for the confusion.

miketaylr:
> @miketaylr I addressed a couple of the comments.

Thanks. I'll get to reviewing the "meat" of the PR today, yesterday was just looking at tests and templates.

tools/archive/model.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+    def save(self, root_dir_path=None):
+        """Save an issue in a specific location.
+
+        root_dir: pathlib.Path
+            full path to where the directory will be saved.
+        return: pathlib.Path
+            it returns the full file location of the archived issue.
+            Probably /static/issue/issue-number.html for webcompat.com
```
miketaylr, line 170: Just a note for the future, I was wondering if we should hard-code this or make it an environment variable -- we can use nginx to serve the content from a different filesystem location or even an s3 bucket (or whatever, see https://cognition.happycog.com/article/deploying-static-websites-to-aws-s3-behind-an-nginx-proxy, etc).

This could be follow-up work, I don't think we should block on it -- up to you.

tests/unit/test_tools_archive.py
```diff
@@ -0,0 +1,167 @@
...
+    fake_r.status_code = 400
+    caplog.set_level(logging.WARNING)
+    model.Issue.recursive_fetch('https://example.org/')
+    assert 'Fetching comments failed' in caplog.text
+
+
+@pytest.mark.skip(reason="TODO: JSON for ALL comments")
+def test_recursive_fetch(mocker):
+    """Test that we receive a JSON of all comments."""
+    # This a recursive function which is probably hard to test.
```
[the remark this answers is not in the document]
miketaylr: (But having to mock out the link header and tying that to fixtures sounds not fun...)

tools/archive/model.py
```diff
@@ -0,0 +1,193 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""WebCompat Issue Model for archives."""
+
+from dataclasses import dataclass
+from dataclasses import field
+from datetime import datetime
+# from distutils.util import strtobool
```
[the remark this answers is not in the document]
karlcow: done.

tools/archive/model.py
```diff
@@ -0,0 +1,193 @@
...
+        """Fetch all comments."""
+        try:
+            r = requests.get(comments_url)
+            r.raise_for_status()
+        except HTTPError as err:
+            logging.warning(f'Fetching comments failed {err}')
+        else:
+            comments = r.json()
+            while 'next' in r.links.keys():
+                r = requests.get(r.links['next']['url'])
```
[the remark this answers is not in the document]
karlcow: This will not work, because this reset comments at each call, but it gave me ideas. :)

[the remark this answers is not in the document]
miketaylr: lol, you're right. writing code during internal meetings... not a great idea

miketaylr:
> Markdown conversion

I was curious what GitHub uses, and found https://github.blog/2017-03-14-a-formal-spec-for-github-markdown/

That references their fork of commonmark/cmark, which appears to still be actively maintained: https://github.com/github/cmark-gfm

There are a few Python bindings around, for example, https://github.com/theacodes/cmarkgfm. But... https://github.com/theacodes/cmarkgfm/issues/22 :/

tools/archive/model.py
```diff
@@ -102,20 +101,40 @@ def comments_as_list(self, comments_json):
...
+        """Fetch all comments.
+
+        * if 0 comments, response is empty
+        * if 1 page of comments, no links http header
+        * if 1+ page of comments
+          1st page has               next, last
+          next page has first, prev, next, last
+          last page has first, prev
+
+        so when there is no more next, we reached the last page
```
miketaylr, line 132: This docstring is a work of art. 👨‍🎨

miketaylr:
@karlcow is this ready for re-review?

karlcow:
> @karlcow is this ready for re-review?

yup

PR #3407, opened 2020-07-20
Fixes #3401 - Don't modify the file_path inside bust_cache

### LINKED ISSUE #3401: Cache busting hash broken
This happened between versions 34 and 35:`<script type="module" src="/dist/ga.js?missing_file"></script>`

miketaylr:
Actually fixes it this time. 

(Modifying the file_path var had bad consequences for some of the files, they could end up as 404)

Screen Shot 2020-07-20 at 5 00 06 PM

r? @ksy36

### DIFF
--- a/webcompat/templates/__init__.py
+++ b/webcompat/templates/__init__.py
@@ -30,9 +30,10 @@ def bust_cache(file_path):
     """
     if app.config['LOCALHOST']:
         return file_path
+    trimmed_file_path = file_path
     if file_path.startswith('/'):
-        file_path = file_path[1:]
-    absolute_path = os.path.join(STATIC_PATH, file_path)
+        trimmed_file_path = file_path[1:]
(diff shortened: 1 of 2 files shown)


webcompat/templates/__init__.py
```diff
@@ -31,8 +31,10 @@ def bust_cache(file_path):
     if app.config['LOCALHOST']:
         return file_path
     if file_path.startswith('/'):
-        file_path = file_path[1:]
-    absolute_path = os.path.join(STATIC_PATH, file_path)
+        trimmed_file_path = file_path[1:]
```
ksy36, line 34: not an expert, but maybe a bit shorter version:
```python
    trimmed_file_path = file_path
    if file_path.startswith('/'):
        trimmed_file_path = file_path[1:]

    absolute_path = os.path.join(STATIC_PATH, trimmed_file_path)

```

miketaylr: oh right, nicer. :)

Issue #3318, opened 2020-05-28
Remove AB testing initialization


karlcow:
This line means

https://github.com/webcompat/webcompat.com/blob/b3e5e1ffb58dc4cc372e4dad94eb390675a9c2d3/webcompat/views.py#L82

we are still going through AB code for each single request. 
We should fence it with an environment flag so that it is running only when the flag is on.

apsychogirl:
Hi, Can I take this up?

karlcow:
@apsychogirl you can try. I haven't tested that it would break anything, but please propose a PR if you want. That will give us code to think about it.

apsychogirl:
@karlcow would you please explain a bit how should I begin this?

karlcow:
@apsychogirl 

ok let's see

```
. env/bin/activate
git status
git fetch upstream
git merge upstream/master
git checkout -b 3318/1
npm install
npm run prod
```

Let's try to create first a flag just for the purpose of testing for now. 
We have a good example with 
https://github.com/webcompat/webcompat.com/blob/363a6eb9b3c61abaa9cff185d9a05ca895533f4b/config/environment.py#L73-L75

```python
AB_EXPERIMENT_ENABLED = strtobool(os.environ.get("AB_EXPERIMENT", "off"))
```
The `AB_EXPERIMENT_ENABLED` will be `True` or `False` with a default to `False`. see [strtobool](https://docs.python.org/3/distutils/apiref.html?highlight=strtobool#distutils.util.strtobool)

This gives us the ability to fence the code. Let's go to `views.py`

https://github.com/webcompat/webcompat.com/blob/363a6eb9b3c61abaa9cff185d9a05ca895533f4b/webcompat/views.py#L76-L83

we can now do something like 

```python
if AB_EXPERIMENT_ENABLED:
   ab_init(response)
```

Then we can start the app and see if we get any breakage

```
% flask run
 * Serving Flask app "webcompat:app" (lazy loading)
 * Environment: development
 * Debug mode: on
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
 * Restarting with stat
Writing logs to: /Users/karl/code/webcompat.com-karlcow/tmp
Statuses Initialization…
Milestones in memory
 * Debugger is active!
 * Debugger PIN: 905-956-639
Writing logs to: /Users/karl/code/webcompat.com-karlcow/tmp
Statuses Initialization…
Milestones in memory
```

[...]

PR #3410, opened 2020-07-21
Fixes #3408 - Remove active `exp` experiment (that was no longer active)

### LINKED ISSUE #3408: Remove `exp` experiment from AB_EXPERIMENTS config
>Cookie “exp” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

Screen Shot 2020-07-21 at 8 29 05 AM

We're not using them, so if we're setting them we should stop doing that.

miketaylr:
I also updated the test to preload the configuration -- it was relying on that configuration being active.

r? @karlcow

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -248,15 +248,8 @@ AB_DEFAULTS = {
 EXP_MAX_AGE = int(os.environ.get('EXP_MAX_AGE', 0))
 
 # AB testing config
-AB_EXPERIMENTS = {
-    'exp': {
-        'variations': {
-            'v1': get_variation('V1_VARIATION', AB_VARIATIONS, AB_DEFAULTS),
-            'v2': get_variation('V2_VARIATION', AB_VARIATIONS, AB_DEFAULTS),
-        },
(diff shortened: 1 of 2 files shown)


miketaylr:
(active, but inactive. what a paradox)

karlcow:
@miketaylr why are we trying to re-add this? We made the choice to remove it. 
see https://github.com/webcompat/webcompat.com/issues/3318

miketaylr:
> @miketaylr why are we trying to re-add this? We made the choice to remove it.

I don't think this PR adds anything, just removes the `exp` experiment. 

But, this commit could be landed in #3318 (I didn't realize that @apsychogirl was working on that) -- my one test fix here would fix the failure you mentioned in https://github.com/webcompat/webcompat.com/issues/3318#issuecomment-647911476.

So let's close here.

Issue #3177, opened 2020-02-05
Rename labeler webhook


miketaylr:
As of https://github.com/webcompat/webcompat.com/pull/3167, this webhook does a lot more than just add labels.

miketaylr:
(I think I was going to address this in some other PR, but am not currently working on it)

PR #3415, opened 2020-07-24
 Fixes #3291 - Make issue moderation slightly more efficient #3379 

### LINKED ISSUE #3291: Make issue moderation more efficient

miketaylr:
OK, this is a larger refactor, so we should probably review carefully so we don't introduce any regressions around moderation especially. 

<s>Still marking as a draft, because I'd like to sleep on it and look again next week.</s>

### DIFF
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -4,14 +4,11 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-"""Tests for our webhooks."""
+"""Tests for our webhooks: HTTP handling and helper methods."""
 
-from dataclasses import asdict
 import json
 import os
 import unittest
-from unittest.mock import ANY
-from unittest.mock import MagicMock
 from unittest.mock import patch
 
 import flask
@@ -23,7 +20,6 @@ import webcompat
 from webcompat.db import Site
 from webcompat.helpers import to_bytes
 from webcompat.webhooks import helpers
-from webcompat.webhooks.model import WebHookIssue
 
 
 # The key is being used for testing and computing the signature.
@@ -380,24 +376,6 @@ class TestWebhook(unittest.TestCase):
             self.assertTrue(webhook_request,
                             'X-GitHub-Event and X-Hub-Signature are correct')
 
-    def test_WebHookIssue_model(self):
-        """Test initializing a WebHookIssue model instance:
-
-        1. from a private new issue
-        2. from a milestoned issue
-        """
-        json_event, signature = event_data('private_issue_opened.json')
-        payload = json.loads(json_event)
-        issue = WebHookIssue.from_dict(payload)
-        expected = self.issue_info4
-        assert expected == asdict(issue)
-
-        json_event, signature = event_data('private_milestone_accepted.json')
-        payload = json.loads(json_event)
-        issue = WebHookIssue.from_dict(payload)
-        expected = self.issue_info2
-        assert expected == asdict(issue)
-
     def test_signature_check(self):
         """Test the signature check function for WebHooks."""
         payload = 'A body'
@@ -410,101 +388,6 @@ class TestWebhook(unittest.TestCase):
         self.assertFalse(helpers.signature_check(key, post_signature, payload))
         post_signature = 'sha1=wrong'
         self.assertFalse(helpers.signature_check(key, post_signature, payload))
-
-    @patch('webcompat.webhooks.model.make_request')
-    def test_WebHookIssue_close_private_issue(self, mock_mr):
-        """Test issue state that is sent to GitHub."""
-        mock_mr.return_value.status_code == 200
-        json_event, signature = event_data('private_issue_opened.json')
-        payload = json.loads(json_event)
-        issue = WebHookIssue.from_dict(payload)
-        issue.close_private_issue()
-        assert issue.state == 'closed'
-
-    @patch('webcompat.webhooks.model.make_request')
-    def test_WebHookIssue_close_private_issue_fails(self, mock_mr):
-        """Test issue state after a simulated GitHub failure."""
-        mock_mr.side_effect = HTTPError()
-        mock_mr.status_code = 400
-        json_event, signature = event_data('private_issue_opened.json')
-        payload = json.loads(json_event)
-        issue = WebHookIssue.from_dict(payload)
-        with pytest.raises(HTTPError):
-            issue.close_private_issue()
-        assert issue.state == 'open'
-
-    @patch('webcompat.webhooks.model.make_request')
-    def test_WebHookIssue_tag_as_public(self, mock_mr):
-        """Test tagging an issue as public."""
-        # A 200 response
-        json_event, signature = event_data('new_event_valid.json')
-        payload = json.loads(json_event)
-        mock_mr.return_value = MagicMock(status_code=200, spec=Response)
-        issue = WebHookIssue.from_dict(payload)
-        try:
-            issue.tag_as_public()
-            # did the milestone get set, and were labels updated?
-            method, uri, data = mock_mr.call_args[0]
-            assert issue.milestone == webcompat.app.config['STATUSES']['needstriage']['id']  # noqa
-            assert all(x in data['labels'] for x in
-                       ['browser-firefox', 'engine-gecko'])
-        except HTTPError:
-            pytest.fail("Unexpected requests.exceptions.HTTPError!")
-        # A 401 response
-        with patch.dict('webcompat.webhooks.helpers.app.config',
-                        {'OAUTH_TOKEN': ''}):
-            mock_mr.side_effect = HTTPError()
-            mock_mr.status_code = 401
-            with pytest.raises(HTTPError):
-                issue.tag_as_public()
-
-    @patch.object(webcompat.webhooks.model.WebHookIssue, 'tag_as_public')
-    def test_new_issue_right_repo(self, mock_proxy):
-        """Test that repository_url matches the CONFIG for public repo.
-
-        Success is:
-        payload: 'gracias amigos'
-        status: 200
-        content-type: text/plain
-        """
-        json_event, signature = event_data('new_event_valid.json')
-        headers = {
-            'X-GitHub-Event': 'issues',
-            'X-Hub-Signature': signature,
-        }
-        with webcompat.app.test_client() as c:
-            mock_proxy.return_value = MagicMock(status_code=200, spec=Response)
-            rv = c.post(
-                '/webhooks/labeler',
-                data=json_event,
-                headers=headers
-            )
-            self.assertEqual(rv.data, b'gracias, amigo.')
-            self.assertEqual(rv.status_code, 200)
-            self.assertEqual(rv.content_type, 'text/plain')
-
-    def test_new_issue_wrong_repo(self):
-        """Test when repository_url differs from the CONFIG for public repo.
-
-        In this case the error message is
-        payload: 'Wrong repository'
-        status: 403
-        content-type: text/plain.
-        """
-        json_event, signature = event_data('wrong_repo.json')
-        headers = {
-            'X-GitHub-Event': 'issues',
-            'X-Hub-Signature': signature,
-        }
-        with webcompat.app.test_client() as c:
-            rv = c.post(
-                '/webhooks/labeler',
-                data=json_event,
-                headers=headers
-            )
-            self.assertEqual(rv.data, b'Wrong repository')
-            self.assertEqual(rv.status_code, 403)
-            self.assertEqual(rv.content_type, 'text/plain')
 
     def test_repo_scope_public(self):
         """Test the public scope of the repository."""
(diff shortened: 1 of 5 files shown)


webcompat/webhooks/model.py
```diff
@@ -0,0 +1,178 @@
...
+        {
+            "milestone": 2,
+            "labels": ['Label1', 'Label2']
+        }
+        """
+        # Grabs the labels already set so they will not be erased
+        # Gets the labels from the body
+        labels = get_issue_labels(self.body)
+        labels.extend(self.original_labels)
+        milestone = app.config['STATUSES']['needstriage']['id']
```
miketaylr, line 141: This should be `self.milestone`.

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,178 @@
...
+        # Preparing the proxy request to the public repo
+        path = f'repos/{PUBLIC_REPO}/{self.number}'
+        payload_request = {'labels': labels, 'milestone': milestone}
```
miketaylr, line 144: This should be `payload_request = {'labels': labels, 'milestone': self.milestone}`

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,178 @@
...
+    @staticmethod
+    def get_public_issue_number(public_url):
+        """Extract the issue number from the public url."""
+        public_number = public_url.strip().rsplit('/', 1)[1]
+        return public_number
+
+    def __post_init__(self, payload):
+        """Build up the model from the initial WebHook payload."""
+        # Extract the title and the body
+        print(type(payload))
```
miketaylr, line 157: Whoops 🤡

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""WebCompat Issue Model for webhooks."""
+
+from typing import Any, Dict, List
+from dataclasses import dataclass, field, InitVar
```
miketaylr, line 10: I guess I should fix the ordering here.

miketaylr:
r? @karlcow 

No rush on the review, within a week would be awesome.

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,177 @@
...
+from webcompat.webhooks.helpers import prepare_rejected_issue
+
+PUBLIC_REPO = app.config['ISSUES_REPO_URI']
+PRIVATE_REPO = app.config['PRIVATE_REPO_URI']
+
+
+@dataclass
+class WebHookIssue:
+    """WebCompat Issue Model for WebHook consumption"""
+    payload: InitVar[Dict[str, Any]]
```
miketaylr, line 25: This is a pretty lazy type annotation for (parsed) JSON, we're trusting that GitHub doesn't send us scary thing.

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,177 @@
...
+        """Mark the private issue as closed."""
+        path = f'repos/{PRIVATE_REPO}/{self.number}'
+        self.state = 'closed'
+        make_request('patch', path, {'state': self.state})
+
+    def comment_public_uri(self):
+        """Publish a comment on the private issue with the public uri."""
+        payload = {'body': self.prepare_public_comment()}
+        # Preparing the proxy request
+        path = f'repos/{PRIVATE_REPO}/{number}/comments'
```
miketaylr, line 48: Oops! This should be `self.number`, corresponding to https://github.com/webcompat/webcompat.com/pull/3415/files#diff-53c7dc964488671110b629cf6885fb3eL435

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,177 @@
...
+            'labels': labels,
+            'title': self.title,
+            'body': self.body
+        }
+        return payload_request
+
+    def prepare_public_comment(self):
+        """Build the comment for the public repo."""
+        # issue number on private repo
+        number = self.number
```
miketaylr, line 102: This is unused.

karlcow:
@miketaylr btw as you are at it, probably we could address #3177

webcompat/webhooks/model.py
karlcow: @rik proposed something else for initialization.

https://gist.github.com/rik/d7d195ee5cc11b16d0d951312905eee3#gistcomment-3396628

which is nice too with less code.

karlcow: (for context)
he proposed it because I talked about `InitVar` on 
https://www.la-grange.net/2020/07/29/plantez

as I was discovering them during the review today.

miketaylr: I see, so the idea is to add a static `from_dict`, which will do the initialization.

I like this solution, it seems less confusing to read (especially for newcomers to dataclasses like me) -- also it means we can ditch the `field(init=False)` stuff I think.

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,175 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""WebCompat Issue Model for webhooks."""
+
+from dataclasses import dataclass, field, InitVar
```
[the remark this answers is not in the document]
miketaylr: Ah right. :)

webcompat/webhooks/__init__.py
```diff
@@ -33,9 +33,9 @@ def hooklistener():
     event_type = request.headers.get('X-GitHub-Event')
     # Treating events related to issues
     if event_type == 'issues':
-        issue_info = get_issue_info(payload)
+        webhook_issue = WebHookIssue(payload)
```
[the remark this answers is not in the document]
miketaylr: Hm, that's a good question. It may have to do with the evolution of the design in my local commits.

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,175 @@
...
+    title: str = field(init=False)
+    original_labels: List[str] = field(init=False)
+    milestone: str = ''
+    milestoned_with: str = ''
+
+    def close_private_issue(self):
+        """Mark the private issue as closed."""
+        path = f'repos/{PRIVATE_REPO}/{self.number}'
+        self.state = 'closed'
+        make_request('patch', path, {'state': self.state})
```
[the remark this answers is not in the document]
miketaylr: > if the request fails, then we have the issue with a closed state when it is still open remotely

Ah, I did not consider this because GitHub *never* fails. 😈

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def comment_public_uri(self):
+        """Publish a comment on the private issue with the public uri."""
+        payload = {'body': self.prepare_public_comment()}
```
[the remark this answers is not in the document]
miketaylr: Forgive me for offending your aesthetic sensibilities, lol. I don't have a strong feeling, I can change it. 

It's maybe easier to scan and understand with your suggested way.

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,175 @@
...
+        Milestone should be already set on needstriage
+
+        we get the destination through the public_url
+        """
+        payload_request = self.prepare_accepted_issue()
+        public_number = WebHookIssue.get_public_issue_number(self.public_url)
+        # Preparing the proxy request
+        path = f'repos/{PUBLIC_REPO}/{public_number}'
+        proxy_response = make_request('patch', path, payload_request)
+        return proxy_response
```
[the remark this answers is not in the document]
miketaylr: very good point, thanks.

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,175 @@
...
+
+    @staticmethod
+    def get_public_issue_number(public_url):
+        """Extract the issue number from the public url."""
+        public_number = public_url.strip().rsplit('/', 1)[1]
+        return public_number
+
+    def __post_init__(self, payload):
```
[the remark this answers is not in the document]
miketaylr: I think there's a good argument to be made that it's easier to scan if you can look at init() or post_init() and their side-effects, without having to read through all the class methods.

webcompat/webhooks/model.py
miketaylr: > also it means we can ditch the `field(init=False)` stuff I think

(no, we can't, because we need to still do the data manipulation in `__post_init__()`)

miketaylr: > (no, we can't, because we need to still do the data manipulation in `__post_init__()`)

I'm a little slow... we can do the data manipulation in the classmethod. :) So, no need for field(init=False). :))

webcompat/webhooks/__init__.py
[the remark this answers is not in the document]
miketaylr: I will resolve this, since we now use `WebHookIssue.from_dict(payload)`.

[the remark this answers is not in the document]
miketaylr: It's also possible to just do `WebHookIssue(action=payload.get('action'), body=issue_body, domain=domain, number=issue.get('number'), public_url=public_url, repository_url=issue.get('repository_url'), state=issue.get('state'), title=full_title, original_labels=original_labels, milestone=milestone, milestoned_with=milestoned_with)` (assuming the right variables are in scope).

But that offends *my* aesthetic sensibilities.

webcompat/webhooks/model.py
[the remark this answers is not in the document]
miketaylr: (just to be clear, this reads as super snotty and rude, but I write it with love ❤️ )

miketaylr:
> @miketaylr btw as you are at it, probably we could address #3177

Yeah, might as well. I'll steal that. We just need to remember to update the public and private bugs repos to point to the new thing.

webcompat/webhooks/__init__.py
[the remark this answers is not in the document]
karlcow: yes that makes it a bit verbose :)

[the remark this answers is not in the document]
rik: > It's also possible to just do `WebHookIssue(action=payload.get('action'), body=issue_body, domain=domain, number=issue.get('number'), public_url=public_url, repository_url=issue.get('repository_url'), state=issue.get('state'), title=full_title, original_labels=original_labels, milestone=milestone, milestoned_with=milestoned_with)` (assuming the right variables are in scope).
> 
> But that offends _my_ aesthetic sensibilities.

I wouldn't recommend using this constructor in the codebase but it can be useful in tests. If you set default values for the fields in the dataclass declaration, this pattern can allow you to express the intention of a test more clearly and succinctly by only providing the fields necessary for a test, eg: `WebHookIssue(public_url='foo')`. (not that I'd recommend doing that in this case)

I think this pattern is generally called "alternate constructor" if you want to look it up.

karlcow:
I have the feeling we will have a very similar model for Issues. And Probably it might be worth thinking when you have finished yours on having an `Issue` model and a `WebHookIssue(Issue)` model, the same way I will get an `ArchiveIssue(Issue)` model.

But let's not block this. I will eventually refactor, when we get to it.

karlcow:
> I have the feeling we will have a very similar model for Issues. And Probably it might be worth thinking when you have finished yours on having an `Issue` model and a `WebHookIssue(Issue)` model, the same way I will get an `ArchiveIssue(Issue)` model.

Starting to do that here, but will be moved.

https://github.com/webcompat/webcompat.com/blob/fa6643004b61dda388251f3a1c42a4b10d845af9/tools/archive/model.py#L81

webcompat/webhooks/helpers.py
```diff
@@ -198,44 +160,14 @@ def get_issue_labels(issue_body):
...
-    proxy_response = proxy_request(
-        'patch',
-        path,
-        headers=headers,
-        data=json.dumps(payload_request))
-    return proxy_response
-
-
-def process_issue_action(issue_info):
+def process_issue_action(issue):
```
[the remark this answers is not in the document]
davidbgk: The `raise_for_status()` method can be handy here, available both in [requests](https://2.python-requests.org/en/master/api/#requests.Response.raise_for_status) and [httpx](https://www.python-httpx.org/quickstart/#response-status-codes).

miketaylr:
> And Probably it might be worth thinking when you have finished yours on having an `Issue` model and a `WebHookIssue(Issue)` model, the same way I will get an `ArchiveIssue(Issue)` model.

Yep, this seems nice to do next. And then use the Issue model during issue creation, etc. 👍

webcompat/webhooks/__init__.py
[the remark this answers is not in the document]
miketaylr: Yeah, that does seem nice to be able to be very explicit about intent in tests.

webcompat/webhooks/model.py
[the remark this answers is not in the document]
miketaylr: OK, I think I've fixed all the instances where we return a `proxy_response` and do nothing with it, replacing with `try, except HTTPError, else`

miketaylr:
I think I have 1 more test to write, then ready for re-review.

webcompat/webhooks/model.py
```diff
@@ -0,0 +1,185 @@
...
+        issue = payload.get('issue')
+        full_title = issue.get('title', 'Weird_Title - Inspect')
+        labels = issue.get('labels', [])
+        issue_body = issue.get('body')
+        domain = full_title.partition(' ')[0]
+        public_url = extract_metadata(issue_body).get('public_url', '').strip()
+        original_labels = [label['name'] for label in labels]
+        # webhook with a milestone already set
+        milestone = ''
+        if issue.get('milestone'):
```
davidbgk, line 55: If the `issue` key can be missing from `payload` you'll get an `AttributeError: 'NoneType' object has no attribute 'get'` at that point. That is probably better to break at line 46 with a direct access `payload['issue']` and the more explicit error: `KeyError: 'issue'`

miketaylr:
r? @karlcow - last 5 commits ready for review!

I sort of made a mess of the tests, I'd like to circle back and simplify those in the future: https://github.com/webcompat/webcompat.com/issues/3436

tests/unit/test_webhook.py
```diff
@@ -489,8 +524,10 @@ def test_repo_scope_unknown(self):
         actual = helpers.repo_scope(url)
         self.assertEqual(expected, actual)
 
-    @patch('webcompat.webhooks.helpers.private_issue_moderation')
-    def test_patch_acceptable_issue(self, mock_proxy):
+    @patch.object(webcompat.webhooks.model.WebHookIssue,
+                  'moderate_private_issue')
+    @patch.object(webcompat.webhooks.model.WebHookIssue, 'close_private_issue')
+    def test_patch_acceptable_issue(self, mock_close, mock_moderate):
```
karlcow, line 530: This probably doesn't test what we want.
 
Probably just need to rewrite what is expected from the test then cover it.

tests/unit/test_webhook.py
```diff
@@ -513,11 +552,13 @@ def test_patch_acceptable_issue(self, mock_proxy):
...
-    @patch('webcompat.webhooks.helpers.private_issue_moderation')
-    def test_patch_acceptable_issue_problem(self, mock_proxy):
+    @patch.object(webcompat.webhooks.model.WebHookIssue,
+                  'moderate_private_issue')
+    @patch.object(webcompat.webhooks.model.WebHookIssue, 'close_private_issue')
+    def test_patch_acceptable_issue_problem(self, mock_close, mock_moderate):
         """Test for accepted issues failed.
 
-        payload: 'ooops'
+        payload: 'oops'
```
karlcow, line 561: hehe.

tests/unit/test_webhook.py
```diff
@@ -527,18 +568,51 @@ def test_patch_acceptable_issue_problem(self, mock_proxy):
...
+            assert rv.data == b'oops'
+            assert rv.status_code == 400
+            assert rv.content_type == 'text/plain'
 
-    @patch('webcompat.webhooks.helpers.private_issue_rejected')
-    def test_patch_not_acceptable_issue(self, mock_proxy):
+    @patch.object(webcompat.webhooks.model.WebHookIssue,
+                  'reject_private_issue')
+    @patch.object(webcompat.webhooks.model.WebHookIssue, 'close_private_issue')
+    def test_patch_acceptable_issue_problem(self, mock_close, mock_reject):
```
karlcow, line 586: Probably not testing what we want. 

There are two parts for these tests.

1. Testing the functions which are in the modules. (easy)
2. Testing the HTTP interactions outcome (aka the thing in the `__init__` which receives the HTTP requests `c.post`)

Probably everything else doesn't really need a POST request to `/webhooks/labeler`. It's more what is the input function and what is the output of the function.

Let's take `tag_as_public`.

1. This is a function. It creates a payload request and answer a response through make_request. 
    - The patch needs to be only about `make_requests` 
    - the tests needs to cover do we send the right payload.
    - The HTTP response variability does NOT really affect the outcome at this level.
2. `tag_as_public` as used in `process_issue_action` 
    - The patch needs to be only about `tag_as_public`
    - the test needs to cover the different HTTP response.
    - the HTTP responses is what create variations here on process_issue_action. So basically the input is an `issue`, this issue depending on its inner criteria will trigger `tag_as_public` which can have two different responses.
   

When thought like this for each part it becomes a lot easier to create all tests. The `c.post`  should only be for testing the route and this happens only in __init__.py. Basically you inherited from **bad tests from me** 😝 

And trying to convert them to the new model doesn't fix the bad.

webcompat/webhooks/helpers.py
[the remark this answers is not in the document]
karlcow: note that we are still sending HTTP responses here, and I'm not sure it's necessary as nothing is really reading them. We could log the success (we do already for the failures).

webcompat/webhooks/model.py
karlcow: I guess the test here would be what is happening if we get a payload which is not the expected one or no payload because GitHub went down or something on the network broke. 

this is tied to https://github.com/webcompat/webcompat.com/blob/77c102721ce05d74c9372a35b14884b62ab45623/webcompat/webhooks/__init__.py#L33-L37

we know already that's a webhook issue event payload. So that would be really the case where `'X-GitHub-Event'` and the payload contradict themselves. so a try would just send back `ooops` instead of a 500.

miketaylr:
Thanks for the comments!

However, it does seem useful to know when `assert 1 == 1` fails (kidding).

tests/unit/test_webhook.py
```diff
@@ -401,26 +411,51 @@ def test_signature_check(self):
...
-    def test_tag_new_public_issue(self):
+    @patch('webcompat.helpers.requests.patch')
+    def test_WebHookIssue_close_private_issue(self, mock_patch):
+        """Test issue state that is sent to GitHub."""
+        with webcompat.app.test_request_context():
+            mock_patch.return_value.status_code == 200
+            json_event, signature = event_data('private_issue_opened.json')
+            payload = json.loads(json_event)
+            issue = WebHookIssue.from_dict(payload)
+            issue.close_private_issue()
```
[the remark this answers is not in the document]
miketaylr: Thanks, this was a useful explanation!

tests/unit/test_webhook.py
```diff
@@ -401,26 +411,51 @@ def test_signature_check(self):
         post_signature = 'sha1=wrong'
         self.assertFalse(helpers.signature_check(key, post_signature, payload))
 
-    def test_tag_new_public_issue(self):
+    @patch('webcompat.helpers.requests.patch')
+    def test_WebHookIssue_close_private_issue(self, mock_patch):
+        """Test issue state that is sent to GitHub."""
+        with webcompat.app.test_request_context():
```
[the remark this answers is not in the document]
miketaylr: Yep, not needed, thanks.

tests/unit/test_webhook.py
```diff
@@ -401,26 +411,51 @@ def test_signature_check(self):
...
         payload = json.loads(json_event)
-        issue_info = helpers.get_issue_info(payload)
-        with patch('webcompat.webhooks.helpers.proxy_request') as proxy:
+        issue = WebHookIssue.from_dict(payload)
+        with patch.object(webcompat.webhooks.model.WebHookIssue,
+                          'tag_as_public') as proxy:
             proxy.return_value.status_code = 200
-            response = helpers.tag_new_public_issue(issue_info)
+            response = issue.tag_as_public()
             self.assertEqual(response.status_code, 200)
```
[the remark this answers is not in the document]
miketaylr: Ah right. And this needs to be re-written entirely, because these methods don't return HTTP responses anymore, but instead throw `HTTPError` if they fail.

tests/unit/test_webhook.py
miketaylr: After fixing this typo, I realized I had to fix all the tests -- protip: never fix typos.

tests/unit/test_webhook.py
miketaylr: Can you provide more info?

These tests are checking the HTTP responses that `process_issue_action` returns, so probably they need to be renamed. I can do that.

Screen Shot 2020-08-04 at 4 44 27 PM

For this one, we need to mock `moderate_private_issue` and tell it to return a 200. And since it doesn't throw, it will also call `close_private_issue`. If we don't mock that out, it will complain at us. 

I think we need to add a test for `moderate_private_issue`, however.

miketaylr: OK, actually, you answered my question a few comments down!

tests/unit/test_webhook.py
miketaylr: Mayyyyybe deleting those tests (or simplifying them) can be what happens in #3436 (or we just close that one).

tests/unit/test_webhook.py
```diff
@@ -401,26 +411,54 @@ def test_signature_check(self):
...
+                       ['browser-firefox', 'engine-gecko'])
+        except HTTPError:
+            pytest.fail("Unexpected requests.exceptions.HTTPError!")
+        # A 401 response
+        with patch.dict('webcompat.webhooks.helpers.app.config',
+                        {'OAUTH_TOKEN': ''}):
+            mock_mr.side_effect = HTTPError()
+            mock_mr.status_code = 401
+            with pytest.raises(HTTPError):
+                issue.tag_as_public()
```
karlcow, line 459: @miketaylr 

I was puzzled by this test because it seemed very complicated. 

When I do not understand I try to go back to the code itself. So let's check what we need to test `tag_as_public`

https://github.com/webcompat/webcompat.com/blob/98412a9a5239724feed3c73a0d707c527a04493d/webcompat/webhooks/model.py#L152-L178

1. it doesn't take any parameters so that's one thing less to test on the input side.
2. it has variability inside the function such as labels and milestones. (All of these are tested elsewhere)
3. it creates a `payload_request` which should probably be named `request_payload` (nit) and a `path` 
4. It returns an HTTP response

So there are only 6 tests for this:

1. assert that the request payload is a `dict`
2. assert that the request payload keys are `label` and `milestone`
3. assert that the `path` is a `str` and exists
4. (Optional) assert that the `path` is the string `api.github… etc.`
5. assert that the call to `make_request` is made with `patch`
6. assert that the return is an HTTP response.

and that's all. there's nothing else to test here. 

ok now let's go to `process_issue_action` in the case of 

**Test Condition Input**: `if issue.action == 'opened' and scope == 'public':`

1. Test 1 (Success)
    - when `issue.tag_as_public()` returns a 200 Response (mocking)
        -  `assert process_issue_action(issue).data == 'gracias, amigo.'`
        - `assert process_issue_action(issue).status_code == '200'`

2. Test 2 (Failure)
    - when `issue.tag_as_public()` returns an HTTPError (mocking)
        - `assert 'public:opened labels failed' in caplog.text`
        - `assert process_issue_action(issue).data == 'oops'`
        - `assert process_issue_action(issue).status_code == 400`

And **that's all** to be tested :) really

Now the `r.raise_for_status()` is **not** part of `process_issue_action`, so it has to be tested only at the level of `make_request` tests.

With the same logic. What are the input possible, what are the outcomes possible for this functions. Probably mocking `proxy_request`  just for the tests about `make_request`

Doing like this will make everything clearer more isolated and testing one thing at a time with less complicated tests. Everything will be more readable. :) 

* Conditions, Inputs, Outputs for each functions :) 


## about testing logging: [caplog](https://docs.pytest.org/en/latest/logging.html#caplog-fixture)

(Don't feel bad, I just discovered it now when I was trying to find out how to do it. )

```python
import logging
LOGGER = logging.getLogger(__name__)

def test_blahblah(caplog):
     caplog.set_level(logging.INFO):
     # Here all the code to set the right condition such as scope and issue value
     process_issue_action(issue)
     assert 'The log output…' in caplog.text
```

karlcow: @miketaylr Here a simple example, I made today:

```python
import logging
LOGGER = logging.getLogger(__name__)


def see(animal):
    if animal == 'snake':
        logging.warning('danger!')
    else:
        logging.info('everything is fine')
    return f'Animal : {animal}'


logging.basicConfig(level=logging.DEBUG)
see('snake')
see('grasshoper')


def test_voir(caplog):
    caplog.set_level(logging.INFO)
    voir('snake')
    assert 'danger' in caplog.text
    voir('grasshoper')
    assert 'everything is fine' in caplog.text
```

Then

```
% python test_deleteme.py
WARNING:root:dangereux
INFO:root:tout va bien
```

And 

```
% COLUMNS=40 pytest test_deleteme.py
========= test session starts ==========
platform darwin -- Python 3.8.5, pytest-5.4.3, py-1.8.1, pluggy-0.13.1
rootdir: /Users/karl/code/webcompat.com-karlcow, inifile: pytest.ini
plugins: mock-3.2.0
collected 1 item                       

test_deleteme.py .               [100%]

========== 1 passed in 0.03s ===========
```

miketaylr: ah! so we can actually test that we're logging things too. very nice.

miketaylr: Thanks for the "think out loud" session here. 

One slight correction - `tag_as_public` doesn't return a HTTP response, it doesn't return anything. Maybe we want it to later? But we're relying on HTTPErrors being thrown as a way to react, rather than check the rv. `make_request` does return an HTTP response, but it isn't returned outside of `tag_as_public`. **But**, that doesn't change what you are saying in terms of how to test `process_action`.

karlcow: > One slight correction - `tag_as_public` doesn't return a HTTP response, it doesn't return anything. 

correct. Thanks.

miketaylr:
OK, getting closer.

TODO:

* [ ] address https://github.com/webcompat/webcompat.com/pull/3415#discussion_r464662594
* [ ] test logs that come from HTTPErrors 
* [ ] have one more pass at the model methods and ensure we're testing the right things (i suspect there's a few tests left to write).

webcompat/webhooks/model.py
miketaylr: Re-reading, we can add some error recovery if `issue` is missing from `payload`, but yes, it would imply that GitHub is totally broken.

We do need to check for `milestone` on the `issue`, however, because it's possible for a payload to not have a `milestone` property on `issue` (or at the root level, in the instanced of a `milestoned` webhook type.

miketaylr: OK, I've made the change in https://github.com/webcompat/webcompat.com/pull/3415/commits/3e5eadcd330400487e778093345bc68cfee6cb22. I wrote a test, but it doesn't provide much value -- it just asserts that if you pass in garbage, you get a KeyError. It's possible that GitHub does weird stuff one day, but we'll have bigger problems if so (everything will fail, not just this webhook). But the KeyError will provide better clues in logs, so that's nice.

miketaylr:
OK, tests are green. r? @karlcow

tests/unit/test_webhook.py
```diff
@@ -4,26 +4,40 @@
...
 
+# Some expected responses as tuples
+accepted = ('Moderated issue accepted', 200, {'Content-Type': 'text/plain'})
+rejected = ('Moderated issue rejected', 200, {'Content-Type': 'text/plain'})
+boring = ('Not an interesting hook', 403, {'Content-Type': 'text/plain'})
+gracias = ('gracias, amigo.', 200, {'Content-Type': 'text/plain'})
+wrong_repo = ('Wrong repository', 403, {'Content-Type': 'text/plain'})
+oops = ('oops', 400, {'Content-Type': 'text/plain'})
+comment_added = ('public url added', 200, {'Content-Type': 'text/plain'})
+
```
karlcow, line 40: super cool and clear ❤️

tests/unit/test_webhook.py
```diff
@@ -722,67 +435,132 @@ def test_prepare_rejected_issue(self):
...
+         'private:moving to public failed', 'moderate_private_issue'),
+        ('private_issue_no_source.json', 'comment failed',
+         'comment_public_uri'),
+        ('new_event_valid.json', 'public:opened labels failed',
+         'tag_as_public'),
+        ('private_milestone_closed.json',
+         'public rejection failed', 'reject_private_issue')
+    ]
+    for scenario in scenarios:
+        issue_payload, expected_log, method = scenario
```
karlcow, line 556: This is **good**. ❤️

tests/unit/test_webhook_model.py
```diff
@@ -0,0 +1,195 @@
...
+def test_close_private_issue_fails(mock_mr):
+    """Test issue state after a simulated GitHub failure."""
+    mock_mr.side_effect = HTTPError()
+    mock_mr.status_code = 400
+    json_event, signature = event_data('private_issue_opened.json')
+    payload = json.loads(json_event)
+    issue = WebHookIssue.from_dict(payload)
+    with pytest.raises(HTTPError):
+        issue.close_private_issue()
+    assert issue.state == 'open'
```
karlcow, line 85: ❤️

karlcow:
Awesome work!

PR #3416, opened 2020-07-27
Fixes #3392 - Refactor css after original form removal

### LINKED ISSUE #3392: Remove HTML and CSS code associated with the old form

ksy36:
In this PR I've deleted unused css and also refactored styles to make wizard input styles be the default input styles for the site.


webcompat/static/css/src/form.css
```diff
@@ -11,52 +11,13 @@
...
   padding: 0;
   position: relative;
 }
 
-.form-button-multiple {
-  align-items: center;
-  display: flex;
-  flex-wrap: wrap;
-  justify-content: space-between;
-  margin-left: -10px;
```
ksy36, line 35: These classes are no longer in use after deleting form.html

miketaylr:
It would be good for @magsout to take a look when he's back from vacation, but no need to block on that. Guillaume, if you have feedback after the fact, please file an issue!

ksy36:
thanks!

PR #3424, opened 2020-07-28
Fixes #3395 - Change copy after filing an issue with "Report site issue" reporter

### LINKED ISSUE #3395: Incorrect copy after filing an issue with "Report site issue" reporter
It seems that there is an incorrect copy on the wizard page after a report is made through "Report site issue" reporter (postMessage).

Expected:
```
Thank you for filing a web compatibility issue.
Do we have the correct site?
```
Actual:
```
Please enter the URL where you encountered the web compatibility issue. 
```

Note that the correct copy is shown when a report is made with get params:

Screen Shot 2020-07-15 at 11 18 23 PM

ksy36:


webcompat/static/js/lib/wizard/steps/submit.js
```diff
@@ -53,15 +53,14 @@ const submitForm = function () {
...
-  sendAnalyticsEvent("success", "end");
-
   formEl.submit();
   dfd.resolve();
   return dfd.promise();
 };
 
 const onFormSubmit = (event) => {
   event.preventDefault();
+  sendAnalyticsEvent("success", "end");
```
ksy36, line 63: I've moved analytics success event to happen a bit earlier, since I've noticed that it's not being tracked due to the url change after form submit. 
(this is not related to the changes addressed in this PR, but it's probably not worth creating a separate issue)

PR #3426, opened 2020-07-29
fix #3425 - Adds a pytest.ini

### LINKED ISSUE #3425: Speed up pytest command
I'm lazy, so I usually invoke pytest via `pytest` from the project root. This can take forever, likely due to it scanning all directories for tests.

I did some digging and found:

https://docs.pytest.org/en/latest/reference.html#confval-testpaths

So if I do `pytest -o "testpaths=tests"`, the tests start immediately. 

@karlcow do you think we should add a `pytest.ini` to the project root to specify this? Or do you have some other tricks you use?

karlcow:
This PR fixes issue #3425

## Proposed PR background

without this pytest takes a lot of times to scan and discover the tests.
This will speed up when we just type pytest.
It might also accelerate CircleCi.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,3 @@
+# configuration file for pytest
+[pytest]
+testpaths = tests/unit


karlcow:
Et voilà @miketaylr 
Freshly served!

PR #3428, opened 2020-07-31
Fixes #3427 - Remove min char requirement from description field

### LINKED ISSUE #3427: Remove min character rule from the description field
Since we're seeing a 26.6% drop at the description field, lets remove the min character rule and make the field optional.  

Screen Shot 2020-07-31 at 11 52 10 AM

ksy36:
I've removed min char requirement, made the description optional and changed the copy:

Screen Shot 2020-07-31 at 6 08 25 PM

### DIFF
--- a/webcompat/templates/issue-wizard-form.html
+++ b/webcompat/templates/issue-wizard-form.html
@@ -230,32 +230,31 @@
     </div>
   </div>
 
-  <div class="step-container step-description col">
+  <div class="step-container step-description col" data-min-char="false">
     <div class="row mobile-col">
       <div class="input-control device-data-hero half">
         <img class="protruding-img" src="{{ url_for('static', filename='img/svg/describe-issue.svg') }}" title="Describe issue" alt="Describe issue" />
         <div class="description-text">
-          OK, now the vital part! <br>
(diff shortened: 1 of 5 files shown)


ksy36:
Also any feedback on the copy is appreciated :) We can change it to something else

webcompat/templates/issue-wizard-form.html
```diff
@@ -230,32 +230,31 @@ <h3 class="page-heading">Filing a web compatibility issue</h3>
     </div>
   </div>
 
-  <div class="step-container step-description col">
+  <div class="step-container step-description col" data-min-char="false">
```
miketaylr, line 233: Cool, so we can toggle it on or off based on this.

I assumed we were going to do some A/B testing based on the existing AB test stuff we have, this approach would be compatible with that -- you set the value in the template based on active experiment.

ksy36:
Thanks! Yeah, the plan is to compare previous two weeks with the week when this change was shipped. I was focusing on reducing drop off at this step, since the rate is pretty high. I think we can still use A/B testing for other steps, where the reason for drop off is not as clear.

PR #3446, opened 2020-08-07
Fixes #3391 - Add ability to close modals with Esc key

### LINKED ISSUE #3391: Remove JS code associated with the old form

miketaylr:
This lets users close the modal with Esc, and adds a test.

r? @ksy36

### DIFF
--- a/webcompat/static/js/lib/issue-wizard-popup.js
+++ b/webcompat/static/js/lib/issue-wizard-popup.js
@@ -7,35 +7,49 @@ import { sendAnalyticsEvent } from "./wizard/analytics.js";
 
 function Popup() {
   this.init = function () {
+    this.closeButtons = document.querySelectorAll(".popup-modal__close");
+    this.modalTriggers = document.querySelectorAll(".popup-trigger");
+    this.overlay = document.querySelector(".popup-overlay");
+
     this.setUpEvents();
   };
 
   this.setUpEvents = function () {
-    this.popupHandler();
+    this.modalTriggers.forEach(
+      function (trigger) {
+        trigger.addEventListener("click", this.openModal.bind(this));
+      }.bind(this)
+    );
+
+    this.closeButtons.forEach(
+      function (btn) {
+        btn.addEventListener("click", this.closeModal.bind(this));
+      }.bind(this)
+    );
   };
 
-  this.popupHandler = function () {
-    var modalTriggers = document.querySelectorAll(".popup-trigger");
-    var overlay = document.querySelector(".popup-overlay");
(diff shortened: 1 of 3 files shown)


webcompat/static/js/lib/issue-wizard-popup.js
```diff
@@ -3,39 +3,58 @@
...
   };
 
-  this.popupHandler = function () {
-    var modalTriggers = document.querySelectorAll(".popup-trigger");
-    var overlay = document.querySelector(".popup-overlay");
-    var closeButtons = document.querySelectorAll(".popup-modal__close");
-    modalTriggers.forEach(function (trigger) {
-      trigger.addEventListener("click", function (e) {
-        e.preventDefault();
-        sendAnalyticsEvent("whatIsCompat", "click");
```
miketaylr, line 24: It turns out we sent this event when opening the "learn more" modal as well. 🤷

ksy36: oops good catch!

webcompat/static/js/lib/issue-wizard-popup.js
```diff
@@ -3,39 +3,58 @@
...
+
+    var popupTrigger = e.target.dataset.popupTrigger;
+    var popupModal = document.querySelector(
+      '[data-popup-modal="'.concat(popupTrigger, '"]')
+    );
+    sendAnalyticsEvent(
+      // transform trigger into camelCase and send as our event
+      popupTrigger.replace(/-([a-z])/g, function (match) {
+        return match[1].toUpperCase();
+      }),
```
ksy36, line 45: Since analytics goal tracking for this step is set up as `whatIsCompat`, we probably need to keep it that way. It's possible to change the label, but then previous weeks data will still have the old label and won't be reflected in the statistics :(

We can probably change `data-popup-trigger="what-is-webcompat"` to `data-popup-trigger="what-is-compat"` (and same with `data-popup-modal`)

miketaylr: oh yeah! i didn't even notice the difference there, thanks :)

miketaylr:
OK, just re-tested after changing the string. Seems to be working locally.

r? @ksy36

miketaylr:
(hm, test failure -- maybe hold off on review)

ksy36:
lol I pressed rerun it before I saw your comment

miketaylr:
OK, tests seem happy, and merge conflict resolved. :)

ksy36:
Thanks!

PR #3461, opened 2020-08-20
Fixes #3458 - remove optional GitHub username field

### LINKED ISSUE #3458: Remove "Optional GitHub username" field
Lets remove "Optional GitHub username" field and make it so the issue submitted upon pressing "Report anonymously"

ksy36:
r? @karlcow @miketaylr

### DIFF
--- a/webcompat/static/js/lib/wizard/steps/submit.js
+++ b/webcompat/static/js/lib/wizard/steps/submit.js
@@ -8,30 +8,15 @@
 
 import $ from "jquery";
 import { showContainer } from "../ui-utils.js";
-import { isGithubUserNameValid } from "../validation.js";
 import { uploadConsoleLogs } from "./upload-helper/console-logs-upload.js";
 import { uploadImage } from "./upload-helper/image-upload.js";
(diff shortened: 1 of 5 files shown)


karlcow (approved):
I searched and didn't find any related tests. Oops. 
We had before. :) I guess we don't need anymore. 

Thanks @ksy36

webcompat/static/js/lib/wizard/steps/submit.js
```diff
@@ -8,31 +8,16 @@
 
 import $ from "jquery";
 import { showContainer } from "../ui-utils.js";
-import { isGithubUserNameValid } from "../validation.js";
```
miketaylr, line 11: Do we want to delete `isGithubUserNameValid` as well? I don't think it's used anywhere else.

ksy36: Good point, missed that!

ksy36: ready for another look @miketaylr :)

PR #3465, opened 2020-08-21
Fixes #3464 - Only reject an issue for action==closed

### LINKED ISSUE #3464: New milestone moderation not working as expected (on staging)
I will investigate:

I filed 2 issues, here's one:

https://github.com/webcompat/webcompat-tests-private/issues/201 -> closed as accepted:invalid
But the public issue was rejected: https://github.com/webcompat/webcompat-tests/issues/2561

miketaylr:
Without this fix, it's possible to reject a previously closed issue, which is not what we want.


karlcow (approved):
@miketaylr Thanks for the fix. 

Could we give better comments for each of the if in process issues actions. 
Where we clearly explain what we do. 

Also in the initial design, we were making sure that we could not by mistakes publish something which was rejected. Is it still the case?

* closing meant rejected
* milestone meant accepted.

karlcow:
> Also in the initial design, we were making sure that we could not by mistakes publish something which was rejected. Is it still the case?

Still the case. Cool.
https://github.com/webcompat/webcompat.com/blob/c7b1fa5a67e46489f6885274577b9712a65adb4c/webcompat/webhooks/model.py#L279-L291

miketaylr:
Thanks Karl. Will add some comments then merge.

PR #3466, opened 2020-08-22
Fixes #3429 - Skip step 1 if a report is made through "Report site issue".

### LINKED ISSUE #3429: "Please enter the URL" step improvements
Given that we're getting a big drop off rate at the first step, lets experiment with making small improvements.

1) Add focus if a user is not coming through a reporter (https://github.com/webcompat/webcompat.com/issues/2964)
2) Change validation error message to be more explicit (Right now it is just `A valid URL is required.`)
3) Skip step 1 if a report is made through "Report site issue". 

This should be done after we collect metrics data on https://github.com/webcompat/webcompat.com/issues/3427.

ksy36:
r? @miketaylr


ksy36:
Thanks!

PR #3472, opened 2020-08-31
Fixes #3467 - Comment the reason why an issue was accepted but closed

### LINKED ISSUE #3467: webhook accepted: invalid/incomplete milestones should add comment rather than replace issue body
I just realized that we probably don't want to replace the body of an accepted invalid or incomplete issue. Rather, we want to add the invalid/incomplete message as a comment. If we re-train our ML bot, we're going to skew results. And we'll also lose possibly interesting reports for future analysis.

For rejected issues, we absolutely do not want report body content -- these are different though.

miketaylr:
r? @karlcow 

No rush on the review.

### DIFF
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -411,32 +411,6 @@ class TestWebhook(unittest.TestCase):
         actual = helpers.repo_scope(url)
         self.assertEqual(expected, actual)
 
-    def test_prepare_incomplete_issue(self):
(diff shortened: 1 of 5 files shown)


miketaylr:
TODO: close the private issue, the rest seems to work.

karlcow (approved):
👍 Good idea. The comments will give a better understanding too. and indeed for the ML will improve things.

PR #3521, opened 2021-01-25
Fixes #3520: Bring back min characters rule for description field

### LINKED ISSUE #3520: Bring back min char requirement for the description field
Since issues without description are harder to triage and it doesn't seem to be positively affecting the success rate of the form, lets bring back the min character requirement.

ksy36:
r? @karlcow


karlcow (approved):
Thanks a lot @ksy36 🌞

ksy36:
Thanks!

PR #3525, opened 2021-01-26
Fixes #3522: Fix outreach comment for auth reports

### LINKED ISSUE #3522: Outreach template gives 404
1. Go to https://webcompat.com/issues/66259
2. Click on the outreach template link

Expected:
Template being generated

Actual:
Get a 404 on https://webcompat.com/outreach/

ksy36:
I've been getting issue id from `public_url` in `WebHookIssue` model, which gets extracted from the issue body:

`public_url = extract_metadata(issue_body).get('public_url', '').strip()`

However `@public_url:` is only added to the body of issues that require moderation. Therefore issues that are reported with github don't have it and the id is empty (ooops 🤦‍♀️). 

To fix this I've used `number` property instead

### DIFF
--- a/tests/fixtures/webhooks/public_milestone_needscontact_auth_report.json
+++ b/tests/fixtures/webhooks/public_milestone_needscontact_auth_report.json
@@ -1,0 +1,48 @@
+{
+  "action": "milestoned",
+  "issue": {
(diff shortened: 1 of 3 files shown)


karlcow (approved):
Thanks @ksy36 for the quick fix and the test! 🧰

ksy36:
Thanks :)

PR #3527, opened 2021-01-29
Fixes #3526: Make get_description regex less greedy

### LINKED ISSUE #3526: Title on webcompat.com is broken (format issue?)
There is a format issue because a bug being transferred from bugzilla to webcompat gives this result.

Capture d’écran 2021-01-18 à 12 32 48

ksy36:
Makes it so the `get_description` accounts only for the first occurrence of "Steps".

r? @karlcow

### DIFF
--- a/tests/unit/test_template.py
+++ b/tests/unit/test_template.py
@@ -60,6 +60,24 @@ def test_get_description():
     assert get_description(body_html_frag2) is None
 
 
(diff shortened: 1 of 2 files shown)


karlcow (approved):
✂️ 🧷 Excellent!
Thanks @ksy36

PR #3534, opened 2021-02-10
fix #3528 - Fix CSS for description textarea

### LINKED ISSUE #3528: Textarea progress bar hides text when typing
Happens on all browsers. It makes me not able to read what I'm writing for example. See screenshots.

Screenshot_20210202-065150

karlcow:
This PR fixes issue #3528

## Proposed PR background

This is fixing an issue with the CSS in the textarea where the green progress bar was hiding the rest of the text.

### DIFF
--- a/webcompat/static/css/src/issue-wizard.css
+++ b/webcompat/static/css/src/issue-wizard.css
@@ -420,7 +420,7 @@
 }
 
 .issue-form .text-field {
-  padding: 4px 32px 5px 8px;
+  padding: 4px 32px 3ch 8px;
 }
 
 .with-validation-icons.is-validated .input-wrapper::after,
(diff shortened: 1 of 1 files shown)


emilio:
That doesn't work on Gecko does it? Because of padding-block-end shenanigans.

emilio:
Why not just changing `.progress-textarea .progress` to `position: relative`? That seems to fix this and as a bonus it doesn't overlap the scrollbars, etc.

emilio:
`content: "";` in that rule is useless btw, and likely to break stuff once browsers implement css-content as per spec.

karlcow:
> That doesn't work on Gecko does it? Because of padding-block-end shenanigans.

It worked on my local build. so that is strange it didn't work for you?

But yes the full thing would benefit of simpler CSS. I don't like very much the way it's done.

karlcow:
Ah long line! gotcha! indeed.

PR #3535, opened 2021-02-10
Fix #3528 - Fix CSS for description textarea

### LINKED ISSUE #3528: Textarea progress bar hides text when typing

emilio:
Make the progress bar not overlap the textarea. While at it, also remove useless `content: ""` rules which do nothing, and an useless `width` declaration which also does nothing because it's the same width as the block would have minus margins.

This PR fixes issue #3528

## Proposed PR background

See #3528

### DIFF
--- a/webcompat/static/css/src/issue-wizard-textarea.css
+++ b/webcompat/static/css/src/issue-wizard-textarea.css
@@ -22,21 +22,16 @@
 .progress-textarea .progress {
   background: #b3f4d7;
   border-radius: 2px;
-  bottom: 0;
-  content: "";
   height: 20px;
-  left: 0;
   margin: 4px;
   overflow: hidden;
-  position: absolute;
-  width: calc(100% - 8px);
+  position: relative;
   z-index: 2;
 }
 
 .progress-textarea .progress .bar {
   background: #22b573;
   bottom: 0;
-  content: "";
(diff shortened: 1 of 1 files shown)


emilio:
cc @karlcow

karlcow:
This change the initial design  (not that I'm fond of it).

#3534 design

Capture d’écran 2021-02-10 à 12 19 53

#3535 design

Capture d’écran 2021-02-10 à 12 20 02

Though yes the full thing could be redone a bit better.

karlcow:
ok indeed in my design long lines are not working.

karlcow:
This will fix it. Let's merge it and we might need to open an issue to solve this design later on. 
At least it's usable now. 

Thanks @emilio

karlcow (approved):
Thanks for the contribution.

emilio:
ah, didn't notice the focus state was in the textarea.

Should be fixable by moving the outline to `.progress-textarea` with something like `.progress-textarea:focus-within`

karlcow:
ok deployed :) https://github.com/webcompat/webcompat.com/compare/v39.0.1...v40.0.0

PR #3570, opened 2021-05-03
Fix #3562 - Adds better detection for Firefox iOS

### LINKED ISSUE #3562: Better label for Firefox ios reports
Currently 
Capture d’écran 2021-04-14 à 12 28 56

webcompat identifies firefox ios issue as

browser-firefox + engine-gecko which is not good. 
browser-firefox + os-ios would be good. 
browser-firefox-ios is ok too.

karlcow:
This PR fixes issue #3562

## Proposed PR background

Currently the labels for Firefox iOS are dependent on the extra_labels from Firefox iOS. We can make sure that we detect them on our side too. It will be less work for @softvision-oana-arbuzov and @softvision-raul-bucata

### DIFF
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -116,6 +116,19 @@ class TestWebhook(unittest.TestCase):
         **Tested Another Browser**: Yes Safari
         """  # noqa
 
+        self.issue_body9 = """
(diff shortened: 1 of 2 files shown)


ksy36 (approved):

ksy36:
Thanks!

Issue #3571, opened 2021-05-10
Use bugbug to classify issues that are coming to web-bugs-private


ksy36:
Once https://github.com/mozilla/bugbug/pull/2330 is deployed we will be able to make a request to bugbug http service for issues classification. 

As a first iteration, lets classify issues that are coming to our private repository and add a label. Once we confirm that it's marking most of the issues correctly, we could close the labelled issues as invalid (after they're moderated and moved to the public repository) and turn off the existing bot.

Existing implementation: https://github.com/mozilla/webcompat-ml-deploy/blob/master/docker/needsdiagnosis/run.py#L107

GET request to get results from bugbug http service is https://bugbug.herokuapp.com/needsdiagnosis/predict/github/webcompat/web-bugs/<issue_id>

PR #3573, opened 2021-05-17
Issue #3571: Use bugbug to classify issues

ksy36:
This PR fixes issue https://github.com/webcompat/webcompat.com/issues/3571

### DIFF
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -455,6 +455,13 @@ class TestWebhook(unittest.TestCase):
         self.assertEqual(type(actual), dict)
         self.assertEqual(actual, expected)
 
+    @patch('webcompat.webhooks.helpers.make_classification_request')
+    def test_get_issue_classification(self, mock_classification):
+        """Test that api is called only once with response code is 200."""
+        mock_classification.return_value.status_code = 200
+        helpers.get_issue_classification(12345)
+        mock_classification.assert_called_once()
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -13,14 +13,13 @@ from unittest.mock import patch
 
 import flask
 import pytest
-from requests.exceptions import HTTPError
-from requests.models import Response
+from requests.exceptions import ConnectionError
 
 import webcompat
+
 from webcompat.db import Site
 from webcompat.helpers import to_bytes
-from webcompat.webhooks import helpers
-from webcompat.webhooks.model import WebHookIssue
+from webcompat.webhooks import helpers, ml
 
 
 # The key is being used for testing and computing the signature.
@@ -455,12 +454,30 @@ class TestWebhook(unittest.TestCase):
         self.assertEqual(type(actual), dict)
         self.assertEqual(actual, expected)
 
-    @patch('webcompat.webhooks.helpers.make_classification_request')
-    def test_get_issue_classification(self, mock_classification):
-        """Test that api is called only once with response code is 200."""
-        mock_classification.return_value.status_code = 200
-        helpers.get_issue_classification(12345)
-        mock_classification.assert_called_once()
+    @patch('webcompat.webhooks.ml.make_classification_request')
+    def test_get_issue_classification(self, mock_class):
+        """Make only one request if it returns 200 status code right away.
+
+        If make_classification_request function returns 200 status code,
+        make sure that get_issue_classification is not calling it again.
+        """
+        mock_class.return_value.status_code = 200
+        ml.get_issue_classification(12345)
+        mock_class.assert_called_once()
+
+    @patch('time.sleep', return_value=None)
+    @patch('webcompat.webhooks.ml.make_classification_request')
+    def test_get_issue_classification_exception(self, mock_class, mock_time):
+        """Poll bugbug and raise an exception if request limit exceeded
+
+        If make_classification_request function returns 202 status code,
+        call get_issue_classification again until exception occurs.
+        """
+        mock_class.return_value.status_code = 202
+        with pytest.raises(ConnectionError):
+            ml.get_issue_classification(12345)
+
+        assert mock_class.call_count == 4
 
 
 if __name__ == '__main__':
--- a/tests/unit/test_webhook_model.py
+++ b/tests/unit/test_webhook_model.py
@@ -30,7 +30,7 @@ boring = ('Not an interesting hook', 403, {'Content-Type': 'text/plain'})
 gracias = ('gracias, amigo.', 200, {'Content-Type': 'text/plain'})
 wrong_repo = ('Wrong repository', 403, {'Content-Type': 'text/plain'})
 oops = ('oops', 400, {'Content-Type': 'text/plain'})
-comment_added = ('public url added', 200, {'Content-Type': 'text/plain'})
+comment_added = ('public url added and issue classified', 200, {'Content-Type': 'text/plain'})  # noqa
 outreach_comment_added = ('outreach generator url added', 200, {'Content-Type': 'text/plain'})  # noqa
 
 issue_info1 = {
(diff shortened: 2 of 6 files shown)


tests/unit/test_webhook_model.py
```diff
@@ -139,7 +139,7 @@ def test_comment_closed_reason(mock_mr):
 @patch('webcompat.webhooks.model.make_request')
 def test_moderate_public_issue(mock_mr):
     """Test issue state and API request that is sent to GitHub."""
-    mock_mr.return_value.status_code == 200
+    mock_mr.return_value.status_code = 200
     json_event, signature = event_data('private_issue_opened.json')
```
ksy36, line 143: not related to this work, but I think we should be setting a status with `= 200`? as right not it's not doing anything

karlcow: Thanks for catching this. Yes definitely right.

tests/unit/test_webhook.py
```diff
@@ -455,6 +455,13 @@ def test_prepare_rejected_issue(self):
         self.assertEqual(type(actual), dict)
         self.assertEqual(actual, expected)
 
+    @patch('webcompat.webhooks.helpers.make_classification_request')
+    def test_get_issue_classification(self, mock_classification):
+        """Test that api is called only once with response code is 200."""
```
karlcow, line 460: nit:
```suggestion
        """Test that api is called only once with a 200 response status code."""
```

karlcow, line 460: What is happening if it's called more than once?
Does it fail? Can't happen?

tests/unit/test_webhook_model.py
```diff
@@ -76,7 +76,7 @@ def test_model_instance():
 @patch('webcompat.webhooks.model.make_request')
 def test_close_private_issue(mock_mr):
     """Test issue state and API request that is sent to GitHub."""
-    mock_mr.return_value.status_code == 200
+    mock_mr.return_value.status_code = 200
```
karlcow, line 79: So good to have caught this. Now I'm worried… because it doesn't change anything. I wonder what was the value then and how it would modify the test or not. Let's see.

Old:

```
tests/unit/test_webhook_model.py::test_close_private_issue <MagicMock name='make_request().status_code' id='4363956624'>
PASSED
```

New:

```
tests/unit/test_webhook_model.py::test_close_private_issue 200
PASSED
```

ok definitely wrong.

After removing them, everything is still passing.
So we don't test the status code in any ways. 

I think we can safely remove all instances of `mock_mr.return_value.status_code = 200`

webcompat/webhooks/helpers.py
```diff
@@ -230,3 +234,35 @@ def prepare_rejected_issue():
     payload_request['state'] = 'closed'
     payload_request['milestone'] = invalid_id
     return payload_request
+
+
+def make_classification_request(issue_number):
```
karlcow, line 239: I wonder if we should put the classification functions in a new file: `webcompat/webhooks/ml.py`?

Just in case in the future, it grows with more functions so we do not end up with a helpers being huge. Maybe it doesn't matter. Opinions?

karlcow, line 239: nit: Missing docstring

webcompat/webhooks/helpers.py
```diff
@@ -230,3 +234,35 @@ def prepare_rejected_issue():
...
+    The service returns 202 status if request is still in process
+    and 200 status if the issue is classified
+    """
+    for _ in range(retry_count):
+        response = make_classification_request(issue_number)
+
+        if response.status_code == 202:
+            time.sleep(retry_sleep)
+        else:
+            break
```
karlcow, line 262: `202` code in this code is a wait instruction. aka the response is not ready. 
`200` and `4**` and `500` will be treated the same. 

Is it what we want?

webcompat/webhooks/helpers.py
```diff
@@ -230,3 +234,35 @@ def prepare_rejected_issue():
...
+    else:
+        total_sleep = retry_count * retry_sleep
+        msg = f"Couldn't classify issue {issue_number} in {total_sleep} seconds, aborting"  # noqa
+        raise Exception(msg)
```
karlcow, line 266: we need a test for this, aka when there's a failure.

webcompat/webhooks/model.py
```diff
@@ -274,14 +293,21 @@ def process_issue_action(self):
...
             except HTTPError as e:
                 msg_log(f'comment failed ({e})', self.number)
                 return oops()
-            else:
-                return make_response('public url added', 200)
+
+            try:
+                self.classify()
+            except HTTPError as e:
+                msg_log(f'classification failed ({e})', self.number)
```
karlcow, line 307: let's test the error message.

webcompat/webhooks/model.py
```diff
@@ -234,6 +236,21 @@ def get_public_issue_number(self):
             url = self.public_url.strip().rsplit('/', 1)[1]
         return url
 
+    def classify(self):
+        """Make a request to bugbug and label the issue.
+
+        Gets issue classification from bugbug and labels
+        the issue if probability is high
+        """
+        data = get_issue_classification(self.number)
```
karlcow, line 245: This here might raise an Exception. 
https://github.com/webcompat/webcompat.com/pull/3573/files#diff-96859bd3238aaaffb2c63f7fd42e9899ba93c7e69bf96dd6c33f6de6f58b42feR266

but this is not handled here, with a try. 
Later on in https://github.com/webcompat/webcompat.com/pull/3573/files#diff-c2434db66ee7efbeb6ebcafb5470cc409f4a0cc61ac35709a32a1a0341c4b71dR306
the code is attempting to catch an HTTPError. That's good. 

but the exception before is very generic and not only HTTP
```
raise Exception(msg)
```

karlcow (changes requested):
thanks @ksy36 
Super good progress in there.
Some comments to fix, some things to remove. 
And a couple of questions to better understand.

webcompat/webhooks/helpers.py
ksy36: Yeah, this code will basically keep retrying if the status is 202 up to a certain limit. If it's 200, 4** or 500, we don't need to retry. And in case of 4** or 500 the error will be raised in `make_classification_request`

tests/unit/test_webhook.py
ksy36: It's making sure that if we receive a 200 status, `make_classification_request` is not called again.

So basically classification in bugbug is asynchronous, the first request submits an issue for classification (in response we get 202 as the result is not ready yet) and we need to make a second one to get the results (and we get 200). 
This needs another test, but I'm not sure how to do it with a mock:
 
1 request -> 202 
2 request -> 200
assert that mock_classification called twice

webcompat/webhooks/helpers.py
ksy36: Yeah, probably a good idea, I'll move it

tests/unit/test_webhook_model.py
ksy36: Yeah I was thinking that too :) I'll remove them

webcompat/webhooks/helpers.py
ksy36: Added a test in test_webhook.py 👍

webcompat/webhooks/model.py
ksy36: Added a test for that in test_webhook_model.py

webcompat/webhooks/model.py
ksy36: I've changed it to be ConnectionError and added it to catch handle in https://github.com/webcompat/webcompat.com/pull/3573/commits/91cc36c48523bc7a1f328078fc2a283fc3253c48#diff-c2434db66ee7efbeb6ebcafb5470cc409f4a0cc61ac35709a32a1a0341c4b71dR303

ksy36:
I've pushed the fixes, could you please take another look @karlcow?

webcompat/webhooks/model.py
karlcow: Wonderful.

webcompat/webhooks/model.py
karlcow: Yes perfect.

karlcow (approved):
Thanks @ksy36 
This is good to go. Let's deploy it.

ksy36:
Thanks!

Issue #3348, opened 2020-06-16
Updating contributing documentation


miketaylr:
Currently there are screenshots and other text that refer to the old form. It should be updated for the new form.

See https://webcompat.com/contributors/report-bug, for example (worth clicking through all those links to verify what needs to be updated).

karlcow:
Closed by https://github.com/webcompat/webcompat.com/pull/3601

PR #3576, opened 2021-05-27
Fixes #3574: Incorporate bugbug ml labelling into moderation process

### LINKED ISSUE #3574: Incorporate bugbug ml labelling into moderation process
In https://github.com/webcompat/webcompat.com/issues/3571 I've added bugbug classification and issues that are most likely invalid will receive `bugbug-probability-high` label in the private repository. We need to incorporate this into the moderation process to close such issues automatically.

There is a problem with posting issues content to the public repository without human moderation as we may end up submitting abusive content and get banned by GitHub again. So as a solution, we could post a placeholder text to the public issue, instead of its content for these invalid issues. Something like the current bot is posting now [as a comment](https://github.com/webcompat/web-bugs/issues/73946#issuecomment-842043683), but we'll put it in the issue body instead.

`issue is labelled invalid by the bot -> do not post the content to the public repository, post a placeholder text instead and close the issue`

If we decide to go this path, users won't be able to see the content of their report if it was labelled invalid (they will only see the placeholder text). It's not ideal (but acceptable?), and I wonder how important this is given the overwhelming amount of issues we're getting. 

Do you have any thoughts on this? 
@softvision-oana-arbuzov  @softvision-raul-bucata @karlcow

ksy36:
In this PR I've added functionality to automatically close issues marked by bugbug as invalid.

1. Issues marked by bugbug as invalid moved to "ml-autoclosed" milestone. 
2. Once milestone event is received, public issue content is replaced with a placeholder text and label "bugbug-probability-high" is added, public issue closed.
3. Private issue closed

Discussed in https://github.com/webcompat/webcompat.com/issues/3574

Example of a public test issue that was closed using this code: https://github.com/webcompat/webcompat-tests/issues/2638

### DIFF
--- a/config/environment.py
+++ b/config/environment.py
@@ -41,6 +41,7 @@ if PRODUCTION:
     )
     BUGBUG_HTTP_SERVER = "https://bugbug.herokuapp.com"
     CLASSIFIER_PATH = "needsdiagnosis/predict/github/webcompat/web-bugs-private"  # noqa
+    AUTOCLOSED_MILESTONE_ID = 15
 
 if STAGING:
     GITHUB_CLIENT_ID = os.environ.get('STAGING_GITHUB_CLIENT_ID')
@@ -59,6 +60,7 @@ if STAGING:
     )
     BUGBUG_HTTP_SERVER = "https://bugbug.herokuapp.com"
     CLASSIFIER_PATH = "needsdiagnosis/predict/github/webcompat/webcompat-tests-private"  # noqa
+    AUTOCLOSED_MILESTONE_ID = 5
 
 if LOCALHOST:
     # for now we are using .env only on localhost
@@ -82,6 +84,7 @@ if LOCALHOST:
     )
     BUGBUG_HTTP_SERVER = "http://0.0.0.0:8000"
     CLASSIFIER_PATH = "needsdiagnosis/predict/github/webcompat/webcompat-tests-private"  # noqa
+    AUTOCLOSED_MILESTONE_ID = 5
 
 # BUG STATUS
 # The id will be initialized when the app is started.
--- a/webcompat/issues.py
+++ b/webcompat/issues.py
@@ -43,6 +43,33 @@ REJECTED_TITLE = 'Issue rejected.'
 REJECTED_BODY = '''<p>The content of this issue doesn't meet our
 <a href="https://webcompat.com/terms#acceptable-use">acceptable use</a>
 guidelines. Its original content has been deleted.</p>'''
+AUTOCLOSE_TITLE = 'Issue closed.'
+AUTOCLOSE_BODY = '''<p>Thanks for the report. We have closed this issue
+automatically as we suspect it is invalid. If we made a mistake, please
+file a new issue and try to provide more context.</p>'''
+
+TEMPLATE_CONFIG = {
+    'rejected': {
+        'title': REJECTED_TITLE,
+        'body': REJECTED_BODY
+    },
+    'invalid': {
+        'title': '',
+        'body': INVALID_BODY
+    },
+    'incomplete': {
+        'title': '',
+        'body': INCOMPLETE_BODY
+    },
+    'autoclosed': {
+        'title': AUTOCLOSE_TITLE,
+        'body': AUTOCLOSE_BODY
+    },
+    'ongoing': {
+        'title': ONGOING_TITLE,
+        'body': ONGOING_BODY
+    }
+}
 
 
 def moderation_template(choice='ongoing'):
@@ -53,22 +80,20 @@ def moderation_template(choice='ongoing'):
     - rejected: the issue has been rejected.
     - incomplete: the issue is incomplete (but not rejected)
     - invalid: the issue is invalid (but not rejected)
+    - autoclosed: the issue is auto closed by ml bot
 
     The default is 'ongoing' even with unknown keywords.
     """
-    if choice == 'rejected':
-        title = REJECTED_TITLE
-        body = REJECTED_BODY
-    elif choice == 'invalid':
-        title = ''
-        body = INVALID_BODY
-    elif choice == 'incomplete':
-        title = ''
-        body = INCOMPLETE_BODY
+    if choice in TEMPLATE_CONFIG:
+        return {
+            'title': TEMPLATE_CONFIG[choice].get('title'),
+            'body': TEMPLATE_CONFIG[choice].get('body')
+        }
     else:
-        title = ONGOING_TITLE
-        body = ONGOING_BODY
-    return {'title': title, 'body': body}
+        temp = (", ".join(TEMPLATE_CONFIG.keys()))
+        raise ValueError(
+            f'Template choice must be one of the following: {temp}'
+        )
 
 
 def report_private_issue(form, public_url):
--- a/webcompat/webhooks/ml.py
+++ b/webcompat/webhooks/ml.py
@@ -8,6 +8,7 @@
 
 import requests
 import time
+import logging
 
 from requests.exceptions import ConnectionError
 
@@ -26,7 +27,7 @@ def make_classification_request(issue_number):
     return response
 
 
-def get_issue_classification(issue_number, retry_count=4, retry_sleep=3):
+def get_issue_classification(issue_number, max_retry_count=4, retry_sleep=7):
     """Get issue classification from bugbug.
 
     As classification happens in the background we need to make a second
@@ -35,16 +36,23 @@ def get_issue_classification(issue_number, retry_count=4, retry_sleep=3):
     The service returns 202 status if request is still in process
     and 200 status if the issue is classified
     """
-    for _ in range(retry_count):
+    log = app.logger
+    log.setLevel(logging.INFO)
+    start = time.monotonic()
+    retry_count = 0
+    for _ in range(max_retry_count):
         response = make_classification_request(issue_number)
 
         if response.status_code == 202:
             time.sleep(retry_sleep)
         else:
+            retry_count = _
             break
     else:
-        total_sleep = retry_count * retry_sleep
-        msg = f"Couldn't classify issue {issue_number} in {total_sleep} seconds, aborting"  # noqa
+        total = round(time.monotonic() - start, 2)
+        msg = f"ML FAIL: issue {issue_number} in {total} seconds with {max_retry_count} retries"  # noqa
         raise ConnectionError(msg)
-
+    total = round(time.monotonic() - start, 2)
+    msg = f"ML OK: issue {issue_number} in {total} seconds with {retry_count} retries"  # noqa
+    log.info(msg)
     return response.json()
--- a/webcompat/webhooks/model.py
+++ b/webcompat/webhooks/model.py
@@ -20,11 +20,13 @@ from webcompat.webhooks.helpers import msg_log
 from webcompat.webhooks.helpers import oops
 from webcompat.webhooks.helpers import prepare_rejected_issue
 from webcompat.webhooks.helpers import repo_scope
+from webcompat.webhooks.helpers import prepare_private_url
 from webcompat.webhooks.ml import get_issue_classification
 from webcompat.issues import moderation_template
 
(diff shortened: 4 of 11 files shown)


webcompat/webhooks/model.py
```diff
@@ -234,19 +242,20 @@ def get_public_issue_number(self):
...
-        Gets issue classification from bugbug and labels
+        Gets issue classification from bugbug and milestones
         the issue if probability is high
         """
         data = get_issue_classification(self.number)
         needsdiagnosis_false = data.get('class')
         proba = data.get('prob')
 
         if needsdiagnosis_false and proba and proba[1] > THRESHOLD:
-            payload = {'labels': ['bugbug-probability-high']}
```
ksy36, line 247: I have decided to not use this label in private repo because in addition to the milestone it seems unnecessary

webcompat/webhooks/ml.py
```diff
@@ -26,7 +26,7 @@ def make_classification_request(issue_number):
     return response
 
 
-def get_issue_classification(issue_number, retry_count=4, retry_sleep=3):
+def get_issue_classification(issue_number, retry_count=4, retry_sleep=7):
```
ksy36, line 29: I have experimented with sending requests using Postman, it seems like sometimes 5 sec is enough, but sometimes it takes 10-12 sec. So I'm going to set 7 for now and see if we still get errors in the logs

webcompat/webhooks/model.py
```diff
@@ -185,12 +189,16 @@ def close_public_issue(self, reason='rejected'):
...
+        'autoclosed'
         'rejected' (default)
         """
         if reason == 'incomplete':
             payload_request = self.prepare_accepted_issue('incomplete')
         elif reason == 'invalid':
             payload_request = self.prepare_accepted_issue('invalid')
+        elif reason == 'autoclosed':
+            payload_request = prepare_rejected_issue(reason)
+            payload_request['body'] += prepare_private_url(self.html_url)
```
ksy36, line 201: I have added a hidden private issue github url to the closed public issue because I need to use it to train bugbug model. Since there will be no way to get the original content of a public issue, the plan is to make a request to fetch data of a private issue and use its contents instead, as it will be useful for training the model.

So it looks like this:

```
<p>Thanks for the report. We have closed this issue 
automatically as we suspect it is invalid. If we made a mistake, please 
file a new issue and try to provide more context or file a report 
<a href="https://github.com/webcompat/web-bugs/issues/new/choose">
via our GitHub repository</a>.</p>
<!-- @private_url: repos/webcompat/webcompat-tests-private/issues/273 -->
```

webcompat/issues.py
```diff
@@ -43,6 +43,31 @@
 REJECTED_BODY = '''<p>The content of this issue doesn't meet our
 <a href="https://webcompat.com/terms#acceptable-use">acceptable use</a>
 guidelines. Its original content has been deleted.</p>'''
+AUTOCLOSE_TITLE = 'Issue closed.'
+AUTOCLOSE_BODY = '''<p>Thanks for the report. We have closed this issue
+automatically as we suspect it is invalid. If we made a mistake, please
+file a new issue and try to provide more context or file a report
+<a href="https://github.com/webcompat/web-bugs/issues/new/choose">
+via our GitHub repository</a>.</p>'''
```
karlcow, line 51: While this is possible, should we advertise this possibility. I'm thinking about the data structure of these bugs which are always a bit different for `body` parsing. 💭 I don't know. I let you decide.

webcompat/issues.py
```diff
@@ -43,6 +43,31 @@
...
+    },
+    'incomplete': {
+        'title': '',
+        'body': INCOMPLETE_BODY
+    },
+    'autoclosed': {
+        'title': AUTOCLOSE_TITLE,
+        'body': AUTOCLOSE_BODY
+    }
+}
```
karlcow, line 70: Cool.

karlcow, line 70: possibly here ongoing could be part of this structure too. 
```suggestion
    'autoclosed': {
        'title': AUTOCLOSE_TITLE,
        'body': AUTOCLOSE_BODY
    }
    'ongoing': {
        'title': ONGOING_TITLE,
        'body': ONGOING_BODY
    }
}
```

Then the `if` in `moderation_template` becomes unecessary. it just extracts the right case.

webcompat/issues.py
```diff
@@ -56,15 +81,9 @@ def moderation_template(choice='ongoing'):
...
-        body = INVALID_BODY
-    elif choice == 'incomplete':
-        title = ''
-        body = INCOMPLETE_BODY
+    if choice in TEMPLATE_CONFIG:
+        title = TEMPLATE_CONFIG[choice].get('title')
+        body = TEMPLATE_CONFIG[choice].get('body')
     else:
         title = ONGOING_TITLE
         body = ONGOING_BODY
```
karlcow, line 89:
```suggestion
    title = TEMPLATE_CONFIG[choice].get('title', 'ERROR_TITLE')
    body = TEMPLATE_CONFIG[choice].get('body', 'ERROR_BODY')
```

if everything is in `TEMPLATE_CONFIG`

webcompat/webhooks/ml.py
karlcow: That's cool. 
To better understand the data though, probably we [should monitor](https://docs.python.org/3/library/time.html#time.monotonic) success and failure. 

just after the docstring we could have something along

``` python
    log = app.logger
    log.setLevel(logging.INFO)
    start = time.monotonic()
    for _ in range(retry_count):
        response = make_classification_request(issue_number)
        if response.status_code == 202:
            time.sleep(retry_sleep)
        else:
            break
    else:
        total = round(time.monotonic() - start, 2)
        msg = f"ML FAIL: issue {issue_number} in {total} seconds with {retry_count} retries"  # noqa
        raise ConnectionError(msg)
    total = round(time.monotonic() - start, 2)
    msg = f"ML OK: issue {issue_number} in {total} seconds with {retry_count} retries"  # noqa
    log.info(msg)
    return response.json()
```

With this we have an opportunity to create a dashboard or to grep the data to graph them and understand in a couple of weeks. 

What do you think? It doesn't have to be exactly this. There might be better way to organize the code.

webcompat/webhooks/model.py
```diff
@@ -185,12 +189,16 @@ def close_public_issue(self, reason='rejected'):
...
         'invalid'
+        'autoclosed'
         'rejected' (default)
         """
         if reason == 'incomplete':
             payload_request = self.prepare_accepted_issue('incomplete')
         elif reason == 'invalid':
             payload_request = self.prepare_accepted_issue('invalid')
+        elif reason == 'autoclosed':
+            payload_request = prepare_rejected_issue(reason)
```
karlcow, line 200: Probably?

```suggestion
            payload_request = self.prepare_rejected_issue(reason)
```

webcompat/webhooks/model.py
karlcow: Maybe we could add a comment explaining just that. :)

webcompat/webhooks/model.py
```diff
@@ -234,19 +242,20 @@ def get_public_issue_number(self):
         return url
 
     def classify(self):
-        """Make a request to bugbug and label the issue.
+        """Make a request to bugbug and milestone the issue.
```
karlcow, line 245: nit
```suggestion
        """Make a request to bugbug and add a milestone to the issue.
```

webcompat/webhooks/model.py
```diff
@@ -234,19 +242,20 @@ def get_public_issue_number(self):
...
 
-        Gets issue classification from bugbug and labels
+        Gets issue classification from bugbug and milestones
```
karlcow, line 247: nit

```suggestion
        Gets issue classification from bugbug and add a milestone to
```

webcompat/webhooks/model.py
```diff
@@ -234,19 +242,20 @@ def get_public_issue_number(self):
...
         needsdiagnosis_false = data.get('class')
         proba = data.get('prob')
 
         if needsdiagnosis_false and proba and proba[1] > THRESHOLD:
-            payload = {'labels': ['bugbug-probability-high']}
-            path = f'repos/{PRIVATE_REPO}/{self.number}/labels'
-            make_request('post', path, payload)
+            path = f'repos/{PRIVATE_REPO}/{self.number}'
+            self.milestone = AUTOCLOSED_MILESTONE_ID
+            payload_request = {'milestone': self.milestone}
```
karlcow, line 257: Maybe these two lines can be one. 
```suggestion
            payload_request = {'milestone': AUTOCLOSED_MILESTONE_ID}
```

karlcow (changes requested):
@ksy36 This is really really really cool.

That will probably help a lot the softvision team. I made some comments and suggestions. 

Only two points to think about:

1. Milestones fetching vs hardcoded
2. Logging success and failures of communicating with bugbug

config/environment.py
```diff
@@ -41,6 +41,7 @@
     )
     BUGBUG_HTTP_SERVER = "https://bugbug.herokuapp.com"
     CLASSIFIER_PATH = "needsdiagnosis/predict/github/webcompat/web-bugs-private"  # noqa
+    AUTOCLOSED_MILESTONE_ID = 15
```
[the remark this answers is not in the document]
ksy36: Ah cool, I didn't know and did it similarly to https://github.com/webcompat/webcompat.com/blob/master/config/environment.py#L34  
But yeah, that's a good point. I will address this in a separate issue, as we're not doing this for the private repository at the moment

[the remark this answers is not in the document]
ksy36: https://github.com/webcompat/webcompat.com/issues/3577

webcompat/issues.py
ksy36: Yeah, I'm not sure either :) The main goal of the message is to get them to report as authorized users (if they ever come back to the issue, that is). Since bugbug is not checking those reports, so they won't be closed. 

It also could be an anchor link to https://webcompat.com/contributors/report-bug and we can have a section "how to report as authorized GitHub user". We could update this in https://github.com/webcompat/webcompat.com/issues/3348, and then add the anchor link :)

ksy36: I've removed the link for now

webcompat/issues.py
ksy36: Yeah, except when a key that is not in the config is passed, it defaults to ongoing too. 

I just saw this test https://github.com/webcompat/webcompat.com/blob/master/tests/unit/test_issues.py#L146 and wasn't sure if I should go this path :) It's probably safe though, just need to check all references of the function

webcompat/webhooks/model.py
ksy36: I've reused `prepare_rejected_issue` here, which is an existing function in `webhooks_helpers.py` file, so this is correct

webcompat/webhooks/model.py
ksy36: done :)

webcompat/issues.py
ksy36: moved it to the config, added an exception for the wrong key and adjusted the test

webcompat/webhooks/ml.py
ksy36: Thanks for providing the code! I've adjusted it a bit, to save the current retry count

ksy36:
Ready for another look @karlcow :)

config/environment.py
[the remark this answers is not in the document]
karlcow: Wonderful.

config/environment.py
```diff
@@ -59,6 +60,7 @@
     )
     BUGBUG_HTTP_SERVER = "https://bugbug.herokuapp.com"
     CLASSIFIER_PATH = "needsdiagnosis/predict/github/webcompat/webcompat-tests-private"  # noqa
+    AUTOCLOSED_MILESTONE_ID = 5
```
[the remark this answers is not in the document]
karlcow: see #3577

karlcow (approved):
Thanks a lot @ksy36 
Awesome work.

ksy36:
Thanks!

PR #3606, opened 2021-07-21
fix #3114 - Fix excessive labels assignment

### LINKED ISSUE #3114: Investigate and fix excessive label assigning
This is related to the plethora of labels applied to:

https://github.com/webcompat/web-bugs/issues/46909
https://github.com/webcompat/web-bugs/issues/46908
https://github.com/webcompat/web-bugs/issues/19195

karlcow:
This PR fixes issue #3114 

## Proposed PR background

This switches the strategy for assigning labels. 
We will control a bit more what is done.

### DIFF
--- a/tests/unit/test_api_urls.py
+++ b/tests/unit/test_api_urls.py
@@ -179,7 +179,7 @@ class TestAPIURLs(unittest.TestCase):
                                 environ_base=headers)
             self.assertEqual(rv.status_code, 403)
 
-    @patch('webcompat.api.endpoints.proxy_request')
+    @patch('webcompat.api.endpoints.api_request')
     def test_api_patch_issue_valid_request(self, github_data):
         """Patching the issue - Valid request."""
         with webcompat.app.app_context():
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -343,9 +343,7 @@ class TestWebhook(unittest.TestCase):
             ({'extra_labels': 'type-media'}, ['type-media']),
             ({'extra_labels': 'browser-focus-geckoview'},
              ['browser-focus-geckoview']),
-            ({'extra_labels': 'cool, dude'}, ['cool', 'dude']),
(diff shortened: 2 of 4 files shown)


karlcow:
Do not merge. Just pushing for the PR to run the tests.

ksy36 (approved):

ksy36:
Thanks, looks good!

ksy36:
There is an error when editing labels/milestone for users with write access (however the change is applied):

```python
  File "/Users/ksenia/Documents/projects/webcompat.com/webcompat/api/endpoints.py", line 69, in edit_issue
    print(edit.status_code, 'test')
AttributeError: 'tuple' object has no attribute 'status_code'
```
I'll create a PR with a fix

karlcow:
ok. I think I understand the issue. 
So we replaced `proxy_request` by `api_request`, but this too doesn't return a regular object (bad design, we need to fix this, but probably in a separate bug).

https://github.com/webcompat/webcompat.com/blob/e5dbd1c8ea33e329dd5e295268c2f0717aa9f7dc/webcompat/helpers.py#L459-L481

https://github.com/webcompat/webcompat.com/blob/e5dbd1c8ea33e329dd5e295268c2f0717aa9f7dc/webcompat/helpers.py#L436-L456

so when we reach, the object is not working 

https://github.com/webcompat/webcompat.com/blob/e5dbd1c8ea33e329dd5e295268c2f0717aa9f7dc/webcompat/api/endpoints.py#L68-L70

The request has been correctly sent but the answers is not in the format we expect.

This is the shape of the content we receive once going through `api_request`, a tuple with 3 values.

Capture d’écran 2021-07-22 à 09 17 14

While the `edit.content, edit.status_code` is trying to access properties of an object, this can be fixed with a quick hack testing the nature, the type of `edit`, but probably we need to make it more regular and that would require another bug.

karlcow:
I created a new issue for fixing this issue in #3607

PR #3609, opened 2021-07-23
Fix #3607 - Make modify_labels and edit_issue expect tuple instead of Request

### LINKED ISSUE #3607: fix api_request broken with #3114
This is a followup on the PR #3606 and issue #3114

So we replaced `proxy_request` by `api_request`, but this too doesn't return a regular object (bad design, we need to fix this, but probably in a separate bug).

https://github.com/webcompat/webcompat.com/blob/e5dbd1c8ea33e329dd5e295268c2f0717aa9f7dc/webcompat/helpers.py#L459-L481

https://github.com/webcompat/webcompat.com/blob/e5dbd1c8ea33e329dd5e295268c2f0717aa9f7dc/webcompat/helpers.py#L436-L456

so when we reach, the object is not working 

https://github.com/webcompat/webcompat.com/blob/e5dbd1c8ea33e329dd5e295268c2f0717aa9f7dc/webcompat/api/endpoints.py#L68-L70

The request has been correctly sent but the answers is not in the format we expect.

This is the shape of the content we receive once going through `api_request`, a tuple with 3 values.

Capture d’écran 2021-07-22 à 09 17 14

While the `edit.content, edit.status_code` is trying to access properties of an object, this can be fixed with a quick hack testing the nature, the type of `edit`, but probably we need to make it more regular and that would require another bug.

_Originally posted by @karlcow in https://github.com/webcompat/webcompat.com/issues/3606#issuecomment-884579305_

ksy36:
r? @karlcow

### DIFF
--- a/tests/unit/test_api_urls.py
+++ b/tests/unit/test_api_urls.py
@@ -49,6 +49,12 @@ def mock_api_response(response_config={}):
     return api_response
 
 
+def get_user_mock():
+    user = MagicMock()
+    user.user_id = 1
+    return user
+
+
 class TestAPIURLs(unittest.TestCase):
     """Tests for all API URLs."""
 
@@ -137,12 +143,6 @@ class TestAPIURLs(unittest.TestCase):
             self.assertEqual(
                 rv.content_type, 'text/html')
 
-    def test_api_set_labels_without_auth(self):
-        """API setting labels without auth returns JSON 403 error code."""
-        rv = self.app.post('/api/issues/1/labels',
-                           environ_base=headers, data='[]')
-        self.assertEqual(rv.status_code, 403)
-
     def test_api_user_activity_without_auth(self):
         """API access to user activity without auth returns JSON 401."""
         rv = self.app.get('/api/issues/miketaylr/creator',
@@ -160,9 +160,16 @@ class TestAPIURLs(unittest.TestCase):
         self.assertEqual(rv.content_type, 'application/json')
         self.assertEqual(json_body['status'], 404)
 
-    def test_api_patch_issue_state_status(self):
+    @patch('webcompat.db.User.query')
+    def test_api_patch_issue_state_status(self, db_user_mock):
         """Patching the issue - Incompatible state and status."""
         with webcompat.app.app_context():
+
+            # mock authenticated user
+            db_user_mock.return_value.get.return_value = get_user_mock()
+            with self.app.session_transaction() as sess:
+                sess['user_id'] = 1
+
             webcompat.app.config.update(STATUSES=STATUSES)
             data = {'state': 'closed', 'milestone': 2}
             patch_data = json.dumps(data)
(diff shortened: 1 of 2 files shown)


tests/unit/test_api_urls.py
```diff
@@ -137,12 +143,6 @@ def test_api_comments_link_header_auth(self):
             self.assertEqual(
                 rv.content_type, 'text/html')
 
-    def test_api_set_labels_without_auth(self):
```
ksy36, line 140: Moved this test to the bottom of the file to have all related tests in one place https://github.com/webcompat/webcompat.com/pull/3609/files#diff-688eb42ddc0072d66da1cbe00c95cdef58f51e6cf2b058186be3215c53eebf1dR240

webcompat/api/endpoints.py
```diff
@@ -52,10 +52,13 @@ def edit_issue(number):
...
     - It only allows change of state and change of milestones.
-    - It is always proxied to allow any logged in user
-      to be able to edit issues.
+    - It's not proxied, so only users with write access are
+      able to edit issues.
       Format: {'milestone': 2, 'state': 'open'}
     """
+    if not g.user:
+        abort(403)
+
```
ksy36, line 61: Added this check here because `api_request` conveniently sends `proxy_request` if `g.user` doesn't exist :D

https://github.com/webcompat/webcompat.com/blob/main/webcompat/helpers.py#L474

tests/unit/test_api_urls.py
```diff
@@ -160,37 +160,89 @@ def test_api_search_wrong_parameter(self):
...
+    @patch('webcompat.api.endpoints.api_request')
+    def test_api_set_labels_valid_request(self, github_data, db_user_mock):
+        """Setting labels - Valid request."""
+        with webcompat.app.app_context():
+
+            # mock authenticated user
+            db_user_mock.return_value.get.return_value = get_user_mock()
+            with self.app.session_transaction() as sess:
+                sess['user_id'] = 1
+
```
ksy36, line 233: Tried mocking `g.user` as the documentation suggests in https://flask.palletsprojects.com/en/2.0.x/testing/#faking-resources-and-context , but it didn't work for some reason :/ 

So here I'm mocking session and `User.query.get` and `g.user` gets assigned a mocked value in `before_request`
https://github.com/webcompat/webcompat.com/blob/main/webcompat/views.py#L65. Maybe there is a better way.

karlcow:
```
Flask==1.1.2
```

and i double check and the doc is similar for 1.1.2
https://flask.palletsprojects.com/en/1.1.x/testing/#faking-resources-and-context

tests/unit/test_api_urls.py
karlcow: good idea.

tests/unit/test_api_urls.py
```diff
@@ -160,37 +160,89 @@ def test_api_search_wrong_parameter(self):
...
+            data = {'state': 'open', 'milestone': 2}
+            patch_data = json.dumps(data)
+            rv = self.app.patch('/api/issues/1/edit', data=patch_data,
+                                environ_base=headers)
+            self.assertEqual(rv.status_code, 403)
+
+    @patch('webcompat.db.User.query')
     @patch('webcompat.api.endpoints.api_request')
-    def test_api_patch_issue_valid_request(self, github_data):
+    def test_api_patch_issue_valid_request(self, github_data, db_user_mock):
```
karlcow, line 207: double mocks are always a bit scary. 
Because basically by faking too many things at the end we do not test anything.

Something to remember if we can improve the way we do things. so we isolate a bit more the test circumstances. To think 🤔

karlcow: Maybe we could work out something along https://github.com/webcompat/webcompat.com/blob/408d8032dcdcdd94893b2ca84e071c9b71ba7926/webcompat/helpers.py#L371

webcompat/api/endpoints.py
karlcow: We probably need to work out a bit the architecture and double check what is still using `proxy_request` and how. 

And then simplify things.

webcompat/api/endpoints.py
```diff
@@ -65,8 +68,9 @@ def edit_issue(number):
     data_check = (patch_data['milestone'], patch_data['state'])
     # The PATCH data can only be of length: 2
     if data_check in valid_statuses and len(patch_data) == 2:
-        edit = api_request('patch', path, data=request.data)
-        return (edit.content, edit.status_code,
+        (content, status_code, headers) = api_request('patch',
+                                                      path, data=request.data)
+        return (content, status_code,
```
karlcow, line 73: neat!

karlcow (approved):
Thanks a lot @ksy36 This should unblock things and preserve us from mistake.

To extensively test on staging.

ksy36:
Thanks! I'll deploy and test on staging

ksy36:
Tested and it works well. We can deploy on Monday

PR #3629, opened 2021-10-28
Fixes #655 - Convert OAuth app to GitHub app to limit permissions we request 

### LINKED ISSUE #655: Excessive permissions required to file a bug using Github
When I tried to file a bug about a website there are two buttons at the bottom "Report via Github" and "Report anonymously", I clicked on Github and then I was prompted to grant too many permissions:

mezer_08-03_12-15-18

I had expected to only allow permissions on the webcompat repository and granting permission to read my identity and file a bug on my behalf seems like should be enough, no need to give full permissions on everything

ksy36:
I've created 3 GitHub apps, for production, staging and local (similarly to OAuth apps we have currently) and installed them on webcompat account. Each app has an access only to a specific repository (`web-bugs` for production app and `webcompat-tests` for staging and local). 

The apps don't have any permissions apart from read and write access to issues of the repository where they're installed AND they request permission from users to act on their behalf through OAuth (again only on the repository where the app is installed). So the OAuth token that the app receives is valid only for a specific interaction (filing issues on `web-bugs` / `webcompat-tests` in our case).

There is an explanation of the process in https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/authorizing-github-apps#to-what-extent-can-a-github-app-know-which-resources-you-can-access--and-act-on-your-behalf 

And this article was helpful too https://tyers.io/posts/4-github-apps (Github app vs OAuth app section).

The changes that will need to be made:
1) Update `GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET` on the server to the one from GitHub app.
2) Change `SECRET_KEY` on the server to invalidate existing sessions / logout users. This is needed so users don't try to submit issues with a token for the old OAuth app.
When they log in again, they will be prompted to give access to the new GitHub app.
3) Revoke tokens from the old OAuth app

### DIFF
--- a/webcompat/views.py
+++ b/webcompat/views.py
@@ -20,6 +20,7 @@ from flask import send_from_directory
 from flask import session
 from flask import url_for
 from flask_firehose import push
+from flask_github import GitHubError
 
 from webcompat.api.endpoints import proxy_issue
 from webcompat.db import session_db
@@ -102,7 +103,7 @@ def login():
             # manually set the referer so we know where to come back to
             # when we return from GitHub
             set_referer(request)
-            return github.authorize('public_repo')
+            return github.authorize()
     else:
         return redirect(g.referer)
 
@@ -319,7 +320,20 @@ def create_issue():
                 url_for('show_issue', number=json_response.get('number')))
         # Authenticated reporting
         if form.get('submit_type') == 'github-auth-report':
-            if g.user:  # If you're already authed, submit the bug.
+            if g.user:
+                # If the user already authed, try to get user info from github.
+                # If this request is failing, the token has been revoked.
+                try:
+                    github.get('user')
+                except GitHubError as e:
+                    # In case of an error, logout the user and try to
+                    # fetch new token
+                    log.info(e)
+                    session.clear()
+                    session['form'] = form
+                    return redirect(url_for('login'))
+
+                # Otherwise proceed with submitting the form
                 json_response = report_issue(form)
                 session['show_thanks'] = True
                 return redirect(url_for('show_issue',


webcompat/views.py
```diff
@@ -102,7 +103,7 @@ def login():
             # manually set the referer so we know where to come back to
             # when we return from GitHub
             set_referer(request)
-            return github.authorize('public_repo')
+            return github.authorize()
     else:
```
ksy36, line 107: This is no longer needed as scopes are set during app creation/installation (and we don't really ask for any permissions apart from "act on your behalf").

webcompat/views.py
```diff
@@ -319,7 +320,20 @@ def create_issue():
                 url_for('show_issue', number=json_response.get('number')))
         # Authenticated reporting
         if form.get('submit_type') == 'github-auth-report':
-            if g.user:  # If you're already authed, submit the bug.
+            if g.user:
```
ksy36, line 323: This is to handle a case in the future when a user revokes a token, but is still logged in on webcompat.com and tries to submit an issue (not really required for this migration to GitHub apps, but nice to handle this exception).

webcompat/views.py
```diff
@@ -20,6 +20,7 @@
 from flask import session
 from flask import url_for
 from flask_firehose import push
+from flask_github import GitHubError
```
karlcow, line 23: *Thoughts*: 

I wonder if this project is still maintained. 
https://github.com/cenkalti/github-flask

Last PR on it was in 2020. and it is said to be compatible with Python 3.4
But I don't have any ideas without researching. This is not related to this PR.

karlcow (approved):
That looks so simple from the code point of view.

Thanks @ksy36 ! 
This is quite cool and one of the oldest requested feature.

miketaylr:
Nice work @ksy36!

webcompat/views.py
ksy36: I was thinking that too. We can revisit the use of this package in the near future. There are two alternatives that I found, https://github.com/singingwolfboy/flask-dance  and authlib (https://docs.authlib.org/en/latest/client/flask.html#flask-client)

ksy36:
Thanks!

PR #3643, opened 2021-11-17
Fixes #3642 - Add `browser-focus` to the EXTRA_LABELS allowlist.

### LINKED ISSUE #3642: Firefox Focus issues are not labeled correctly
The [Firefox Focus team added the webcompat-reporter](https://github.com/mozilla-mobile/focus-android/blob/7f8879347df74c3b56a0d18185ec582ec8b7d697/app/src/main/java/org/mozilla/focus/Components.kt#L105) and unfortunately did not tell us. The `browser-focus` label won't be applied because of this, so let's add it.

denschub:
This PR fixes issue #3642

## Proposed PR background

The [Firefox Focus team added the webcompat-reporter](https://github.com/mozilla-mobile/focus-android/blob/7f8879347df74c3b56a0d18185ec582ec8b7d697/app/src/main/java/org/mozilla/focus/Components.kt#L105) and unfortunately did not tell us. The `browser-focus` label won't be applied because of this.

This PR allows the label to be set, and also marks it as a Gecko browser.

r? @ksy36

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -211,9 +211,10 @@ for cat_label in cat_labels:
 EXTRA_LABELS = [
     'browser-android-components',
     'browser-fenix',
-    'browser-focus-geckoview',
     'browser-firefox-ios',
     'browser-firefox-reality',
+    'browser-focus',
+    'browser-focus-geckoview',
     'device-tablet',
     'type-fastclick',
     'type-google',
--- a/webcompat/webhooks/helpers.py
+++ b/webcompat/webhooks/helpers.py
@@ -28,6 +28,7 @@ GECKO_BROWSERS = ['browser-android-components',
                   'browser-firefox-mobile',
                   'browser-firefox-reality',
                   'browser-firefox-tablet',
+                  'browser-focus'
                   'browser-focus-geckoview',
                   'browser-geckoview', ]
 IOS_BROWSERS = ['browser-firefox-ios', ]


ksy36:
Thanks for creating this PR :) Do you think it would be better to change it to `browser-focus-geckoview` in the focus code, so we don't have two different labels for the same browser? (`browser-focus-geckoview` and `browser-focus`)

There is a handful amount of `browser-focus` now that were added manually: 
https://github.com/webcompat/web-bugs/issues?q=is%3Aissue+label%3Abrowser-firefox-focus+is%3Aopen
(And among those there are even some issues that have both 😀)

I don't have a strong opinion on this, `browser-focus-geckoview` can be the "old" focus and `browser-focus` can be the new one?

denschub:
That `-geckoview` suffix was a fragment for when there was a non-Gecko Firefox Focus, during the very early phases. These days, all Focus browsers are GeckoView, so we could consider that label obsolete. However, I also have no strong opinion here.

I'll let you make the call. If you prefer `borwser-focus-geckoview`, then I'm happy to submit a Pull Request to Focus to have that changed. Either way works for me. :)

denschub:
... but if we decide to go with `browser-focus`, then we should probably close all issue for non-Gecko Focus (that's no longer a thing), re-label all `browser-focus-geckoview` issues to use the label without the suffix, and remove the `-geckoview` label entirely. Not having two labels for "the same" product makes statistics much more easy. :)

So if you prefer the "new" label, then I'll update this PR to remove the old label.

ksy36:
After talking to Dennis we decided to keep `browser-focus-geckoview` as there are 6800 issues with this label, and change the label on the Focus side, so let's close this :)  We'll need to delete `browser-firefox-focus` from existing issues and user `browser-focus-geckoview` instead for better statistics.

PR #3645, opened 2021-11-18
Fixes #3644 - Add a "moved" milestone

### LINKED ISSUE #3644: Add a "moved" milestone
As per Slack,

> When we file a Gecko bug, or a Chromium bug or whatever, we close the web-bug as "duplicate". However, we also close actual duplicate webcompat reports with that tag, and we don't really have a good way to distinguish the two states. Bugzilla recently introduced the "CLOSED MOVED" state, to indicate that an issue has been moved out of bugzilla into another bugtracker.
>
> How do y'all feel about adding a "moved" milestone to web-bugs as well, so we can get a better understand of how many issues actually end up in another bugtracker - vs them just disappearing into the "invalid" pile as they do now?

and Ksenia, Tom, and James agreed that this sounds like a good idea. So this is my feature request to add support for an additional milestone to webcompat.com :)

ksy36:
This PR adds a "moved" milestone for issues that were moved to other bug trackers or repositories (bugzilla, fenix, firefox-ios, etc.)


webcompat/static/js/lib/models/issue.js
```diff
@@ -43,6 +43,7 @@ export const Issue = Backbone.Model.extend({
     var milestoneClass;
     if (response.milestone) {
       milestone = response.milestone.title;
+      milestoneClass = response.milestone.title;
     } else {
```
ksy36, line 47: I've noticed that milestone labels were not coloured on the issue page, because a css class was missing, this is a fix for that.

currently:
Screen Shot 2021-11-18 at 5 53 12 PM
will be:
Screen Shot 2021-11-18 at 5 53 32 PM

webcompat/templates/issue/issue-milestones.jst
```diff
@@ -15,5 +15,5 @@
     <span class="label label-<%= data.milestoneClass %> js-Milestone">
       <%= data.milestone %>
     </span>
-  </span>
+  </div>
 </script>
```
ksy36, line 19: There seems to be closing `</span>` instead of a `</div>`

config/environment.py
```diff
@@ -100,7 +100,8 @@
     'invalid': {'id': 0, 'order': 4, 'state': 'closed'},
     'non-compat': {'id': 0, 'order': 5, 'state': 'closed'},
     'wontfix': {'id': 0, 'order': 6, 'state': 'closed'},
-    'worksforme': {'id': 0, 'order': 7, 'state': 'closed'}}
+    'worksforme': {'id': 0, 'order': 7, 'state': 'closed'},
+    'moved': {'id': 0, 'order': 1, 'state': 'closed'}}
```
karlcow, line 104: see comment https://github.com/webcompat/webcompat.com/pull/3645#commitcomment-60534190

webcompat/static/js/lib/models/issue.js
```diff
@@ -43,6 +43,7 @@ export const Issue = Backbone.Model.extend({
     var milestoneClass;
     if (response.milestone) {
       milestone = response.milestone.title;
+      milestoneClass = response.milestone.title;
```
karlcow, line 46: good catch!

webcompat/templates/issue/issue-milestones.jst
```diff
@@ -15,5 +15,5 @@
     <span class="label label-<%= data.milestoneClass %> js-Milestone">
       <%= data.milestone %>
     </span>
-  </span>
+  </div>
```
karlcow, line 18: ooops

karlcow (changes requested):
Thanks @denschub for the issue and @ksy36 for the implementation

This is really cool.

A simple change on the order.

Also do not forget to create the actual milestones.

config/environment.py
ksy36: Good to know, I'll change that :)

ksy36:
I've made the change, could you please take another look @karlcow?

karlcow (approved):
Thanks! 🌊

PR #3690, opened 2022-04-18
Fixes #3656 - Top sites priority change

### LINKED ISSUE #3656: Alexa sunset. Priority rethinking.
On May 1st, 2022, Amazon will sunset Alexa.
https://support.alexa.com/hc/en-us/articles/4410503838999

The priority flag for our bug is defined according to Alexa ranking. We need to rethink this strategy.

ksy36:
The new script fetches results from both Tranco (global) and Alexa (regional) and saves results in 2 dbs. Regional db only includes domains whose priority is higher than in global db, therefore it's much smaller. I have decided to store the rankings in 2 dbs since once Alexa's API is deprecated, we can only update the global db and preserve the regional one for some time.

The script can be run is this way: 
`python3 ./tools/fetch_topsites.py --retrieve-regional --ats_access_key=<access_key> --ats_secret_key=<secret_key>`

and works as follows:

1. Gets top 10k csv from Tranco and saves it to the "global sites" db and cache.
2. Optionally fetches top 100 sites from specified regions if `--retrieve-regional` parameter is passed and saves the results in "regional sites" db.


tools/topsites/alexa.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/python
```
ksy36, line 1: This class is based on the code from the original `topsites.py`

webcompat/webhooks/helpers.py
```diff
@@ -106,8 +108,12 @@ def extract_priority_label(body):
     hostname = domain_name(extract_url(body))
     if hostname:
         priorities = ['critical', 'important', 'normal']
-        # Find host_name in DB
-        for site in site_db.query(Site).filter_by(url=hostname):
+        # Find host_name in regional DB first,
+        # if it doesn't exist, look in global
+        for site in regional_site_db.query(SiteRegional).filter_by(url=hostname):       # noqa
+            return f'priority-{priorities[site.priority - 1]}'
```
karlcow, line 114: I don't remember why the code is using a for loop. Maybe an opportunity to fix it. 

Do we expect `regional_site_db.query(SiteRegional).filter_by(url=hostname)` to return more than one domain? and it would mean that even if there was more than one domain, only the first hit would be taken into consideration, because of the return.

Basically the code is currently the equivalent of 

```python
sites = ['A', 'B', 'C', 'A2', 'C']
def extract(sites):
    for site in sites:
        if site.startswith('A'):
            return site
```

which would return 'A' always.

Should we just have something like.

```suggestion
        site = regional_site_db.query(SiteRegional).filter_by(url=hostname).first()
        if not site:
            site = global_site_db.query(SiteGlobal).filter_by(url=hostname).first()
        if site:
            return f'priority-{priorities[site.priority - 1]}'
        #… and so on for the subdomains.   
```

`.first()` returns None if the row doesn't exist.

or something else. 


Maybe there is a function to extract here. because the code is repeating the code in a loop. 

Or maybe I totally misunderstood the code 🍭 Long time I have touched this part.

webcompat/webhooks/helpers.py
```diff
@@ -116,7 +122,10 @@ def extract_priority_label(body):
                    for i, subpart in enumerate(subparts)
                    if 0 < i < hostname.count('.')]
         for domain in domains:
-            for site in site_db.query(Site).filter_by(url=domain):
+            for site in regional_site_db.query(SiteRegional).filter_by(url=domain):     # noqa
+                return f'priority-{priorities[site.priority - 1]}'
+
+            for site in global_site_db.query(SiteGlobal).filter_by(url=domain):
```
karlcow, line 128: The fact that we do it here again makes me think there is an opportunity for a function to reduce it and test it.

tests/unit/test_topsites_siterank.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Tests for Siterank class."""
```
karlcow, line 7: For the future, maybe there should be two mock DBs covering what we want instead of mocking. or in the setup a dictionary with data creating the mock DB. to think. Nothing to do now.

tools/fetch_topsites.py
```diff
@@ -0,0 +1,85 @@
...
+import argparse
+
+from tools.topsites.tranco import Tranco
+from tools.topsites.alexa import Alexa
+from tools.topsites.siterank import SiteRank
+from tools.topsites.siterank import DATA_PATH
+from tools.topsites.utils import parse_site_dom_element
+
+REGIONS = ['US', 'FR', 'IN', 'DE', 'TW', 'ID', 'HK', 'SG', 'PL',
+           'GB', 'RU']
```
karlcow, line 24: Another thought, this was initially created like this I guess because of priority market for Firefox, but we can imagine that other browsers could have a desire to extend this. Nothing to do now.

tools/fetch_topsites.py
```diff
@@ -0,0 +1,85 @@
...
+    parser.add_argument(
+        "--ats_secret_key",
+        help="AWS secret key.",
+        type=str,
+    )
+
+    args = parser.parse_args()
+
+    if args.retrieve_regional:
+        print('Warning: Alexa APIs will be deprecated on December 15, 2022.')
```
karlcow, line 51: Should the code have a failsafe feature. Either based on the date. "Hey it's past December 15, 2022. We will ignore the request." Or based on failing the request to Alexa.

karlcow (changes requested):
Thanks @ksy36 
Big update! Impressive. 

I'm a bit concerned by the amount of mock we do for testing the DB. It's code inherited, so probably it would be good to open a new issue on this for the future. That could help us to probably test some weird scenarios.

There's also probably an optimization which would be good to do on the queries in a for loop which will return only one result. 

Request Changes just for getting the discussion rolling and see your thoughts.

webcompat/webhooks/helpers.py
```diff
@@ -106,8 +108,12 @@ def extract_priority_label(body):
...
-        for site in site_db.query(Site).filter_by(url=hostname):
+        # Find host_name in regional DB first,
+        # if it doesn't exist, look in global
+        for site in regional_site_db.query(SiteRegional).filter_by(url=hostname):       # noqa
+            return f'priority-{priorities[site.priority - 1]}'
+
+        for site in global_site_db.query(SiteGlobal).filter_by(url=hostname):
             return f'priority-{priorities[site.priority - 1]}'
         # No host_name in DB, find less-level domain (>2)
         # If host_name is lv4.lv3.example.com, find lv3.example.com/example.com
```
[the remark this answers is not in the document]
ksy36: Yeah, I've left this part untouched from the previous version. Updated now though :)

webcompat/webhooks/helpers.py
ksy36: The domain names are unique, so using `.first()` will work indeed. Thanks for the suggestion, I've changed the function

tests/unit/test_topsites_siterank.py
ksy36: Yeah, good idea, I'll file an issue for that.

tools/fetch_topsites.py
ksy36: Thanks, I've added the check 👍

karlcow (approved):
Thanks a lot @ksy36 for addressing the changes. 🔝 🥳

ksy36:
Thanks!!

PR #3691, opened 2022-04-28
Fixes #3016 - Remove current browser from list of tested browsers

### LINKED ISSUE #3016: When reporting with another browser the list of other browsers needs to be updated
The current workflow assumes always Firefox.
webcompat.com is about any browsers.

Here I'm reporting on the left with Safari, and I'm getting proposed… Safari.

Capture d’écran 2019-10-23 à 16 17 31

ksy36:
Following the rule from this comment https://github.com/webcompat/webcompat.com/issues/3016#issuecomment-545686016


ksy36:
As an example, in Firefox:

Screen Shot 2022-04-28 at 4 52 44 PM

Chrome:
Screen Shot 2022-04-28 at 4 53 20 PM

When the current browser is not in the list OR when setting custom browser, all icons are displayed:

Screen Shot 2022-04-28 at 4 56 45 PM

karlcow (approved):
That looks good. 
Thanks @ksy36

ksy36:
Thank you :)

PR #3704, opened 2022-06-14
Fixes #3703 - Automatically label nsfw issues

### LINKED ISSUE #3703: Automatically label nsfw issues and delete a link to a screenshot
Let's automatically label adult, gore, etc. sites based on a block list and delete a link to the screenshot in the issue body.

Possible lists to consider:

https://oisd.nl/
https://github.com/cbuijs/oisd (Optimized version of OISD BlockLists)
https://github.com/StevenBlack/hosts
https://github.com/blocklistproject/Lists
https://github.com/EnergizedProtection/block

ksy36:
This is a first iteration of NSFW moderation based on a blocklist. I've looked at a few lists and decided to go with OISD (https://oisd.nl/). Most of these lists come in txt format, so I parsed and saved it to a DB on the server. The plan is to periodically fetch updates for this list, but for now it is static. I'll file a follow up issue for that.

Once an anonymous/auth issue is filed, the script looks whether the reported domain is in NSFW db and deletes the image from the issue body before it's posted to GitHub. Also `nsfw` label is added to both private and public issue (after it's moderated).

After testing this a bit, I'm thinking it probably makes sense to block image uploading in the ui, instead of post upload. However it might still be helpful to allow uploading and keep images link in the private repo, in case they're helpful for triage. That would require some feedback from Oana and Raul, so we can revisit this approach after this change is "live" for some time.

### DIFF
--- a/webcompat/helpers.py
+++ b/webcompat/helpers.py
@@ -785,6 +785,15 @@ def get_filename_from_url(uri):
     return script_path
 
 
+def get_domains(hostname):
+    """Extract subdomains"""
+    subparts = hostname.split('.')
+    domains = ['.'.join(subparts[i:])
+               for i, subpart in enumerate(subparts)
+               if 0 < i < hostname.count('.')]
+    return domains
+
+
 @app.context_processor
 def register_get_filename_from_url():
     return dict(get_filename_from_url=get_filename_from_url)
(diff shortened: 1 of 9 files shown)


webcompat/helpers.py
```diff
@@ -785,6 +785,15 @@ def get_filename_from_url(uri):
     return script_path
 
 
+def get_domains(hostname):
```
ksy36, line 788: moved this function from webhooks helpers, so it can be used in multiple places

denschub (approved):
Looks good to me, thanks for working on this. :)

ksy36:
Thanks!

PR #3729, opened 2022-08-27
Fixes #3681 - Remove analytics for the form wizard

### LINKED ISSUE #3681: Disable google analytics
We've enabled GA to run a few experiments to try and improve the drop-off rate of the web compact form. As this work is now done, we can turn off tracking for the time being.

ksy36:
This PR removes detailed analytics from the form wizard that we added to track the success rate of the form.

The standard "pageview" analytics still remain. We don't use the data it collects at the moment, though it could be useful if we need it at some point in the future (to see the spread between android/ desktop reports, for example). With that being said, I'm not opposed to removing it altogether :)

### DIFF
--- a/webcompat/static/js/lib/issue-wizard-popup.js
+++ b/webcompat/static/js/lib/issue-wizard-popup.js
@@ -4,7 +4,6 @@
 
 import $ from "jquery";
 import Mousetrap from "Mousetrap";
-import { sendAnalyticsEvent } from "./wizard/analytics.js";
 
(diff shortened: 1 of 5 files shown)


denschub (approved):
I have no objections in removing analytics altogether if we don't have use for the data, but this can be done at a later time as well! This looks good to me.

ksy36:
Thanks!

LOHAYBOY999 (approved):

Issue #3735, opened 2022-10-05
Honor `prefers-reduced-motion`


denschub:
As per [this Reddit post](https://www.reddit.com/r/firefox/comments/xvx3yo/if_animation_gives_you_migraines_you_may_need_to/), we should probably set the animation duration to 0 in [`showContainer`/`hideContainer`](https://github.com/webcompat/webcompat.com/blob/408d8032dcdcdd94893b2ca84e071c9b71ba7926/webcompat/static/js/lib/wizard/ui-utils.js#L33-L72) based on a  `window.matchMedia('(prefers-reduced-motion: reduce)')` media-query check.

ksy36:
Fixed in https://github.com/webcompat/webcompat.com/pull/3747

MarjaE2:
Still broken in Firefox.

PR #3747, opened 2023-01-20
Fixes #3505 - Remove animations/scrolling for prefers-reduced-motion

### LINKED ISSUE #3505: [A11Y] Bug report form can trigger migraines and motion sickness for users with visual or vestibular issues
Visiting https://webcompat.com/issues/new

1. Type url into 1st field.

2. Click confirm.

3. Get hit with blinding pain. Get migraine and motion sickness. It's hard to see as it happens, but I think there's flashing and smooth scrolling. Flashing often gives me migraines, and smooth scrolling, zooming, etc. always reliably do.

See also https://github.com/webcompat/webcompat.com/issues/3134

ksy36:
Fixes https://github.com/webcompat/webcompat.com/issues/3505 (and https://github.com/webcompat/webcompat.com/issues/3735)

### DIFF
--- a/webcompat/static/js/lib/wizard/helpers/postMessageParser.js
+++ b/webcompat/static/js/lib/wizard/helpers/postMessageParser.js
@@ -6,9 +6,7 @@ import notify from "../notify.js";
 import { isSelfReport, convertToDataURI } from "../utils.js";
 import { blobOrFileTypeValid, isImageDataURIValid } from "../validation.js";
 import { sendAnalyticsCS } from "../analytics.js";
-import { extractPrettyUrl } from "../utils.js";
 
-const showStep = (id, data) => notify.publish("showStep", { id, data });
 const updateStep = (id, data) => notify.publish("updateStep", { id, data });
 const updateUrl = (url) => updateStep("url", { url });
 const updateHidden = (data) => updateStep("hidden", { data });
@@ -20,9 +18,6 @@ const handleMessage = (message) => {
 
   sendAnalyticsCS(utm_campaign, utm_source);
   if (url) {
-    showStep("category", {
-      url: extractPrettyUrl(url),
-    });
     updateUrl(url);
   }
   updateHidden(additional);
--- a/webcompat/static/js/lib/wizard/ui-utils.js
+++ b/webcompat/static/js/lib/wizard/ui-utils.js
@@ -8,6 +8,11 @@ import anime from "animejs/lib/anime.es.js";
 const HEADER_HEIGHT = 130;
 const SPACING = 24;
 
+const isReducedMotion = () => {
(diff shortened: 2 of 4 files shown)


webcompat/static/js/lib/wizard/helpers/postMessageParser.js
```diff
@@ -20,9 +18,6 @@ const handleMessage = (message) => {
 
   sendAnalyticsCS(utm_campaign, utm_source);
   if (url) {
-    showStep("category", {
-      url: extractPrettyUrl(url),
-    });
     updateUrl(url);
```
ksy36, line 21: I've also removed the "skip the url and scroll to the category step" when reporting through "Report site issue". It was part of the experiment to try and improve the form drop off. Thinking it might be confusing for new users as it happens immediately, skipping the part that we're looking for web compatibility issues (and "Learn more about web compatibility" pop up) .

webcompat/templates/issue/thanks.jst
```diff
@@ -8,4 +8,3 @@
   <p>You can return to this web page at any time to track the progress of this report.</p>
 </section>
 </script>
-˚
```
denschub, line 11: 😆

webcompat/static/js/lib/wizard/ui-utils.js
```diff
@@ -52,8 +69,19 @@ export const showContainer = (container, cb) => {
...
+  domElement.style.display = "none";
+  container.removeClass("open");
+};
+
 export const hideContainer = (container) => {
   const domElement = container[0];
+
+  if (isReducedMotion()) {
+    return hideContainerReduced(container);
+  }
```
denschub, line 83: Is there a reason to have a second hide-method instead of setting the `duration` in the animation-call below to 0? Intuitively, it feels like it would be simpler to do that, but my intuition might be wrong.

ksy36: I've tried it yeah, but the function inside `anime.begin` callback doesn't run with duration: 0 :(
```js
    begin: () => {
      domElement.style.display = "flex";
    },
```
(this is basically what's making next step visible).

Tried to increase duration to 20 or 40, then it works most of the time and sometime doesn't, which seemed pretty fragile, so I just decided to have a separate function to have more control.

webcompat/templates/issue/thanks.jst
ksy36: 🙈

webcompat/static/js/lib/wizard/ui-utils.js
denschub: Ah, that's a shame. Thanks for explaining!

denschub (approved):
Thanks for working on this!

ksy36:
Thank you!

PR #3759, opened 2023-04-13
Fixes #3758 - Fix a problem with issue getting lost when filing an auth report

### LINKED ISSUE #3758: Fix a problem with issue getting lost when filing an auth report
STR: 

1. Log out on https://webcompat.com/ if logged in (but remain logged in on GitHub)
2. Visit https://www.benefits.ml.com/Login/Login
3. Click on Help > Report site issue
4. Fill in the form and click in Report via GitHub

Expected:
Issue filed and redirected to the issue page, i.e. `https://webcompat.com/issues/<issue-id>`

Actual:
Redirect to the main page, issue gets lost

The problem seems to be that the form is not saved to session cookie, likely due to it containing too much data:

`
projects/webcompat.com/env/lib/python3.9/site-packages/werkzeug/sansio/response.py:235: UserWarning: The 'session' cookie is too large: the value was 6970 bytes but the header required 26 extra bytes. The final size was 6996 bytes but the limit is 4093 bytes. Browsers may silently ignore cookies larger than this.
`

The saving is happening here:

```python
        # Authenticated reporting
        if form.get('submit_type') == 'github-auth-report':
            if g.user:
                ...
            else:
                # Stash form data into session, go do GitHub auth
                session['form'] = form
                return redirect(url_for('login'))
```

And at retrieval `session.get('form', None)` is actually `None`:

```python
# OAuth2 callback handler that GitHub requires.
# If this moves, it needs to change in GitHub settings as well
@app.route('/callback')
@github.authorized_handler
def authorized(access_token=None):
    """Set the callback route for oauth2 with GitHub."""
    ...
    ...
    if session.get('form', None) is not None:
        return redirect(url_for('file_issue'))
    else:
        return redirect(g.referer)
```

This is a variation of https://github.com/webcompat/webcompat.com/issues/3508. 

It only happens on some websites and I think the reason is the large amount of console logs. To fix this, we can strip the console logs from the form, since it's being saved in a separate file anyways.

ksy36:
There is a bit of investigation in https://github.com/webcompat/webcompat.com/issues/3758#issue-1666894730, but in short, stripping the form of console logs after they're saved in a separate file helps reduce the amount of serialized data stored in the session cookie.

### DIFF
--- a/webcompat/static/js/lib/wizard/steps/upload-helper/console-logs-upload.js
+++ b/webcompat/static/js/lib/wizard/steps/upload-helper/console-logs-upload.js
@@ -35,6 +35,9 @@ export const uploadConsoleLogs = () => {
   const formdata = new FormData();
   formdata.append("console_logs", JSON.stringify(details.consoleLog));
 
+  delete details.consoleLog;
+  detailsInput.val(JSON.stringify(details));
+
   return $.ajax({
     contentType: false,
     processData: false,


denschub (approved):
Ohh, good catch!

ksy36:
Thanks!

PR #3771, opened 2023-06-20
Fixes #3769 - Dual form reporter

### LINKED ISSUE #3769: Create a simplified UI for the dual flow form
The new UI should contain:

"Url"
"Description"
“Submit” button (simple report) - user can submit the report right away without providing additional details. This report won't be stored in GitHub, only BigQuery.
“Add more details” button (current reporting flow) - user will proceed to the detailed report, i.e. Category, Tested browsers, Screenshot, Submit. This report will be stored in both GitHub and BigQuery.

ksy36:
In this PR I've added the following:
- Dual flow for the reporter
- Reports are saved to BQ for both short and detailed reports


denschub (approved):
My Python skills aren't the strongest, but this looks reasonable - and it works. r+ :)

ksy36:
Thanks!

Issue #3775, opened 2023-07-11
Add UA string to the details, if it's missing 


ksy36:
It appears that UA string is missing in some reports stored in BQ, we should add it to be able to detect the browser/device a user is encountered the issue.

PR #3776, opened 2023-07-12
Issue #3775 - Add UA string to details (if it's missing in the additional data from the reporter)

ksy36:
... or report site issue is not used :)


denschub (approved):
Thanks for the quick change!

ksy36:
ty!

PR #3777, opened 2023-07-18
BigQuery-related changes

denschub:
Two things,

- Since I've had to recreate tables anyway, I renamed `user_reports_test` to `user_reports_staging` so that it is very clear what that table is, and that it's used on staging (as opposed to "this is a temporary table used for some testing")
- `reported_at` was not meant to be `REPEATABLE`. Whoops.

r? @ksy36


ksy36 (approved):

ksy36:
Thanks :)

